Issue #18, opened 2023-01-16
Arhcitecture document coordinator


bonomat:

bonomat:
closed in favor of [#17](https://github.com/get10101/10101/issues/17)

Issue #17, opened 2023-01-16
Architecture document app


bonomat:
Create a highlevel architecture overview showing how `mobile` and `coordinator` are interacting with each other.

- [x] #36 
- [x] #48 
- [x] #49 
- [x] #55 
- [x] #80
- [x] #81
- [x] #82
- [x] #83
- [x] #84
- [x] #85

// edit: 31.01.2023 split this ticket up in multiple smaller ones to address the discussions started in https://github.com/get10101/10101/pull/33 as individual ADRs

holzeis:
Started this ticket, with looking at how we could reach the phone from the outside. I had a look at the breez app, which offers a very nice developer menu in their app. They are running an lnd node on the phone and expose the full lncli api to you.

`getinfo` shows that there is no uri to reach the phone.. however, if i connect to my lightning node I can find the lnd node running on my phone (breez) in my list of peers. including an ip address.

Using that ip address I was able to open a channel from my lnd node to the Breez app

https://mempool.space/tx/067a571e7b8a863d93358a0a133dfff2902e7d4001367318c0c99629a4615cc3

so, we don't need any special handling other than the lightning protocol that's already there with rust-lightning. I guess it did not work for us during the poc for other reasons related to how threads are handled in the frb bridge.

holzeis:
However you need to have the breez app open to receive sats.. That's a bit annoying. Would be great if we could improve that.

Restioson:
[Syncthing](https://syncthing.net/) has some kind of service running in the background on android to make this work but it seems to be very annoying. For instance, some android distributions apparently require a constant notification for it not to kill the app in the background.

bonomat:
All tickets either closed or done 😅

PR #20, opened 2023-01-17
CI Testing guidelines

### LINKED ISSUE #11: [mobile] define build pipeline document: instead of just creating multiple workflows for all the things we should get a clear picture of what and how things should be build

### LINKED ISSUE #13: CI pipeline document for our coordinator
We want to have a clear picture of what and how things should be build. 

The goal of this document is to describe the CI workflows we need and ensure that we think about them in advance. 

Note: this issue covers only the coordinator while #11 covers the mobile app.

bonomat:
The below shouldn't be that new to you as it is based on past experiences and decisions. What is new is that we should aim for a sub 10 minutes test execution. Let's see if we can achieve this :)

Initially I expected this to be more detailed but I stopped and went high-level.
Once we have https://github.com/get10101/10101/issues/18 and https://github.com/get10101/10101/issues/17 we can add more.

Resolves #11 
Resolves #13 

Note: I opted for asciidoc because I like the format more, however, I could not find a formatter for it which is a bid sad. 
So I'm happy to reconsider.


docs/001-ci-strategy.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+This document outlines the reasoning behind our CI strategy.
+
+== Decision
+
+* We have 2 stages of tests
+** Unit tests:
+In unit tests we stub all DB/io/networking/etc to achieve a sub *10 minute* runtime.
+Unit tests can run on every commit.
+** Integration tests:
```
holzeis, line 16: I guess this better applies to system tests - we we are actively testing the integration between multiple systems and their non-functional requirements, like performance, security, etc.

docs/001-ci-strategy.adoc
```diff
@@ -0,0 +1,101 @@
...
+Integration tests run only occasionally and *only* if all unit tests have passed.
+We run integration tests only on `bors r+`.
```
holzeis, line 18: In my opinion, running tests after the change has been merged is cumbersome. In case they fail you have a broken `main`.

So, my suggestion would be to work on the integration tests and unit tests to run below 10 minutes, by focusing less on system tests. 

System tests are then executed on an regular basis, but not after every merge.

docs/001-ci-strategy.adoc
```diff
@@ -0,0 +1,101 @@
...
+Once this branch builds, it is automatically merged into master.
+* We only run tests for what has been changed. E.g. If a user only does frontend changes, we only run tests for the frontend.
+
+== Context
+
+A continuous integration and continuous delivery platform (CI/CD) should help us to deliver high quality software to our users.
+In the past we have been heavily using https://github.com/features/actions[GitHub Actions] but we had our problems with performance and maintenance of the CI suite.
```
holzeis, line 25: Just looked at old integration test runs from 10101-poc. Android tests actually didn't took too long, but the ios integration tests took ages. I suspect there is room for improvement, though.

docs/001-ci-strategy.adoc
```diff
@@ -0,0 +1,101 @@
...
+If we keep following our well-known `bors r+`-workflow, we can execute stage 2 tests only when merging into the `staging` branch, i.e. on or when `bors r+` is executed.
+
+The advantages of this two stage approach are:
+
+1. separation of test files: in order to be able to differentiate between unit tests and integration tests, we will need to differentiate them on a file level as well. This will result in cleaner code and cleaner interfaces as integration tests should only _know_ about external api.
+2. separation of test setup: we could theoretically easily deploy the integration test runner onto a custom machine
+3. by-passing integration tests: while not favorable, the advantage of this approach is that we can bypass the second stage if needed and merge directly into master without running the whole integration test suite but at least the unit test suite.
+
+[IMPORTANT]
+.Tests should be executed locally first
```
holzeis, line 64: I have to object to that statement a bit. Of course tests should be run locally, but our process should not depend on a dev ensuring locally all checks are met.

I understand that costs are coupled to using CI resources, but I'd rather follow an approach to reduce the required resources for running tests than to get dependent on a manual task. The proposed approach goes against a high degree of automation in my point of view.

docs/001-ci-strategy.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+1. Stage 1: Run a build and unit tests with the database/io/network/etc completely stubbed out.
+2. Stage 2: Run larger scale interaction tests.
+
+Stage 1 is run on the last commit of a PR.
+Stage 2 is only run when merging into `staging`, i.e. after running `bors r+`
+
+==== Option 2 (so far)
+
+So far we have run all tests on every commit. This is slow and cumbersome.
```
holzeis, line 101: Did we already looked at all possibilities to make it faster? I wouldn't want give up automated tests on the CI so easily just because it was slow in the poc.

(I don't get the cumbersome part 😅 - it's more cumbersome to have the test run on your local machine, as that will block me from working on something else)

dprint.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://dprint.dev/schemas/v0.json",
+  "projectType": "openSource",
+  "incremental": true,
+  "includes": [
+    "**/*.{md}",
+    "**/*.{json}"
+  ],
+  "plugins": [
+    "https://plugins.dprint.dev/markdown-0.9.2.wasm",
```
holzeis, line 10: I guess there is no dprint plugin for ascii doc? 🙈

holzeis (commented):
Thanks for writing this up 🙂

docs/001-ci-strategy.adoc
bonomat: You are right, I was actually talking about system tests in this context.

docs/001-ci-strategy.adoc
bonomat: > I have to object to that statement a bit. Of course tests should be run locally, [...]

I think you are not objecting :) 

The CI is a backup system to ensure that the tests don't just run on your machine but also on others. It's not a replacement for running tests.

The problem is that each push to _try to make tests succeed_ on CI results in a new test runner to be spun up. This results in other actions having to wait because a new test runner cannot be allocated which slows down our overall development process.

docs/001-ci-strategy.adoc
bonomat: > Did we already looked at all possibilities to make it faster? I wouldn't want give up automated tests on the CI so easily just because it was slow in the poc.

It's based on years and years of experience :P Just kidding. I'm sure others will agree that slow running CI was always our problem in the past.

I don't want to give up on automation neither. I hope this was not the impression of this document. I think that some tests (the slow ones) don't have to run that regularly. 

> (I don't get the cumbersome part 😅 - it's more cumbersome to have the test run on your local machine, as that will block me from working on something else)

Your machine is 10x more powerful than what we get on CI. 
Make a copy of your workspace and you can run tests on one project while continuing with something else :)

docs/001-ci-strategy.adoc
bonomat: Which test did you look at?

This is a random build from 10101-poc running android integration tests which took >30 minutes

https://github.com/get10101/10101-PoC/actions/runs/3677365461/jobs/6219226765

docs/001-ci-strategy.adoc
bonomat: > In my opinion, running tests after the change has been merged is cumbersome. In case they fail you have a broken main.

This was not my proposal. We need to run the tests before merging but after everything else was approved on. Which means, in the ideal case we have a bunch of revisions on the PR, hit `bors r+` and then the slow tests are run. If they fail (which they shouldn't because you tested it locally already), then we merge into master.

docs/001-ci-strategy.adoc
holzeis: on your example the actual integration tests took only 12-15 minutes.. still longer than the test I looked at, but it might be possible to improve that?

One additional note, we are running on shared instances so it might very well be that we do not get a lot of resources and that will have an impact on how fast our pipeline will run. We could consider paying for dedicated resources. e.g. setup a dedicated VM on google cloud as runner.

docs/001-ci-strategy.adoc
holzeis: ok I see what you mean. so if I understand you correctly, that would mean that on bors r+ the heavy tests will be run and the action (merge) will only succeed if the tests succeed. Meaning we will not run into the issue I feared of potentially getting a broken main. correct?

docs/001-ci-strategy.adoc
holzeis: > Make a copy of your workspace and you can run tests on one project while continuing with something else :)

That sounds cumbersome 😝

docs/001-ci-strategy.adoc
bonomat: This is correct :)

docs/001-ci-strategy.adoc
bonomat: Ah, I was counting the build time as well.

dprint.json
bonomat: As mentioned, unfortunately not.

1010Tom (approved):
LGTM

bonomat:
This PR has been open for a week now. It assume this ADR is not controversial and will go ahead and merge it.

luckysori (commented):
I never reviewed it explicitly, but we did take it into account when working on #30.

I say we give this a go. We can always amend this decision in the future if this new approach doesn't work so well for us. The risk here is smaller than committing to event sourcing, for example 😉

I also had the idea to schedule a retro specifically about testing (including CI) a month after development starts, so that we give ourselves a chance to consider how we are doing before continuing with the same process.

PR #30, opened 2023-01-19
Add documentation about testing

### LINKED ISSUE #12: Testing document for mobile
A document which describes what and how we are testing our mobile app.
E.g. do we need to run a simulator, do we write unit tests for the front end, do we write integration tests, etc.

### LINKED ISSUE #14: Testing document for our coordinator
A document which describes what and how we are testing our mobile app.

luckysori:
Fixes #12.
Fixes #14.

This is just our take on the relatively open-ended tasks listed above. @da-kami and I went into lengthy discussions to arrive at this already, but we can take this PR review as an opportunity to continue the discussion until we reach some kind of consensus.


docs/diagrams/testing-component-diagram.puml
```diff
@@ -0,0 +1,46 @@
+@startuml
+
+package "Rust Common" {
+    [Orderbook]
+    [Execute]
+    [Rust Lightning]
+}
+
+node "App" {
```
holzeis, line 9: Would be great if we could align on a naming - in https://github.com/get10101/10101/pull/31 we use a different terminology.

```
App = mobile
FRB / App Backend = native
```

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Integration tests
+
+These are your traditional "UI tests".
+The clicks on the UI are simulated as if someone was using the application.
+
+They remain lightweight as they only call to a mocked version of the API provided by `flutter_rust_bridge`.
+We do not currently specify how this API shoud be mocked.
+We considered (1) [mocking the dependency in Dart](https://docs.flutter.dev/cookbook/testing/unit/mocking) and (2) mocking the dependency in Rust.
+We would prefer (1) as we think it would lead to simpler code than (2), but we are not sure if it's a feasible approach.
```
holzeis, line 86: I prefer (1) as well - since we not aim to test the generated flutter_rust_bridge code.

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+Thusly, we recommend that system tests involving the `app` backend be written in Dart.
+These tests will be run using the [`dart test` command](https://dart.dev/tools/dart-test).
+These tests will target the `flutter_rust_bridge` Dart API generated by the `app` backend.
+
+The details of how the test environment should bet set up or consumed are not yet specified.
+Regardless, we should aim to be able to conveniently run these tests locally by providing `docker-compose` files and/or setup scripts or by connecting to an always-available semi-public environment.
+
+This document does not specify what type of Bitcoin blockchain these tests should run against.
+We do believe it would be important to be able to run them against `signet` and `testnet`.
```
holzeis, line 113: Why is it important that these tests run against testnet and signet? in other words, what insights do we hope to get what we can't get in the other?

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Load tests
+
+We should leverage a semi-public environment to assess the scalability of our solution.
+This environment should not be misused for traditional system testing.
+
+## Closing thoughts
+
+Here we list what we think are the most important takeaways from this document:
```
holzeis, line 122: A good practice when it comes to testing is to look at it like a pyramid. 

```
              -- system tests --
      ------- integration test -------
---------------- unit tests ---------------- 
```

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+### The `app` frontend
+
+This includes Flutter code (written in Dart) and the generated `flutter_rust_bridge`.
+In practice, there shouldn't be a strong need to write tests for the `flutter_rust_bridge` code we write (Rust) since it wil be a thin layer.
+
+#### Unit tests
+
+We can apply similar principles as the ones mentioned for Rust code.
+
+#### Widget tests
```
holzeis, line 72: In my point of view widget tests are also unit tests.

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+Additionally, this kind of tests usually involve mocked state.
+
+We are aiming to not spend too much time maintaining Rust integration tests, in favour of more comprehensive tests.
+
+### The `app` frontend
+
+This includes Flutter code (written in Dart) and the generated `flutter_rust_bridge`.
+In practice, there shouldn't be a strong need to write tests for the `flutter_rust_bridge` code we write (Rust) since it wil be a thin layer.
+
+#### Unit tests
```
holzeis, line 68: This section appears twice. I guess we can combine it with the one above.

docs/diagrams/testing-component-diagram.puml
```diff
@@ -0,0 +1,46 @@
+@startuml
```
holzeis, line 1: I like that diagram, but it has little to do with the testing approach described below. I would move that to another section e.g. component diagram etc.

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+Here we list a non-exhaustive set of assumptions that influenced our findings:
+
+### Common Rust code between `app` and `coordinator`
+
+There will be a number of crates that will be used as dependencies in both `app` and `coordinator`.
+
+### No initial effort to keep different versions of `app` and `coordinator` compatible
+
+The `coordinator` will not be backwards-compatible with older versions of the `app` until later in the development of 10101.
```
bonomat, line 15: 👍

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+Unit tests will ideally only be written for pure functions.
+
+They can be handy as API snapshot tests. For example, to ensure that a `serde` implementation doesn't change.
+
+#### Integration tests
+
+These are tests that depend on multiple components (perhaps across module boundaries).
+Additionally, this kind of tests usually involve mocked state.
+
+We are aiming to not spend too much time maintaining Rust integration tests, in favour of more comprehensive tests.
```
bonomat, line 61: What do you mean with `in favor of more comprehensive tests`?

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+The details of how the test environment should bet set up or consumed are not yet specified.
+Regardless, we should aim to be able to conveniently run these tests locally by providing `docker-compose` files and/or setup scripts or by connecting to an always-available semi-public environment.
+
+This document does not specify what type of Bitcoin blockchain these tests should run against.
+We do believe it would be important to be able to run them against `signet` and `testnet`.
+
+### Load tests
+
+We should leverage a semi-public environment to assess the scalability of our solution.
```
bonomat, line 117: Which component/part do you want to assess exactly here?

docs/testing.md
bonomat, line 61: The way how I interpret this: 
- you prefer writing system tests over integration tests (with mocked data).

Is this not the path we were taking in the past where we had an extensive test suit of long and slowly running tests? 

I would be in favor of having maximum a handful of system tests which need a test environment to be spun up (e.g. docker-compose). For example, these could cover only the main paths (aka happy path). Any edge case scenario I would cover on a lower level with mocked data for fast test execution.

docs/testing.md
```diff
@@ -0,0 +1,130 @@
...
+
+### The `app`'s state will be controlled by its Rust backend
+
+Thanks to this we should be able to write system tests against the `app`'s API without involving the frontend.
+
+### The architecture will be similar to this component diagram
+
+This diagram does include some of the assumptions listed above:
+
+![Assumed component diagram](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/get10101/10101/d7619613324df3ce5c0f0a35d97bf24323f9515b/docs/diagrams/testing-component-diagram.puml)
```
bonomat, line 37: 🙃 If you use mermaid you can have this code inline :)

bonomat (commented):
Thanks for starting this discussion. I've added some comments in line and want to discuss the relation between integration tests and system tests :)

docs/diagrams/testing-component-diagram.puml
luckysori: It's a fair point. We only used it to be explicit about our assumptions at this stage. Only after we imagined the components were we able to actually think about how they would be tested.

We should eventually remove this diagram. Maybe once #17 is specified?

docs/testing.md
luckysori: In the past, we used to only test using `regtest`. manually trying our software on `testnet` then revealed significant issues.

In ItchySats, we exclusively "tested" against `testnet` when we actually involved a blockchain of any kind.

Here I am assuming that `signet` takes the old role of `regtest`: tests that can be run locally or on CI, without depending on semi-public infrastructure. These tests are valuable because you have control of the blockchain, so you can make them fast and offline.

With `testnet` I am referring to a public test blockchain, so a public or semi-public `signet` would also work in theory!

docs/testing.md
luckysori: This section is not fleshed out, but we were referring to the whole system (`app` + `composer` + `blockchain`).

Daniel pointed out that our testnet setup was actually just load testing.

> Which component/part do you want to assess exactly here?

The question still stands though. I think it would make sense to eventually load test the `composer`, to figure out how many users it can serve. Like we used to do in ItchySats. In ItchySats we never quite solved all the problems related to load, so it would be good to avoid that situation in 10101.

docs/testing.md
luckysori: Fair enough. I think we were going off Flutter's own documentation, but it's true that more generally they belong to another category.

docs/testing.md
luckysori: We discussed this topic briefly offline. These are my takeaways from the meeting:

- Here we are defining a testing strategy, but we should give ourselves an opportunity to revise it over time. With this in mind, I take responsibility to schedule a "Testing strategy retro" in about a month's time.
- As context, we agree that our ItchySats actor tests were not good enough: flaky, expensive to maintain and with long execution times.
- We agree that we must not impose long test execution times on devs or CI.
- We agree that we should consider not running all tests of all kinds on every push to the repository. Although this concern can be addressed independently in #20, it can guide us in choosing how we should write the test themselves.
- We want to explore the long-discussed possibility of running system tests that start from an in-between state, in the interest of shortening text execution times. This could severely help in allowing us to test edge case scenarios via system tests, perhaps addressing Phillip's concern mentioned in the previous comment.
- We agree that this document does not represent a clear plan of action in its current state. @da-kami volunteered to create a document that more clearly and succinctly specifies our current testing strategy.

As a side note, Daniel expressed that the way we model things has a very big effect on the quality and usability of our unit tests. He used the `Cfd` struct as an example of a bad unit, which was very hard to unit test.

docs/testing.md
da-kami: Unfortunately mermaid does not have component diagrams - I *wanted* to use it but had to fallback to PlantUML :/

da-kami: we might be able to use one of their other diagram types, but so far I did not find a good fit, see:

https://github.com/mermaid-js/mermaid/issues/1462

https://mermaid.js.org/intro/

Restioson (approved):
This is very detailed and all seems good to me! 

One thing which is not strictly related to testing itself: it would be nice to be able to reproduce the deployed setup locally (maybe using docker-compose for instance?) for poking around in case of issues post-deployment.

docs/testing.md
bonomat: You can create a component diagram using mermaid: https://mermaid.js.org/syntax/c4c.html. 
Eitherway, plantuml is fine.

docs/testing.md
da-kami: It depends on your definition of a unit test - if we talk about one distinct atomic unit the widget test is more on an integration test level because it might mock a whole lifecycle of  the widget and it's behavior throughout. I guess it depends how to write that test.

Here's the docs we read when writing this document: https://docs.flutter.dev/testing#widget-tests

docs/diagrams/testing-component-diagram.puml
da-kami: I agree that this diagram should live somewhere else but could be referenced here! Given that it was included in https://github.com/get10101/10101/pull/33 I think we can remove it from this PR.

docs/diagrams/testing-component-diagram.puml
da-kami: I think `mobile` and `native` are good in context of the app, but we have to be clear that this is in context of the app and not the coordinator.

docs/testing.md
da-kami: oh sweet, I missed that C4 diagrams include component diagrams! We should switch to mermaid then!

docs/testing.md
holzeis: whats the `composer`? do you mean the `coordinator`?

bonomat:
@da-kami / @luckysori : what's the status of this please?

luckysori:
> @da-kami / @luckysori : what's the status of this please?

I think we can just merge it for reference and move on. We could also create a ticket to tackle what @da-kami already pointed out during our last get-together about this topic:

> We agree that this document does not represent a clear plan of action in its current state. @da-kami volunteered to create a document that more clearly and succinctly specifies our current testing strategy.

da-kami:
> > @da-kami / @luckysori : what's the status of this please?
> 
> I think we can just merge it for reference and move on. We could also create a ticket to tackle what @da-kami already pointed out during our last get-together about this topic:
> 
> > We agree that this document does not represent a clear plan of action in its current state. @da-kami volunteered to create a document that more clearly and succinctly specifies our current testing strategy.

I don't remember volunteering for this, slipped through my brainnet 😅
I'll see to have a go at this tomorrow - have a look at this document and decide if it's worth to adapt it so it's a plan of attack.

luckysori:
> > > @da-kami / @luckysori : what's the status of this please?
> > 
> > 
> > I think we can just merge it for reference and move on. We could also create a ticket to tackle what @da-kami already pointed out during our last get-together about this topic:
> > > We agree that this document does not represent a clear plan of action in its current state. @da-kami volunteered to create a document that more clearly and succinctly specifies our current testing strategy.
> 
> I don't remember volunteering for this, slipped through my brainnet 😅 I'll see to have a go at this tomorrow - have a look at this document and decide if it's worth to adapt it so it's a plan of attack.

You don't need to tackle the new document yourself necessarily! I just wanted to refer back to the observation that we did want to write something a bit more actionable.

luckysori:
We already merged a different, more actionable document. We can close this.

PR #33, opened 2023-01-22
docs: document architecture topics

holzeis:
Cael and I discussed on Friday architecture topics in the context of https://github.com/get10101/10101/issues/17. We looked at it from a very broad angle.

I documented our discussion points and views in this document.

Please look at it as a discussion document, the topics in there will eventually result in ADRs.

Once we have a better understanding of the target architecture we should also spend more time on drawing the architecture logical and deployment view.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,122 @@
+# 10101 Architecture
+
+This is a working document to design the architecture drafted in the below diagram. It will not remain in that structure but rather will be eventually documented in ADRs.
+
+![Assumed component diagram](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/get10101/10101/d7619613324df3ce5c0f0a35d97bf24323f9515b/docs/diagrams/testing-component-diagram.puml)
+
+The following represents a non exhaustive list of topics relevant to be discussed when designing the 10101 architecture and is primarily based on the assumed component diagram introduced in https://github.com/get10101/10101/pull/30. Please also note that the order does not represent a priority of topics.
+
+## Database
+
+Both app and coordinator need to persist data. This section addresses the corresponding database decisions.
+
+### What data needs to be stored?
+
+ - Trading relevant data, like Offers, Orders, etc.
+ - Lightning channel related data.
+ - DLC related data - lightning channel extension data.
+
+Assumption: The app and coordinator hold mostly the same kind of data.
+
+### Should we use the same database / schemas for the app and coordinator?
+
+- **Option 1: Use the *same database***: The data relevant to be stored is structurally very similar for the app and the coordinator. In the poc (as well as in itchysats) we've used the same database abd schemas for both databases (taker and maker).
+  - **Advantage**: Maintaining migrations as well as the data access layer code is simplified.
+
+- **Option 2: Use *separate databases***: While using the same database makes sense from a maintenance point of view it might restrict us in the technology choices since a mobile app and backend service do have completely differnt constraints. It might as well hinder us from optimising our backend service as we have to consider mobile constraints.
+  - **Advantage**: Databases can be optimised according to the platform specific needs.
+
+### What kind of database should be used?
+
+- **Option 1: Use a *relational database***: In the past we have used relational databases, promoting a clear data structure and model. A lot of care has been taken to migrate old records to adhere to data model changes. No data redundancy.
+- **Option 2: Use a *no sql database***: A much more flexible option in terms of data model, with performance prioritised over data integrity. Data redundancy.
+
+
+## Lightning node
+
+Our design principles promote self-custody and self-sovereignty - as a result the users lightning node must run on her phone.
+
+### How should the lightning node run on the phone?
+
+- **Option 1: *Foreground process***: The lightning node will be started only when the app is opened.
+  - **Advantages**: low battery consumption, and simple implementation
+- **Option 2: *Background process***: The lightning node runs as background task on the phone, even if the app is closed.
+  - **Advantages**: payments can be received at any time
+
+Another alternative has been discussed if push notifications could be used to notify the user about an incoming payment asking to accept it by opening the app. We have not addressed the question of privacy.
+
+However, this feature might not be important for the MVP.
+
+### How do we extend the lightning channel for DLCs?
+
+DLCs are not supported by the lightning channels outof the box and a custom extension needs to be implemented. Downwards compatibility should be granted though.
+
+ - **Option 1: *Virtual Channels***: ..
+ - **Option 2: ...**:
+
+## State management
+
+How are state changes managed in 10101.
+
+ - **Option 1: *Events***: State changes are reflected with immutable events allowing to recreate a state by applying all events in the correct order. While this architectural pattern introduces some complexities it allows for a much more flexible, async processing. In itchysats an event driven architecture has been implemented for parts of the solution.
+ - **Option 2: *API***: State changes are promoted through synchronous api calls with a blocking wait to check if the call has been processed successfully. This is a simple paradigm an less complex.  
+
+*Note, these options are not exhaustive nor very general for state management. Also the target architecture could make use of a combination of both approaches. Further elaboration and discussion is required here.* 
+
+## Coordinator API
+
+The coordinator will have to expose an API along the p2p API for the lightning (including dlc) communication e.g. order book. **Assumption**: 10101 will not exclusively talk via the p2p protocol with the app. As DLCs should get integrated into upstream rust-lightning a more generalised approach would be beneficiary.  
+
+- **Option 1 *Restfull API***: Simple API to expose CRUD operations through an HTTP. Broadly used and well understood.
+- **Option 2 *Websocket***: Only one API, exposing all sorts of data the client can subscribe to.
+
+A combination of both options seem to be reasonable as the approach in itchysats with only web sockets seemed artificial and introduced unneeded complexities. However using websockets to push updates to subscribed data objects (instead of polling) seems to be a good practice for multiple reasons (bandwidth, freshness, etc.).
+
+*We also shortly discussed graphql, but probably figured it to be not beneficiary (as too powerful and complex) for our use case.*
+  
+### RPC vs REST: 
+
+It seems RPC is broadly used nowadays. We might want to consider evaluating RPC for our Coordinator APIs.
+
+## Security
+
+One of our most important features is Security. We must not only think of it from a point of view that the protocol is secure. Equally we need to look at the way we derive and manage crypto materials. 
+
+### How to secure the cryptographic key on the phone?
+
+- **Option 1 *HSM***: iPhone and Android support hardware security modules on the phone. For iPhone its Secure Element - We could use such module (isolated from the os) to securily perfom cryptographic algorithms without ever exposing the private key. *Note, its unclear what that would mean for exporting the key if required* 
+- **Option 2 *Encrypt keys on disk***: Simply encrypting the keys on disk e.g. in combination with an entered pin to unlock the phone.
+
+*Note, this again not a exhaustive list on all the options for managing cryptographic keys on the phone. More work is required here.*
+
+### How to authenticate the user 
+
+**Assumption**: The coordinator api does not require any authentication - however we might want to ensure that it is only used from the 10101 app.
+
+The coordinator api should though only be exposed through HTTPS.
+
+## Backup and restore (remember punish transactions etc.)
+
+The user needs to be able to backup not only his keys but also his channel (dlc) data so that she can restore it on another phone. 
+
+- **Option 1 *Closing and reopening all channels and positions***: A costly, but straight forward simplified backup process. However also costly (as on chain transactions are required) and probably not feasible as the market price for a position will shift and might not get matched when reopening.   
+- **Option 2 *Backup to google cloud, icloud, nas backup, SW3 Bucket, Zip File, etc.***: Also quite straight forward, but might produce race conditions (complexities) as the apps state could live on multiple phones.
+
+## On chain monitoring 
+
+- **Option 1 *Electrum***: Not a standard, but broadly used and accepted.
+- **Option 2 *Neutrino (BIP-157)***: A bitcoin standard to run a lightweight node. (used by Breez - full nodes can be replaced (similarly how you would do it with electrum server, but with one component less))
+
+
+## Open Topics:
+
+### Deployment View
+- Reverse proxy for ssl offloading
+- Containerized
+- VM (docker-compose, podman) + cloud conf, Kubernetes, AWS Container Services
+- what envrionments?
+  - Test
+    - signet node (ourselves, electrum server (depends on neutrino?), blockchain explorer (esplora)?)
+  - Prod
+    - full node? (electrum server (depends on neutrino), blockchain explorer (esplora)?)
+


docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Lightning node
+
+Our design principles promote self-custody and self-sovereignty - as a result the users lightning node must run on her phone.
+
+### How should the lightning node run on the phone?
+
+- **Option 1: *Foreground process***: The lightning node will be started only when the app is opened.
+  - **Advantages**: low battery consumption, and simple implementation
```
bonomat, line 42:
```
 - **Disadvantage**: if the app was closed for a long time, starting it and catching-up with the blockchain will take a while.
````

File: docs/architecture.md [hunk 2]
```diff
@@ -0,0 +1,122 @@
...
+## State management
+
+How are state changes managed in 10101.
+
+ - **Option 1: *Events***: State changes are reflected with immutable events allowing to recreate a state by applying all events in the correct order. While this architectural pattern introduces some complexities it allows for a much more flexible, async processing. In itchysats an event driven architecture has been implemented for parts of the solution.
+ - **Option 2: *API***: State changes are promoted through synchronous api calls with a blocking wait to check if the call has been processed successfully. This is a simple paradigm an less complex.  
+
+*Note, these options are not exhaustive nor very general for state management. Also the target architecture could make use of a combination of both approaches. Further elaboration and discussion is required here.* 
+
+## Coordinator API
```
bonomat, line 66: I would make a difference between the orderbook and the coordinator.

I'm aiming for a separation of concern from a legal perspective in the long run, meaning, from a technical perspective I wold like to seperate these two things from the getgo to simplify a legal separation later on.

For example, _a way_ how this could play out: 
- orderbook contains orders
- taker grabs an order which includes information of the counter party
- taker _tells_ the coordinator he wants to take this order
- coordinator opens positions with both parties simultaneously.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+It seems RPC is broadly used nowadays. We might want to consider evaluating RPC for our Coordinator APIs.
+
+## Security
+
+One of our most important features is Security. We must not only think of it from a point of view that the protocol is secure. Equally we need to look at the way we derive and manage crypto materials. 
+
+### How to secure the cryptographic key on the phone?
+
+- **Option 1 *HSM***: iPhone and Android support hardware security modules on the phone. For iPhone its Secure Element - We could use such module (isolated from the os) to securily perfom cryptographic algorithms without ever exposing the private key. *Note, its unclear what that would mean for exporting the key if required*
```
bonomat, line 87: Interesting idea. Is this then portable? E.g. can you create a back-up to recover your funds in case you lose your phone?

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+**Assumption**: The coordinator api does not require any authentication - however we might want to ensure that it is only used from the 10101 app.
+
+The coordinator api should though only be exposed through HTTPS.
+
+## Backup and restore (remember punish transactions etc.)
+
+The user needs to be able to backup not only his keys but also his channel (dlc) data so that she can restore it on another phone. 
+
+- **Option 1 *Closing and reopening all channels and positions***: A costly, but straight forward simplified backup process. However also costly (as on chain transactions are required) and probably not feasible as the market price for a position will shift and might not get matched when reopening.   
+- **Option 2 *Backup to google cloud, icloud, nas backup, SW3 Bucket, Zip File, etc.***: Also quite straight forward, but might produce race conditions (complexities) as the apps state could live on multiple phones.
```
bonomat, line 103: Most platforms explicitly say that using the same backup on different devies at the same time is not supported and might result in a loss of funds.

bonomat (approved):
Great write-up and a good starting point for discussions. I've added some thoughts first and will then add my opinion to each point.

docs/architecture.md
luckysori, line 66: As far as I can remember, in ItchySats the maker:

- Communicated with the taker via `libp2p`.
- Exposed a HTTP API and a WebSocket API so that a separate component could automate its behaviour.

This section seems to be conflating both aspects, which I found a bit confusing. It might be good to treat them separately.

da-kami (commented):
I assume that we are discussing the long term vision for the architecture and will derive what to do for the MVP in a separate step.
In the best case the long term vision can be applied directly to the MVP - but I'm not sure we want to do this for all decisions.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+How are state changes managed in 10101.
+
+ - **Option 1: *Events***: State changes are reflected with immutable events allowing to recreate a state by applying all events in the correct order. While this architectural pattern introduces some complexities it allows for a much more flexible, async processing. In itchysats an event driven architecture has been implemented for parts of the solution.
+ - **Option 2: *API***: State changes are promoted through synchronous api calls with a blocking wait to check if the call has been processed successfully. This is a simple paradigm an less complex.  
+
+*Note, these options are not exhaustive nor very general for state management. Also the target architecture could make use of a combination of both approaches. Further elaboration and discussion is required here.* 
+
+## Coordinator API
+
+The coordinator will have to expose an API along the p2p API for the lightning (including dlc) communication e.g. order book. **Assumption**: 10101 will not exclusively talk via the p2p protocol with the app. As DLCs should get integrated into upstream rust-lightning a more generalised approach would be beneficiary.
```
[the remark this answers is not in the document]
bonomat: > In terms of the assumption, I believe that the intention is to reuse the p2p layer provided by the Lightning node to exchange DLC-specific data. Is that correct, @bonomat?

Imho this correct: rust-dlc was built _on top of_ the lightning p2p layer.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
+# 10101 Architecture
+
+This is a working document to design the architecture drafted in the below diagram. It will not remain in that structure but rather will be eventually documented in ADRs.
+
+![Assumed component diagram](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/get10101/10101/d7619613324df3ce5c0f0a35d97bf24323f9515b/docs/diagrams/testing-component-diagram.puml)
```
da-kami, line 5: Please consider changing this diagram to use mermaid's C4 diagrams - https://mermaid.js.org/syntax/c4c.html - as discussed here:
https://github.com/get10101/10101/pull/30/files#r1082034689

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+The following represents a non exhaustive list of topics relevant to be discussed when designing the 10101 architecture and is primarily based on the assumed component diagram introduced in https://github.com/get10101/10101/pull/30. Please also note that the order does not represent a priority of topics.
+
+## Database
+
+Both app and coordinator need to persist data. This section addresses the corresponding database decisions.
+
+### What data needs to be stored?
+
+ - Trading relevant data, like Offers, Orders, etc.
```
1010Tom, line 15: I believe the term 'offer' will be phased out in favor of 'limited order' and 'market order'.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+- **Option 1: Use the *same database***: The data relevant to be stored is structurally very similar for the app and the coordinator. In the poc (as well as in itchysats) we've used the same database abd schemas for both databases (taker and maker).
+  - **Advantage**: Maintaining migrations as well as the data access layer code is simplified.
+
+- **Option 2: Use *separate databases***: While using the same database makes sense from a maintenance point of view it might restrict us in the technology choices since a mobile app and backend service do have completely differnt constraints. It might as well hinder us from optimising our backend service as we have to consider mobile constraints.
+  - **Advantage**: Databases can be optimised according to the platform specific needs.
+
+### What kind of database should be used?
+
+- **Option 1: Use a *relational database***: In the past we have used relational databases, promoting a clear data structure and model. A lot of care has been taken to migrate old records to adhere to data model changes. No data redundancy.
```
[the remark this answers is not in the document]
1010Tom: I'm in favor of using `Diesel` as it is more highly praised than `sqlx` in regards to performance and performance will be our limiting factor.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+### How should the lightning node run on the phone?
+
+- **Option 1: *Foreground process***: The lightning node will be started only when the app is opened.
+  - **Advantages**: low battery consumption, and simple implementation
+- **Option 2: *Background process***: The lightning node runs as background task on the phone, even if the app is closed.
+  - **Advantages**: payments can be received at any time
+
+Another alternative has been discussed if push notifications could be used to notify the user about an incoming payment asking to accept it by opening the app. We have not addressed the question of privacy.
+
+However, this feature might not be important for the MVP.
```
1010Tom, line 48: Agreed, this feature is not needed for the MVP and an awesome feature to announce later on :)

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+### How do we extend the lightning channel for DLCs?
+
+DLCs are not supported by the lightning channels outof the box and a custom extension needs to be implemented. Downwards compatibility should be granted though.
+
+ - **Option 1: *Virtual Channels***: ..
+ - **Option 2: ...**:
```
1010Tom, line 55:
```suggestion
 - **Option 2: Extend commit transaction**: see https://10101.substack.com/p/noncustodial-trading-with-10101
```

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+
+Another alternative has been discussed if push notifications could be used to notify the user about an incoming payment asking to accept it by opening the app. We have not addressed the question of privacy.
+
+However, this feature might not be important for the MVP.
+
+### How do we extend the lightning channel for DLCs?
+
+DLCs are not supported by the lightning channels outof the box and a custom extension needs to be implemented. Downwards compatibility should be granted though.
+
+ - **Option 1: *Virtual Channels***: ..
```
1010Tom, line 54:
```suggestion
 - **Option 1: *Virtual Channels***:  see https://medium.com/crypto-garage/dlc-on-lightning-cb5d191f6e64
```

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+ - DLC related data - lightning channel extension data.
+
+Assumption: The app and coordinator hold mostly the same kind of data.
+
+### Should we use the same database / schemas for the app and coordinator?
+
+- **Option 1: Use the *same database***: The data relevant to be stored is structurally very similar for the app and the coordinator. In the poc (as well as in itchysats) we've used the same database abd schemas for both databases (taker and maker).
+  - **Advantage**: Maintaining migrations as well as the data access layer code is simplified.
+
+- **Option 2: Use *separate databases***: While using the same database makes sense from a maintenance point of view it might restrict us in the technology choices since a mobile app and backend service do have completely differnt constraints. It might as well hinder us from optimising our backend service as we have to consider mobile constraints.
```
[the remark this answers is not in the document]
holzeis: it will also mean that we have to implement two adaptors. Using the same schemas makes definitely sense, however we will not be able to use the exact same data access layer in case we mix databases.

That being said, I am also in favor of separate databases, as the requirements are completely different and we don't want to be limited on what could run on the phone.

That benefit will though come with more operational and maintenance costs / efforts.

docs/architecture.md
holzeis: @bonomat I guess we never experienced that issue since we tested against regtest? thats actually a good reason to use testnet (also signet will probably not help there)

docs/architecture.md
holzeis: We can build the orderbook api as a separate module, but I would not build it as a separate service from the "get-go". because coordinating multiple micro services is hard, just think of testing, deployment (CI/CD), etc. If we stick to a proper modularization of the components within one service, we can still easily separate them afterwards if we really need to.

docs/architecture.md
holzeis: > *Note, its unclear what that would mean for exporting the key if required* 

I haven't looked into that yet. But I think it's worthwhile to look into that in more detail.

docs/architecture.md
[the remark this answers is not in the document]
holzeis: I agree on the arguments for a relational database!

docs/architecture.md
holzeis: **silly thought**: we could also misuse keysend to exchange custom messages.

While this is definitely a (costly) misuse of the feature, it would mean we stay compatible with all lightning implementations.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+One of our most important features is Security. We must not only think of it from a point of view that the protocol is secure. Equally we need to look at the way we derive and manage crypto materials. 
+
+### How to secure the cryptographic key on the phone?
+
+- **Option 1 *HSM***: iPhone and Android support hardware security modules on the phone. For iPhone its Secure Element - We could use such module (isolated from the os) to securily perfom cryptographic algorithms without ever exposing the private key. *Note, its unclear what that would mean for exporting the key if required* 
+- **Option 2 *Encrypt keys on disk***: Simply encrypting the keys on disk e.g. in combination with an entered pin to unlock the phone.
+
+*Note, this again not a exhaustive list on all the options for managing cryptographic keys on the phone. More work is required here.*
+
+### How to authenticate the user
```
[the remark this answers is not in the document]
holzeis: its a separate question in my point of view, both are valid.

imho, the lightning node identity (or more specifically its public key / node id) should be enough. I wouldn't bind it to the connection though, as we have done it in itchysats.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+## Lightning node
+
+Our design principles promote self-custody and self-sovereignty - as a result the users lightning node must run on her phone.
+
+### How should the lightning node run on the phone?
+
+- **Option 1: *Foreground process***: The lightning node will be started only when the app is opened.
+  - **Advantages**: low battery consumption, and simple implementation
+- **Option 2: *Background process***: The lightning node runs as background task on the phone, even if the app is closed.
+  - **Advantages**: payments can be received at any time
```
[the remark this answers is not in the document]
holzeis: A quick google indicates that it should work (https://developer.apple.com/documentation/backgroundtasks) however I haven't read much more than the headers 😜

docs/architecture.md
1010Tom: It also wasn't so significant because we used an electrum-client which goes against an already synced node.

docs/architecture.md
bonomat: I actually think it is easier because we can evolve and deploy it independently.

Restioson: Small note: if we are going to build toward a microservice based architecture it may be worth considering some kind of distributed tracing setup again, though we may only want to track very high level spans like REST calls from service to service to reduce the amount of data (an issue we had in itchysats)

docs/architecture.md
Restioson: https://www.thecodeside.com/2020/09/14/android-auto-backup-keystore-encryption-broken-heart-love-story/

This article seems to suggest that it is not possible to back up data encrypted using a key stored in an Android HSM

docs/architecture.md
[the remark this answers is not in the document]
Restioson: As far as diesel vs sqlx goes, it should be noted that diesel is sync whereas sqlx is async (correct me if I am wrong). However, in the conversation @holzeis and I had, the database being async was considered a pain point as opposed to a useful feature.

docs/architecture.md
```diff
@@ -0,0 +1,122 @@
...
+However, this feature might not be important for the MVP.
+
+### How do we extend the lightning channel for DLCs?
+
+DLCs are not supported by the lightning channels outof the box and a custom extension needs to be implemented. Downwards compatibility should be granted though.
+
+ - **Option 1: *Virtual Channels***: ..
+ - **Option 2: ...**:
+
+## State management
```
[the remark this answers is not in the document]
Restioson: > In the business logic we have to make sense of the data and build up the current state of the application; usually we represent this in some form of state machines.

Would this mean that instead of having tables like closed_cfds vs the normal cfds tables a la itchysats we would have just one table and the data on the cfd (open/closed flag) would represent whether it's open or closed? To use a different example, would we be treating the data more how the check_cfds project did it than how itchysats did?

docs/architecture.md
Restioson: In that case is the term "offer" a generalisation of "limited order" and "market order"?

docs/architecture.md
holzeis: so, is this a disadvantage then?

docs/architecture.md
holzeis: > Small note: if we are going to build toward a microservice based architecture it may be worth considering some kind of distributed tracing setup again, though we may only want to track very high level spans like REST calls from service to service to reduce the amount of data (an issue we had in itchysats)

That's exactly the kind of complexity i fear if we start building the coordinator and orderbook as separate micro services. We also need to consider retry mechanisms, distributed transactions, etc.

docs/architecture.md
1010Tom: Yea, maybe it's not worth mentioning.

docs/architecture.md
1010Tom: > We also need to consider retry mechanisms, distributed transactions, etc.

Eventually yes, but we don't have to solve all problems in distributed computing at day 0.

docs/architecture.md
1010Tom: Then let's rule it out for now.

docs/architecture.md
1010Tom: No need to misuse keysend here. One can send custom messages over the p2p protocol.

docs/architecture.md
1010Tom: Mhmm... I would not make this generalisation. Offer is more a limited order from the point of view of the person who created the limited order but reserves the right to cancel it anytime.

bonomat:
@holzeis : what's the status of this please?

PR #75, opened 2023-01-31
Add `ln-dlc-node` crate

### LINKED ISSUE #45: Create `ln-dlc-node` sub-crate
The purpose of this library crate is to provide the API that an application such as the 10101 app would want to use to be able to implement the features described in #34.

Even though it is a library, we need to remember that at this stage we only expect two consumers: app and coordinator. Thus, let's avoid premature generalisations.

### Observations

- Eventually we should be able to move most of the code we write to `p2pderivates/rust-dlc`, but that requires more coordination than what we can afford right now.
- Depends on our forks of `rust-dlc` and `rust-lightning`.

luckysori:
Fixes #45.


holzeis (approved):
Looks good! thank you for getting this started.

Only odd thing though is that you had to update the `pubspec.lock`. After investigating offline, we should also be careful of not to update our local flutter version as this may have an impact on the `pubspec.lock`.

However, I agree of upgrading flutter to the latest version being `3.7.0`.

PR #79, opened 2023-01-31
Import correct generated files

Restioson:
The project would not build without importing from the `dart_generated` directory for me. Does it work as-is for anyone else or was this failure unique to me?


mobile/pubspec.lock
```diff
@@ -129,16 +146,18 @@ packages:
...
   flutter_rust_bridge:
     dependency: "direct main"
     description:
       name: flutter_rust_bridge
-      url: "https://pub.dartlang.org"
+      sha256: "5aea0f3980dcd314f1890ef0d2392263817899cc15e543734b5d4dbe66b761eb"
+      url: "https://pub.dev"
     source: hosted
-    version: "1.61.1"
+    version: "1.62.0"
```
holzeis, line 160: 🤔 It's strange that this is shown as diff. The main branch is already locked on 1.62.0.

https://github.com/get10101/10101/blob/16cf62e9bc073fdb78f7521c174d603c8437e33d/mobile/pubspec.lock#L153-L160

bonomat (approved):
Thanks for fixing this. The problem came from moving the generated code into `bridge_generated` which was only added later-on :/

bonomat:
I go ahead and merge this. 

bors r+

PR #87, opened 2023-01-31
docs: Add 003 onchain monitoring ADR

### LINKED ISSUE #81: Document on chain monitoring ADR

holzeis:
resolves #81

### DIFF
--- a/docs/003-onchain-monitoring.adoc
+++ b/docs/003-onchain-monitoring.adoc
@@ -1,0 +1,29 @@
+= ADR 003 - Onchain monitoring
+Richard Holzeis <richard@10101.finance>
+3.0, July 29, 2022: AsciiDoc article template
+:toc:
+:icons: font
+:attributes: 2023-01-31
+
+This document outlines the decision how onchain transactions are monitored.
+
+== Decision
+
+10101 will use a public electrum server to monitor onchain transactions. Electrum is broadly accepted and used. We will use public infrastructure as the trust model will have no impact to the user whether or not 10101 would run the infrastructure components or delegates that task to public infrastructure. A corresponding setting to the user to choose her infrastructural components will be added within the app to allow the user to be fully self-sovereign.
+
+== Context
+
+We need a way to monitor on chain transaction to firstly reflect the wallet balances correctly and secondly ensure that off-chain transactions have not been published malicously. 
+
+=== What standard should be used to monitor for onchain transactions
+
+- *Option 1 _Electrum_*: Not a standard, but broadly used and accepted. Indicies the UTXO set and exposes a simple API to be consumed by the app.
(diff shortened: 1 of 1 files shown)


docs/003-onchain-monitoring.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+== Context
+
+We need a way to monitor on chain transaction to firstly reflect the wallet balances correctly and secondly ensure that off-chain transactions have not been published malicously. 
+
+=== What standard should be used to monitor for onchain transactions
+
+- *Option 1 _Electrum_*: Not a standard, but broadly used and accepted. Indicies the UTXO set and exposes a simple API to be consumed by the app.
+- *Option 2 _Neutrino (BIP-157)_*: A bitcoin standard to run a lightweight node, connecting directly to a full node.
+
```
bonomat, line 22: - *Option 3 _Blockexplorer client_*: Not a standard, but the goto solution for most users to check for transactions is to go to mempool.space or blockchain.info. We could write a client going against these services.

bonomat (approved):
Great summary. 👍

docs/003-onchain-monitoring.adoc
da-kami: I suppose by _Blockexplorer client_ you mean something like mempool.space? 
We could even mention them by name as example referencing their API, and mentioning that they offer to run the explorer self-hosted for scaling up. 

This  brings us back to the discussion how much we want to "rely" on public infrastructure. I think in the beginning it is fine, but with scaling up we need to plan for being able to provide our own infrastructure in case the free, public capacity is exhausted. We might want to mention this and take it into account for decision making down the road :)

docs/003-onchain-monitoring.adoc
```diff
@@ -0,0 +1,28 @@
...
+3.0, July 29, 2022: AsciiDoc article template
+:toc:
+:icons: font
+:attributes: 2023-01-31
+
+This document outlines the decision how onchain transactions are monitored.
+
+== Decision
+
+10101 will use a public electrum server to monitor onchain transactions. Electrum is broadly accepted and used. We will use public infrastructure as the trust model will have no impact to the user whether or not 10101 would run the infrastructure components or delegates that task to public infrastructure. A corresponding setting to the user to choose her infrastructural components could be added within the app.
```
da-kami, line 12:
```suggestion
10101 will use a public electrum server to monitor onchain transactions. Electrum is broadly accepted and used. We will use public infrastructure as the trust model will have no impact to the user whether or not 10101 would run the infrastructure components or delegates that task to public infrastructure. A corresponding setting to the user to choose her infrastructural components will be added within the app to allow the user to be fully self-sovereign.
```

I would be explicit here and I would add this from the get-go.
In the long run it might become less important, but for our initial target users I think it is, and it should be a low-hanging fruit!

da-kami (approved):

PR #89, opened 2023-02-01
Support web ffi

da-kami:
In order to properly support the `ffi_web.dart` file that is part of the project we have to generate the bridge code accordingly.

This resolves dart analyser errors in the project.

---

alternatively we can also remove the `ffi_web.dart` file if we don't want to support web target.

### DIFF
--- a/justfile
+++ b/justfile
@@ -32,7 +32,8 @@ gen:
         --rust-output native/src/bridge_generated/bridge_generated.rs \
         --dart-output lib/bridge_generated/bridge_generated.dart \
         --dart-decl-output lib/bridge_generated/bridge_definitions.dart \
-        --dart-format-line-length {{line_length}}
+        --dart-format-line-length {{line_length}} \
+        --wasm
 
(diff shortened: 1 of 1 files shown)


luckysori:
> alternatively we can also remove the ffi_web.dart file if we don't want to support web target.

Should we just do this then? I think we don't mean to support the web atm.

justfile
```diff
@@ -32,7 +32,8 @@ gen:
         --rust-output native/src/bridge_generated/bridge_generated.rs \
         --dart-output lib/bridge_generated/bridge_generated.dart \
         --dart-decl-output lib/bridge_generated/bridge_definitions.dart \
-        --dart-format-line-length {{line_length}}
+        --dart-format-line-length {{line_length}} \
+        --wasm
```
bonomat, line 36: This was removed because we do not wan to build for web.

da-kami:
superseded by https://github.com/get10101/10101/pull/93

PR #90, opened 2023-02-01
Cocoapods update to overcome error when running the app

da-kami:
I ran into this error when trying to run the application:

```
The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.
```

I overcame this error by running `pod install` in the `macos` and `ios` folders. The files were updated with this content.

### DIFF
--- a/mobile/ios/Runner.xcodeproj/project.pbxproj
+++ b/mobile/ios/Runner.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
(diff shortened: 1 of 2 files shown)


luckysori (approved):
I just want to point out that it does build for Linux without these changes. That might be obvious but at least it should mean that we don't need to make any other changes to make it work for that (development) platform.

bonomat (approved):
Interestingly it was working for me and after I run your PR I get changes proposed in these two files but it keeps working. 

```
        modified:   mobile/macos/Podfile
        modified:   mobile/macos/Runner.xcodeproj/project.pbxproj
```

luckysori:
> Interestingly it was working for me and after I run your PR I get changes proposed in these two files but it keeps working.
> 
> ```
>         modified:   mobile/macos/Podfile
>         modified:   mobile/macos/Runner.xcodeproj/project.pbxproj
> ```

That's a bit more concerning. Could it be because you're running different versions of something?

da-kami:
Probably better to close this PR - These changes happened before I upgraded to the latest Flutter version and I cannot reproduce them anymore.
Thing is, that when running the `pod install` command it did pull two casks, but if nobody else runs into the error it was probably just an outdated Flutter version that caused it.

PR #93, opened 2023-02-01
delete: ffi_web target

da-kami:
dart analyzer keeps complaining about the `ffi_web.dart` file so we should remove it if we don't want to support the web target.


1010Tom (approved):
Thanks for fixing this.

PR #94, opened 2023-02-01
update: podfile

da-kami:
With the latest flutter version `stable, 3.7.0, on macOS 13.1 22C65 darwin-x64, locale en-AU` and the latest Cocoapod version `1.11.3` I run into these changes.

--- 

@bonomat could you run `just deps` and `just run` from `main` and see if you get the same? 

I am not sure how relevant these changes are, but eventually it will happen that they slip into a PR.

### DIFF
--- a/mobile/macos/Podfile
+++ b/mobile/macos/Podfile
@@ -1,4 +1,4 @@
-platform :osx, '10.11'
+platform :osx, '10.14'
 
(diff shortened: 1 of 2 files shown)


bonomat (approved):
Can confirm, I'm getting the same changes.

PR #95, opened 2023-02-02
docs: Add 005 mobile app architecture ADR

da-kami:
This is an attempt to add some more structure and agree on terminology within the mobile project. 

This ADR is inspired by:

* https://codewithandrea.com/articles/flutter-project-structure/
* https://github.com/get10101/10101-PoC/discussions/122

Note that I did not try to answer *all* questions and decide all things, but try to give us a structure to start with and then see how we go. The structure should not be set in stone but should be a guideline. We should update it from our learnings.


docs/005-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,98 @@
...
+* `domain` - domain model files
+* `data` - storage related files; we will likely not need this in Dart
+
+Initially it is expected that almost all of our files except from commonly share widgets / models will be in the `presentation` folder, so I would not introduce this folder structure, but rather add all files to `wallet`.
+
+We can opt to further structure a features subfolder into sub-feature folders (e.g. `send` and `receive` for the `wallet` folder), but I would do that once the application grows.
+
+Recommended naming conventions for files:
+
+- `[...]_route`: composition of widgets that represents a route within the application
```
da-kami, line 84: This is somewhat controversial - most projects still use `screen` here. 
Unfortunately `screen` is ambiguous - but `route` might be somewhat misleading / out of context here, because in the end it's still a composition of widgets. 
If you find `screen` more intuitive we can also use that.

da-kami:
After working on this further I feel this still needs some adjustments, to I am downgrading it to draft and plan to open it again with the initial navigation ramped up.

docs/006-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,109 @@
...
+General rule of thumb:
+
+* If it is structured data / relational data it should be in the database.
+* If it is a simple flag for decision making in Dart it can be stored in the `shared_preferences`.
+
+=== Data Model
+
+The data model is mostly managed in the rust code.
+It is recommended to separate the domain model used in the rust code from the model exposed to the Dart code.
+Very simple, basic domain objects can be shared between the BL and the API layer.
```
1010Tom: What does `BL` stand for?

docs/006-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,109 @@
...
+** features
+*** wallet
+*** trade
+
+This image shows the boundaries of the Flutter layers:
+
+.Flutter Layer Overview (https://codewithandrea.com/articles/flutter-project-structure/[source])
+image::assets/006-mobile-application-architecture/flutter-layers.png[alt=FlutterLayers,width=400,height=480]]
+
+Within the `wallet` and `trade` features we can introduce subfolders:
```
1010Tom: The linked article is really good. Thanks for sharing. One thing comes to mind is the question of [`what is a feature`](https://codewithandrea.com/articles/flutter-project-structure/#what-is-a-feature?). Wallet doesn't sound like feature to me but a group of feature. Maybe that's good but maybe we will need to slice wallet/trade into multiple features. But that's something we can do later-on as well and don't have to document here :) 

Thoughts @da-kami ?

1010Tom (approved):
Thanks for the write-up. 
Using `feature first` structure sounds good to me 👍

docs/006-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,109 @@
...
+* Route / Screen / [.line-through]#Page#: Flutter recommends to use https://docs.flutter.dev/cookbook/navigation/navigation-basics[Routes] when referring to a screen or page. However, "route" is not an intuitive term to refer to a user interface.
+** When we are talking about the app it's OK to use screen. We define "screen" as a composition of widgets that are presented to the user.
+*** There is some ambiguity if e.g. the top/bottom navigation is part of a screen or not. From a usability perspective the navigation is likely part of the screen; from a composition perspective it might not be.
+** The route *identifies* a screen. Each screen should have a route, i.e. if we name a file `[...]_screeen` we expect it to be routable.
+** There is no clear definition that distinguishes pages from screens. I would recommend not to use page because it is yet another term that adds confusion.
+* Service: A service is on the application level. Given our setup with a native rust backend we will likely not need services in Dart code.
```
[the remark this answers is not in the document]
da-kami: That's a great question - I think my statement that we won't have services is wrong. 

One thing that makes our native rust "backend" different is that there is code generation involved, the consumer side dart code is generated including domain objects.
But we should still treat the rust code as backend to some extent; the fact that code is generated by flutter rust bridge is an implementation detail. From the Dart perspective we are fetching data from the outside.

I will reword this section to make it clear that we should also use `[...]_service` for files that manage fetching / processing data from the native rust code.

docs/006-mobile-app-architecture.adoc
da-kami: I totally agree but after some thought last week I don't have a perfect split in mind. 

We could go more with *actions* such as `send`, `receive`, but I find it hard to find the right abstraction. Would the transaction history be a separate feature? What about the scanner? 

The wallet itself definitely is a feature; there will be some sort of wallet screen.
I came to the conclusion that it would be easier to have `wallet` and `trade` and then define *sub*-features. The article mentions this:

> When building Flutter apps, it's very common to have a ratio of 5:1 (or more) between UI code and business logic. If your presentation folder ends up with many files, don't be afraid to group them into sub-folders that represent smaller "sub-features".

I am not sure that is the best approach - maybe it would be better to flatten the the feature hierarchy out.

That said: I would not overthink it and roll with `wallet` and `trade` for now - that's what we ended up having in the POC. I think `wallet` and `trade` is not wrong - but might be too broad. But I think a refactoring round where we split this into smaller features can be done later.

If you want we can talk about this - I am happy to change it if we can come up with a better split now already!

mobile/pubspec.yaml
```diff
@@ -36,6 +36,7 @@ dependencies:
   ffi: ^2.0.1
   flutter_rust_bridge: ^1.61.1
   meta: ^1.8.0
+  go_router: ^6.0.2
```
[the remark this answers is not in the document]
da-kami: that must have slipped in by accident, I'll remove it

docs/006-mobile-app-architecture.adoc
da-kami: business logic - I can remove the abbreviation

docs/006-mobile-app-architecture.adoc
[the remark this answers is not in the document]
da-kami: I added a fixup that amends the section.

luckysori (commented):
I appreciate the effort that went into investigating and documenting these topics. I just don't have much to contribute at this stage. Thanks for putting this together!

da-kami:
> I appreciate the effort that went into investigating and documenting these topics. I just don't have much to contribute at this stage. Thanks for putting this together!

That's all good, the PR is currently waiting for an update because some constraints change slightly when we implemented. I'll convert it into a draft and will see to update this in the next week and get it in :)

da-kami:
Now that @klochowicz is joining I have a reason to finally give this a quick update ;)

docs/005-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,153 @@
...
+Initially it is expected that almost all of our files except from commonly share widgets / models will be in the `presentation` folder, so I would not introduce this folder structure, but rather add all files to the `wallet` folder.
+
+We can opt to further structure a features subfolder into sub-feature folders (e.g. `send` and `receive` for the `wallet` folder), but I would do that once the application grows.
+
+It is recommended to name widgets semantically by context (feature, task) and a logical suffix that is associated with the composition of the widget.
+Examples:
+
+- `[...]_sceen`: composition of widgets that represents a screen (identified by a route)
+    - Examples: `wallet_screen`, `trade_screen`, `scanner_screen`
+    - There is not clear guideline for screens that are part of a wizard yet, I would recommend numbering hem (e.g. `send_screen_1`)
```
holzeis, line 116: How about creating a sub folder e.g. funding wizard. All screens within that folder belong to that wizard.

docs/005-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,153 @@
...
+This document contains the reasoning for high level architecture and project structure decisions for the mobile app.
+
+== Decision
+
+* Application State and data storage are handled in the mobile app's native rust code.
+** Whenever data is required the app should ask the native backend for data.
+** Ephemeral (widget) state should be kept on the presentation layer.
+* The Dart presentation layer is encapsulated as presentation layer
+** We have a Dart and a rust model and map between them
+** No rust ffi code should be use in the presetnation layer, i.e. no ffi includes in `ChangeNotifier`s and `Widget`s.
```
holzeis, line 17: Typo: presentation

holzeis (approved):
Thanks for putting this together! 👍

docs/005-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,153 @@
...
+** Ffi bindings should be encapsulated in `Service`s in Dart
```
klochowicz, line 18: nit: consider using `FFI`  (here and other spots) - it's an acronym, just like API and other acronyms that are uppercase in this doco

klochowicz (approved):
A great step forward; can confirm, having such docs is invaluable when someone joins the project! ☺️
 Let's get this in, we can always update it if our understanding evolves.

docs/005-mobile-app-architecture.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+== Decision
+
+* Application State and data storage are handled in the mobile app's native rust code.
+** Whenever data is required the app should ask the native backend for data.
+** Widget state should be kept on the presentation layer.
+* The domain model is mostly defined in rust.
+** It is recommended to keep the internal domain model used in the business logic encapsulated in the native code and expose a separate domain model on the API.
+** There must be a separate DTO model for the app's database.
+** We will likely still need some models in Dart, but they should be added for presentation and presentation controller purposes only. Examples are: a model to cache data until it is complete and can be handed over to the API to be stored, or a model that aggregates API data for display purposes.
```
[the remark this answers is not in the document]
da-kami: We are not inventing a pattern, but at the same time we are not following one pattern *specifically*. Those patterns are defining how to manage application state; which is not clearly defined in Flutter - there are [multiple approaches](https://docs.flutter.dev/development/data-and-backend/state-mgmt/options) and they don't all map to a specific pattern 1:1.

I think our approach is closer to MVC than to MVP (the controller notifies the view; if we regard the `ChangeModifier` as controller); but it's not *pure* MVC.
I have never taken a closer look at MVVM; and don't have the time to do so now. If you see a clear pattern in our approach feel free to update the ADR.

Note: For MVC there are libraries for flutter, that would help enforce the MVC pattern. I looked into them briefly in the beginning, but then decided to go with [`Provider`](https://pub.dev/packages/provider) and the recommendations around that. I did not see them map to one specific architecture pattern.

Issue #62, opened 2023-01-31
Create main navigation


1010Tom:
In order to be able to distribute work this ticket sets out to create the main navigation flow and empty pages according to our figma mocks. 
This ticket focuses on the navigation between pages and not the page content. We should not focus on page details, but only the main elements needed for navigation should be added and then the actual navigation - so we can click through the main pages of the app without having content there. Once we have the main pages we can fill them with content quite independently.

Scope of this ticket: 

- [x] Decide which routing library to use, see: https://github.com/get10101/10101/pull/91
- [x] Add the initial page (loading page) with the 10101 logo (can be placeholder)
- [x] Add the bottom navigation and the the two main pages (Wallet and Trade)
- [x] Add the top bar to navigate to Settings and Scanner in wallet, Settings in Trade
- [x] Placeholder for the Settings and Scanner page
- [x] Buy and Sell buttons that open an (empty) overlay
- [x] Add FAB (floating action button) to the Wallet page that navigates to a child-screen representing the funding flow (i.e. receive funds flow). This is to showcase the relationship between the parent and child screen in the app and how we are planning to handle it in the navigation (router)
    - Note: The FAB does not have to be an expandable FAB yet, we can do that later.

Feel free to re-use code from the POC: https://github.com/get10101/10101-poc

---

Below are screenshots of the main screens in the Figma prototype. The final prototype can be found [here](https://www.figma.com/proto/Sr3SfZ46sZstKjSwftmTtc/10101-v1-(incl-bitcoin.design)?page-id=4254%3A62685&node-id=5042%3A80076&viewport=2018%2C1608%2C0.14&scaling=scale-down&starting-point-node-id=5042%3A80076&show-proto-sidebar=1).

Initial pages:

![image](https://user-images.githubusercontent.com/5557790/215983674-eb7d3b89-e2bb-48d9-a377-96af195cee6a.png)

Bottom Tab navigation and main Wallet and Trade screen - top navigation to settings and scanner:

![image](https://user-images.githubusercontent.com/5557790/215983306-80e0b4c4-0b8a-4afa-b205-63717a289f9d.png)

![image](https://user-images.githubusercontent.com/5557790/215986762-f415f1aa-0d30-4566-9cd4-652bd4f991ac.png)

Note: We can initially model the trading screen in a state where the `Buy` and `Sell` button is available and later change the state to only showing these buttons once the wallet was funded.

da-kami:
[...]

PR #96, opened 2023-02-02
Chore/62 create main navigation

da-kami:
resolves https://github.com/get10101/10101/issues/62

Please note that, as stated in #62 some components are just placeholders and will be replaced with actual content. Some of the screen will eventually be multiple screens (e.g. send / receive). Some of the components will be more complex than currently depicted (e.g. but/sell overlay that needs a component where one can also switch context once the overlay is open, ...). We can also opt to re-style some of the components.

Note that the folder structure and file naming conventions are summed up in https://github.com/get10101/10101/pull/95
If you want to dispute that let's please discuss it on the ADR :)

When reviewing please don't focus on small details in the UI, and keep the above in mind. 
Please focus on the overall concept (screen navigation, ...) - I'm also happy about code-style related improvement comments :)

Some noteworthy decisions:

- I opted to use a `ShellRoute` for our main screens, wrapping `Wallet` and `Trade` screen in a shell (The bottom and top navigation act as the shell). 
    - This allows us to display child screens within the shell - or display them using the `rootNavigator`, which displays them overlaying the shell. 
    - So far I configured all child screens on the `rootNavigator` level. The shell is still useful for wrapping the main screens; and we could consider having child screens that are wrapped within the shell. To be honest I did not really consider this pattern during the design phase - but I would not waste too much time on it now. 

- I invested some time into understanding how to *properly* add icons; the recommended way is to convert the icons to fonts. This reduces the size of the bundle because of great compression, and it is neat to use. It requires some extra work to set up initially though. I used https://www.fluttericon.com/ and [this guide](https://github.com/fontello/fontello/wiki/How-to-use-custom-images#importing-svg-images) to get this done. 
    - Somehow the `receive` icon (used through the generated font) is not centered. I have not figured out why yet. (Just in case you are wondering why that looks a bit weird in the expandable FAB).
- I also tried to use `go_router_builder` for using [Typed Routes](https://pub.dev/documentation/go_router/latest/topics/Type-safe%20routes-topic.html) but I don't think there is a way to combine that with a more complex `go_router` configuration, because I would not know how to instruct the code generation. Unfortunately there is not too much documentation and example around this, so I just stopped trying and went with simple named routes and a similar pattern to what we did in the POC.

---


mobile/native/Cargo.toml
```diff
@@ -10,4 +10,9 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 anyhow = "1"
+atty = "0.2"
```
bonomat, line 13: I'd like to be more picky with dependencies. Do we really need this?

mobile/native/Cargo.toml
```diff
@@ -10,4 +10,9 @@ crate-type = ["cdylib", "staticlib"]
...
 flutter_rust_bridge = "1"
+state = "0.5.3"
+time = { version = "0.3", features = ["serde", "parsing", "std", "formatting", "macros", "serde-well-known"] }
```
bonomat, line 16: Do we need all these features?

bonomat: Second question: can we use of `time` which is included in `tracing-subscriber`?

mobile/native/src/logger.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    let fmt_layer = tracing_subscriber::fmt::layer()
+        .with_writer(std::io::stderr)
+        .with_ansi(is_terminal);
+
+    let fmt_layer = if json_format {
+        fmt_layer.json().with_timer(UtcTime::rfc_3339()).boxed()
+    } else {
+        fmt_layer
+            .with_timer(UtcTime::new(format_description!(
```
bonomat, line 147: If this is the only place where we need the `time` crate then I would just switch to `UtcTime::rfc_3339()` here aswell.

mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class _WalletScreenState extends State<WalletScreen> {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+        body: ListView(
+      padding: const EdgeInsets.only(left: 25, right: 25),
+      children: [
+        const Center(child: Text("Wallet Screen")),
+        ElevatedButton(
```
bonomat, line 24: I'm not sure how Flutter/Material works nor if it is relevant yet, but what do you think of wrapping Material types into our own types so that we easily adjust all instances at once if we want to redesign them?

bonomat (commented):
Added some initial comments. Is there anything particularly you want review of?

mobile/lib/features/wallet/wallet_screen.dart
da-kami: I would configure global (theme) changes using the theme. You can overwrite e.g. the default styles for all Elevated Buttons of the App. We did that in the POC already, see:

https://github.com/get10101/10101-PoC/blob/main/lib/main.dart#L103-L123

Sidenote: For very specific instances of buttons we can wrap them into our own type - or just configure them in place as needed.

da-kami:
> Added some initial comments. Is there anything particularly you want review of?

Sorry if there was confusion: This PR is still in Draft state.
I don't want to have feedback yet. This PR was opened so that others that work on the ticket can have a look and see what is going on.

Did you get a request for review? I don't think I clicked it...

mobile/native/Cargo.toml
da-kami: Fair enough - I might remove this commit from this PR and move it to a separate one. 
My intention was to clean up the "template" code that was initially there and replace it with something useful. I opted for adding the logging because that will come in handy. 

But I mostly copied things over from the POC for this, and it is just not super clean yet. 
Since this PR is about ramping up the Dart side I don't think it should be blocked by rust related "cleanup" - I'll see to remove this commit.

da-kami: > I'd like to be more picky with dependencies. Do we really need this?

As far as I understand this was used to pretty print std-err; I don't care much about this. 
I removed it.

mobile/lib/features/trade/settings_screen.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:flutter/material.dart';
+import 'package:get_10101/features/trade/trade_screen.dart';
+
+class TradeSettingsScreen extends StatelessWidget {
```
1010Tom, line 4: If we split the settings screens between `TradeSetting` and `Wallet Settings`, what can you configure in either?

mobile/lib/features/trade/trade_screen.dart
```diff
@@ -15,10 +13,81 @@ class TradeScreen extends StatefulWidget {
...
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: <Widget>[
+              SizedBox(
+                  width: fabWidth,
+                  child: FloatingActionButton.extended(
+                    heroTag: "btnBuy",
+                    onPressed: () {
+                      showModalBottomSheet<void>(
+                        useRootNavigator: true,
+                        backgroundColor: Colors.green.shade50,
```
1010Tom, line 37: Shall we introduce constants so that we can change them all at once, e.g. `GoLongGreen` (or similar) to indicate it's a green used for everything related to go long/buy in the trade screen.

mobile/lib/util/send_receive_icons.dart
```diff
@@ -0,0 +1,29 @@
+/// Flutter icons SendReceiveIcons
+/// Copyright (C) 2023 by original authors @ fluttericon.com, fontello.com
+/// This font was generated by FlutterIcon.com, which is derived from Fontello.
```
1010Tom, line 3: Is this needed?

1010Tom (approved):
Looks good. 

2 Points: 
1. I've added a comment regarding the split settings screen
2. We need to align the logo colors. Also see here: https://github.com/get10101/meta/issues/130

mobile/lib/features/trade/settings_screen.dart
da-kami: This split is indeed a bit tricky - and is not really what we want.

I added another commit that adds a common settings screen that is used in both routes; this allows us to have the same settings screen for both wallet and trade (including general settings and app info), but potentially highlight the settings that are relevant when opening the screen:



Note: This is not totally in line with "each screen should be identified by a route"; but I did not find a better solution for now.

mobile/lib/features/trade/trade_screen.dart
da-kami: I invested a timebox to understand how to properly configure such kind of values in the `ThemeData`, but the timebox expired. I think it might still be a good idea to invest a bit of time into making this better, but that probably goes to far for this PR. 

I create https://github.com/get10101/10101/issues/98 to tackle this.

mobile/lib/util/send_receive_icons.dart
da-kami: I removed it.

bonomat:
@da-kami : do you mind deleting the branches if you don't need them anymore please?

Issue #65, opened 2023-01-31
Wallet screen


1010Tom:
![wallet](https://user-images.githubusercontent.com/88702308/215654184-b0516836-9ae2-4ab9-89cc-e68e1f09a5af.gif)

Issue #66, opened 2023-01-31
Wallet history item details


1010Tom:
When clicking on a transaction in the wallet the user wants to see some details . 

We can have atleast 3 different transactions in the wallet: 
- lightning transactions
- on-chain transactions
- trading related activities

da-kami:
It would be great if we can achieve a similar look and feel throughout all the list items - and have similar behavior for them when e.g. clicking on them. 

I propose three steps of information:

1. The list tile shows the most important information
2. When taping on the list tile, the tile [expands](https://pub.dev/packages/expandable) (downwards) to show additional information
3. (optional) we show a button `More details` when expanded that would jump to a child-screen with even more information.

For reference, the `PositionListItem` already has an `ExpandablePanel`, see https://github.com/get10101/10101/blob/main/mobile/lib/features/trade/position_list_item.dart

---

Look of the `OrderListItem`:

![image](https://user-images.githubusercontent.com/5557790/221098610-5524b108-84ba-415f-a88d-83a730c64850.png)

Ticket that tracks adding order information in `ExpandablePanel`: https://github.com/get10101/10101/issues/163

da-kami:
As stated in https://github.com/get10101/10101/issues/67#issuecomment-1442828486 I propose to have three levels of information:

1. The list tile (most important information)
2. The expanded list tile (some more information in expansion + link to jump to "more information"); this can be either a separate detail screen or jumping off to a different screen.
3. (optional) Separate detail screen that one can navigate to from buttons in the expansion.

The following designs are not complete, but should give an idea of how I imagine this:

<img width="445" alt="image" src="https://user-images.githubusercontent.com/5557790/221104331-3d5b15c3-beac-4650-9e34-853dc2e654a9.png">

Previously we did not talk about `Position` in the wallet histroy, but just had `Order Filled` entries, but I think that is not great because we cannot easily distinguish between in- and outgoing flow. I think this reflecting the position is more useful for the user (but this might be subjective).

notes on the data (model):

- For the lightning related items it is important to have a fee-breakdown in the expansion
- For the trade items in the wallet history ...
    - ... there are three items:
        1. `Position Opened`: This item is created when there was no position before; funds flow *out* of the wallet
             - this item being the latest trade related item in the wallet histroy indicates that there must be an *open position*
        2. `Position Changed` This item is create when there was a position already and we changed it; funds flow *out* of the wallet
            - this item being the latest trade related item in the wallet histroy indicates that there must be an *open position*
[...]

Issue #60, opened 2023-01-30
Play with the color gradient of the wallet


bonomat:
Our current wallets look like this: 

<img width="609" alt="image" src="https://user-images.githubusercontent.com/224613/215404026-200e717f-bc2e-4a30-9a4e-0b2954005b64.png">

<img width="609" alt="image" src="https://user-images.githubusercontent.com/224613/215404055-9ca681c3-fc03-429c-8be3-2e0c3213af0f.png">

This ticket is about playing around with this gradient. 

For reference the initial feedback on this was that it is too colorful

bonomat:
Note: please play with the styles: `WalletTileGradien_Yellow`

da-kami:
To finally do something about this and close this Figma issue (for now) I opted to remove the Gradient and use the low-saturation colors that we used for the stable-sats mock. 

I updated the Figma mock accordingly.

For reference this is how the colors look:

![image](https://github.com/get10101/10101/assets/5557790/20336308-09ff-4766-ad06-5c848a632185)

PR #99, opened 2023-02-06
Add wallet screen

Restioson:
resolves https://github.com/get10101/10101/issues/65

As per @da-kami's request I am creating this draft pull request very early (after my first commit) so that he may take over if I am unavailable.

The goal of this draft PR is to add the wallet screen and display the balances based on data which is mocked in Rust and retrieved through the FFI API. That is, I am aiming to implement the layout and plumbing between the Rust's UI model and Flutter view.

To do:
- [x] Call into Rust to retrieve balances
  - [ ] ~~Stream API plumbing~~
- [x] Style properly
  - [ ] ~~Make it faithful to Figma~~
  - [x] Gradient
  - [x] Keep colours in a theme for configurability
  - [x] Lighting & on-chain icons
- [x] Send/receive buttons on click of different balance types


Restioson:
I see that there is a lot of code in the 10101 POC repository for handling wallet state and how it changes. It also has things like transaction history and a distinction between confirmed, trusted, and untrusted UTXOs. As per the figma prototype, the distinctions between these UTXO types are not currently displayed. Therefore, I am leaving them out for now and simplifying everything into just a number of satoshis available on lighting and on chain. Please let me know if I should change this. I guess the crux of the matter is whether we want to build now for what we think we will need (i.e for what the prototype needed) or just get something small up and running for these initial PRs.

da-kami:
> I see that there is a lot of code in the 10101 POC repository for handling wallet state and how it changes. It also has things like transaction history and a distinction between confirmed, trusted, and untrusted UTXOs. As per the figma prototype, the distinctions between these UTXO types are not currently displayed. Therefore, I am leaving them out for now and simplifying everything into just a number of satoshis available on lighting and on chain. I am also simplifying the history as per the prototype to simply show direction, time, amount, address, and whether or not it was on chain.

This is indeed missing in the Figma prototype, thanks for pointing it out!
I created https://github.com/get10101/10101/issues/101 so that we can tackle it before implementing the details.

Note that there is also a pending `Figma` ticket to add a history-detail bottom sheet, see #66 

Also note that there is still a ticket to display details when tapping a history item: https://github.com/get10101/10101/issues/66

> Please let me know if I should change this. I guess there are two decisions at play here. The first is whether we want to build now for what we think we will need (i.e for what the prototype needed) or just get something small up and running for these initial PRs. The second is whether we want these models to be reused for other parts of the app or if they should have different, more detailed models when more detailed data is required - presumably the experience of writing the POC will help with deciding on this.

I think it's OK to model this is figma first to get a better idea how it can look (see #101 ), then we can create specific tickets for the implementation. I think the states there should already be somewhat complete - we might learn some more once we implement the transaction state handling in the Rust.

Your approach to create a first, simplified, implementation is fine. We can add follow-up tickets for https://github.com/get10101/10101/issues/65

da-kami:
Note: I liked this PR to ticket https://github.com/get10101/10101/issues/65 and assigned you to https://github.com/get10101/10101/issues/65 so that it's clear you are working on this. 
I will see to focus on the Trade screen for now and keep following the progress on the wallet screen here :)

mobile/lib/features/wallet/wallet_change_notifier.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:f_logs/f_logs.dart';
+import 'package:flutter/material.dart' hide Flow;
+import 'package:get_10101/ffi.dart';
+
+class WalletChangeNotifier extends ChangeNotifier {
+  WalletInfo walletInfo = WalletInfo(
+      balances: Balances(onChain: 0, lightning: 0),
```
da-kami, line 7: In the POC we opted for depicting the lightning balance separate from the on-chain balance, because they might change completely separately.
In the mobile app we might also only be concerned about one of of the two - e.g. in `Trade` we will only be concerned about he `Lightning` balance, so it might be better to split that up. Otherwise refresh might also take long, even if we are only interested in one.

da-kami: I will need the `Lightning` balance in the `Trade` widgets, I will just use a placeholder change notifier for now, and we will have to merge that eventually.

mobile/native/src/api.rs
```diff
@@ -5,3 +5,56 @@ use flutter_rust_bridge::StreamSink;
...
+    pub history: Vec<Transaction>
+}
+
+pub struct Balances {
+    pub on_chain: u64,
+    pub lightning: u64,
+}
+
+#[tokio::main(flavor = "current_thread")]
+pub async fn refresh_wallet_info() -> WalletInfo {
```
da-kami, line 20: Is it on purpose that we poll for these values?

In the PoC we used streams for such kind of updates in the end. So, instead of calling `refresh` from the UI we subscrubed to updates from Dart and pushed updates into a stream. 

I think we can use the same pattern here, it would be great if we can already set this up on the rust side (with just an initial mocked value being used to initialize the stream), so that we can reflect it in Dart code.

mobile/lib/features/wallet/wallet_change_notifier.dart
Restioson: > In the POC we opted for depicting the lightning balance separate from the on-chain balance, because they might change completely separately.

Are you suggesting having the WalletInfo have two separate fields - onChain and lighting - instead of them being bundled inside of Balances? Or, having a separate change notifier for both?

Restioson:
I've now added a gradient to mimic the one in the Figma. It is not totally faithful - neither in colour nor shape. I was not sure how I should be handling the exact theme colours that we are using - is there a theme definition file somewhere I should be referring to? How are we handling icons (for the bitcoin and lightning icons)?

screenshot of the wallet balances view

Until those questions are resolved, I will work more on the functionality and FFI, e.g making the send/receive buttons appear when tapped.

da-kami:
> 


You can have a look at https://github.com/get10101/10101/pull/104 to see how I extended the theme with `TradeTheme` - I would have done the same for the wallet and define the colors there. I felt it was fair to separate those instead of adding one big theme file. 

Note on the gradient: We will likely want to change that, you don't have to invest more time into replicating the gradient; see #60 

I will see to link dependencies on Figma tickets relevant for the implementation in future tickets, it's hard to know these dependencies at the moment.

da-kami:
Note: I will see to get https://github.com/get10101/10101/pull/104 into a mergeable state today. 
I already extraced several widgets into `common` that will be handy for the wallet side as well, such as:

- `submission_status_dialog.dart`
- `amount_text.dart`

mobile/lib/features/wallet/wallet_change_notifier.dart
da-kami: Initially I thought a separate change notifier would be a good idea, but that was before I learned that `Provider` is capable of updating upon *selected* values of the `ChangeNotifier` changing (using `Selector`; I used it quite often in the trade widgets).

That said: It does not matter that much if it's one `ChangeNotifier` or not because we can selectively update on changes.
You can keep the code as is for now The backend can decide the update frequency regardless of the balance that is returned being combined in one `struct` or not.

da-kami:
@Restioson can we get this work into a mergable state soon and then add follow up tickets? 
I think the scope of this ticket might be too big and it would be better to merge some work early and then iterate on it - otherwise we will likely run into merge conflicts now that the integration work as has started with https://github.com/get10101/10101/pull/146

Restioson:
I agree :) Shall we leave out the more detailed FFI plumbing and Figma faithfulness and just finish up the tx/rx buttons? Can adjust todo list accordingly.

Restioson:
I've now added the buttons and they are animated and routed properly! I think this may need to be squashed up and reviewed and then it will be ready.

mobile/lib/features/wallet/wallet_theme.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:flutter/material.dart';
+
+@immutable
```
da-kami, line 3: 💡 O cool, I did not know one can mark an entire class as immutable.

mobile/lib/features/wallet/wallet_theme.dart
```diff
@@ -0,0 +1,73 @@
...
+  final Color dividerColor;
+  final ButtonStyle iconButtonStyle;
+
+  WalletTheme(
+      {this.lightning = Colors.yellow,
+        this.onChain = Colors.orange,
+        this.borderColor = Colors.grey,
+        this.bgColor = Colors.white,
+        this.dividerColor = Colors.black,
+        ColorScheme? colors,
```
da-kami, line 19: 🔧 I am a bit confused by this code.

`colors` is optional, but we use it with `!` below - without a null check. Or am I missing something?
I think it needs a null-check or it has to be mandatory. 

I like the idea to pass in the swatch to derive colors from it.

mobile/lib/main.dart
```diff
@@ -119,12 +122,16 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
...
+    MaterialColor swatch = Colors.blue;
+
     return MaterialApp.router(
       title: "10101",
       theme: ThemeData(
-        primarySwatch: Colors.blue,
-        extensions: const <ThemeExtension<dynamic>>[
-          TradeTheme(),
+        primarySwatch: swatch,
+        useMaterial3: true,
```
da-kami, line 131: 💡 Interesting, I did not know one has to configure that to use the latest version.
❓What's the effect of setting this? Are there any compatibility concerns here, i.e. is material3 only supported by android >x ?

da-kami (changes requested):
Thanks for the great work!

I think it'd be great to get a service(s) in for the calls into Rust.

mobile/lib/features/wallet/balance_row.dart
```diff
@@ -0,0 +1,187 @@
...
+import 'package:get_10101/features/wallet/receive_screen.dart';
+import 'package:get_10101/features/wallet/send_screen.dart';
+import 'package:get_10101/features/wallet/wallet_change_notifier.dart';
+import 'package:get_10101/features/wallet/wallet_theme.dart';
+import 'package:get_10101/ffi.dart';
+import 'package:go_router/go_router.dart';
+import 'package:provider/provider.dart';
+
+class BalanceRow extends StatefulWidget {
+  // TODO(restioson): use dart version of this?
```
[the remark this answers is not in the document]
Restioson: The ffi.dart import covers this

mobile/lib/features/wallet/wallet_theme.dart
Restioson: That is copied from TradeTheme :sweat_smile:

mobile/lib/features/wallet/wallet_theme.dart
Restioson: It's done this way to allow either ColorScheme or ButtonStyle to be passed. First is passed when WalletTheme is created - the ColorScheme comes directly from the main ThemeData. Second is only used for copyWith and lerp. I am not sure how to express "must have one or the other" in Dart.

mobile/lib/main.dart
Restioson: This was a quick and dirty solution to get the send/receive buttons working as I copied them from a flutter example which relies on the material3 ColorScheme properties being set up properly. It can definitely be changed but I am not 100% sure how we want to style the buttons. However, apparently Flutter will move to having useMaterial3: true as default soon anyway [(see docs(](https://api.flutter.dev/flutter/material/ThemeData/useMaterial3.html). Apparently it may lead to an inconsistent look and feel for unmigrated widgets.

mobile/lib/features/wallet/wallet_theme.dart
da-kami: hahaha, cool - then I did apparently know at some point and forgot it again

mobile/lib/features/wallet/wallet_theme.dart
da-kami: I think the easiest solution for this example would be to give `ColorScheme? colors` a default value to ensure that it is always set:

```
  WalletTheme(
      {this.lightning = Colors.yellow,
        this.onChain = Colors.orange,
        this.borderColor = Colors.grey,
        this.bgColor = Colors.white,
        this.dividerColor = Colors.black,
        ColorScheme colors = ColorScheme.fromSwatch(primarySwatch: Colors.blue),
        ButtonStyle? iconButtonStyle,}):
        iconButtonStyle = iconButtonStyle ?? IconButton.styleFrom(
          foregroundColor: colors!.onPrimary,
          backgroundColor: colors.primary,
          disabledBackgroundColor: colors.onSurface.withOpacity(0.12),
          hoverColor: colors.onPrimary.withOpacity(0.08),
          focusColor: colors.onPrimary.withOpacity(0.12),
          highlightColor: colors.onPrimary.withOpacity(0.12),
        );
```

We don't use `  ColorScheme colors` internally, so I don't see a reason not to default it. But maybe I'm missing something.

---

Alternatively, you could add an `assert!` to ensure that *either* of the values is set; here's an example where I use an `assert` in a constructor in a different context:

https://github.com/get10101/10101/blob/e038857285b25c28db66d8798a666ce7e0955c7f/mobile/lib/features/trade/trade_tabs.dart#L35

mobile/lib/main.dart
da-kami: Ok, thanks for the link. 
It seems that the transition from m2 to m3 is still in progress - but mostly for the m3.web components, as far as I could see the mobile components are stable already.

---

I think it's OK to stick to m3 and accept that some widgets might not behave perfect; m2 will not be supported eventually, so better stick to m3 early.

da-kami:
@Restioson when are you planning to work in the changes? 
I will focus on driving the `position` and `order` forward today, and will drive by here tomorrow again. It'd be great if we can get this in a mergable state by the end of the iteration tomorrow.

Restioson:
Hi, I am working on these changes now :) Just fixed the animation and will tend to the rest of the changes you've asked for shortly.

mobile/lib/features/wallet/wallet_theme.dart
Restioson: The change you've given as-is with 
```
ColorScheme colors = ColorScheme.fromSwatch(primarySwatch: Colors.blue)
```

will not work as fromSwatch is not const, so I think the assert solution may be best

Restioson:
I have converted everything to services now. I have also switched to the new AmountText widget, which is very nice!

mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -46,15 +46,15 @@ class _ReceiveScreenState extends State<ReceiveScreen> {
...
-                  await api.openChannel();
-                  FLog.info(text: "Open Channel successfully started.");
-                } catch (error) {
-                  FLog.error(text: "Error: $error", exception: error);
-                }
-              },
-              child: const Text("Open Channel!"))
+            onPressed: () async {
+              try {
+                await api.openChannel();
```
da-kami, line 51: I think this call should be in the `WalletService` as well.

da-kami (approved):
Looks good! One more API call that I would move into the `WalletService` :)

mobile/lib/features/wallet/receive_screen.dart
Restioson: Can move this. Shall I rename WalletInfoService to WalletService as opening a channel is not really getting info from the wallet?

Restioson:
Rebasing this on main now; have to go for a class but should finish by the end of my day today. There is a bit more work involved in the rebase than just fixing conflicts unfortunately but I'll be done soon

Restioson:
From my end this is done - rebased on main and squashed too.

mobile/native/src/api.rs
holzeis: In the PoC we were able to refresh the wallet info manually (i think it was also triggering a sync) - so its good we keep that.

mobile/native/src/api.rs
```diff
@@ -17,6 +17,66 @@ pub fn init_logging(sink: StreamSink<logger::LogEntry>) {
...
+                amount_sats: 104,
+                wallet_type: WalletType::Lightning,
+            },
+        ],
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct Transaction {
+    // TODO(Restioson): newtype?
```
holzeis, line 58: 🤔 should we act on this todo?

holzeis (approved):
LGTM 👍

mobile/native/src/api.rs
Restioson: I think it is out of scope for this PR

mobile/lib/features/wallet/wallet_change_notifier.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  Future<void> refreshWalletInfo() async {
+    update(await service.getWalletInfo());
+  }
+
+  Amount total() => Amount(onChain().sats + lightning().sats);
+  Amount onChain() => walletInfo.balances.onChain;
+  Amount lightning() => walletInfo.balances.lightning;
+
+  // TODO: This is not optimal, because we map the WalletInfo in the change notifier. We can do this, but it would be better to do this on the service level.
```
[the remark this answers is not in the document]
Restioson: No, I don't think so. This is the same as what exists for the order notifier and that isn't fixed either so the fix can go together

da-kami:
Note: I merged this one without `bors`, because I don't think I rebase the Fork. Since there were no conflicts this should be fine.

Issue #98, opened 2023-02-06
Extract the colors of the `Buy` and `Sell` button into the Theme


da-kami:
We should generally learn to use the `ThemeData` to our needs so we can scale custom colors / styles once the application grows.

Here is some inspiration how to do that: https://stackoverflow.com/questions/49172746/is-it-possible-extend-themedata-in-flutter
I think the proper way, as of Flutter 3 is this approach: https://lakshay-parnami.medium.com/flutter-3-themedata-finally-extendeddd-aa0069cc9996

We could then create a `TradeTheme` where we define specific styles related to the trade feature.

PR #102, opened 2023-02-07
refactor: add trade theme

da-kami:
resolves https://github.com/get10101/10101/issues/98

Using `ThemeExtension` was inspired by [this blogpost](https://lakshay-parnami.medium.com/flutter-3-themedata-finally-extendeddd-aa0069cc9996)

### DIFF
--- a/mobile/lib/features/trade/trade_theme.dart
+++ b/mobile/lib/features/trade/trade_theme.dart
@@ -1,0 +1,48 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class TradeTheme extends ThemeExtension<TradeTheme> {
+  // Unfortunately shades of Colors in flutter are not constant; see https://github.com/flutter/flutter/issues/31351
+  // Workaround to make is constant: Initialize the color from HEX
+  // For color codes see: https://api.flutter.dev/flutter/material/Colors/green-constant.html
+  static const Color green600 = Color(0xFF43A047);
+  static const Color red600 = Color(0xFFE53935);
+
+  const TradeTheme({
+    this.buy = green600,
+    this.sell = red600,
+  });
+
+  final Color buy;
+  final Color sell;
+
+  @override
+  TradeTheme copyWith({
+    Color? buy,
+    Color? sell,
+    ShapeBorder? tradeButtonShape,
+    double? tradeButtonWidth,
+  }) {
+    return TradeTheme(
+      buy: buy ?? this.buy,
+      sell: sell ?? this.sell,
+    );
+  }
+
+  @override
+  TradeTheme lerp(ThemeExtension<TradeTheme>? other, double t) {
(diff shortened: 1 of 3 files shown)


mobile/lib/features/trade/trade_theme.dart
```diff
@@ -0,0 +1,66 @@
...
+        Radius.circular(8),
+      ),
+    ),
+    this.tradeButtonWidth = 100.0,
+  });
+
+  final Color buy;
+  final Color sell;
+  final ShapeBorder tradeButtonShape;
+  final double tradeButtonWidth;
```
da-kami, line 25: note: Adding the button's width and shape to the theme could be disputed - usually the theme is used for Colors and Fonts; see: https://docs.flutter.dev/cookbook/design/themes

My idea is to create Themes per feature for now, and that's what motivated me to add more to the theme. We can review this practice once the application grows.

bonomat: Do you expect this being used for multiple buttons? 
If only used for the buy/sell FABs I don't think this should be in here.

mobile/lib/features/trade/trade_theme.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class TradeTheme extends ThemeExtension<TradeTheme> {
+  // Unfortunately shades of Colors in flutter are not constant; see https://github.com/flutter/flutter/issues/31351
+  // Workaround to make is constant: Initialize the color from HEX
+  // For color codes see: https://api.flutter.dev/flutter/material/Colors/green-constant.html
+  static const Color green600 = Color(0xFF43A047);
+  static const Color red600 = Color(0xFFE53935);
```
bonomat, line 9: Random thought: we might want to document these colors outside of the code base so that we also use them elsewhere.

bonomat (approved):
LGTM :)

Restioson (approved):
LGTM, I am not 100% sure what is idiomatic in flutter though

mobile/lib/features/trade/trade_theme.dart
da-kami: yes, I think we should eventually store this "somewhere" globally. I don't have a clear vision on what the best place is yet and how far we should go. One idea I have is to define a design system that is available publicly online (we had this for COMIT at some point but I can't find the page where we managed this), and then pull all the information from there into the code / Figma (...). 
We can also store them in the `Brand Assets` in Notion.

For now I'd say:

We will have to align the colors that we use in Figma with those in the MVP. If that means that we adapt some of the Figma colors to Material codes, or vice-versa I leave open. 

Once the MVP is shipped the code becomes law, and changes to colors in Figma should always be reflected in code.
Let's come back to this once we are closer to having a more complete color pallet.

PR #104, opened 2023-02-07
Trade screen order bottom sheets until submission

da-kami:
Since this is rapidly growing and I already extracted some useful common widgets let's get this in so that we can build on top.

Notes:

- I did not really invest much time into tests yet - I will see to tackle that in follow ups as well. The `Buy/Sell` logic should be captured in widget tests.
- I decided to manage the `TradeValues` state in Dart and only get the native Rust backend involved once an `Order` is sent. I had a version where I tried to manage this state in Rust, but that was cumbersome during the development phase (re-generating code, the rust model being part of the UI leading to cumbersome changes, ...). I think this part is better encapsulated in Rust. 
- The`Order` that will be displayed in the `Order` tab (that part is not implemented yet) will come from the Rust code, and its state will be managed in Rust. For that I will likely depend on the `Rust` model directly. Note that we could also decide to create a fully encapsulated model and service in Dart, but that would go against what is outlined in https://github.com/get10101/10101/pull/95 - since we don't expect the Rust API to change often, and we don't expect to ever change the Rust code to something it's fair to depend on the API model directly as outlined in https://github.com/get10101/10101/pull/95

Done:

- [x] Buy / sell bottom sheets 
- [x] Success screen popup after successfully submission of the order

To be done in follow up; I will create specific tickets so we can tackle in parallel as well:

- Order API: Post/Get orders to/from backend
- Positions API: Post/Get orders to/from backend
- Implement Close
- Implement the candlestick chart or create a ticket for it
- Form validation for the Buy/Sell bottom sheet (so one can only press the button when
- Use the actual wallet balance in the `TradeValuesChangeNotifier` (depends on #65 )
- Price feed and actual price in `TradeValuesChangeNotifier`
- Funding Rate feed and actual funding rate in `TradeValuesChangeNotifier`

---


mobile/lib/common/amount_text.dart
```diff
@@ -0,0 +1,32 @@
...
+String formatAmount(AmountDenomination denomination, Amount amount) {
+  final formatterSat = NumberFormat("#,###,###,###,###", "en");
+  final formatterBtc = NumberFormat("##,##0.00000000", "en");
+
+  switch (denomination) {
+    case AmountDenomination.bitcoin:
+      return "${formatterBtc.format(amount.btc)} BTC";
+    case AmountDenomination.satoshi:
+      return "${formatterSat.format(amount.sats)} sats";
+  }
```
bonomat, line 31: Nice :)

mobile/lib/features/trade/domain/trade_values.dart
```diff
@@ -0,0 +1,91 @@
...
+      required double price,
+      required double fundingRate,
+      required Direction direction}) {
+    Amount margin =
+        Amount(api.calculateMargin(price: price, quantity: quantity, leverage: leverage.leverage));
+    double liquidationPrice = api.calculateLiquidationPrice(
+        price: price, leverage: leverage.leverage, position: direction.intoPosition());
+
+    // TODO: Calculate fee based on price, quantity and funding rate
+    Amount fee = Amount(30);
```
bonomat, line 39: What fee is this?

mobile/native/src/api_calculations/mod.rs
```diff
@@ -0,0 +1,73 @@
+use crate::api_model::Position;
+use bdk::bitcoin;
+use bdk::bitcoin::Denomination;
+use rust_decimal::prelude::ToPrimitive;
+use rust_decimal::Decimal;
+
+/// Calculate the margin in BTC.
+pub fn calculate_margin(opening_price: f64, quantity: f64, leverage: f64) -> u64 {
```
bonomat, line 8: Tests would be nice :)

mobile/lib/features/trade/trade_screen.dart
```diff
@@ -32,6 +31,66 @@ class _TradeScreenState extends State<TradeScreen> {
...
+                                type: ValueType.amount,
+                                value: submitOrderChangeNotifier.pendingOrderValues!.fee,
+                                label: "Fee")
+                          ],
+                        ),
+                      ),
+                      Padding(
+                        padding: const EdgeInsets.only(top: 20, left: 10, right: 10, bottom: 5),
+                        child: Text(
+                            "Your Position will be shown automatically in the Orders tab once your buy order has been filled!",
```
bonomat, line 70: 🙃
The text is the same for `buy` and `sell` orders.

mobile/lib/features/trade/trade_screen.dart
```diff
@@ -32,6 +31,66 @@ class _TradeScreenState extends State<TradeScreen> {
...
+
+                  switch (state) {
+                    case PendingOrderState.submitting:
+                      return SubmissionStatusDialog(
+                          title: title, type: SubmissionStatusDialogType.pending, content: body);
+                    case PendingOrderState.submittedSuccessfully:
+                      return SubmissionStatusDialog(
+                          title: title, type: SubmissionStatusDialogType.success, content: body);
+                    case PendingOrderState.submissionFailed:
+                      // TODO: This failure case has to be handled differently; are we planning to show orders that failed to submit in the order history?
```
bonomat, line 84: Even though, an order should not _fail_ in general but be pending, I think we should depict for now that a `market order` was not able to be filled immediately as discussed in our MVP scope.

bonomat (approved):
Most of the highlevel things were discussed offline. I've had a quick look and couldn't find any major issues, so I think it's fine to merge and iterate over it.

mobile/lib/features/trade/buy_sell_bottom_sheet.dart
```diff
@@ -37,25 +121,72 @@ class BuySellContents extends StatelessWidget {
...
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      crossAxisAlignment: CrossAxisAlignment.center,
+      children: [
+        Wrap(
+          runSpacing: 15,
+          children: [
+            Row(
+              children: const [
+                Flexible(child: Text("Available Balance:")),
+                Flexible(child: Text("2,000 sats"))
```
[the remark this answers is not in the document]
da-kami: Sorry, seems the commits were still not clean - this is an `Amount` that is displayed using `AmountText`.

mobile/lib/features/trade/domain/trade_values.dart
da-kami: Note: I just depicted this as it is depicted in Figma.
My assumption is, that it is the fee for the first 24h.

In ItchySats we defined two fees when opening a position:

1. The first 24h of "rollovers" (can be a positive or negative "fee")
2. An optional opening fee charged by the maker (always a positive value on the taker side); that we never used in production

Given that we have an orderbook now, I am not sure what this fee should represent; I suppose it could be the fee for the "orderbook service" as well. This might need some more thought!

mobile/native/src/api_calculations/mod.rs
da-kami: yeah, we had them in ItchySats but never pulled them into the PoC (same here...).
I'll add a ticket to re-add them!

mobile/lib/features/trade/trade_screen.dart
da-kami: Thanks for pointing that out! I'll add a specific ticket for that task.

mobile/lib/features/trade/buy_sell_bottom_sheet.dart
```diff
@@ -37,25 +121,72 @@ class BuySellContents extends StatelessWidget {
...
+              ],
+            ),
+            TextFormField(
+              controller: TextEditingController()..text = '19,900.0',
+              enabled: false,
+              decoration: const InputDecoration(
+                border: OutlineInputBorder(),
+                hintText: "Price",
+                labelText: "Market Price",
```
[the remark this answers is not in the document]
da-kami: Sorry, I am not sure what you are referring to - you mean pull the text into localization files so that all app text can be easily translated into other languages?

mobile/lib/features/trade/domain/trade_values.dart
bonomat: Created a ticket for that: https://github.com/get10101/10101/issues/139

PR #108, opened 2023-02-08
Fix/android build and running the application on Android phone

da-kami:
The first commit updates the manifest files so we are able to build and deploy.
The second commit changes the the strategy to remove the splashscreen - on my phone I keep running into an error when initializing the rust logging, resulting in the app being stuck on the splashscreen. 

Note that several sections of the old `Manifest` file were not actually needed - I tried to reduce what is needed to fix this to a minimum. If something looks like it slipped in unnecessarily nonetheless please point it out!

I also removed a whole section of the `build.gradle` file because we did not have that in the POC, and it does not seem to be needed to build and deploy - but maybe I'm missing something. @bonomat might know more :)

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -26,8 +26,7 @@ apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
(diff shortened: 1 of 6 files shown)


bonomat (approved):
LGTM

PR #109, opened 2023-02-08
Use a global test subscriber and switch to traces for most printlns

Restioson:
Global test subscriber was needed to capture traces from threads spawned by the rust-lightning libraries (e.g lightning-background-processor spawns its own threads, which we cannot install the local subscriber in).

I did leave some printlns  hat start with `EVENT` as they seemed to take control of stdout itself and do other things, which I did not understand and want to meddle with yet.

### DIFF
--- a/crates/ln-dlc-node/src/lib.rs
+++ b/crates/ln-dlc-node/src/lib.rs
@@ -1,3 +1,4 @@
+use crate::logger::TracingLogger;
 use dlc_manager::custom_signer::CustomKeysManager;
 use dlc_manager::custom_signer::CustomSigner;
 use dlc_messages::message_handler::MessageHandler as DlcMessageHandler;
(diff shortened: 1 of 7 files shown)


holzeis:
> I did leave some printlns hat start with EVENT as they seemed to take control of stdout itself and do other things, which I did not understand and want to meddle with yet.

what printlns do you mean exaclty?

Restioson:
For instance:
```
println!("\nEVENT: received payment from payment hash {} of {} millisatoshis", util::hex_str(&payment_hash.0), amount_msat.to_string());
print!("> ");
io::stdout().flush().unwrap();
```

in event_handler.rs

holzeis:
> For instance:
> 
> ```
> println!("\nEVENT: received payment from payment hash {} of {} millisatoshis", util::hex_str(&payment_hash.0), amount_msat.to_string());
> print!("> ");
> io::stdout().flush().unwrap();
> ```
> 
> in event_handler.rs

large parts of this code was copied from the `ldk-sample` which exposed these functions through the cli. Since we are wrapping it in automated tests, we can safely replace them.

holzeis (approved):
Thank you!

PR #116, opened 2023-02-09
fix: align all product identifier strings to `finance.get10101.app`

da-kami:
Took my brain overnight that it might be better to align these - I only changed some of them yesterday :)
Tested deploy to my Android phone and iOS simulator after a `flutter clean`


1010Tom (approved):
👍

PR #117, opened 2023-02-09
feat: run cargo tests on ci

### LINKED ISSUE #113: Add cargo tests to ci pipeline
With https://github.com/get10101/10101/pull/105 we introduced two important tests for single and multi hop payments. These, and other cargo tests should get executed automatically on the pipeline according to the [ci strategy](https://github.com/get10101/10101/blob/main/docs/001-ci-strategy.adoc).

Note, the tests are currently dependent on a filesystem lock, due to the wallet using the sled db. Thus the tests can not be executed in parallel as otherwise the single and multi hop payment tests will compete over the database lock (and one test will always fail) The only workaround at the moment is to run the tests in a single thread.

```
cargo test -- --test-threads=1
```

bonomat:
we can't just make use of nigiri on CI unfortunately, hence, we copied the docker-compose file from https://github.com/vulpemventures/nigiri and changed it slightly to use docker-managed containers.

resolves #113


.github/workflows/ci.yml
```diff
@@ -51,3 +51,30 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: extractions/setup-just@v1
+      - name: Setup rust toolchain
+        run: rustup show
+      - uses: Swatinem/rust-cache@v2.2.0
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: "stable"
+      - name: Install FFI bindings
+        run: cargo install flutter_rust_bridge_codegen
```
holzeis, line 67: 🔧 We should fix the version of the `flutter_rust_bridge_codegen` here. Otherwise it will always install the latest, which will get out of sync with our flutter dependency which is currently locked to `1.62.0`

.github/workflows/ci.yml
```diff
@@ -51,3 +51,30 @@ jobs:
...
+      - name: Start containers
+        run: |
+          docker-compose up -d
+          sleep 10 # We need to give docker a bit of time to startup
+
+      - name: Test containers are up
+        run: |
+          curl -d '{"address":"bcrt1qylgu6ffkp3p0m8tw8kp4tt2dmdh755f4r5dq7s", "amount":"0.1"}' -H "Content-Type: application/json" -X POST http://localhost:3000/faucet
```
holzeis, line 75: 🤔 How about adding a retry logic here? by checking if the returned `http_code` equals `200`

```
$(curl -LI http://google1.com -o /dev/null -w '%{http_code}\n' -s)
```

This might also allow us to remove the `sleep 10` from the step before.

.github/workflows/ci.yml
```diff
@@ -51,3 +51,30 @@ jobs:
...
+
+      - name: Generate FFI bindings
+        run: just gen
+      - name: Running cargo tests
+        run: cargo test
```
holzeis, line 80: The job is called `integration-test`, but here we are running all tests. Also, given our definition in the [CI strategy](https://github.com/get10101/10101/blob/main/docs/001-ci-strategy.adoc) I think our single and multi hop payment tests rather qualify as system tests as "They need a whole infrastructure to be set up to run."

holzeis (approved):
Thank you!

I am happy to merge that, but would like to note that this PR does not adhere to our [ci strategy ADR](https://github.com/get10101/10101/blob/main/docs/001-ci-strategy.adoc) yet. And follow up PRs are required to change that.

* System tests run only occasionally and only if all unit tests have passed. We run integration tests only on bors r+. Once this branch builds, it is automatically merged into master. 
* We only run tests for what has been changed. E.g. If a user only does frontend changes, we only run tests for the frontend.

docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+      - bitcoin
+    ports:
+      - 50000:50000
+      - 30000:30000
+    volumes:
+      - bitcoin:/config
+    restart: unless-stopped
+
+  # Block explorer frontend
+  esplora:
```
holzeis, line 61: 🔧 We don't have to start a blockchain explorer for the tests.

bonomat: I'll leave it in this file so that we can use it locally (instead of nigiri)

bonomat:
> * System tests run only occasionally and only if all unit tests have passed. We run integration tests only on bors r+. Once this branch builds, it is automatically merged into master.

Fixed :) [`6d17d41` (#117)](https://github.com/get10101/10101/pull/117/commits/6d17d414f8c51b2778190c8d3be9064274ee4365) and [`fa22cdf` (#117)](https://github.com/get10101/10101/pull/117/commits/fa22cdfa713f389c2a07f271a6739e8d4099d8c6)

> * We only run tests for what has been changed. E.g. If a user only does frontend changes, we only run tests for the frontend.

I'll create a follow-up ticket for that

bonomat:
@holzeis : feel free to take over and `bors` it any time.

.github/workflows/ci.yml
Restioson: Specifically we can use --version to set it

bonomat:
I'm giving the new github action with bors a try :) 

bors r+

bonomat:
Integration tests are being run: https://github.com/get10101/10101/actions/runs/4139289572/jobs/7156693794 but failed. Trying to fix it in #121

PR #120, opened 2023-02-09
chore: use github action to cache cargo isntall

bonomat:


holzeis (approved):
We should use the same `flutter_rust_bridge_codegen` version everywhere. Currently our pubspec.lock is still pointing to `1.62.0`. That shouldn't be a big issue, but should be fixed in a follow up PR.

https://github.com/get10101/10101/blob/d89c055dcf4381b7e251238cbfb6cdf6f06d5210/mobile/pubspec.lock#L193-L200

bonomat:
@holzeis : I've bumped it to 1.63.1 and unified it everywhere

PR #121, opened 2023-02-09
CI/Bors fixes

bonomat:


.github/workflows/system_tests.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: cargo install flutter_rust_bridge_codegen --version 1.62.1
       - name: Start containers
         run: |
-          docker-compose up bitcoin electrs -d
+          docker-compose up
```
holzeis, line 28: The services have to come after the options - `docker compose up [OPTIONS] [SERVICE...]`

Also I guess we need chopsticks for the system tests. So `docker-compose up -d chopsticks` should be fine. Note, we can skip the `bitcoin` and `electrs` service as these are dependencies of `chopsticks` and will get started anways.

1010Tom (approved):
Imho this is fine if it works.

1010Tom:
Going ahead and merge this is it is not controversial.

bors r+

1010Tom:
Going ahead and merge this is it is not controversial.

@bonomat : I hope this is fine.

bors r+

PR #122, opened 2023-02-13
Just-in-time channels

### LINKED ISSUE #55: ADR for initial funding of Lightning wallet
These options are also outlined in #34:

#### 1. Route payment all the way to the coordinator first
1.  User tries to create invoice.
2.  App generates an invoice *via the coordinator* with a payment preimage that only the app knows.
3.  ~App shares payment hash with coordinator.~[^1]
4.  Payer pays the invoice, which eventually reaches the coordinator as a HTLC.
5.  Coordinator matches the received HTLC to the one registered by the app.
6.  Coordinator opens channel with user, including HTLC with amount minus flat routing fee and liquidity-based routing fee.
7.  App claims payment.

#### 2. Open channel between coordinator and user first
1.  User tries to create invoice.
2.  Coordinator opens channel with user with desired invoice amount as inbound liquidity for user (plus whatever inbound liquidity they would want).
3.  Invoice then generated, including route hint for final hop between coordinator and user.

### Tasks

We want to:

- [x] Test that option 1 is feasible, as it is the preferred option.
- [ ] Write an ADR to explain findings, decision and any other insights.

[^1]: The coordinator shouldn't need to know about the invoice. It should be sufficient to learn about the incoming payment, match it with a peer with a connection, but without a channel, and proceed to create the channel with the forwarded payment minus the liquidity fees.

bonomat:
This PR adds the feature of paying an invoice to a 10101 user without having a channel open first. 
The ADR for this protocol is also included

Resolves #55


crates/ln-dlc-node/src/tests/system_tests/channel_less_payment.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::tests::system_tests::init_tracing;
+use crate::tests::system_tests::log_channel_id;
+use crate::tests::system_tests::setup_ln_node;
+
+#[tokio::test]
+async fn given_no_channel_with_coordinator_when_invoice_generated_then_can_be_paid_through_coordinator(
+) {
+    init_tracing();
+
+    let test_dir = create_tmp_dir("multi_hop_test");
```
holzeis, line 17: 🔧 `multi_hop_test` that should probably be something like `channel_less_payment`. Although I think it doesn't matter what we provide here as an argument, as it will always be within in a random temp dir. We could as well remove that argument from the function.

crates/ln-dlc-node/src/tests/system_tests/channel_less_payment.rs
```diff
@@ -0,0 +1,154 @@
...
+    let _alice_bg = alice.start().await.unwrap();
+    let _coordinator_bg = coordinator.start().await.unwrap();
+    let _bob_bg = bob.start().await.unwrap();
+
+    // 2. Connect the two nodes.
+
+    tokio::time::sleep(Duration::from_secs(2)).await;
+    alice.keep_connected(coordinator.info).await.unwrap();
+    bob.keep_connected(coordinator.info).await.unwrap();
+    alice.keep_connected(bob.info).await.unwrap();
```
holzeis, line 37: 🤔 I guess `alice` and `bob` don't have to be connected for this test to work. do they?

crates/ln-dlc-node/src/tests/system_tests/channel_less_payment.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    tracing::info!("alice: {}", alice.info);
+    tracing::info!("coordinator: {}", coordinator.info);
+    tracing::info!("bob: {}", bob.info);
+
+    let _alice_bg = alice.start().await.unwrap();
+    let _coordinator_bg = coordinator.start().await.unwrap();
+    let _bob_bg = bob.start().await.unwrap();
+
+    // 2. Connect the two nodes.
```
holzeis, line 32: nit: the comment seems a bit off.

crates/ln-dlc-node/src/tests/system_tests/channel_less_payment.rs
```diff
@@ -0,0 +1,154 @@
...
+            .unwrap();
+        coordinator
+            .fund(bitcoin::Amount::from_sat(1_000_000))
+            .await
+            .unwrap();
+
+        // we need to wait here for the wallet to sync properly
+        tokio::time::sleep(Duration::from_secs(5)).await;
+
+        alice.sync();
```
holzeis, line 52: nit: if we do not fund alice, we don't have to sync her wallet.

crates/ln-dlc-node/src/tests/system_tests/channel_less_payment.rs
```diff
@@ -0,0 +1,154 @@
...
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    // Add 1 confirmation required for the channel to get usable.
+    // we are mining to a random address to no pollute the users wallets
+    let address = Address::from_str("bcrt1qylgu6ffkp3p0m8tw8kp4tt2dmdh755f4r5dq7s")
+        .expect("To be a valid address");
+    fund_and_mine(address.clone(), bitcoin::Amount::from_sat(1000)).await;
+
+    // Add 5 confirmations for the channel to get announced.
+    for _ in 1..6 {
```
holzeis, line 79: ❓ since we change the channel config to forward also on private channels, do we still need these confirmations?

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -400,7 +400,32 @@ impl EventHandler {
                     expected_outbound_amount_msat,
                     "Intercepted HTLC"
                 );
-                // TODO: check if there is a channel present
+
+                // if we have already a channel with them, we try to forward the payment.
+                // TODO: here we would need to increase the channel size if the channel is too small
+                if let Some(channel) =
```
holzeis, line 406: 🤔 Isn't the `requested_next_hop_scid` the fake channel id that we have generated with the invoice? In that case, wouldn't that `scid` different for every invoice? If yes, I guess that check should rather check if there already exists a channel with the target_node.

docs/006-just-in-time-channels.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+We want to offer to the users of the 10101-app a great UX when initially transferring funds into the wallet.
+For example, Alice uses 10101 and wants to receive a payment from Bob via the coordinator but does not yet have a channel open with the coordinator.
+To receive the payment, the coordinator will have to intercept the payment, create a channel on-the-fly to Alice and then forward the payment.
+
+The same feature slightly adapted can be used later-on when receiving funds which exceed the in-bound channel capacity.
+
+=== Options considered
+
+We initially thought that for the described feature above, no communication is needed between the coordinator and the payee (Alice in the scenario above).
```
holzeis, line 57: 🤔 so there is actually no other option, than the way you have implemented it, correct?

holzeis (approved):
Very nice! 👍

crates/ln-dlc-node/src/node.rs
```diff
@@ -461,6 +481,53 @@ impl Node {
         .map_err(|e| anyhow!(e))
     }
 
+    ///  Creates an invoice which is meant to be intercepted
+    ///
+    /// Doing so we need to pass in `intercepted_channel_id` which needs to be generated by the
+    /// intercepting node. This information, in combination with `hop_before_me` is used to add a
+    /// routing hint to the invoice. Otherwise the sending node does not know how to pay the
+    /// invoice
```
luckysori, line 489: :+1: Thanks for the documentation!

docs/006-just-in-time-channels.adoc
```diff
@@ -0,0 +1,59 @@
+= ADR 006 - Just-in-time channels
+Philipp Hoenisch <philipp@10101.finance>
+3.0, July 29, 2022: AsciiDoc article template
+:toc:
+:icons: font
+:attributes: 2023-02-13
+
+Status: Draft
+
+This document contains a description about the protocol we use for channel-less payments.
```
luckysori, line 10: :upside_down_face: Channel-less was a funny name at the time, but it is a bit misleading for anyone reading who doesn't already know what this is.

docs/006-just-in-time-channels.adoc
```diff
@@ -0,0 +1,59 @@
...
+We want to offer to the users of the 10101-app a great UX when initially transferring funds into the wallet.
+For example, Alice uses 10101 and wants to receive a payment from Bob via the coordinator but does not yet have a channel open with the coordinator.
+To receive the payment, the coordinator will have to intercept the payment, create a channel on-the-fly to Alice and then forward the payment.
+
+The same feature slightly adapted can be used later-on when receiving funds which exceed the in-bound channel capacity.
+
+=== Options considered
+
+We initially thought that for the described feature above, no communication is needed between the coordinator and the payee (Alice in the scenario above).
+However, due to technical constraints in https://github.com/lightningdevkit/rust-lightning/blob/a1f7cbb572fa50dcb933bb162facb5ce2cd82576/lightning/src/util/events.rs#L753-L784[rust-lightning], we do not learn about the next hop ID.
```
luckysori, line 58: I wonder if there is a way to tie the fake SCID to the payee public key without the communication step. After all, the coordinator knows the public keys of all the peers that they're connected to. Maybe we could come up with a deterministic algorithm to derive the fake SCID from the public key?

In any case, this is for sure good enough at this stage, and perhaps forever.

luckysori (approved):
Nice work!

I would vote to immediately work to define different (1) Lightning event handlers and (2) Lightning user configurations for coordinator and apps. Eventually these probably will live in separate crates, but it's much easier to reason about this stuff if we don't introduce incorrectly/unnecessarily generic code.

luckysori:
Build passes and we want to keep building on top of this. Let's try to address any comments in follow-up PRs. @bonomat feel free to continue replying to our comments here.

docs/006-just-in-time-channels.adoc
bonomat: Potentially, but we would need to dig into rust-lightning for that.

docs/006-just-in-time-channels.adoc
bonomat: Not unless we want to change rust-lightning.

docs/006-just-in-time-channels.adoc
luckysori: I thought that wouldn't be necessary!

1. Alice generates fake SCID with a deterministic algorithm based on her PK.
2. The coordinator eventually learns about it.
3. The coordinator applies the same algorithm to the PKs of all the connected peers without a channel.
4. Only one result matches the fake SCID from the event.
5. The coordinator then attempts to open the channel with the identified peer.

In my mind, steps (3), (4) and (5) happen in `ln_dlc_node`. Obviously, step (4) has to work reliably.

bonomat: ah, I think now I get it. Sorry, I was confused. 

I think you are right, this should be possible. 

Note that the currently used `get_intercept_scid` is doing quite some stuff to generate a _realistic looking_ scid.

https://github.com/lightningdevkit/rust-lightning/blob/ac6e0b3fed4119c196a5e1e864239087422727ea/lightning/src/util/scid_utils.rs#L100-L104

luckysori: Wow, okay. Good point. Let's not mess with this for now. It might not even be a real problem anyway.

Issue #111, opened 2023-02-08
Subscribe to lightning and dlc events


holzeis:
With https://github.com/get10101/10101/pull/105 we introduced an `event_handler` that is processing all lightning events. dlc will introduce a similar handler.

There are multiple situations where it would be handy to be able to subscribe to theses events so that follow up processing can happen.

1. Testing: Currently we poll for a certain state to imply that an event might have happened. By simply subscribing to the expected event this could be done much cleaner.
2. Payment notifications: Informing the user about an incoming payment, when the event actually arrives.

Additionally it would be nice if the `event_handler` could be separated into individual handlers for e.g. larger events. Done correctly we could leverage that for a more flexible testing approach. e.g. mocking certain events, etc.

bonomat:
Handing over this to @luckysori and @holzeis . :)

luckysori:
Removing assignment of @holzeis and me since we have not worked on it and it's low priority.

luckysori:
To be fair to us, this has already been done to some extent, so I will close it as done haha.

PR #124, opened 2023-02-13
feat: introduce event subscriber

bonomat:
this allows us to subscribe to ln events from the outside

attempts to solve https://github.com/get10101/10101/issues/111

@holzeis / @luckysori : is this what you had in mind?


holzeis:
> this allows us to subscribe to ln events from the outside
> 
> attempts to solve #111
> 
> @holzeis / @luckysori : is this what you had in mind?

cool, thank you for looking into this.

Can you change it so that we can simply subscribe to e.g. `ChannelReady` events and we don't have to filter on the subscriber side? Also it would be great if we could register multiple subscribers for different events.

PR #127, opened 2023-02-13
Follow-up PR to #122

1010Tom:


holzeis (approved):
Thank you 👍 

No sure about the `new_app` and `new_coordinator` apis, but we can change that later if needed.

crates/ln-dlc-node/src/node.rs
```diff
@@ -112,17 +112,70 @@ impl Display for NodeInfo {
...
+            network,
+            data_dir,
+            address,
+            electrs_origin,
+            seed,
+            ephemeral_randomness,
+            user_config,
+        )
+        .await
+    }
```
luckysori, line 164: I suppose we'll eventually move this code to the specific crates, but this is a good first step. Thank you!

crates/ln-dlc-node/src/tests/system_tests/mod.rs
```diff
@@ -139,19 +139,29 @@ pub(crate) async fn setup_ln_node(test_dir: &Path, node_name: &str) -> Node {
...
-        node_name.to_string(),
-        Network::Regtest,
-        data_dir.as_path(),
-        address,
-        ELECTRS_ORIGIN.to_string(),
-        seed,
-        ephemeral_randomness,
-    )
-    .await
+    if is_coordinator {
```
luckysori, line 142: Hahahaha nooooo. Two functions better than a boolean flag? :stuck_out_tongue:

luckysori (approved):
Thanks for addressing the suggestions!

crates/ln-dlc-node/src/tests/system_tests/mod.rs
bonomat: With two functions we would have to duplicate so much stuff, so I went for the easier one :D

PR #140, opened 2023-02-15
Refactor: contract symbol instead of trading pair

da-kami:
Introduce `ContractSymbol` in the UI and harmonize the name.

---

Not sure we want to open that can of worms again: Terminology.
At first I thought about opening a discussion, but I am not sure it's worth it. In the end we just have to agree what we use - I think there is no golden path here.

I opted for using `contract symbol` because that's what we agreed upon in the past.
Personally, I find `trading pair` more intuitive - because it does not limit us to "contracts" - but I'm happy roll with `contract symbol` for now.


bonomat (approved):
For the time being we will only deal with derivatives and contracts, hence, `ContractSymbol` is the correct term to use :)

PR #141, opened 2023-02-15
refactor: direction and position

da-kami:
Order with direction buy or sell; we can either buy or sell with an order - which adds or reduces a position. 
If a position is "negative" it is a short position, if a position is positive it is a long position.

note on the specific PR: It is disputable what the calculation API for the liquidation price should expose, but since we are only using it in the context of order value calculations I opted for exposing `Direction` and map to position in the API. This is a small implementation detail.

---

When introducing `Order` I had a hard time using `Position` in it - it just does not make sense. We have `buy` and `sell` orders. For me the order indicates the *action* of buying or selling something - to either add to or reduce the position. 

But once we have a position and we are "exposed to the market" we are either long or short; where short could be regarded as negative and long as positive. Buy buying and selling we are adding or reducing the position.

For me, once we have a position, `Long` and `Short` is more relatable, and not `buy` and `sell`. But essentially this is the same, so if we want to use just `Direction` throughout that's fine with me.

### DIFF
--- a/mobile/ios/Runner/bridge_generated.h
+++ b/mobile/ios/Runner/bridge_generated.h
@@ -27,7 +27,7 @@ WireSyncReturn wire_calculate_margin(double price, double quantity, double lever
 
 WireSyncReturn wire_calculate_quantity(double price, uint64_t margin, double leverage);
 
-WireSyncReturn wire_calculate_liquidation_price(double price, double leverage, int32_t position);
+WireSyncReturn wire_calculate_liquidation_price(double price, double leverage, int32_t direction);
 
 void free_WireSyncReturn(WireSyncReturn ptr);
 
--- a/mobile/lib/features/trade/domain/direction.dart
+++ b/mobile/lib/features/trade/domain/direction.dart
@@ -1,14 +1,14 @@
-import '../../../bridge_generated/bridge_definitions.dart';
+import '../../../bridge_generated/bridge_definitions.dart' as rust;
 
-enum Direction { buy, sell }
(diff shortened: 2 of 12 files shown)


bonomat (commented):
You asked for the 🪱, I open the 🗑️!

Personally I would just roll with `Long`/`Short` everywhere, and only make a difference in the label of the buttons shown to the user if deemed necessary. 
For me it works thinking of a `Long`-Order and a `Short`-Order. Which is equivalent of creating a `Buy`-Order or a `Sell`-Order. 

Long story short: I would merge your proposed `Direction` and `Position` into a single enum (`Direction` with `long`,`short` makes most sense to me) and add a rust doc if needed explaining how it used. 

This should result in less code and less conversion between them :)

da-kami:
> You asked for the 🪱, I open the 🗑️!
> 
> Personally I would just roll with `Long`/`Short` everywhere, and only make a difference in the label of the buttons shown to the user if deemed necessary. For me it works thinking of a `Long`-Order and a `Short`-Order. Which is equivalent of creating a `Buy`-Order or a `Sell`-Order.
> 
> Long story short: I would merge your proposed `Direction` and `Position` into a single enum (`Direction` with `long`,`short` makes most sense to me) and add a rust doc if needed explaining how it used.
> 
> This should result in less code and less conversion between them :)

Fair enough, as long as we are aligned! 
I'm happy to use `Direction` with `Long` and `Short` throughout then - I'll amend the PR!

bonomat (approved):
Nice

Issue #129, opened 2023-02-14
Handling Orders: Post/Get orders to/from backend + UI


da-kami:
Replace the placeholder code in `SubmitOrderChangeNotifier` with actually posting an order to Rust.

This includes:

- [x] API to create order
- [x] API to fetch order
- [x] Stream to get notifications that orders changes
- [x] Display of `Order` in UI

Make sure `Order` has a unique identifier.
There has to be a unique identifier that is assigned by the mobile's native Rust backend - additionally there *can* be a separate identifier that is assigned to the `Order` by the orderbook, but that is out of scope for this ticket. 
The order has to be identifiable without the `Orderbook` as well, because we might want to keep track of "failed" orders that never make it into the orderbook.

PR #143, opened 2023-02-15
Feat/order UI and api

da-kami:
resolves https://github.com/get10101/10101/issues/129

I ramped up the complete order model in Dart, the API and the business logic (BL). 

We have a roundtrip of:

- Dart subscribes to Order update notification
- submit `NewOrder`
- BL handles stuff (yet to be implemented) and send update notification back to Dart
- Upon notification Dart fetches the orders (currently always all of them; we can optimize that later).

There is not persistence yet - the `get_orders` API always returns the same dummy. Since we initially fetch that dummy you won't the update after creating an order unless you turn off the initial fetch in the `OrderChangeNotifier`.

Notes:

- Flutter Rust Bridge somehow has trouble with same struct names in different modules - so I gave the BL classes different names. This is quite annoying, but I have not found a workaround for this yet.

---

It is quite a log of boilerplate code - especially on the Dart side. But it is proper encapsulated now, and the boundaries are (quite) clear - all the points where we communicate with the Rust API are bundled in the `OrderService`; the Dart model is completely independent from the Rust API.


mobile/lib/features/trade/application/order_service.dart
```diff
@@ -7,6 +9,10 @@ import 'package:get_10101/features/trade/domain/direction.dart';
 import 'package:get_10101/features/trade/domain/leverage.dart';
 import 'package:get_10101/features/trade/domain/api_response.dart';
 
+import 'package:get_10101/features/trade/domain/order.dart';
+
+import '../order_change_notifier.dart';
```
holzeis, line 14: 🔧 can we stick to using the package import instead of the relative.. that makes it easier to move files.

mobile/lib/features/trade/domain/order.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:get_10101/features/trade/domain/contract_symbol.dart';
+import 'package:get_10101/features/trade/domain/leverage.dart';
+
+import 'package:get_10101/features/trade/domain/direction.dart';
+import '../../../bridge_generated/bridge_definitions.dart' as rust;
```
holzeis, line 5: 🤔 It looks inconsistent that we use the `rust` alias here as mostly we use the rust alias for the `ffi.dart` package. How about `bridge` as alias here?

Also can you change the path to a package import?

mobile/native/src/api_model/order/notifications.rs
```diff
@@ -0,0 +1,29 @@
+use crate::api_model::order::OrderNotification;
+use anyhow::anyhow;
+use anyhow::Result;
+use flutter_rust_bridge::support::lazy_static;
+use flutter_rust_bridge::StreamSink;
+pub use std::sync::Mutex;
+
+lazy_static! {
```
holzeis, line 8: 🔧 Wouldn't it be simpler to manage subscriptions in flutter? In the poc we had a single event listener in flutter that dispatched events to the corresponding subscribers.

Also I would prefer a single event stream instead of specific streams per event / notification type.

mobile/native/src/common/mod.rs
```diff
@@ -0,0 +1,14 @@
+use flutter_rust_bridge::frb;
+
+#[frb]
```
bonomat, line 3: Given this has `frb` on it, should this model only live in `api_model`?

mobile/native/src/trade/order.rs
```diff
@@ -0,0 +1,44 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum OrderTypeTrade {
+    Market,
+    Limit { price: f64 },
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum OrderStatus {
+    Open,
+    Filled,
```
bonomat: Can you add a comment explaining these states please?

mobile/lib/features/trade/application/order_service.dart
```diff
@@ -24,4 +30,30 @@ class OrderService {
...
+  Future<List<Order>> fetchOrders() async {
+    List<rust.Order> apiOrders = await rust.api.getOrders();
+    List<Order> orders = apiOrders
+        .map((order) => Order(
+            leverage: Leverage(order.leverage),
+            quantity: order.quantity,
+            contractSymbol: ContractSymbol.fromApi(order.contractSymbol),
+            direction: Direction.fromApi(order.direction),
+            status: OrderStatus.fromApi(order.status),
+            type: OrderType.market))
```
bonomat, line 43: Why are you not using `order.order_type`?

mobile/native/src/api_model/order/mod.rs
```diff
@@ -0,0 +1,111 @@
+use crate::common::ContractSymbol;
+use crate::common::Direction;
+use crate::trade;
+use flutter_rust_bridge::frb;
+
+pub mod notifications;
+
+#[frb]
```
bonomat, line 8: Do you think it makes sense if we can keep flutter related types isolated somehwere?

bonomat (approved):
I don't see any bigger issues. One thing to address might be @holzeis comment here: [`d242fa6` (#143)](https://github.com/get10101/10101/pull/143/commits/d242fa609c427818b699df65bda93ca01e0a96d6#r1108268192). Otherwise looks good.

mobile/lib/features/trade/application/order_service.dart
da-kami: It's annoying that the IDE keeps importing them like this - I tried to find a way to change the default but could not find a setting for it - did you manage to configure package imports as default?

mobile/lib/features/trade/submit_order_change_notifier.dart
```diff
@@ -16,16 +18,17 @@ class PendingOrder {
...
     notifyListeners();
 
-    // TODO: Actually submit the pending order...
-    await Future.delayed(const Duration(seconds: 1));
-    // TODO: Handle submit failure
+    await orderService.submitMarketOrder(
+        tradeValues.leverage, tradeValues.quantity, ContractSymbol.btcusd, tradeValues.direction);
     _pendingOrder!.state = PendingOrderState.submittedSuccessfully;
 
     notifyListeners();
```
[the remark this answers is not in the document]
da-kami: Yes, because the model changes - when we set the `PendingOrder` it is in status `pending` and once we have a reply from the api call we update the status.

First, we notify that a `PendingOrder` exists, then we notify about the status change after the call to the backend.
I an try to add another function to make it more clear :)

mobile/native/src/api.rs
```diff
@@ -30,3 +33,10 @@ pub fn calculate_liquidation_price(
         price, leverage, direction,
     ))
 }
+
+#[tokio::main(flavor = "current_thread")]
+pub async fn submit_order(order: NewOrder) -> Result<()> {
+    process_new_order(order.into()).await?;
```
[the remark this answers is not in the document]
da-kami: >  is there a reason why the api is called submit_order but the internal function is called process_new_order. Seems inconsistent.

It's from a time when I was not sure what exactly the order handler would do with the order - now that we basically submit the order to the orderbook in that call it makes sense to name it `submit_order` in the `OrderHandler` as well, I'll rename it!

mobile/lib/features/trade/domain/api_response.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:get_10101/features/trade/domain/response_status.dart';
```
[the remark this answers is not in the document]
da-kami: Fair enough - tbh I have not invested much thought into this model; not sure it's such a great idea to use it in that way. We could also stick to exceptions here instead of creating artificial types. 
I'll give it another thought

mobile/native/src/api_model/order/notifications.rs
da-kami: Why is a single event stream preferred? 
I deliberately used several separate ones for different concerns, otherwise we might encounter locks in cases where we want to send multiple events.

Note that in the poc we did not notify the UI about changes, but we managed all subscriptions in an event loop that periodically checked for data changes in the database.
We could do that again, but this can never result in a UX that is as good as notifying the UI after an update.
The backend does work, and the user interface should be notified of updates right after the work is done, that's why I used this pattern.

I find these single even streams for specific purposes a better design then the single event loop / one single event stream. Why would you prefer one stream over multiple? Do you have concerns that this will not scale, or we will run into resource issues?

da-kami: I do agree that the notification service could be encapsulated better and sending ntofications out in all kinds of places in the business logic is not optimal. But I would still like to achieve that the UI is notified when the work is done and not just on an interval.

mobile/native/src/trade/order.rs
da-kami: I added all the states we defined yesterday from #149 and rustdoc for them. Note that I introduced separate states for `Rejected` (orderbook rejects submission) and `Failed` (trade execution failed) in the state machine, so we can properly distinguish them. Note that not all internal states are exposed to the UI via the API.

I added the rustdoc on the business logic class, because that one is more exhaustive in states and represents the actual state machine.

mobile/native/src/common/mod.rs
da-kami: I did a cleanup round and removed the `common` module, not worth having, better encapsulate *all* the types from the get-go. If we feel the boileplate is not worth it we can change that again easily.

Note that `api_model` is nicely encapsulated, besides the notifications - the BL imports types from the API layer there which is not optimal (higher layer should know lower layer, not the other way round...). But I would merge this for now and iterate on it :)

mobile/lib/features/trade/submit_order_change_notifier.dart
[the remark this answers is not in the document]
da-kami: I opted for adding comments

mobile/lib/features/trade/domain/api_response.dart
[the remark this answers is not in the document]
da-kami: removed these classes in favor of exception handling in the change notifier

mobile/native/src/api_model/order/notifications.rs
da-kami: I'm merging this PR for now - but let's discuss this further!

Issue #145, opened 2023-02-15
Create lightning faucet


holzeis:
Our wallet onboarding path expects the user to transfer funds from another lightning wallet. However this won't be feasible for our tests (regtest, signet or testnet). Consequently we need some kind of lightning faucet that has an open channel with our coordinator.

The flow could be the following.

1. Tester creates an invoice in the app
2. Tester copies the decoded invoice string and navigates his browser to our public lightning faucet.
3. Tester enters pastes the invoice to the faucet and submits it.
4. The faucet automatically pays the invoice and the tester gets an channel with the coordinator.

PR #146, opened 2023-02-15
Hook ln dlc node into mobile and coordinator

### LINKED ISSUE #138: Hook ln-dlc-node into mobile app and coordinator
The aim is to get a working setup where a just-in-time payment can be executed from one phone to another over the coordinator.

holzeis:
This PR provides the first cut through for performing a channel-less payment. I've added a rudimentary open channel button - that needs to be replaced by the lightning faucet. #145 

just-in-time-channel-payment

resolves #138


mobile/lib/main.dart
```diff
@@ -136,8 +137,17 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
     try {
       await setupRustLogging();
 
-      setState(() {
-        FLog.info(text: "10101 is ready!");
+      final appSupportDir = await getApplicationSupportDirectory();
+      FLog.info(text: "App data will be stored in: $appSupportDir");
+
+      api.run(appDir: appSupportDir.path).listen((event) {
```
da-kami, line 155: 🙃 It might be a good idea to encapsulate these calls in a `service`.
For the `trade` feature I have an `order_service.dart` where the API calls are bundled; see https://github.com/get10101/10101/pull/143

Not sure it's worth it for these high-level calls, but it would make the dependencies more clear - otherwise it might be hard to manage these rust `api` calls over time, because they are sprinkled all over the code.

mobile/native/src/api_lndlc/lndlc.rs
```diff
@@ -47,6 +51,18 @@ pub fn run(stream: StreamSink<Event>, data_dir: String) -> Result<()> {
 
         let background_processor = node.start().await?;
 
+        // todo: should the library really be responsible for managing the task?
```
da-kami, line 82: ❓What are you referring to here? The fact that we use `keep_connected` and don't manage this here? Or the fact that we manage the connection here and not in Dart? 

My perspective on Rust vs. Dart:
Stateful presentation components such as Change notifiers to track value changes are OK; but managing the conneciton is on an application state level, and I would have seen that in the responsibility of the rust code.

mobile/lib/features/wallet/balance_change_notifier.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:flutter/material.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart';
+
+class BalanceChangeNotifier extends ChangeNotifier {
+  Balance balance = Balance(onChain: 0, offChain: 0);
```
da-kami, line 5: 🔧 I would avoid using API model code in a change notifier. I think this should be encapsulated in a `service`, and we should map it into a Dart class that handles the balance. That way we can also use the Dart model `Amount` for representing the balance.

mobile/lib/features/wallet/balance_change_notifier.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:flutter/material.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart';
+
+class BalanceChangeNotifier extends ChangeNotifier {
```
da-kami, line 4: Note: This will likely result in conflicts with https://github.com/get10101/10101/pull/99 but should not be too hard to resolve

mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -23,6 +24,16 @@ class ReceiveScreen extends StatelessWidget {
           const SizedBox(height: 50),
           SelectableText("Address: ${api.getNewAddress()}"),
           Text("Balance: ${balance.offChain} / ${balance.onChain}"),
+          ElevatedButton(
```
da-kami, line 34: Note: The commit that adds this as `Open Channel` button is outdated and I think it should be removed. 

The sentiment of `Create Invoice` is correct; that's in line with the flow where the channel gets automatically established upon initial payment.

mobile/lib/features/wallet/send_screen.dart
```diff
@@ -1,17 +1,46 @@
...
+          const SizedBox(height: 50),
+          TextFormField(onChanged: (text) {
+            setState(() {
+              encodedInvoice = text;
+            });
+          }),
+          ElevatedButton(
+              onPressed: () async {
+                try {
+                  await api.sendPayment(invoice: encodedInvoice);
```
da-kami, line 35: 🔧 Same here: I would like to avoid these API calls in presentation layer code and add a dedicated service.

da-kami (commented):
Great stuff!

I think it would be great to introduce `service`s for the API calls into Rust before merging; and a dedicated Dart model to be used in Dart. 
If you feel that's not worth it at this stage we can also merge it and iterate on it later - but the encapsulation will be different then https://github.com/get10101/10101/pull/143

crates/ln-dlc-node/src/lib.rs
```diff
@@ -28,7 +28,7 @@ mod ln;
 mod ln_dlc_wallet;
 mod logger;
 mod on_chain_wallet;
-mod seed;
+pub mod seed;
```
luckysori, line 31: As discussed, perhaps the `ln-dlc-node` library shouldn't mandate what kind of seed to use. We could just let consumers provide a seed for LDK and another for the on-chain wallet.

In practice, I don't think it matters too much because we will want to use the same BIP39 seed for app and coordinator.

data/coordinator/regtest/seed
```diff
@@ -0,0 +1 @@
+hؤ�SfNv��Z���
```
luckysori, line 1: :question: Do we want to commit this seed?

mobile/native/src/api.rs
```diff
@@ -30,3 +33,24 @@ pub fn calculate_liquidation_price(
         price, leverage, direction,
     ))
 }
+
+pub fn run(stream: StreamSink<Event>, app_dir: String) -> Result<()> {
+    anyhow::ensure!(!app_dir.is_empty(), "app_dir must not be empty");
```
luckysori, line 38: :question: What does this condition mean? The message is unclear to me.

mobile/native/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 mod api;
 pub mod api_calculations;
+pub mod api_lndlc;
 pub mod api_model;
```
luckysori, line 4: :wrench: This module structure and naming seems a bit weird. Why is there a private `api` module and public `api_*` modules. Why not sub-modules of `api` leading to a more natural naming and structure e.g. `api::calculations`.

Also, as a nitpick, I would use `ln_dlc`  or `ln_dlc_node` as opposed to `lndlc` to be consistent with the name of the crate.

mobile/native/src/logger.rs
```diff
@@ -19,7 +19,16 @@ static LOG_STREAM_SINK: Storage<StreamSink<LogEntry>> = Storage::new();
...
 // Tracing log directives config
 pub fn log_base_directives(env: EnvFilter, level: LevelFilter) -> Result<EnvFilter> {
-    let filter = env.add_directive(Directive::from(level));
+    let filter = env
+        .add_directive(Directive::from(level))
+        .add_directive("hyper=warn".parse()?)
+        .add_directive("sqlx=warn".parse()?) // sqlx logs all queries on INFO
+        .add_directive("reqwest=warn".parse()?)
+        .add_directive("rustls=warn".parse()?)
+        // set to debug to show ldk logs (they're also in logs.txt)
```
luckysori, line 28: :wrench: This comment seems out of place.

mobile/native/src/api_lndlc/mod.rs
```diff
@@ -0,0 +1,33 @@
+use anyhow::Result;
+use state::Storage;
+use tokio::runtime::Runtime;
+
+const ELECTRS_ORIGIN: &str = "tcp://localhost:50000";
+
+pub mod lndlc;
```
luckysori, line 7: :thinking: I don't get why there is an `lndlc` sub-module to the `api_lndlc` module. The separation doesn't seem to mean anything in this case.

luckysori (approved):
Thanks for working on this!

mobile/native/src/api.rs
holzeis: I removed the assertion - as not really needed.

data/coordinator/regtest/seed
holzeis: yes, otherwise we can not fix the coordinators node id for regtest.

coordinator/src/routes.rs
```diff
@@ -0,0 +1,60 @@
...
+pub struct Balance {
+    offchain: u64,
+    onchain: u64,
+}
+
+#[rocket::get("/get_balance")]
+pub async fn get_balance(node: &State<Arc<Node>>) -> Result<Json<Balance>, HttpApiProblem> {
+    let offchain = node.get_ldk_balance().map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Invalid public key")
```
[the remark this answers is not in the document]
holzeis: https://github.com/get10101/10101/pull/146/commits/60120fa9fc2d5075a9aab79d3dcfbbc102115d14

coordinator/src/routes.rs
```diff
@@ -0,0 +1,60 @@
...
+use http_api_problem::StatusCode;
+use ln_dlc_node::node::Node;
+use rocket::serde::json::Json;
+use rocket::serde::Deserialize;
+use rocket::serde::Serialize;
+use rocket::State;
+use std::sync::Arc;
+
+#[rocket::get("/get_fake_scid/<target_node>")]
+pub async fn get_fake_scid(
```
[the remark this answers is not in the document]
holzeis: https://github.com/get10101/10101/pull/146/commits/b94c67899e70af67dad03b37c86a1f993708abc2

holzeis:
> Great stuff!
> 
> I think it would be great to introduce `service`s for the API calls into Rust before merging; and a dedicated Dart model to be used in Dart. If you feel that's not worth it at this stage we can also merge it and iterate on it later - but the encapsulation will be different then #143

@da-kami thanks for your review. I am going to address it in a follow up PR.

da-kami:
> > Great stuff!
> > I think it would be great to introduce `service`s for the API calls into Rust before merging; and a dedicated Dart model to be used in Dart. If you feel that's not worth it at this stage we can also merge it and iterate on it later - but the encapsulation will be different then #143
> 
> @da-kami thanks for your review. I am going to address it in a follow up PR.

Did you already start with this? What is the status? 
The code is quite messy at the moment - no big deal, but it would be good to know what is the current status so I know how I should proceed. 

I assume you will action moving code into services; I will start to refactor having only one EvenStream as discussed on Friday, so I can go on with my work on the `position`.

Note: I noticed that you removed the `api_` prefix from `model` and `calculations`. I think we should make the boundary between the API and the business logic clearer - we don't have to do that with the prefix (I did not like that - I hard tried to create an `api` modules folder, but ran into trouble with flutter rust bridge); but at the moment the boundaries are not easy to spot.

Maybe we could opt to create a module `business_logic` to make it more explicit, because we have another model conversion there (the internal model is different than what is exposed on the API). Thoughts?

Issue #144, opened 2023-02-15
Just-in-time channels are not "useable"


holzeis:
**How to reproduce**

1. Start 3 nodes, alice, coordinator and bob.
2. Fund coordinator and bob with sufficient bitcoin.
3. Connect alice with the coordinator and bob with the coordinator.
4. Open a channel between bob and the coordinator. (ensure that the coordinator get inbound and outbound liquidity)
5. Create an interceptable invoice for alice.
6. Pay invoice with bob.
7. Create an interceptable invoice with bob (or a regular invoice, it shouldn't matter)
8. Pay invoice with alice.

**Expected outcome:**
- Alice is able to use the just-in-time channel created to receive the payment from bob to pay bob a smaller amount back.

**Observed outcome:**
```
Failed to find route err=LightningError { err: "Failed to find a path to the given destination", action: IgnoreError }
```

holzeis:
Attempts to fix that issue.

1. The just-in-time channel needs balanced inbound and outbound liquidity so I tried to double the channel capacity when processing the `HTLCIntercepted` event.

```
.create_channel(*target_node_id, expected_outbound_amount_msat * 2, 0, 0, None)
```
--> That is definitely required as otherwise the channel wouldn't have enough liquidity to send the money.
2. Mine additional blocks: Even mining more than 6 blocks did not help, however I noticed that the short_channel_id is never set (regardless of the amount of mined blocks)

holzeis:
If I log the channel I get the following output for the just-in-time channel created between the `coordinator` and `alice`

```
ln_dlc_node::tests::system_tests: alice-coordinator channel_id="0586bb5b87a32ae4244b14e047665d5fddcc2efc5b8f3ebdfa446cd54b9e6d93" short_channel_id=0 is_ready=true is_usable=true inbound=9895000000 outbound=0
```

- as already mentioned above the `short_channel_id` is not set, which I remember from the poc to be a problem. 
- there is no outbound liquidity although the channel capacity has been created with twice the amount of the `expected_outbound_amount_msat`. <- that seems to be the problem, but creating the channel with twice the amount did not help.

bonomat:
I think I got a bit closer: it might be a configuration problem due to unconfirmed channels?

- `alice` pay to `coordinator`: no problem
- `alice` pay to `bob`: problem

... unless we have 6+ confirmations on the channel opening from `bob-coordinator`-channel

So again, it looks like the route towards `bob` is not know...

bonomat:
[...]

PR #148, opened 2023-02-16
showcase: do not merge

bonomat:
#144 is no bug
The problem seems to be related to the channels not being confirmed enough.

By adding a few confirmations to the channel I was able to send a small amount back to bob from alice


bonomat:
Closing this as it only served as a showcase.

Issue #130, opened 2023-02-14
Handle Positions: Position updates in the UI and relationship to oder


da-kami:
Note: Position is automatically created in the Rust backend once an order was filled; the UI is notified in the positions stream that there is a new position.

- [ ] API to fetch position 
- [ ] Stream to get notifications for new position or position changes
- [ ] Display of Position in UI (excluding close logic - but `Close` button to be modeled as placeholder)

Note on data model:

The position is basically a view on the filled orders; so filled orders should be *aggregated* to represent the position + additional time-relevant data such as e.g. fees that incurred through rollovers.

This ticket does not have to tackle the data model in depth, because it does not include the representation of the position (and its relation to order) in the database; but in order to mocked UI coherent in state we will need some form of business logic that relates the position to the order. This should already be tackled in this ticket.

Likely, the data persistence model will become clearer once we tackle this.

da-kami:
Some interesting questions concerning the position:

- Who should be in charge of calculating the position values?
 If e.g. the price changes do we expect a position change event from the backend or do we just consume the price change and the PnL calc is triggered by Dart?

da-kami:
I had some more thoughts about the order lifecylce in the UI, I am writing this up here for now. I will see to reflect this in code once we have worked on the event model.

Relevant events for trade from Rust to UI:

- new (open) order
- order rejecte by orderbook - unclear if we should handle this case upon submission (response) or via event; see section below
- order filled (MVP is always complete fill, so one event should be good enough)
- order trade execution failed
- position opened
- position changed (a newly filled order has changed the position)
- position closed

---

## Initial success/failure when submitting order

Order rejection at submission into the orderbook can be handled with return value of request against the orderbook. In the UI this requires to lock the "success dialog" until the pending order is in either a `success` or `failure` state; otherwise the user might close the dialog and never receives information about the order failure.

- Advantage:
    - We don't have to deal with a failed order state for rejection
- Disadvantage:
    - We have to lock the dialog so that the user does not close it and misses information
    - We still have to depict a "failed" order in case the trade execution fails after the orderbook probides the trade parameters / the orderbook providing trade params times out.

It might be better to handle this as event as well, and make the dialog dependent on the event. This is not that easy to achieve in the user interface, because the events are detached from the `async` call the the dialog lifecycle is currently bound to - we would have to re-design the dialog lifecycle in the user interface.

da-kami:
This ticket is basically implemented with #164 
[...]

PR #149, opened 2023-02-16
From order to position

da-kami:
This sums up a pairing session with @bonomat to outline the communication with the orderbook and the coordinator in the app Rust code.

---

Notable decisions:

- The orderbook handles matching the order
    - When the taker (app) submits an order the orderbook checks:
         - can the order be filled? 
              - yes: return success
                  - The orderbook returns an identifier so the order can be identified in the orderbook 
              - no: return failure (MVP: results in cancellation of the order on the app side, the user has to "try again")
         - note: if success it returned the app can still decide to "remove" the order of no filling event is returned within `x` time.
    - The orderbook has an event stream that the app subscribes to, Events:
        - Latest order(s) / price - this is only for display purposes, no decision making based on it
        - `CompleteFillWith(TradeParams) - when the app receives this event it will trigger trade execution with the `coordinator` resulting in:
            - success: results in a DLC and either an update or creation of a position in the database. Order is update to `filled`
            - failure: Order is updated to `failed`
     
we opted for the stream setup because it detaches the execution from the submission - which will eventually allow us to e.g. cancel an order (...). This is the more sound model, changing this later might be cumbersome, so we decided to introduce it from the getgo.
 
---

Order lifecycle:

- `initial`: before being submitted into the orderbook; this is so we can track failed orders (we can decide to remove this at some point, it will be useful to gather this data initially)
- `open`: the orderbook returned success upon submitting the order
- `failed`: the orderbook returned failed upon submitting the order OR the trade execution (i.e. handing `CompleteFillWith` failed)
- `filled`: the order was completely filled (no partial filling in MVP)

---

Position lifecycle: 

tbd.


da-kami:
Closing this because it is outdated; the `TODO` comments might still be useful, but we don't need a Draft PR for that - and some decisions will likely change!

See https://github.com/get10101/10101/issues/130 for tracking

PR #150, opened 2023-02-16
Add test that clicks through the buy process until the order is submitted

da-kami:
I am not sure how valuable the test itself is - I will likely do another iteration on that.
The test passes and I think all the other work is valuable to go in; so I thought I'll open a PR now and potentially iterate once merged, so we have a "template" :)

This PR shows how to:

- use keys to identify widgets for testing
- use mockito to mock services
- how proper encapsulation is important for testing - with moving the rust API calcs into a dedicated service we are able to mock it

### DIFF
--- a/mobile/lib/util/constants.dart
+++ b/mobile/lib/util/constants.dart
@@ -1,0 +1,56 @@
+// This file contains global constants.
+// Note that global constants should only be use sparingly.
+// They make sense for e.g. keys to identify widgets for testing.
+// Hint: If you want to define a color it is recommended to use extend a Theme rather than adding a constant.
+
+import 'package:flutter/material.dart';
+
+// ############################################################################
+// Common keys
+
+const _tabs = "tabs/";
+const _button = "button/";
+const _slider = "slider/";
+
+// ############################################################################
+// Main keys
+
+const _wallet = "wallet/";
+const _trade = "trade/";
+
+// path on screen
+
+const _bottomSheet = "bottom_sheet/";
+const _confirmSheet = "confirm/";
+
+// concrete selectors
+
+const _buy = "buy";
+const _sell = "sell";
+const _positions = "positions";
+const _orders = "orders";
+
+const tradeScreenTabsOrders = Key(_trade + _tabs + _orders);
+const tradeScreenTabsPositions = Key(_trade + _tabs + _positions);
+
+const tradeScreenButtonBuy = Key(_trade + _button + _buy);
+const tradeScreenButtonSell = Key(_trade + _button + _sell);
+
+const tradeScreenBottomSheetTabsBuy = Key(_trade + _bottomSheet + _tabs + _buy);
+const tradeScreenBottomSheetTabsSell = Key(_trade + _bottomSheet + _tabs + _sell);
+
+const tradeScreenBottomSheetButtonBuy = Key(_trade + _bottomSheet + _button + _buy);
+const tradeScreenBottomSheetButtonSell = Key(_trade + _bottomSheet + _button + _sell);
+
+const tradeScreenBottomSheetConfirmationSliderBuy =
+    Key(_trade + _bottomSheet + _confirmSheet + _slider + _buy);
+const tradeScreenBottomSheetConfirmationSliderSell =
+    Key(_trade + _bottomSheet + _confirmSheet + _slider + _sell);
+
+const tradeScreenBottomSheetConfirmationSliderButtonBuy =
+    Key(_trade + _bottomSheet + _confirmSheet + _slider + _button + _buy);
+const tradeScreenBottomSheetConfirmationSliderButtonSell =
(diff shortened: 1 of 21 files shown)


mobile/lib/util/constants.dart
```diff
@@ -0,0 +1,56 @@
...
+
+const _wallet = "wallet/";
+const _trade = "trade/";
+
+// path on screen
+
+const _bottomSheet = "bottom_sheet/";
+const _confirmSheet = "confirm/";
+
+// concrete selectors
```
bonomat, line 26: Are we going to have a constant for every single clickable item?

mobile/test/trade_test.mocks.dart
```diff
@@ -0,0 +1,208 @@
+// Mocks generated by Mockito 5.3.2 from annotations
+// in get_10101/test/trade_test.dart.
+// Do not manually edit this file.
```
bonomat, line 3: If this file is generated, should it be in git then?

mobile/test/trade_test.dart
```diff
@@ -0,0 +1,114 @@
...
+
+    expect(find.byKey(tradeScreenBottomSheetConfirmationSliderButtonBuy), findsOneWidget);
+
+    // TODO: This is not optimal because if we re-style the component this test will likely break.
+    // Drag to confirm
+    await tester.timedDrag(find.byKey(tradeScreenBottomSheetConfirmationSliderButtonBuy),
+        const Offset(275, 0), const Duration(seconds: 2),
+        pointer: 7);
+
+    verify(orderService.submitMarketOrder(any, any, any, any)).called(1);
```
bonomat, line 112: Should we verify that `submitMarketOrder` was submitted with the correct values?

mobile/test/trade_test.dart
```diff
@@ -0,0 +1,114 @@
...
+          // Need the trade theme otherwise the trade widgets that rely on it can't find it on the context and fail to render
+          TradeTheme(),
+        ],
+      ),
+    );
+  }
+}
+
+void main() {
+  testWidgets('Buy order flow until submission', (tester) async {
```
bonomat, line 58: Is this the approach to call tests like this in flatter?

I personally would like to have the naming simular to what we do in rust, e.g. `when user fills out order submission form, orderservice gets invoced with correct values.`

bonomat (approved):
Looks good. 2 points: 

- I think we can iterate on the test itself later if you want
- it would be good to execute these tests in CI aswell.

mobile/test/trade_test.mocks.dart
da-kami: I think it depends on where we want to run these tests - here's a [stackupverflow discussion](https://stackoverflow.com/questions/67927401/should-we-source-control-the-mocks-dart-files-created-by-build-runner-for-mocki) where the recommendation is `No`.

If we want to run these tests e.g. on CI I would check the files in, because another `build` step to generate them would add time to CI runs.
If we only run these tests locally then 

So far these widget tests are not really long-running, but we were not planning to have extensive CI runs with them.

If we only use these tests locally or in dedicated test runs we can ensure to run the build step ` flutter pub run build_runner build` by ourselves or as part of the task.
So, for now I'm happy to add them to the ignore file - we just have to be aware that an additional, manual step might be necessary to run the tests locally.

da-kami:
> it would be good to execute these tests in CI aswell.

If we want to do that; should I add a build step that generates the mocks or should we version control the mocks? 😅

mobile/test/trade_test.dart
da-kami: That's a horrible name, I will amend it using the `given` `when` `then` pattern :)

mobile/test/trade_test.dart
da-kami: I had that in mind at some point, but I would probably test that in a different test because it's not the concern here.

This test basically clicks through the flow to ensure that the models pop up correctly and we trigger submitting the order after confirmation.
I would see a smaller unit test in charge of checking on the actual values, unless we fear that the values might be wrong in the end because of something happening in the flow.

mobile/lib/util/constants.dart
da-kami: I would make it a rule to identify widgets that we use in tests with a specific key.
I would not add them to "everything" but add them as needed when we write tests.

mobile/test/trade_test.mocks.dart
da-kami: I opted for ignoring it and automatic re-generation with `just flutter-test`

mobile/test/trade_test.dart
da-kami: > Looks good. 2 points:
> 
>     * I think we can iterate on the test itself later if you want

sweet

>     * it would be good to execute these tests in CI aswell.

Let's give this a try. If it's not as trivial as I think it is I will post here and do a follow up PR :)

da-kami: image

easy :)

Issue #126, opened 2023-02-13
Add DLC channel next to Lightning channel


luckysori:
We want to be able to use `rust-dlc` to "upgrade" an already-open Lightning to have a DLC channel next to it.

This task involves writing a passing test in `ln_dlc_node`.

PR #152, opened 2023-02-20
Support creation of DLC channel

luckysori:
Fixes https://github.com/get10101/10101/issues/126.

This currently single-patch PR includes too many things that I have not managed to split into smaller patches. I can spend some time doing so if the review is too hard.

Namely:

- Add new test to prove that we can open a DLC channel.
- Start refactoring the "production" code in `ln-dlc-node`.
- Rework all the tests in `ln-dlc-node`. The objective was to simplify and better understand why things work. This led to fixing a couple of bugs, and it made the tests faster and easier to read.
- Remove `system_tests` feature in favour of `#[ignore]` attributes.

The DLC functionality still needs to be tested more e.g. can we make a payment after opening a DLC channel.


crates/ln-dlc-node/src/seed.rs
```diff
@@ -8,17 +8,30 @@ use bip39::Language;
...
     pub fn new() -> Result<Self> {
-        let mut rng = rand::thread_rng();
-        let mnemonic = Mnemonic::generate_in_with(&mut rng, Language::English, 12)?;
+        let word_count = 12;
+
+        let entropy_bytes = (word_count / 3) * 4;
+        let mut entropy = [0u8; (MAX_NB_WORDS / 3) * 4];
+        thread_rng().fill_bytes(&mut entropy[0..entropy_bytes]);
+
+        let mnemonic = Mnemonic::from_entropy_in(Language::English, &entropy[0..entropy_bytes])?;
```
bonomat, line 34: Can you explain this please?

crates/ln-dlc-node/src/tests/add_dlc.rs
```diff
@@ -1,4 +1,98 @@
...
+use crate::tests::wait_until;
+use anyhow::anyhow;
+use bitcoin::Amount;
+use dlc_manager::sub_channel_manager::SubChannelState;
+use dlc_manager::Oracle;
+use dlc_manager::Storage;
+use std::time::Duration;
+
+#[tokio::test]
+#[ignore]
```
bonomat, line 13: Is this being ignored on purpose?

crates/ln-dlc-node/src/tests/just_in_time_channel.rs
```diff
@@ -0,0 +1,162 @@
...
+    // `max_inbound_htlc_value_in_flight_percent_of_channel`
+    // configuration value for said channel.
+    //
+    // At this point in time, we fix the channel size to
+    // `JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT` (10_000 sats) and
+    // the default value of
+    // `max_inbound_htlc_value_in_flight_percent_of_channel` is set to
+    // 10 percent. This means that we won't succeed in routing the
+    // payment through the just-in-time channel if the amount is
+    // greater than 1_000 sats.
```
bonomat, line 73: Love it 😅

crates/ln-dlc-node/src/node.rs
```diff
@@ -679,9 +681,107 @@ impl Node {
...
+            .accept_sub_channel(channel_id)
+            .map_err(|e| anyhow!(e.to_string()))?;
+
+        self.dlc_message_handler
+            .send_subchannel_message(node_id, SubChannelMessage::Accept(accept_sub_channel));
+
+        Ok(())
+    }
+
+    pub fn process_incoming_messages(&self) -> Result<()> {
```
bonomat, line 734: Does this have to be a manual step?

crates/ln-dlc-node/src/tests/mod.rs
```diff
@@ -1,6 +1,368 @@
...
+                {
+                    break details.clone();
+                }
+
+                tracing::debug!(
+                    %peer,
+                    temp_channel_id = %hex::encode(temp_channel_id),
+                    "Waiting for channel to be usable"
+                );
+                tokio::time::sleep(Duration::from_millis(500)).await;
```
bonomat, line 163: I'm not sure I understand why this work. Is a channel not only usable once we have X confirmation? And this function doe snot create blocks, so why does it work?

crates/ln-dlc-node/src/node.rs
```diff
@@ -679,9 +681,107 @@ impl Node {
             .insert(intercept_scid, target_node);
         intercept_scid
     }
+
+    pub async fn initiate_dlc_channel(
```
bonomat, line 685: Should we rename this to something like `propose_dlc_channel`? If I understand it correctly then the other party needs to manually accept this step.

crates/ln-dlc-node/src/tests/mod.rs
```diff
@@ -1,6 +1,368 @@
...
+    oracle_pk: XOnlyPublicKey,
+) -> ContractInput {
+    let total_collateral = offer_collateral + accept_collateral;
 
-#[cfg(feature = "system_tests")]
-mod system_tests;
+    ContractInput {
+        offer_collateral,
+        accept_collateral,
+        maturity_time: 1653378780,
```
bonomat, line 296: Should we dynamically compute this value to ensure it is always in the future?

crates/ln-dlc-node/src/tests/mod.rs
```diff
@@ -1,6 +1,368 @@
...
+                        PayoutFunctionPiece::PolynomialPayoutCurvePiece(
+                            PolynomialPayoutCurvePiece::new(vec![
+                                PayoutPoint {
+                                    event_outcome: 50_000,
+                                    outcome_payout: 0,
+                                    extra_precision: 0,
+                                },
+                                PayoutPoint {
+                                    event_outcome: 60_000,
+                                    outcome_payout: total_collateral,
```
bonomat, line 326: Is this code somehow creating the curve?

bonomat (approved):
Very nice clean-ups for the tests :)
Nothing obvious, just a few comments.

crates/ln-dlc-node/src/seed.rs
holzeis: We copied this peace of code over from bip39 as the bip39 dependency uses an outdated rand lib. This was the fastest way to bypass the conflicting rand dependency. @luckysori wanted to look into creating our own fork of bip39 fixing that dependency issue the right way.

crates/ln-dlc-node/src/node.rs
holzeis: i fear so yes. have a look at our discussion with tobi. 

https://matrix.to/#/!YOUZSBmuJuaGzmxzaA:matrix.org/$AEqhGweYmktqLYTCUBR8ZT13coxfOJzcpYCd-pFe2MU?via=matrix.org

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.8"
+version: "3.7"
```
holzeis, line 1: 🤔 why the downgrade?

holzeis (approved):
👍

crates/ln-dlc-node/src/node.rs
```diff
@@ -561,9 +561,11 @@ impl Node {
             .private_route(RouteHint(vec![RouteHintHop {
                 src_node_id: hop_before_me,
                 short_channel_id: intercepted_channel_id,
+                // QUESTION: What happens if these differ with the actual values
```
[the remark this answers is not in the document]
luckysori: Why let you pass these values here if the `ChannelConfig` effectively overrules them?

crates/ln-dlc-node/src/seed.rs
luckysori: I've updated this now so that we don't have to inline this code.

crates/ln-dlc-node/src/tests/add_dlc.rs
luckysori: I think the right approach for these tests is to:

- Ignore them by default (at least until we mock the dependencies).
- Only run them (`cargo test -p ln-dlc-node -- --ignored`) on CI if `ln-dlc-node` (and/or other dependencies) actually change.

Were we actually running them on CI before this PR?

crates/ln-dlc-node/src/tests/mod.rs
luckysori: I believe that all our configs support 0-conf channels at the time of writing this code:

```rust
UserConfig {
    // ...
    // we want to accept 0-conf channels from the coordinator
    manually_accept_inbound_channels: true,
    // ...
}
```

I think that should not be the case in general, so a nice thing to do here might be to mine as many blocks as the channel config says and then start looping.

luckysori: I've updated the documentation and added a TODO.

crates/ln-dlc-node/src/node.rs
luckysori: Yep!

crates/ln-dlc-node/src/tests/mod.rs
luckysori: It doesn't seem to matter right now, but I've set it to `now + 24 hours` just in case it starts mattering.

crates/ln-dlc-node/src/tests/mod.rs
luckysori: I think this code _is_ the piecewise polynomial curve. Later, `rust-dlc` uses this to interpolate the values for intermediate payouts.

One thing we need to figure out is how to configure this to generate a set number of payouts. Because atm the test generates 7.5k CETs.

docker-compose.yml
luckysori: Because it didn't work for me locally and I was lazy:

```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
```

I'll try to fix this.

luckysori: I have fixed this now! Turns out `docker compose != docker-compose`.

crates/ln-dlc-node/src/tests/add_dlc.rs
luckysori: > Were we actually running them on CI before this PR?

I've confirmed that we were _not_ running them on CI before this PR.

Nonetheless, I've created an issue that I will tackle right after merging this PR: #157.

luckysori: Ended up doing it with this PR.

crates/ln-dlc-node/src/seed.rs
bonomat: Thanks for the explanation

crates/ln-dlc-node/src/tests/add_dlc.rs
bonomat: I initially solved it with a feature flag. What's the reason why you didn't go with this approach?

crates/ln-dlc-node/src/node.rs
bonomat: > ... tobi 

--> Tibo ;)

crates/ln-dlc-node/src/tests/mod.rs
bonomat: > One thing we need to figure out is how to configure this to generate a set number of payouts. Because atm the test generates 7.5k CETs.

ufff...

crates/ln-dlc-node/src/tests/add_dlc.rs
luckysori: I didn't like the feature flag because the tests aren't even _built_ with `cargo clippy --all-targets`.

I think using the `#[ignore]` attribute to ignore tests is much more predictable too.

PR #153, opened 2023-02-21
Fix/app startup

da-kami:
I was not able to run the app without errors, these are some quick-fixes. More permission config might be needed for mobile.

The seconds commit is to retry the connection to the coordinator instead of failing in `run`.
Eventually we might want to communicate errors back to the user instead of just logging... but for now this should do?

### DIFF
--- a/crates/ln-dlc-node/src/node.rs
+++ b/crates/ln-dlc-node/src/node.rs
@@ -439,13 +439,22 @@ impl Node {
 
     // todo: That might be better placed in a dedicated connection manager file.
     pub async fn keep_connected(&self, peer: NodeInfo) -> Result<()> {
-        let connection_closed_future = Self::connect(self.peer_manager.clone(), peer).await?;
-
         let peer_manager = self.peer_manager.clone();
         tokio::spawn({
             async move {
-                let mut connection_closed_future = connection_closed_future;
+                // initial connection retry
+                let mut connection_closed_future = loop {
+                    match Self::connect(peer_manager.clone(), peer).await {
+                        Ok(fut) => break fut,
+                        Err(e) => {
+                            tracing::error!(%peer, "Establishing connection failed because {e:#}");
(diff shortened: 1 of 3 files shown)


crates/ln-dlc-node/src/node.rs
```diff
@@ -439,13 +439,22 @@ impl Node {
...
         let peer_manager = self.peer_manager.clone();
         tokio::spawn({
             async move {
-                let mut connection_closed_future = connection_closed_future;
+                // initial connection retry
+                let mut connection_closed_future = loop {
+                    match Self::connect(peer_manager.clone(), peer).await {
+                        Ok(fut) => break fut,
+                        Err(e) => {
+                            tracing::error!(%peer, "Establishing connection failed because {e:#}");
```
da-kami, line 450: Note: I somehow don't see this error in the logs in flutter, do we not see the logs from tasks?

da-kami: Not true, this was just because I did not run `just gen` 😅

mobile/macos/Runner/DebugProfile.entitlements
```diff
@@ -8,5 +8,7 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+	<key>com.apple.security.network.client</key>
+    <true/>
```
bonomat, line 12: The formatting seams to be broken here.

bonomat (approved):
👍

holzeis (approved):
LGTM 👍 

We might want to separate the connection stuff from this PR and open a dedicated PR for it.

da-kami:
> LGTM 👍
> 
> We might want to separate the connection stuff from this PR and open a dedicated PR for it.

Do we want to do another iteration on the connection handling now? If not I would merge this and we can do another iteration later.

PR #154, opened 2023-02-21
Refactor/event service

da-kami:
@holzeis opening this for keeping track of the work


da-kami:
CC @holzeis 

This event service is quite a timesink. I hope we are not over-engineering this 😅

Some discussion will be needed:

- The generic `Subscriber` would limit us to only be able to subscribe to one event. I tried to implement multiple of it (i.e. `implement Subscriber<A>, Subscriber<B>`) - but while one can implement multiple interface one cannot implement the same one twice, and the generic does not make any difference. 
Hence, I don't think this generic Subscriber makes sense - I removed it. You find more reasoning in the commits: https://github.com/get10101/10101/pull/154
- I added explicit overwrites for `PartialEq` and `Hash` for `Event` in rust and tests that shows that it works.
- The Dart side was more tricky - the only way to actually properly derive the runtime `Type` I found is to pass the constructed event in (i.e. a "dummy" event so we can properly derive the type from it). This works - and it's only "ugly" upon subscription. Probably acceptable for now.

da-kami:
the `trade_test` is actually not concerned by these changes because we don't deal with the notification lifecycle there - the API for submitting the order did not change.

@holzeis that means we don't have to adapt the test for this work. 

Note that it would still be good to encapsulate the layers properly in some places, notably:

should:

-  Rust: we use the same `Event` internally and on the API at the moment - I think for the internal `EventHub` we should use an `Event` that is *not* exposed on the API, and the `FlutterSubscriber` should then map into the `Event` that is exposed.

could:
 
- Dart: We could consider introducing an `InitService` or similar, that handles the initialization to encapsulate the bridge code better; similarly it would be great if we can achieve that the `EventService` would map the incoming data to the internal types. But we cannot easily achieve that if we want to keep the `EventService` as generic as it is. But that means that we leak the bridge definitions into the `ChangeNotifier`. We can accept that for now 🤷‍♂️

holzeis (approved):
Great stuff! Definitely an improvement!

I am not sure about the `api_model` structure, but we can address that in a follow up PR if we want to.

PR #155, opened 2023-02-21
Orderbook foundation

bonomat:
This little project makes use of

- actix: so that we can use websockets
- diesel: we wanted to use diesel over sqlx due to performance issue and other annoyances
- postgres: we want to use postgres for our services to be able to use db as a service
- docker: we need to run a postgres db
- testcointainers: we also need to run a db for our tests, hence, we use our beloved testcontainers for this

addresses: https://github.com/get10101/10101/issues/156


holzeis:
@bonomat Are now going for a separate order book service after all? or is this intended to get merged with the coordinator?

bonomat:
> @bonomat Are now going for a separate order book service after all? or is this intended to get merged with the coordinator?

I would merge it with the coordinator once we are happy with it. I think it's so nicely decopelled that we can work in it in a separate binary :).

bonomat:
@da-kami / @Restioson / @luckysori :if you want you can have already a look. 

I don't expect an indepth review. Afaik it's the first time we are using actix and diesel, so there might be a few things which we can do better. Also error handling has not been taken care yet. 

I plan to add websockets tomorrow then it's ready for full review.

Restioson:
Have not used actix-web in a long time but I have used it for a web socket app in the past. I think nowadays people also like to use warp and axum which also handle websockets. Back then I switched off of actix-web because of the unsoundness in actix's codebase but I believe the new maintainers have addressed that now.

orderbook/src/bin/orderbook.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::sync::atomic::AtomicUsize;
+use std::sync::Arc;
+use tracing::level_filters::LevelFilter;
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
+    orderbook::logger::init_tracing(LevelFilter::DEBUG, false).unwrap();
+
+    // set up database connection pool
+    let conn_spec = "postgres://postgres:mysecretpassword@localhost:5432/orderbook".to_string();
```
Restioson, line 27: "mysecretpassword" should contain a special character, uppercase, and number - try "My$3cretpassword" :wink:

orderbook/src/bin/orderbook.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // start price feed server actor
+    let ws_server = websocket::PriceFeed::new(app_state.clone(), pool.clone()).start();
+
+    // Start HTTP server
+    HttpServer::new(move || {
+        App::new()
+            // set up DB pool to be used with web::Data<Pool> extractor
+            .app_data(web::Data::new(pool.clone()))
+            .wrap(middleware::Logger::default())
```
Restioson, line 50: I wonder if there is any tracing middleware for actix :thinking: if we want to go that route again, we should consider ahead of time how much detail we want to include in the traces so we avoid explosive disk usage

orderbook/src/routes/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+        let res = HttpResponse::NotFound().body(format!("No order found with id: {order_uid}"));
+        Ok(res)
+    }
+}
+
+/// Get all orders
+#[get("/orders")]
+pub async fn get_orders(pool: web::Data<DbPool>) -> Result<HttpResponse, Error> {
+    // use web::block to offload blocking Diesel code without blocking server thread
```
Restioson, line 76: This is something we need to be very careful of in general; in the past I've gone as far as only allowing access to the DB through a `database` model which wraps every DB call in `async` first. Otherwise maybe we can lint or something. It is just very easy to forget one DB call on the reactor thread and then get weird performance issues elsewhere (as seen in ItchySats)

orderbook/src/routes/session.rs
```diff
@@ -0,0 +1,137 @@
...
+        // future within context, but context waits until this future resolves
+        // before processing any other events.
+        // HttpContext::state() is instance of SessionState, state is shared
+        // across all routes within application
+        let addr = ctx.address();
+        self.addr
+            .send(websocket::Connect {
+                addr: addr.recipient(),
+            })
+            .into_actor(self)
```
Restioson, line 75: Does actix still not allow Handlers and lifecycle methods to be async blocks? :grimacing:

Restioson (commented):
Looks good so far! :rocket:

orderbook/src/bin/orderbook.rs
bonomat: 😅

orderbook/src/bin/orderbook.rs
bonomat: Looks like there is a lib for it. 
I agree, better look into this sooner than later :)

orderbook/src/routes/session.rs
bonomat: Does not look like it.

1010Tom:
@Restioson : I've reimplented it using Axum and it feels much faster :)

1010Tom:
@luckysori / @Restioson  / @da-kami : if we are happy with this approach then I would merge it into `coordinator`.

da-kami (approved):
Looks good, happy to roll with `axum` - don't have too much experience with different web frameworks, but it looks good and is developed by the tokio team :)

da-kami:
Note: On the `coordintator` side we are using rocket at the moment. I don't think it is efficient to use two different web frameworks so I would adapt that if we decide to use `axum` - CC @holzeis

bonomat:
I'll merge this for now, then go ahead and change the coordinator to use `axum` and then migrate the orderbook to the coordinator :)

bonomat:
@1010Tom : I'll take over

orderbook/src/routes/session.rs
Restioson: Very sad

Issue #44, opened 2023-01-25
re add flutter lint


bonomat:
This commit (https://github.com/get10101/10101/pull/20/commits/12e25a2cce4563f31f409b8d727c85654d7c78b3) removes flutter lint from CI because it was failing and I was not able to fix it in this PR.

We want to lint flutter though, so let's re add it.

bonomat:
@da-kami : this might be a good idea to add?

PR #158, opened 2023-02-21
Chore/flutter lint

da-kami:
resolves https://github.com/get10101/10101/issues/44


da-kami:
hm... interesting that the last commit makes `lint-flutter` be skipped - we have changes in files that should trigger it being executed. I will see to move this commit to a separate PR, because it's just an optimization.

bonomat (approved):
👍 nice fix. Thanks for looking into that.

PR #160, opened 2023-02-22
Display the order according to it's state

da-kami:
- Some beautifying
- Display the execution price if it is set

image


da-kami:
For a nicer display of double values (i.e. no `.0` display) I created a separate ticket: https://github.com/get10101/10101/issues/161

da-kami:
Expandable Order tracked here: https://github.com/get10101/10101/issues/163

bonomat (approved):
Looking really nice already :)

PR #162, opened 2023-02-22
Only run `lint-flutter` if flutter files changed

da-kami:
Separate PR, still needs testing if this logic actually works; let's see how it behaves once the branch that this is based on merges and it's getting rebased onto master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,27 @@ jobs:
         run: just gen
       - run: cargo clippy --all-targets --all-features -- -D warnings
 
+  lint-flutter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: extractions/setup-just@v1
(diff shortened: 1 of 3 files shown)


da-kami:
Given that this adds complexity to our CI setup that we currently might not want/need I close this for now and we can come back to it later. If we feel CI is slow we can optimize! 

@luckysori has removed the `change` step for now and introduced a separate action (in separate file) for long running tasks. That might be a better pattern to use!

luckysori:
> @luckysori has removed the `change` step for now and introduced a separate action (in separate file) for long running tasks. That might be a better pattern to use!

I only removed it because it was unused on `main` (I added it in 2 places and only used it in one :facepalm:). I'm not against using the same job in several files! Although it will need to be duplicated because afaik it's not easy (possible?) to reuse code between files.

bonomat:
Reusing code is not easy but you can reuse the whole job.

PR #164, opened 2023-02-22
Position user interface and simulation dummies in rust

da-kami:
Position user interface + wired into rust with rust simulating an update with dummies (upon an event from the orderbook).

This does not include closing the position yet.

---

This shows the change of the position after 10 seconds, simulating the lifecycle to some extent:

1. We fetch the position initially 
2. We subscribe to notifications for the position handler from the orderbook event in `run`
3. We subscribe to notifications for position updates for the flutter event subscriber
4. We simulate an orderbook event after 10 seconds
5. We "handle" the event in the position handler and send out a different dummy position
6. The position is updated in the UI :)

note: I don't know why the color green is broken in the gif 😅 

---

This is how the close button is currently implemented; we can iterate on that if it's too hidden:

### DIFF
--- a/mobile/lib/features/trade/domain/position.dart
+++ b/mobile/lib/features/trade/domain/position.dart
@@ -2,12 +2,22 @@ import 'package:get_10101/features/trade/domain/contract_symbol.dart';
 import 'package:get_10101/features/trade/domain/direction.dart';
 import 'package:get_10101/features/trade/domain/leverage.dart';
 import 'package:get_10101/common/domain/model.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
 
 enum PositionState {
   open,
 
   /// once the user pressed button to close position the button should be disabled otherwise the user can click it multiple times which would result in multiple orders and an open position in the other direction
-  closing
+  closing;
+
+  static PositionState fromApi(bridge.PositionState positionState) {
+    switch (positionState) {
+      case bridge.PositionState.Open:
+        return PositionState.open;
+      case bridge.PositionState.Closing:
+        return PositionState.closing;
+    }
+  }
 }
 
 class Position {
@@ -17,7 +27,7 @@ class Position {
   final Direction direction;
   final double averageEntryPrice;
   final double liquidationPrice;
-  final Amount unrealizedPnL;
+  final Amount unrealizedPnl;
   final PositionState positionState;
 
   Position(
@@ -28,5 +38,29 @@ class Position {
       required this.contractSymbol,
       required this.direction,
       required this.positionState,
-      required this.unrealizedPnL});
+      required this.unrealizedPnl});
+
+  static Position fromApi(bridge.Position position) {
+    return Position(
+        leverage: Leverage(position.leverage),
+        quantity: position.quantity,
+        contractSymbol: ContractSymbol.fromApi(position.contractSymbol),
+        direction: Direction.fromApi(position.direction),
+        positionState: PositionState.fromApi(position.positionState),
+        averageEntryPrice: position.averageEntryPrice,
+        liquidationPrice: position.liquidationPrice,
+        unrealizedPnl: Amount(position.unrealizedPnl));
+  }
+
+  static bridge.Position apiDummy() {
+    return bridge.Position(
+        leverage: 0,
+        quantity: 0,
+        contractSymbol: bridge.ContractSymbol.BtcUsd,
+        direction: bridge.Direction.Long,
+        positionState: bridge.PositionState.Open,
+        averageEntryPrice: 0,
+        liquidationPrice: 0,
+        unrealizedPnl: 0);
+  }
 }
--- a/mobile/lib/features/trade/domain/position.dart
+++ b/mobile/lib/features/trade/domain/position.dart
@@ -1,0 +1,32 @@
+import 'package:get_10101/features/trade/domain/contract_symbol.dart';
+import 'package:get_10101/features/trade/domain/direction.dart';
+import 'package:get_10101/features/trade/domain/leverage.dart';
+import 'package:get_10101/common/domain/model.dart';
+
+enum PositionState {
+  open,
+
+  /// once the user pressed button to close position the button should be disabled otherwise the user can click it multiple times which would result in multiple orders and an open position in the other direction
+  closing
+}
+
+class Position {
+  final Leverage leverage;
+  final double quantity;
+  final ContractSymbol contractSymbol;
+  final Direction direction;
+  final double averageEntryPrice;
+  final double liquidationPrice;
+  final Amount unrealizedPnL;
+  final PositionState positionState;
+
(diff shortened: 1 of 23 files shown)


mobile/lib/features/trade/domain/position.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'package:get_10101/features/trade/domain/direction.dart';
+import 'package:get_10101/features/trade/domain/leverage.dart';
+import 'package:get_10101/common/domain/model.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
+
+enum PositionState {
+  open,
+
+  /// once the user pressed button to close position the button should be disabled otherwise the user can click it multiple times which would result in multiple orders and an open position in the other direction
+  closing;
```
Restioson, line 11: May be worth copying the comment in the rust model to explain why there is no "closed" variant. Also may consider a different name as "PositionState" seems to imply the state of a position,which seems to include closed ones, as opposed really just representing manipulable positions / non-closed positions :man_shrugging:  Maybe I am still just thinking in the itchysats way, though

mobile/lib/features/trade/position_change_notifier.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import 'domain/position.dart';
+
+class PositionChangeNotifier extends ChangeNotifier implements Subscriber {
+  HashMap<ContractSymbol, Position> positions = HashMap();
+
+  Future<void> _create(PositionService positionService) async {
+    List<Position> positions = await positionService.fetchPositions();
+    for (Position position in positions) {
+      this.positions[position.contractSymbol] = position;
```
Restioson, line 17: Is it worth checking that there are no duplicate positioned returned by the position service? Maybe just in debug. Unless this is actually allowed?

mobile/native/src/api_model/position.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#[frb]
+#[derive(Debug, Clone)]
+pub struct Position {
+    pub leverage: f64,
+    pub quantity: f64,
+    pub contract_symbol: ContractSymbol,
+    pub direction: Direction,
+    pub average_entry_price: f64,
```
Restioson, line 36: What units are these numbers in? I assume unrealised pnl is in sats, but how about the prices here? Are these the price of BTC in the trading pair? If so, does this model still work for quanto CFDs like ETH/USD? I see no obvious reason aside from that there is no BTC side anymore so which side is which becomes arbitrary.

Restioson (commented):
LGTM, just a few queries :)

mobile/lib/features/trade/domain/position.dart
da-kami: For me this specific comment contains an implementation detail for the UI - we are talking about the behavior of a button here. 

There is rustdoc in the rust model (in `api_model` - we could move the comments to the internal model because that represents the state machine in more detail).
I will see to harmonize that in a follow up, i.e. always describe the state machine in the feature model and reference it from the api model.

```rust
pub enum PositionState {
    /// The position is open
    ///
    /// Open in the sense, that there is an active position that is being rolled-over.
    /// Note that a "closed" position does not exist, but is just removed.
    /// During the process of getting closed (after creating the counter-order that will wipe out
    /// the position), the position is in state "Closing".
    ///
    /// Transitions:
    /// Open->Closing
    Open,
    /// The position is in the process of being closed
    ///
    /// The user has created an order that will wipe out the position.
    /// Once this order has been filled the "closed" the position is not shown in the user
    /// interface, so we don't have a "closed" state because no position data will be provided to
    /// the user interface.
    Closing,
}
```

mobile/lib/features/trade/position_change_notifier.dart
da-kami: I would not bother for now - I would see such a check more on the database level. If we implement the data model right we should not have this problem. 
Checking on the consumer side for sanity can get out of hand quite fast.

One thing that's a bit neglected at the moment is tests - I would see most unit tests on the DB and business logic layers though; so far it is mostly UI and glue code.

mobile/native/src/api_model/position.rs
da-kami: > I assume unrealised pnl is in sats

It can also be represented as percentage

> Are these the price of BTC in the trading pair? 

The BTC/USD price to be precise - we could also price our CFDs in e.g. Euro - but that's a smaller market.

> If so, does this model still work for quanto CFDs like ETH/USD?

I am not sure I can perfectly explain the relationship between the price and the margin (in BTC) in a quanto product - but in a nutshell you take bother assets price based on the common price denominator (USD) and the rest is calculation. But that is not really relevant for the `average_entry_price`.

The average entry price would be the ETH/USD price. This average entry price is just an accumulation of prices when the position was opened / updated by orders, so it's an aggregate of the filled order prices that defined the position.

CC @bonomat mayb have a quick read over this just in case, maybe I'm missing something ;)

bonomat: What you said makes sense. 

- the `average_entry_price` is in USD (as long as the contracts are against USD, e.g. ETH/USD, XMR/USD, TESLA/USD). We have an average because there might be multiple trades to make this position.
- `unrealized_pnl` can be in sats or in percent. @da-kami : it might make sense to specify the type in the name, e.g. `unrealized_pnl_sats`. Then it is clear what it is. One can then easily be calculated from the other.

mobile/native/src/trade/position/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    ///
+    /// The user has created an order that will wipe out the position.
+    /// Once this order has been filled the "closed" the position is not shown in the user
+    /// interface, so we don't have a "closed" state because no position data will be provided to
+    /// the user interface.
+    Closing,
+}
+
+#[derive(Debug, Clone)]
+pub struct PositionTrade {
```
bonomat, line 28: We might want to add an identifier for the position in question. This is relevant once we have multiple positions (e.g. ETH/USD, BTC/USD).

bonomat (approved):
I think this is good to go :)

mobile/native/src/trade/position/mod.rs
da-kami: Depending on how we want to handle the lifecycle the `ContractSymbol` might be enough.

For one `ContractSymbol` there can only be one positions at any time.

At least from the UI perspective a `Position` is only displayed as long as it is open (with the caveat of also displaying it in out `closing` state, but that's basically still `open` - we just track that there is an order that wants to close it in the state).

Internally we *could* give each position an identifier and also represent a `closed` state. I'm not sure about that yet and would decide once we introduce the DAL.

1010Tom: > For one contract symbol there can only be one positions at any time.

Ah right, way easier :) 👍

PR #165, opened 2023-02-22
Allow `ln-dlc-node` tests to be skipped on `bors`

luckysori:
`Bors` treats skipped required jobs as _failed_. We want to be able to selectively skip certain tests in `ln-dlc-node` because they're slow.

We define a new job whose only purpose is to report a job success to `bors` whether the `run-tests-ln-dlc-node` job is skipped or succeeds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,25 +35,6 @@ jobs:
       - name: Disallow squash! commits
         run: git log --pretty=format:%s origin/main..HEAD | grep -zv squash!
 
-  # job to run change detection
-  changes:
-    runs-on: ubuntu-latest
-    # Set job outputs to values from filter step
-    outputs:
-      ln-dlc-node: ${{ steps.filter.outputs.ln-dlc-node }}
-    steps:
-      - uses: actions/checkout@v3
-      - uses: dorny/paths-filter@v2
-        id: filter
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          filters: |
-            ln-dlc-node:
-              - '.github/workflows/**'
-              - 'crates/bdk-ldk/**'
-              - 'crates/ln-dlc-node/**'
-              - 'Cargo.*'
-
   clippy:
     runs-on: ubuntu-latest
     needs: formatting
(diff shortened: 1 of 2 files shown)


luckysori:
Gonna bors this to see if it works.

bors r+

luckysori:
bors retry

luckysori:
bors ping

luckysori:
bors ping

luckysori:
bors retry

da-kami:
Have you considered using a pattern like the one we used in the POC?
We checked against *not containing* `failure` for the results of the jobs as far as I understand:

https://github.com/get10101/10101-PoC/blob/main/.github/workflows/ci.yml#L152-L166

da-kami:
I think the problem is that the name of the outputs is not aligned with the name used for checking the result.

The name of the output in `changes` is defined as `ln-dlc-node` - I think this is what we have to use as name for checking on the outputs in `tests-ln-dlc-node`.

I'll add a fixup, let;s see if that fixes it :)

da-kami:
OK, no point in adding a fixup here :D
I force pushed it with the changes - will try to bors it

da-kami:
I think the `outputs.result` might also be wrong in this context - it might just be `.result` here, because we select the output by name already. If it fails that's the next thing I would change :)

da-kami:
bors retry

da-kami:
> I think the problem is that the name of the outputs is not aligned with the name used for checking the result.
> 
> The name of the output in `changes` is defined as `ln-dlc-node` - I think this is what we have to use as name for checking on the outputs in `tests-ln-dlc-node`.
> 
> I'll add a fixup, let;s see if that fixes it :)

OK, this is bogus, what was there before is correct - because we are in the context the `needs` of `tests-ln-dlc-node`.

It is surprising that the version that was there did not work; one difference is that `changes` is not part of the needs of `tests-ln-dlc-node` - maybe that makes a difference? I would not see how though.

da-kami (approved):

da-kami:
bors retry

da-kami:
self approval is the best!

PR #167, opened 2023-02-23
Refactor/move api model in feature modules

da-kami:
Based on a discussion with @holzeis last week we agreed that it would feel more ergonomic to move the `api` tyoes of the respective feature into an `api` module of the feature instead of having `api_model` as a module that replicates the feature structure internally. 

This is a pure refactor; no logic changes should be in there. I mostly used the IDE to refactor->move things so I hope this is clean :)

Note: I stacked this on top of https://github.com/get10101/10101/pull/164 because otherwise it would be a pain to rebase that branch later. We should get *both* PRs in fast, please also review https://github.com/get10101/10101/pull/164 when reviewing this!


Restioson (approved):
LGTM, I think this organisation makes sense

bonomat (approved):
👍

holzeis (approved):
Much better 👍 Thank you!

PR #173, opened 2023-02-24
Add system integration test for onboarding

### LINKED ISSUE #170: Add system integration test for onboarding from external node
Wtih https://github.com/get10101/10101/pull/169 we introduced a lightning faucet, that is also integrating an lnd node to onboard a 10101 mobile user to the lightning wallet.

This is useful from two perspectives.

1. It enables our testers to get to funds easily, without manipulating the actual flow.
2. It ensures that we stick compatible with lnd.

Both capabilities are critical and should be protected from regressions.

holzeis:
Adds a system test for onboarding the 10101 lightning wallet from an external lnd node.

This will also help us to ensure that we stay compatible with lnd in the future.

I've replaced the chopsticks faucet with our proxy faucet as we will also need to reach lnd and it doesn't make sense to use different faucets for that. However, we can't simply remove chopsticks yet, as it also initialises the regtest network with a wallet and blocks.

resolves #170


crates/ln-dlc-node/src/tests/bitcoind.rs
```diff
@@ -0,0 +1,56 @@
...
+        let response =
+            query(r#"{"jsonrpc": "1.0", "method": "getnewaddress", "params": []}"#.to_string())
+                .await?;
+        let response: BitcoindResponse = response.json().await.unwrap();
+
+        query(format!(
+            r#"{{"jsonrpc": "1.0", "method": "generatetoaddress", "params": [1, "{}"]}}"#,
+            response.result
+        ))
+        .await?;
```
bonomat, line 35: What's the response of this? If it is the TX-id then we can get rid of the sleep below and return the tx ids. The caller can then do a wait-loop and check if the wallet learned about the tx-id.

bonomat (approved):
Nice

crates/ln-dlc-node/src/tests/bitcoind.rs
holzeis: It's the block hash.

https://developer.bitcoin.org/reference/rpc/generatetoaddress.html

holzeis:
bors retry

holzeis:
I guess for this test to run on github actions we need to address the `macaroons.db`. It shouldn't be mounted directly for two reasons.

1. It will make your workspace dirty everytime you run `docker-compose up`
2. Github Actions seems to mount the file in a special way, which will results into the error shown below.

```
lnd           | 2023-02-27 09:28:09.993 [INF] CHDB: Swapping old DB file from /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db to /root/.lnd/data/chain/bitcoin/regtest/macaroons.db
lnd           | 2023-02-27 09:28:09.993 [ERR] LTND: Shutting down because error in main method: unable to open databases: unable to obtain database backends: error opening macaroon DB: rename /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db /root/.lnd/data/chain/bitcoin/regtest/macaroons.db: device or resource busy
lnd           | unable to open databases: unable to obtain database backends: error opening macaroon DB: rename /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db /root/.lnd/data/chain/bitcoin/regtest/macaroons.db: device or resource busy
```

bonomat:
> I guess for this test to run on github actions we need to address the `macaroons.db`. It shouldn't be mounted directly for two reasons.
> 
>     1. It will make your workspace dirty everytime you run `docker-compose up`
> 
>     2. Github Actions seems to mount the file in a special way, which will results into the error shown below.
> 
> 
> ```
> lnd           | 2023-02-27 09:28:09.993 [INF] CHDB: Swapping old DB file from /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db to /root/.lnd/data/chain/bitcoin/regtest/macaroons.db
> lnd           | 2023-02-27 09:28:09.993 [ERR] LTND: Shutting down because error in main method: unable to open databases: unable to obtain database backends: error opening macaroon DB: rename /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db /root/.lnd/data/chain/bitcoin/regtest/macaroons.db: device or resource busy
> lnd           | unable to open databases: unable to obtain database backends: error opening macaroon DB: rename /root/.lnd/data/chain/bitcoin/regtest/temp-dont-use.db /root/.lnd/data/chain/bitcoin/regtest/macaroons.db: device or resource busy
> ```

@holzeis : I've fixed it by removing the macaroons. See the commit messages :)

bonomat:
@luckysori / @da-kami : would be cool if one of you could give this a try locally so that we can get it merged :)

da-kami:
The added test passes for me locally with:

1. Fresh docker container pull of the complete environment after removing everything.
2. Calling `curl -d '{"address":"bcrt1qylgu6ffkp3p0m8tw8kp4tt2dmdh755f4r5dq7s", "amount":"0.1"}' -H "Content-Type: application/json" -X POST http://localhost:3000/faucet` after 10 secs wait
3. Running the test.

Note that the mulit-hop test failed:

```
test tests::dlc_collaborative_settlement::given_dlc_channel_present_when_dlc_settled_collaboratively_then_sibling_channel_operational ... ok
test tests::dlc_non_collaborative_settlement::given_dlc_channel_present_when_dlc_settled_non_collaboratively_then_sibling_channel_operational ... ok
test tests::single_hop_payment::single_hop_payment ... ok
test tests::multi_hop_payment::multi_hop_payment ... FAILED
test tests::just_in_time_channel::just_in_time_channel ... ok
test tests::add_dlc::given_lightning_channel_then_can_add_dlc_channel ... ok
test tests::onboard_from_lnd::onboard_from_lnd has been running for over 60 seconds
test tests::onboard_from_lnd::onboard_from_lnd ... ok
```

I can provide the logs, but I think this failure is unrelated to this PR.

bonomat:
@da-kami : did you also try to use the faucet manually?

da-kami (approved):
> I ran into a problem where the fee estimate using the public server was way too high and I could not open a channel.
> Hence, we hardcode a fee estimate file and serve it using our ngnix server. with this we have a reliable fee estimation of 1sat/vbyte

Note: I ran into this problem as well when running the test against a setup that had been running for a while. When using a fresh setup I did not encounter this problem. Not sure if there was actually a correlation though - because I only saw this error once.

da-kami:
> @da-kami : did you also try to use the faucet manually?

I did the three steps I listed https://github.com/get10101/10101/pull/173#issuecomment-1449322383 - and I think step two is calling the faucet manually. Or are you referring to something else? 

I tried to replicate what the ci action does locally.

bonomat:
Right, then my question doesn't matter :) 

bors r+

bonomat:
bors retry

1010Tom:
I've fixed the startup of the containers. Looks like `-d` was removed by accident. 

bors retry

PR #176, opened 2023-02-27
chore: replace rocket with axum

bonomat:
this is a preparation for adding the orderbook

### DIFF
--- a/coordinator/Cargo.toml
+++ b/coordinator/Cargo.toml
@@ -11,10 +11,10 @@ bdk = { version = "0.24.0", features = ["key-value-db"] }
 bitcoin = "0.29"
 clap = { version = "4", features = ["derive"] }
 dlc-manager = { version = "0.3.0", features = ["use-serde"] }
-http-api-problem = { version = "0.55.0", features = ["rocket"] }
 ln-dlc-node = { path = "../crates/ln-dlc-node" }
 rand = "0.8.5"
 serde = "1.0.147"
+serde_json = "1"
 time = { version = "0.3", features = ["serde", "parsing", "std", "formatting", "macros", "serde-well-known"] }
 tokio = { version = "1", features = ["full"] }
 tracing = "0.1.37"
--- a/coordinator/Cargo.toml
+++ b/coordinator/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 atty = "0.2.14"
+axum = "0.6.8"
 bdk = { version = "0.24.0", features = ["key-value-db"] }
 bitcoin = "0.29"
 clap = { version = "4", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


coordinator/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 anyhow = { version = "1", features = ["backtrace"] }
 atty = "0.2.14"
+axum = "0.6.8"
```
luckysori, line 9: ❓ Can we remove `rocket` then?

coordinator/src/main.rs
```diff
@@ -68,24 +69,16 @@ async fn main() -> Result<()> {
...
-        )
-        .manage(node)
-        .launch()
-        .await?;
+    let addr = SocketAddr::from((http_address.ip(), http_address.port()));
+    tracing::debug!("listening on http://{}", addr);
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
```
luckysori, line 79:
```suggestion
        .await?;
```

coordinator/src/main.rs
```diff
@@ -68,24 +69,16 @@ async fn main() -> Result<()> {
...
 
-    tracing::trace!(?mission_success, "Rocket has landed");
+    // tracing::trace!(?mission_success, "Rocket has landed");
```
luckysori, line 81: :wrench: Can we add a similar log message without the rocket analogy?

luckysori (approved):
Let's give `axum` a go!

holzeis (approved):
I think you mentioned it on the orderbook PR that you want to replace `rocket` as it does not support websockets. Is that the only reason or are there other advantages of `axium` over `rocket`?

Regarding `websockets`. In itchysats we used `Server Sent Event (SSE)`, because we did not need a bidirectional communication with the client. Do we need that in 10101?

https://blog.logrocket.com/server-sent-events-vs-websockets/

bonomat:
> I think you mentioned it on the orderbook PR that you want to replace `rocket` as it does not support websockets. Is that the only reason or are there other advantages of `axum` over `rocket`?
> 

Yes, websocket is the main reason but also it seems like `axum` has also a better performance as it is just a thin layer on top of `hyper` while `rocket` is a bit _bloated_. 


> Regarding `websockets`. In itchysats we used `Server Sent Event (SSE)`, because we did not need a bidirectional communication with the client. Do we need that in 10101?
> 

SSE caused a lot of pain in the past because of _seemingly_ dropping connections and the missing client support. 
I think `bidirectional` will be useful in the future when we want to support multi-match orders, i.e. a market order which gets executed by multiple times. 

Fun fact: https://github.com/flosse/rust-web-framework-comparison states rocket as outdated framework :D

coordinator/Cargo.toml
bonomat: definitely

PR #183, opened 2023-02-28
Revert "chore: Remove open channel from api"

holzeis:
@da-kami: I've reverted the commit, removing the open channel button.

I had to change one additional configuration, being that the coordinator must not enforce his channel announcement preferences (public inbound channels only), since the app only accepts private channels.

This will now allow us again to open a private channel from the app - I hope we will only work with that temporarily, as this does not reflect the real flow.

This reverts commit 3e2fac2f

### DIFF
--- a/crates/ln-dlc-node/src/node.rs
+++ b/crates/ln-dlc-node/src/node.rs
@@ -169,7 +169,15 @@ impl Node {
         seed: Bip39Seed,
         ephemeral_randomness: [u8; 32],
     ) -> Self {
-        let user_config = coordinator_config();
+        let mut user_config = coordinator_config();
+
+        // TODO: The config `force_announced_channel_preference` has been temporarily disabled
+        // for testing purposes, as otherwise the app is not able to open a channel to the
+        // coordinator. Remove this config, once not needed anymore.
+        user_config
+            .channel_handshake_limits
+            .force_announced_channel_preference = false;
+
         Self::new(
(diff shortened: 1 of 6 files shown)


da-kami (approved):

crates/ln-dlc-node/src/node.rs
```diff
@@ -818,9 +850,14 @@ pub(crate) fn coordinator_config() -> UserConfig {
...
-            // only want to open private channels to our 10101 app.
-            force_announced_channel_preference: true,
+
+            // TODO: The config `force_announced_channel_preference` has been temporarily disabled
+            // for testing purposes, as otherwise the app is not able to open a channel to the
+            // coordinator. Uncomment belows comment once `force_announced_channel_preference` is
+            // set back to `true`
+            // // Enforces incoming channels to the coordinator to be public! We
+            // // only want to open private channels to our 10101 app.
+            force_announced_channel_preference: false,
```
luckysori, line 860: 🙃 I would prefer to instead use `Node::new` in the `coordinator` and pass in a mutated version of the `coordinator_config` with the `force_announced_channel_preference` set to false.

luckysori (approved):
Thanks for tackling this!

crates/ln-dlc-node/src/node.rs
holzeis: agreed.

PR #185, opened 2023-02-28
chore: Use constant for min final cltv expiry

holzeis:
LDK will reject any final (incoming) HTLC if the available `final_cltv_expiry` smaller than 23. This is the reason why LDK has a `MIN_FINAL_CLTV_EXPIRY` constant set to 24. 20 was probably working, because of some buffers added by lnd or ldk. Not entirely sure about that. However, using the LDK defined constant for that seems to be the reasonable choice.


bonomat (approved):
Interesting, thanks for documenting it.

luckysori (approved):

PR #186, opened 2023-02-28
chore: Read channel data from file system upon restart

holzeis:
The ln_dlc node did save channel data to the disk, but never recovered any channel date from disk. This PR fixes that issues and will allow the node to restart without loosing its channel data.

More specifically this PR will fix the following error (which will be printed by the coordinator or app, if a channel was opened before a restart).
```
ERROR lightning::ln::channelmanager: Failed to find corresponding channel
```

Additionally, the network graph is also read from disk. Eventually, all this data should be stored into and read from the database.

### DIFF
--- a/crates/ln-dlc-node/src/disk.rs
+++ b/crates/ln-dlc-node/src/disk.rs
@@ -1,5 +1,6 @@
 use crate::NetworkGraph;
 use crate::TracingLogger;
+use bitcoin::BlockHash;
 use lightning::routing::scoring::ProbabilisticScorer;
 use lightning::routing::scoring::ProbabilisticScoringParameters;
 use lightning::util::ser::ReadableArgs;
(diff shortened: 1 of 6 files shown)


crates/ln-dlc-node/src/node.rs
```diff
@@ -259,21 +267,93 @@ impl Node {
...
+                    keys_manager.clone(),
+                    ln_dlc_wallet.clone(),
+                    chain_monitor.clone(),
+                    ln_dlc_wallet.clone(),
+                    logger.clone(),
+                    ldk_user_config,
+                    channel_monitor_mut_references,
+                );
+                <(BlockHash, ChannelManager)>::read(&mut f, read_args)
+                    .map_err(|e| anyhow::anyhow!("{e:?}"))
```
bonomat: Given you unwrap here I don't think it is useful to map the error to anyhow.

crates/ln-dlc-node/src/disk.rs
```diff
@@ -22,3 +23,16 @@ pub(crate) fn read_scorer(
...
+
+pub(crate) fn read_network(
+    path: &Path,
+    genesis_hash: BlockHash,
+    logger: Arc<TracingLogger>,
+) -> NetworkGraph {
+    if let Ok(file) = File::open(path) {
+        if let Ok(graph) = NetworkGraph::read(&mut BufReader::new(file), logger.clone()) {
+            return graph;
+        }
```
bonomat, line 35: Should we log the error at least?

bonomat (approved):
Nice

crates/ln-dlc-node/src/node.rs
holzeis: this needs to be done as the trait bound `lightning::ln::msgs::DecodeError: std::error::Error` is not satisfied. Mapping the error to an anyhow error allows me to use the `?`. Any suggestions on how this can be done better?

crates/ln-dlc-node/src/disk.rs
holzeis: yes, I think the network graph is also updated through the gossip protocol.

crates/ln-dlc-node/src/node.rs
bonomat: It was only because you unwraped right aftewards. In a later commit you changed it to `?`, so this comment is actually outdated :)

PR #190, opened 2023-03-01
fix: Replace hardcoded available balance with lightning wallet info

holzeis:


da-kami (approved):

da-kami:
The `TradeScreen` widget now has a dependency on the `WalletInfoChangeNotifier`; so we have to add this in the test. We should make it a mocked ChangeNotifier there similar to the other mocked ones.

PR #192, opened 2023-03-01
Make `ln-dlc-node` tests a bit more resilient

luckysori:
This was some work that I've been collecting as I've been trying to verify that the LN channel is usable after a DLC channel has been opened beside it.

### DIFF
--- a/crates/ln-dlc-node/src/tests/just_in_time_channel.rs
+++ b/crates/ln-dlc-node/src/tests/just_in_time_channel.rs
@@ -27,12 +27,12 @@ async fn just_in_time_channel() {
     coordinator.fund(Amount::from_sat(100_000)).await.unwrap();
 
     let coordinator_outbound_liquidity_sat = 25_000;
-    let payer_inbound_liquidity_sat = 25_000;
+    let payer_outbound_liquidity_sat = 25_000;
     let payer_coordinator_channel_details = coordinator
         .open_channel(
             &payer,
             coordinator_outbound_liquidity_sat,
-            payer_inbound_liquidity_sat,
+            payer_outbound_liquidity_sat,
         )
         .await
         .unwrap();
--- a/crates/ln-dlc-node/src/tests/just_in_time_channel.rs
+++ b/crates/ln-dlc-node/src/tests/just_in_time_channel.rs
@@ -6,7 +6,6 @@ use anyhow::Context;
 use anyhow::Result;
 use bitcoin::Amount;
 use rust_decimal::Decimal;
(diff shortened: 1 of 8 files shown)


crates/ln-dlc-node/src/tests/just_in_time_channel.rs
```diff
@@ -22,17 +22,17 @@ async fn just_in_time_channel() {
...
+    let coordinator_outbound_liquidity_sat =
+        min_outbound_liquidity_channel_creator(&payer, payer_outbound_liquidity_sat);
+
     let payer_coordinator_channel_details = coordinator
         .open_channel(
-            &payer.info,
+            &payer,
             coordinator_outbound_liquidity_sat,
-            payer_inbound_liquidity_sat,
+            payer_outbound_liquidity_sat,
```
holzeis, line 35: 🙃 depends on the point of view 😜

crates/ln-dlc-node/src/tests/mod.rs
```diff
@@ -192,6 +204,40 @@ async fn fund_and_mine(address: Address, amount: Amount) -> Result<()> {
...
 
+/// Calculate the "minimum" acceptable value for the outbound liquidity
+/// of the channel creator.
+///
+/// The value calculated is not guaranteed to be the exact minimum,
+/// but it should be close enough.
+///
+/// This is useful when the channel creator wants to push as many
+/// coins as possible to their peer on channel creation.
+fn min_outbound_liquidity_channel_creator(peer: &Node, peer_balance: u64) -> u64 {
```
holzeis, line 215: :+1: very good.. I think we will need something like this as a validation function on the frontend as well. I tested to open a position with a very small amount and run into the reserve issue.

holzeis (approved):
Nice!

crates/ln-dlc-node/src/tests/just_in_time_channel.rs
luckysori: The thing is the coordinator's outbound liquidity is the payer's inbound liquidity, so we can't use both variables together.

crates/ln-dlc-node/src/tests/mod.rs
luckysori: That is true, we can apply a similar logic when opening any channel.

What channel were you thinking off specifically. I suppose the coordinator-app channel.

holzeis: I was referring to the dlc channel.

PR #193, opened 2023-03-01
feat: Calculate cfd fees

holzeis:
This adds the correct fees to the trading confirmation screen. However, these fees should probably be already shown on the trade screen. As you these fees are not that small.

This change uses the rust-dlc implementation for its calculation as this will implicitly be used when we are proposing a sub channel.

### DIFF
--- a/mobile/native/src/calculations/mod.rs
+++ b/mobile/native/src/calculations/mod.rs
@@ -1,4 +1,5 @@
 use crate::common::api::Direction;
+use dlc::channel::sub_channel::LN_GLUE_TX_WEIGHT;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 use trade::cfd;
@@ -30,4 +31,14 @@ pub fn calculate_liquidation_price(price: f64, leverage: f64, direction: Directi
     tracing::trace!("Liquidation_price: {liquidation_price}");
 
     liquidation_price
+}
+
+pub fn calculate_fees() -> u64 {
+    // TODO: this should probably come from a configuration.
+    let fee_rate = 2;
+    let total_fee = (dlc::channel::sub_channel::dlc_channel_and_split_fee(fee_rate)
+        + dlc::util::weight_to_fee(LN_GLUE_TX_WEIGHT, fee_rate)) as f64;
+
+    // the party that offers the trade will pay the extra sat in case the fees are uneven.
+    (total_fee / 2.0).ceil() as u64
 }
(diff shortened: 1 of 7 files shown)


mobile/native/src/calculations/mod.rs
```diff
@@ -31,3 +32,13 @@ pub fn calculate_liquidation_price(price: f64, leverage: f64, direction: Directi
 
     liquidation_price
 }
+
+pub fn calculate_fees() -> u64 {
```
luckysori, line 36: :wrench: I think the transaction fees should be calculated by the ln-dl-node crate.

luckysori (commented):
I would imagine that the transaction fee is only one of the few things that need to be calculated to compute the total fee for opening a position.

This PR addresses the tx fee (at least one of them), but I think the removed TODO was referring to other kinds of fees that the app's trader has to take into account.

bonomat:
> This PR addresses the tx fee (at least one of them), but I think the removed TODO was referring to other kinds of fees that the app's trader has to take into account.

Agree, the fees here should be the fees when opening a position which is the maker/taker fee

mobile/native/src/calculations/mod.rs
```diff
@@ -31,3 +32,13 @@ pub fn calculate_liquidation_price(price: f64, leverage: f64, direction: Directi
...
+    // TODO: this should probably come from a configuration.
+    let fee_rate = 2;
+    let total_fee = (dlc::channel::sub_channel::dlc_channel_and_split_fee(fee_rate)
+        + dlc::util::weight_to_fee(LN_GLUE_TX_WEIGHT, fee_rate)) as f64;
+
+    // the party that offers the trade will pay the extra sat in case the fees are uneven.
+    (total_fee / 2.0).ceil() as u64
```
bonomat, line 43: If I understand this correctly than this is the _worst case_ fee which only applies if we force close the channel.

holzeis: yes, so in the best-case (or regular) scenario, these fees will never be paid. So I guess it would make sense to depict them somehow differently.

It's basically an additional collateral that we have to lock, in order to securely resolve the dlc channel in case the counterparty is trying to cheat.

This could even help the trader to understand that this is self-custodial.

holzeis:
> I would imagine that the transaction fee is only one of the few things that need to be calculated to compute the total fee for opening a position.
> 
> This PR addresses the tx fee (at least one of them), but I think the removed TODO was referring to other kinds of fees that the app's trader has to take into account.

You are absolutely right, and it's probably not even a fee that you'll have to pay. We just need to lock this amount as well, in case the counterparty tries to cheat and we need to publish the split / glue transaction, etc.

holzeis:
Closing this PR as it is outdated and will have to be addressed in the trading fees breakdown ticket. see also https://github.com/get10101/10101/issues/224

PR #194, opened 2023-03-01
fix: Switch roles in dlc channel

holzeis:
I am unsure if this is intended to stay that way, but if the mobile app is proposing the dlc channel, the mobile app is offering and the coordinator is accepting. Correspondingly, these values are currently switched.

This change fixes that issue.

### DIFF
--- a/crates/ln-dlc-node/src/node.rs
+++ b/crates/ln-dlc-node/src/node.rs
@@ -1056,8 +1056,8 @@ impl Node {
         // The contract input to be used for setting up the trade between the trader and the
         // coordinator
         let contract_input = ContractInput {
-            offer_collateral: margin_coordinator,
-            accept_collateral: margin_trader,
+            offer_collateral: margin_trader,
+            accept_collateral: margin_coordinator,
             maturity_time: maturity_time as u32,
             fee_rate: 2,
             contract_infos: vec![ContractInputInfo {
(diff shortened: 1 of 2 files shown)


crates/ln-dlc-node/src/node.rs
```diff
@@ -1056,8 +1056,8 @@ impl Node {
         // The contract input to be used for setting up the trade between the trader and the
         // coordinator
         let contract_input = ContractInput {
-            offer_collateral: margin_coordinator,
-            accept_collateral: margin_trader,
+            offer_collateral: margin_trader,
+            accept_collateral: margin_coordinator,
```
luckysori, line 1060: 🤔 I'm confused about this `trade` API being in this crate in the first place. I think this should be defined by the app crate (if such a thing exists).

luckysori (commented):
I agree that this fixes something, but the code shouldn't be there imo.

bonomat (approved):

crates/ln-dlc-node/src/node.rs
da-kami: To clarify: You mean this should not be in the `ln-dlc-node` but it should be in the `coordinator`, correct? If that's what you mean then I agree :)

da-kami (approved):
Given that the taker and maker are the ones that trigger offering the contract and the coordinator accepts this reasoning makes sense.

PR #196, opened 2023-03-01
chore: Refactor node.rs into smaller builder files

holzeis:
An attempt to split the node creation into separate builder files. Still not perfect, but a little bit better?


mobile/pubspec.lock
```diff
@@ -950,5 +950,5 @@ packages:
     source: hosted
     version: "2.0.3"
 sdks:
-  dart: ">=2.19.0 <4.0.0"
```
holzeis, line 953: I don't know why this is changing all the time from 4.0.0 to 3.0.0 and vice versa. 😅 Could it be related to our locally installed flutter version?

luckysori: I also saw this! It might be my local version, yes. I'll check if I can do something about it.

luckysori: Hmmm, I think this is related to the `pubspec.yaml`.

https://github.com/get10101/10101/blob/57f6085f8cda650ce14fa23348bee3aecc1862ea/mobile/pubspec.yaml#L7

Every time I run `flutter pub get` it reverts back to what is specified there. Not sure how it doesn't happen for others.

luckysori: The `stable` channel for the Dart SDK is on version 2.19.3, so I think it makes sense to keep the `pubspec.yaml` as is. 

The lockfile should respect that, so I encourage the other devs to run `flutter pub get` in `/mobile` and report back. Clearly it's not working as I would expect for some of us.

cc @holzeis @bonomat @da-kami @Restioson

Restioson: Maybe related: the first time that I run just deps per boot it says that my flutter is out of date

luckysori: Umm, it could be. Although it would be strange that an outdated version of Flutter would lead to accepting an even newer version of the Dart SDK.

luckysori (approved):
I approve since we've collaborated on this and agreed that the outcome is a clear improvement over `main`.

PR #200, opened 2023-03-02
feat: introduce database for mobile app 

### LINKED ISSUE #26: Database for mobile
In the past (for itchysats and 10101-poc) we used [sqlx](https://github.com/launchbadge/sqlx) which we found more than painful to work with as it had performance issues. 
This time we want to go with a different implementation, e.g. [diesel](https://diesel.rs/)

This ticket covers a simple setup for a database layer.

bonomat:
We make use of sqlite+diesel for the mobile app. This allows us to write higher order queries.
Note the file changes in
- mobile/ios/Runner.xcodeproj, and
- mobile/native/Runner.xcodeproj
We need to add the LD flag `"-lsqlite3" as otherwise on MacOS neither the iOS App nor native app for MacOs will start up.

resolves #26 

What do you think of introducing an extra layer for models: i.e. have the db models private and pass in application-layer models. This way we can keep application and db logic completely separated and should be able to easily swap the db if needed. It would however add more code. This has not yet been done though.


mobile/native/src/api.rs
```diff
@@ -154,3 +158,8 @@ pub fn create_invoice() -> Result<String> {
 pub fn send_payment(invoice: String) -> Result<()> {
     ln_dlc::send_payment(&invoice)
 }
+
+pub fn insert_last_login() -> Result<LastLogin> {
+    let last_login = db::update_last_login()?;
```
holzeis, line 163: nit: You could write this in one line.

```
db::update_last_login()
```

Also it looks strange that the api is named `insert_last_login` while the db function is named `update_last_login`. Why not use the same name?

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use diesel_migrations::embed_migrations;
+use diesel_migrations::EmbeddedMigrations;
+use diesel_migrations::MigrationHarness;
+use state::Storage;
+use std::sync::Arc;
+use time::OffsetDateTime;
+
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
+
+/// Wallet has to be managed by Rust as generics are not support by frb
```
holzeis, line 21: 🙃  That comment does not make any sense.

mobile/native/src/api.rs
```diff
@@ -136,6 +137,9 @@ pub fn subscribe(stream: StreamSink<event::api::Event>) {
 }
 
 pub fn run(app_dir: String) -> Result<()> {
+    let db_path = format!("sqlite://{app_dir}/trader.sqlite");
```
holzeis, line 140: 🔧 Wouldn't it be better to simply pass the `app_dir` to `db::init_db` and deriving the `database_url` within the function, instead of outside? It feels out of place to do that on the outside, as it implies knowledge about the inner workings of `db::init_db`

set the sqlite specific database_url within the

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+        .map_err(|e| anyhow!("could not run db migration: {e:#}"))?;
+    tracing::trace!("Database migration run - db initialized");
+
+    DB.set(Arc::new(pool));
+
+    Ok(())
+}
+
+fn get_db() -> Result<Arc<Pool<ConnectionManager<SqliteConnection>>>> {
+    DB.try_get().context("DB uninitialised").cloned()
```
holzeis, line 43: nit: you could write that shorter by simply writing `DB.get().cloned()`. It is calling `try_get().context(..)` under the hood.

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    let pool = get_db()?;
+
+    pool.get()
+        .map_err(|e| anyhow!("cannot acquire database connection: {e:#}"))
+}
+
+pub fn update_last_login() -> Result<LastLogin> {
+    let mut db = connection()?;
+    let now = OffsetDateTime::now_utc();
+    let last_login = LastLogin::update_last_login(now, &mut db)?;
```
holzeis, line 56: nit: the last two lines could be combined to one.

```
LastLogin::update_last_login(now, &mut db)
```

holzeis (approved):
Nice 👍

mobile/native/src/db/mod.rs
bonomat: lol 🙈

mobile/native/src/api.rs
bonomat: > Why not use the same name?

Because of reasons! 😬

mobile/native/src/api.rs
bonomat: Ah, got it, you would only pass in the `app_dir` and add the `taker.sqlite` inside. Can do :)

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
+pub mod models;
```
luckysori, line 1: 🙃 I think we usually leave the use statements at the top.

I think it usually goes:

```rust
use foo;

mod bar;

pub use baz;

const  QUX;
```

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
+
+/// Wallet has to be managed by Rust as generics are not support by frb
+static DB: Storage<Arc<Pool<ConnectionManager<SqliteConnection>>>> = Storage::new();
+
+pub fn init_db(database_url: String) -> Result<()> {
+    let manager = ConnectionManager::<SqliteConnection>::new(database_url);
+    let pool = r2d2::Pool::builder()
+        .build(manager)
+        .expect("Failed to create pool.");
```
luckysori, line 28: ❓ Why not bubble up the error since we are already returning a `Result`?

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    let mut connection = pool.get()?;
+
+    connection
+        .run_pending_migrations(MIGRATIONS)
+        .map_err(|e| anyhow!("could not run db migration: {e:#}"))?;
+    tracing::trace!("Database migration run - db initialized");
```
luckysori, line 35: 🙃 I'd usually only relegate to the trace level logs which can be spammy.

```suggestion
    tracing::debug!("Database migration run - db initialized");
```

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    DB.set(Arc::new(pool));
+
+    Ok(())
+}
+
+fn get_db() -> Result<Arc<Pool<ConnectionManager<SqliteConnection>>>> {
```
luckysori, line 42: 🔧 If we only call this once in the function below we should just inline it.

mobile/native/src/db/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    DB.try_get().context("DB uninitialised").cloned()
+}
+
+pub fn connection() -> Result<PooledConnection<ConnectionManager<SqliteConnection>>> {
```
luckysori, line 46: ❓ Does this need to be public?

mobile/native/src/db/mod.rs
luckysori: You're right, although historically we've preferred to end functions that return `Result`s with `Ok(value)`.

luckysori (approved):

mobile/native/src/db/mod.rs
bonomat: Unfortunately this does not seem to work.

mobile/native/src/db/mod.rs
bonomat: Unfortunately it does not work because `LastLogin::update_last_login(now, &mut db)` returns a `diesel::result::Error`.

mobile/native/src/db/mod.rs
bonomat: This will be used more often later on.

mobile/native/src/db/mod.rs
bonomat: nope :)

bonomat:
bors retry

PR #201, opened 2023-03-02
feat: Configurable node announcement

### LINKED ISSUE #191: Make the mobile app configurable
Currently we are still using hardcoded configuration in the mobile app. 

The flutter way of fixing that seems to be through `--dart-define`, which can be consumed by dart through environment variables. The issue of using environment variables directly (as we have done in the PoC) is that they are not available to the flutter app and can only be parsed by our native implementation. However that approach is only feasible for desktop target platforms, but not for ios or android.

see also - https://github.com/get10101/10101-PoC/pull/215

holzeis:
I added quite some descriptions to the commit message. Please refer to https://github.com/get10101/10101/pull/201/commits/c1c02ebb1b1e2674d7d84a41d6c2206a1ea2682d for more details on this PR.

**tl;dr**: This PR fixes an issue with the lnd node and the coordinator not being connected anymore after a restart and allows for configuring the mobile app.

The critical aspect here is that the address the `ln_dlc node` will be bound to is also the address that is used for the node announcement. So you need to start your coordinator with the ip address of your local network.

```
cargo run --bin coordinator -- --p2p-address=10.0.0.20:9045
```
or use the just file analogous.

In order to start the mobile app you need to specify the coordinator p2p-address as well, as the coordinator does not listen to localhost anymore.

```
just run --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@10.0.0.20:9045"
```

resolves #191 

----

Note, alternatively we could have added the local ip to the node announcement addresses also automatically in the code so that we wouldn't have to provide any arguments. But this is only relevant for our testing environment and might not justify adding a new dependency for that. (local_ip) Also it would be confusing if we'd publish externally unreachable addresses with our node announcements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,9 @@ Please refer to the [docker](https://docs.docker.com/) docs for more information
 
 #### Setup
 
-1. Start the coordinator with `cargo run --bin coordinator` or `just coordinator`
+1. Start the coordinator with `cargo run --bin coordinator -- --p2p-address=<your-local-ip>:9045` or `just coordinator --p2p-address=<your-local-ip>:9045`.
+
+   _Ensure that you are using your network ip address and not localhost. This is critical as the docker container will otherwise not be able to reach the coordinator._
 2. Open `http://localhost:8080/faucet/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
 3. Ensure you have enough balance on your bitcoin wallet. Hit the mine button a couple of times if not.
 4. Get a new address of your coordinator by running `curl http://localhost:8000/api/newaddress`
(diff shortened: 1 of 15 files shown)


holzeis:
The network is not configurable yet. I've created a separate ticket for that - https://github.com/get10101/10101/issues/202

holzeis:
The startup could / should be improved to not require any input for the dev environment. This could be done in the two possible ways.

1. Enhance justfile to derive local ip and add the argument automatically
2. Bind the docker-compose network to the host network so that lnd is able to find the coordinator on localhost.

luckysori (approved):

Issue #205, opened 2023-03-06
Clicking "back" from the lightning invoice goes back to main screen instead of back to the previous step (with the amounts)


klochowicz:
it a nutshell, clicking the "back" button (left arrow) on the screen with QR code of the invoice behaves like "home" instead of "back".

PR #206, opened 2023-03-06
Fix/nest wizard screen

da-kami:
fixes https://github.com/get10101/10101/issues/205
 

---

Note: This is a bit cumbersome because of the declarative nature of `go_router`. I did not find a way to navigate "back to the previous screen" without nesting the screens. It would be cool to have the possibility to just navigate by an index, but that requires a bit more thought, because the index would have to be assigned to the screens dynamically, and not as a static route name. I did not find an easy way to achieve this yet, so I decided that nesting the screens is the fastest way to fix this. Once we have wizards with more then ~3 screens (e.g. wallet backup...) we could consider implementing something more elaborate.

### DIFF
--- a/mobile/lib/features/wallet/share_invoice_screen.dart
+++ b/mobile/lib/features/wallet/share_invoice_screen.dart
@@ -2,6 +2,7 @@ import 'dart:developer';
 
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:get_10101/features/wallet/create_invoice_screen.dart';
 import 'package:get_10101/features/wallet/domain/wallet_info.dart';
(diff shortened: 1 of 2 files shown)


mobile/macos/Podfile.lock
```diff
@@ -3,20 +3,26 @@ PODS:
   - path_provider_foundation (0.0.1):
     - Flutter
     - FlutterMacOS
+  - share_plus (0.0.1):
+    - FlutterMacOS
```
klochowicz, line 7: is this package related to this changeset? we're not doing anything with the share menu here?

klochowicz (approved):

holzeis (approved):

PR #207, opened 2023-03-06
chore: re-export external types in frb

1010Tom:
By using flutter rust bridge - mirror, we can re-export foreign types. This allows us to remove various duplicated types, their conversions and extensions such as `*Trade`.

Read more here:
https://cjycode.com/flutter_rust_bridge/feature/lang_external.html#types-in-other-crates

### DIFF
--- a/crates/trade/src/lib.rs
+++ b/crates/trade/src/lib.rs
@@ -65,7 +65,7 @@ pub struct TradeParams {
     pub oracle_pk: XOnlyPublicKey,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub enum ContractSymbol {
     BtcUsd,
 }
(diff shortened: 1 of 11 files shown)


da-kami (approved):
As long as we are only doing this for types that will (likely) not change (i.e. enums with fixed values, ...) I think this is a good idea.
For types where the internal representation is different - or may be different in the future - to the representation that we expose on the API we should not do this.

---

I think it would be good to go one step further and move these types into the `trade` crate so we can share them across multiple systems.

1010Tom:
> As long as we are only doing this for types that will (likely) not change (i.e. enums with fixed values, ...) I think this is a good idea.
For types where the internal representation is different - or may be different in the future - to the representation that we expose on the API we should not do this.

I think as long as the types are always exactly the same we should use the `mirrored` feature. 
We will need to pull field-changes through but that's about it, the `_*`-named types are not used anywhere, it's just for the compiler.

> I think it would be good to go one step further and move these types into the trade crate so we can share them across multiple systems.

Which types are you referring to? I'll create a follow-up ticket for this.

bors r+

holzeis (approved):
👍 I like it, these *Trade copies seemed off anyways. Shouldn't the `Direction` enum also be in the trade crate?

PR #208, opened 2023-03-06
[app] Store orders in db

### LINKED ISSUE #134: Persist orders and positions in the database
- [ ] foundation for storing orders https://github.com/get10101/10101/pull/208
- [ ] foundation for storing positions
- [x] #223
- [ ] update position in db when necessary

1010Tom:
This PR is draft but I'm asking for feedback already please. 

It's quite cumbersome to write those custom types but it works, hence, I want to get feedback now before we decide to go for a different approach. 
If we are happy with this then I will squash most of the commits and finish this PR.

Let me know :)

Related to #134


holzeis:
These custom_types look really cumbersome. You are asking for feedback if we should continue with them, is there an alternative?

bonomat:
> These custom_types look really cumbersome. You are asking for feedback if we should continue with them, is there an alternative?

Not that I'm aware of, but maybe someone else knows more?

bonomat:
> > These custom_types look really cumbersome. You are asking for feedback if we should continue with them, is there an alternative?
> 
> Not that I'm aware of, but maybe someone else knows more?

correction: for some types (enums without variant fields) there is a library https://github.com/adwhit/diesel-derive-enum. I tried it first but decided against it because we would need to manually edit schema.rs which is a pita.

mobile/native/src/db/custom_types.rs
```diff
@@ -0,0 +1,216 @@
...
+use diesel::sql_types;
+use diesel::sql_types::Text;
+use diesel::sqlite::Sqlite;
+
+impl ToSql<sql_types::Text, Sqlite> for OrderType {
+    fn to_sql(&self, out: &mut Output<Sqlite>) -> serialize::Result {
+        let text = match *self {
+            OrderType::Market => "market".to_string(),
+            OrderType::Limit { price } => {
+                format!("limit:{price}")
```
da-kami, line 21: 🤔 Do I understand this correctly, that we are basically serializing the enum variant's field  as pseudo-json into the database column \ here? 

If we want to go for serialization of complex enums then I think we should actually use `serde_json` to achieve that and not write custom serialization code.

---

I wonder if we could represent complex enums (of the business logic) different in the DAL so they reflect the schema in the db. 

For example, the DAL could represent the `OrderType` like this:

```
struct OrderType {
    inner: OrderTypeInner,
    price: Option<f64>
}

enum OrderTypeInner {
    Limit,
    Market
}
```

This flattens the data out into a table and we would avoid serialization. It's still ugly though, and I am not sure it's less cumbersome 😅
Note: It's also not optimal because it's not actually normalized properly; the more complex the enum gets the more nullable fields we have...
Complex enums and relational schemas don't go well together.

We could also make saving the data of the complex enum *implicit* - i.e. we safe the flattend enum variant in a table, and if the enum is complex then we load the data from a separate table, something like this in the DAL:

```
struct OrderTypeLimitData {
    price: f64
}

enum OrderType {
    Limit,
    Market
}
```

There would be two database table for this, `OrderType` and `OrderTypeLimitData`. Upon loading from the `OrderType` table and we have `Limit` we would additionally load from `OrderTypeLimitData` and then build the complex enum type for the business logic.

(note: The enum names can be different, this is just an example ;)

---

We could also re-think our extensive use of complex enum types in the business logic, but re-thinking the whole model might be a bigger step.

klochowicz: > If we want to go for serialization of complex enums then I think we should actually use serde_json to achieve that and not write custom serialization code.

This sounds like a reasonable starting point. 

> We could also re-think our extensive use of complex enum types in the business logic, but re-thinking the whole model might be a bigger step.

As a step towards not getting complex enums go out of hand, I would like to propose not ignoring clippy when it says "this enum is **too** complex" 😅

1010Tom: > > If we want to go for serialization of complex enums then I think we should actually use serde_json to achieve that and not write custom serialization code.
> 
> This sounds like a reasonable starting point.

What I'm doing here is not json and I'd like to avoid json in the db if possible. 

I like the approach of the extra table though. 
This should work for enums without field values but if they have one, I wouldn't know how to do it nicely.

klochowicz: would any other format suit your needs? `serde` is capable of a whole range of formats, not only JSON - any point where we can avoid custom serialisation saves time when writing / refactoring / maintaining.

1010Tom: A problem I see with using `serde` (or other 3rd party serialization libraries) is that it may hide when we break backwards compatibility: we rely on the compiler to tell us that all types have been changed, but loading from an existing DB during runtime will fail, preciously because of relying on others.

E.g. Let's say we store 

```
pub enum Direction {
    Long,
    Short,
}
```

if we use `serde` to serialize/deserialize form the DB, then renaming `Long` to `Buy` will work without a problem. However, during runtime, when loading from an existing DB, it may fail. 

These problems are really hard to detect.

With a custom implementation renaming `Long` to `Buy` requires us to also update the custom implementation which should make the dev think about breaking changes. 

-----

What do you think about flatten the enums out completely and allow null values? 
I think this could be the best option.

E.g. instead of having this:

```
struct Order {
	state: OrderState
} 

enum OrderState {
    Open,
    Filled { execution_price: f64 },
}
```

we go for 

```
struct Order {
	state: OrderState,
	execution_price: Option<f64>
} 

enum OrderState {
    Open,
    Filled,
}
```

Then when loading we can construct our business logic structs again.


I would be in favor of the last proposal as it is easy to implement, easy to maintain and easy to spot errors if types change.

1010Tom: Like this [`8bd4e5a` (#208)](https://github.com/get10101/10101/pull/208/commits/8bd4e5aad275457eb00e57d1295ccf92afdbd242). What do you think?

mobile/native/migrations/2023-03-06-005404_orders/up.sql
```diff
@@ -6,5 +6,9 @@ CREATE TABLE IF NOT EXISTS orders (
     contract_symbol TEXT NOT NULL,
     direction TEXT NOT NULL,
     order_type TEXT NOT NULL,
-    STATUS TEXT NOT NULL
+    STATUS TEXT NOT NULL,
```
da-kami, line 9: ❓unrelated question: why is this column caps when the others are not?

da-kami (commented):
I think adding nullable fields is a fair approach for now - it's not perfect, but it has the best effort/value rating for now 😬

Let's roll with this for now, we can always refactor it later!

da-kami (approved):
I think we can squash the wip commit; looks complete to me otherwise but maybe I'm missing something :)

mobile/native/migrations/2023-03-06-005404_orders/up.sql
bonomat: I was wondering as well.. I think it is dprint :/

1010Tom: Yes, it is dprint. I'll leave it as is as it doesn't matter much :)

1010Tom:
I split the initial ticket into multiple tasks so that we can merge this for now. 
It is only the foundation and stores the initial order but that's about it. 

bors r+

PR #209, opened 2023-03-06
chore: Announce coordinator node in local network

### LINKED ISSUE #203: Default startup arguments so that it will work with the dev setup
The startup could / should be improved to not require any input for the dev environment. This could be done in the two possible ways.

1. Enhance justfile to derive local ip and add the argument automatically
2. Bind the docker-compose network to the host network so that lnd is able to find the coordinator on localhost.

_Originally posted by @holzeis in https://github.com/get10101/10101/pull/201#issuecomment-1453091860_

holzeis:
This change sets all configurations to defaults that will work with our development setup.

 - p2p_address: changed to `0.0.0.0` so that the coordinator is reachable on localhost as well as local ip. This is needed as the phone communicates on localhost and the lnd node on the local ip.
 - skip_local_network_announcement: If this flag is set the local network ip announcement is skipped. This should be set in a productive environment. If not set it will add the local ip to the addresses announced for the coordinator. This is needed so that the lnd node is able to reach the coordinator from the docker network.

resolves #203 

Note, I decided to go for Option 4, announcing the node on the local_ip, because it seemed to me the most practical solution. 

- **Option 1**: Too failure prune, as its pretty hard to know which network interface to use and it will be different on every platform.
- **Option 2**: Binding to the host network feels odd and would require a different `docker-compose` for the github actions pipeline. We might still want that eventually, but probably for other reasons.
- **Option 3**: Only works for mac and windows. DNS hostname is not supported by lnd yet.


coordinator/src/cli.rs
```diff
@@ -34,4 +40,40 @@ impl Opts {
...
+            let local_ip = local_ip().expect("to get local ip address");
+            tracing::info!("Adding node announcement within local network {local_ip}. Do not use for production!");
+            addresses.push(build_net_address(local_ip, self.p2p_address.port()));
+        }
+
+        addresses
+    }
+}
+
+pub fn build_net_address(ip: IpAddr, port: u16) -> NetAddress {
```
bonomat, line 68: I think this does not need to be pub.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -382,12 +388,12 @@ impl Node {
...
                 // error.
-
-                let address = util::build_net_address(address.ip(), address.port());
                 let mut interval = tokio::time::interval(BROADCAST_NODE_ANNOUNCEMENT_INTERVAL);
 
                 loop {
-                    peer_manager.broadcast_node_announcement([0; 3], alias, vec![address.clone()]);
+                    tracing::debug!("Announcing node on {:?}", announcements);
+                    let announcements = announcements.clone();
+                    peer_manager.broadcast_node_announcement([0; 3], alias, announcements);
```
bonomat, line 396: I think this loop should actually be removed and we only announce once.

README.md
```diff
@@ -155,12 +155,14 @@ Please refer to the [docker](https://docs.docker.com/) docs for more information
...
+
+   _Note, if you're on mac or windows you can use `host.docker.internal` as coordinator ip._
 9. Mine a few blocks (at least 6) so that the channel gets announced.
 
 #### Fauceting your lightning wallet
 
 10. Create an invoice in your 10101 app by navigating to the receive screen.
 
-    _Note, that you have to provide the coordinator host to the mobile app like that `just run --dart-define="COORDINATOR_HOST=<your-local-ip>"`_
+    _Note, that you have to provide the coordinator host to the mobile app like that `just run`_
```
bonomat, line 167: Did you mean to remove this part?

README.md
```diff
@@ -155,12 +155,14 @@ Please refer to the [docker](https://docs.docker.com/) docs for more information
 5. Faucet some coins to your coordinator wallet. Hit the mine button afterwards so the transaction gets into a block.
 6. Open `http://localhost:8080/channel/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
 7. Copy the address of the lnd node and faucet that wallet as described in step 5.
-8. Copy the node id (_pubkey@host:port_) from your coordinator logs and instruct lnd to open a channel with your coordinator. Set a reasonable channel capacity. Note, this capacity will be only inbound for your coordinator.
+8. Open a channel with your coordinator (02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@[coordinator ip]:9045) and set a reasonable channel capacity.
+
+   _Note, if you're on mac or windows you can use `host.docker.internal` as coordinator ip._
```
bonomat, line 160: To make things easy to c&p :) 
```suggestion
   _Note, if you're on mac or windows you can use `host.docker.internal` as coordinator ip._
`02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@host.docker.internal:9045`
```

coordinator/src/cli.rs
```diff
@@ -34,4 +40,40 @@ impl Opts {
 
         Ok(data_dir)
     }
+
+    /// Returns a list of addresses under which the node can be reached. Note this is used for the
+    /// node announcements.
+    pub fn p2p_announcement_addresses(&self) -> Vec<NetAddress> {
+        let mut addresses: Vec<NetAddress> = vec![];
+        if !self.p2p_address.ip().is_unspecified() {
```
bonomat, line 48: I believe this works for a local setup but not for production where we are either behind a NAT or inside a docker container.

klochowicz (commented):
we can also remove the requirement to pass in coordinator host to the mobile app (the other README in /mobile directory)

README.md
holzeis: yes, you don't need to define it anymore, as the default values will work now. You can still specify if you want to.

coordinator/src/cli.rs
holzeis: yes, I thought so too.. so, lets add it when we need it. 😉

README.md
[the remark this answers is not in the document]
holzeis: I wasn't able to test it, but from what I've researched it only works on mac and windows. 

Looking again I found this on stack overvlow. Looks like you can make it work for linux, but not out-of-the-box.
https://stackoverflow.com/questions/48546124/what-is-linux-equivalent-of-host-docker-internal

crates/ln-dlc-node/src/node/mod.rs
holzeis: Well, I thought about it, but left it for the following reasons.

1. We use the same code for our mobile node and there it is useful to announce the nodes ip address in a loop. (E.g. the network changes)
2. In case we want to use a domain instead of an ip to announce our node, we could a) change the ip underneath without having to change the config on the coordinator and b) also work with dyndns.

README.md
holzeis: I've added this node id directly to the `open_channel.html` that makes it even more easier to copy & paste 😉

README.md
bonomat: Ah right, thanks for clarification

crates/ln-dlc-node/src/node/mod.rs
bonomat: > We use the same code for our mobile node and there it is useful to announce the nodes ip address in a loop. (E.g. the network changes)

Do we need to announce the app node? 

> In case we want to use a domain instead of an ip to announce our node, we could a) change the ip underneath without having to change the config on the coordinator and b) also work with dyndns.

I thought LN does not work with domains?

klochowicz:
> > I don't mind this approach, however if you will get into trouble with testing in prod (as @bonomat outlined), then option 2) is likely our next best bet. Then we have just have local-dev specific `docker-compose.yml`, which is not too bad, especially when we're exposing all the ports anyway.
> 
> Maybe we shouldn't test in prod then 😅 . However, the whole idea of this PR was to make the defaults work for our local development. You can still change the default configuration to work for any other environment. I just did not add yet, the configuration to add additional announcement addresses.
> 
> > also: we can also remove the requirement to pass in coordinator host to the mobile app (the other README in /mobile directory)
> 
> sorry, I don't find the `README.md` you are referring to.

scratch that, I think I had something on a wrong branch. sorry for making you look for something that wasn't there! 😅

coordinator/src/cli.rs
```diff
@@ -34,4 +40,40 @@ impl Opts {
...
+            addresses.push(build_net_address(
+                self.p2p_address.ip(),
+                self.p2p_address.port(),
+            ));
+        } else {
+            // Announcing the node on an unspecified ip address does not make any sense.
+            tracing::warn!("Skipping node announcement on '0.0.0.0'.");
+        }
```
luckysori, line 56: 🔧 Why don't we just error and crash the application. It's probably pointless to run the app if it's misconfigured.

coordinator/src/cli.rs
```diff
@@ -34,4 +40,40 @@ impl Opts {
...
+
+        if !self.skip_local_network_announcement {
+            let local_ip = local_ip().expect("to get local ip address");
+            tracing::info!("Adding node announcement within local network {local_ip}. Do not use for production!");
+            addresses.push(build_net_address(local_ip, self.p2p_address.port()));
```
luckysori, line 61: ❓ Won't you end up pushing the same `LOCAL_IP:port` address _twice_ if `skip_local_network_announcement == false` and the `p2p_address` is the default one?

crates/ln-dlc-node/src/node/mod.rs
holzeis: > Do we need to announce the app node?

It looks like it. see also 

https://github.com/get10101/10101/blob/a045dc942e22282a90b89cba9b195a0ca9e21194/crates/ln-dlc-node/src/node/mod.rs#L380-L382

But it shouldn't be required, I guess?

> I thought LN does not work with domains?

Well LN Bolt-07 supports it

```
The following address descriptor types are defined:
...
    5: DNS hostname; data = [1:hostname_len][hostname_len:hostname][2:port] (length up to 258)
        hostname bytes MUST be ASCII characters.
        Non-ASCII characters MUST be encoded using Punycode: https://en.wikipedia.org/wiki/Punycode
```
https://github.com/lightning/bolts/blob/master/07-routing-gossip.md

It's just not supported by LND yet - https://github.com/lightningnetwork/lnd/issues/6337

However, we don't have to announce on the hostname, we could still do the announcement on ip and resolve the dns ourselves.

coordinator/src/cli.rs
holzeis: It does make sense to run the app on an unspecified interface (e.g. listen on any address, local host, local ip, public ip). However, this ip does not make sense to be used when announcing our node - as this address can not be used by a client. Here we need to specify the interface. (localhost, local ip or public ip).

Since I am implying the announcement node from the p2p-address, I wouldn't fail here, as this is a valid setup.

coordinator/src/cli.rs
holzeis: No sure if I follow, the default address of p2p_address is `0.0.0.0`. I guess if you set the p2p_address to your local_ip and do not `skip_local_network_announcement` you will end up pushing the same address twice.

Do suggest to add a handling for that? Looks to me more like a miss configuration.

coordinator/src/cli.rs
```diff
@@ -17,6 +20,9 @@ pub struct Opts {
     /// Where to permanently store data, defaults to the current working directory.
     #[clap(long)]
     data_dir: Option<PathBuf>,
+
+    #[clap(long)]
```
bonomat, line 24: If you add a doc comment here it will be printed when providing an invalid argument/input  :)

bonomat (approved):

coordinator/src/cli.rs
luckysori: Okay, thanks for the explanation. I think that makes sense.

Could we maybe not print a warning then? This seems like an info or a debug log to me.

coordinator/src/cli.rs
luckysori: I misread what we were doing above and forgot that it's pointless to announce on `0.0.0.0`. I get it now! Sorry for the confusion.

luckysori (commented):
Approved after the fact.

coordinator/src/cli.rs
holzeis: @luckysori fyi:
https://github.com/get10101/10101/pull/240/commits/efbd04ff1e687bfe9784a2d9bfd611781a3a7cc2

PR #210, opened 2023-03-06
feat: Configurable network

### LINKED ISSUE #202: Make the network configurable
Currently the network is hardcoded to `regtest` throughout the coordinator and mobile app. The network should be configurable through a cli argument resp. --dart-define argument. `regtest` should remain the default value.

holzeis:
This will allow to pass the network as argument on the coordinator as well as on the mobile app. Both default to regtest.

Set on coordinator like this.

```
just coordinator signet
```

Set on mobile app like this.
```
just run --dart-define="NETWORK=signet"
```

resolves #202

### DIFF
--- a/mobile/lib/util/environment.dart
+++ b/mobile/lib/util/environment.dart
@@ -10,6 +10,7 @@ class Environment {
     int httpPort = const int.fromEnvironment("COORDINATOR_PORT_HTTP", defaultValue: 8000);
     String electrsEndpoint =
         const String.fromEnvironment("ELECTRS_ENDPOINT", defaultValue: "127.0.0.1:50000");
+    String network = const String.fromEnvironment('NETWORK', defaultValue: "regtest");
 
     String p2pEndpoint = const String.fromEnvironment('COORDINATOR_P2P_ENDPOINT');
     if (p2pEndpoint.contains("@")) {
@@ -26,6 +27,7 @@ class Environment {
         electrsEndpoint: electrsEndpoint,
         coordinatorPubkey: coordinatorPublicKey,
         p2PPort: lightningPort,
-        httpPort: httpPort);
+        httpPort: httpPort,
+        network: network);
   }
 }
--- a/mobile/native/src/config/api.rs
+++ b/mobile/native/src/config/api.rs
@@ -1,4 +1,5 @@
 use crate::config::ConfigInternal;
+use bdk::bitcoin::Network;
 use flutter_rust_bridge::frb;
 
 #[frb]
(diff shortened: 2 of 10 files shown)


mobile/native/src/config/api.rs
```diff
@@ -1,4 +1,5 @@
 use crate::config::ConfigInternal;
+pub use bdk::bitcoin::Network;
```
klochowicz, line 2: `pub use` of internal library is a bit concerning, especially when we have a custom type for it.

it would be easier if the`Network` that we expose (our type) could parse a string itself.

mobile/lib/util/environment.dart
```diff
@@ -26,6 +27,21 @@ class Environment {
...
+    switch (network) {
+      case 'signet':
+        return Network.Signet;
+      case 'testnet':
+        return Network.Testnet;
+      case 'mainnet':
+        return Network.Bitcoin;
+      default:
+        return Network.Regtest;
+    }
```
klochowicz, line 45: consider doing this part in rust (`from_str()`), and then we don't need to expose BDK enum at all.

mobile/native/src/config/api.rs
```diff
@@ -9,6 +10,17 @@ pub struct Config {
     pub host: String,
     pub p2p_port: u16,
     pub http_port: u16,
+    pub network: Network,
+}
+
+#[allow(dead_code)]
+#[frb(mirror(Network))]
```
klochowicz, line 17: this would disappear if we removed the need to expose the other Network type.

klochowicz (approved):
nice work! my only concern is exposing `bdk::Network` through the FFI boundary - happy to throw in a patch to fix that if you don't mind ;D

klochowicz (approved):
awesome! 🥇

mobile/pubspec.lock
```diff
@@ -508,10 +508,10 @@ packages:
     dependency: "direct main"
     description:
       name: meta
-      sha256: "12307e7f0605ce3da64cf0db90e5fcab0869f3ca03f76be6bb2991ce0a55e82b"
+      sha256: "6c268b42ed578a53088d834796959e4a1814b5e9e164f147f580a386e5decf42"
       url: "https://pub.dev"
     source: hosted
-    version: "1.9.0"
+    version: "1.8.0"
```
klochowicz, line 514: all these downgrades/upgrades are a bit unfortunate..... they don't align if you run `flutter pub get`?

da-kami: This is indeed weird, because there was no change to the `pubspec.yml` - why do we have changes in the `pubspec.lock` file? 😅

I reckon this could be because somebody a different PR did not update the `pubspec.lock` correctly and merged - I think our CI action might not cover that properly. I created: https://github.com/get10101/10101/issues/230

da-kami (approved):

PR #212, opened 2023-03-07
Bump flutter package versions

klochowicz:
When I try to run the app for the first time, these changes appeared.


luckysori (approved):
I think `flutter pub get` just defaults to updating the dependencies as much as possible according to the constraints specified in the `pubspec.yaml`. 

For the `meta` dependency we specify `^1.8.0`, which means updating the minor version is fine. For the transitive dependencies perhaps it just defaults to only allowing patch version bumps?

bonomat (approved):

PR #213, opened 2023-03-07
Update to latest `rust-dlc`

luckysori:
The new `rust-dlc` [branch](https://github.com/p2pderivatives/rust-dlc/pull/97) is still WIP, but it's good to keep this dependency up-to-date as much as possible.


mobile/pubspec.lock
```diff
@@ -1022,5 +1022,5 @@ packages:
     source: hosted
     version: "2.0.3"
 sdks:
-  dart: ">=2.19.0 <4.0.0"
+  dart: ">=2.19.0 <3.0.0"
```
da-kami, line 1025: 🛠️ I don't think this should be in here. I keep running into this problem as well on my machine (somehow it is always changed back; but I think for @bonomat it is then changed back to 4.0.0 😅)

da-kami (approved):
LGTM

mobile/pubspec.lock
luckysori: I can remove it, but according to the `pubspec.yaml`, this change is correct:

```yaml
environment:
  sdk: ">=2.17.5 <3.0.0"
```

If you play around with those constraints and you run `flutter pub get` the `pubspec.lock` changes as expected.

luckysori:
I pushed https://github.com/get10101/10101/pull/213/commits/09d9710c5606b70a08d289fb0f242a5ec71a0113 by mistake here, but since https://github.com/get10101/10101/pull/214 has already been approved and the build is going I'll let it be and close the other PR once this one merges.

PR #214, opened 2023-03-07
Minimise need to depend on `rust-dlc` in leaf crates

luckysori:
This dependency should ideally only be known by core crates such as `ln-dlc-node`.

We define a method `get_new_address` on `Node` so that we can control how we call the underlying on-chain wallet, rather than letting consumers call the wallet through a trait that is designed for a completely different purpose.

We also turn the generic `get_dlcs` API into the more specific `get_confirmed_dlcs`, as this is the only thing that is currently needed. This way we can avoid consumers needing to understand types defined by `rust-dlc`. The `get_confirmed_dlcs` API and the matching `Dlc` type should evolve (and perhaps even disappear) as we meet new requirements.

We cannot remove the `dlc-manager` dependency from the `coordinator` just yet because we haven't yet defined our own type for the `ContractInput`. Once we do so, we should remove this dependency and finally ensure that `rust-dlc` is an implementation detail from the perspective of our leaf crates.


da-kami (approved):
LGTM

da-kami:
> We cannot remove the `dlc-manager` dependency from the `coordinator` just yet because we haven't yet defined our own type for the `ContractInput`. Once we do so, we should remove this dependency and finally ensure that `rust-dlc` is an implementation detail from the perspective of our leaf crates.

I thought about doing so, but it is quite cumbersome because it's a quite complex type with a lot of nested other types inside. Nonetheless, it would be cleaner to create our own and map.

luckysori:
Merged with #213.

PR #222, opened 2023-03-07
Add APIs to perform DLC collaborative settlement

### LINKED ISSUE #50: Collaborative settlement of the DLC
Implementation contingent on #49.

Two properties to be verified:

- [x] Funds correctly awarded to each wallet.
- [ ] Can still create another DLC using those funds without going on-chain.

luckysori:
Fixes #50.


crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -136,6 +136,50 @@ impl Node {
         Ok(())
     }
 
+    pub fn propose_dlc_channel_collaborative_settlement(
```
holzeis, line 139: nit: I'd prefer an aligned naming here. either use `collaborative_settlement` or `close_offer`

--
`initiate_propose_dlc_channel_close_offer` and `initiate_accept_dlc_channel_close_offer`
or
`propose_dlc_channel_collaborative_settlement` and `initiate_accept_dlc_channel_collaborative_settlement`

holzeis (approved):
Nice 👍

Issue #220, opened 2023-03-07
Be able to close position


bonomat:
When the trader is `100 long` and goes `100 short` the position should be closed. 

The idea is that when the trader post `api/trade` the coordinator does the math and then tells the trader to close the DLC collaboratively. 

For this ticket to work we will need to implement things on the coordinator and the app at the same time.

Related to #181

holzeis:
I started to look into this ticket. Here are my two cents.


### Closing a Position on the mobile app

Closing a position basically means creating a "reversed" order of the position which is currently open. e.g. if the existing position is long then the order will be short.

This will mean we will be able to close a Position two ways through the UI. Either you hit the `Close Position` button on the Position itself, or you open a Short Order through the trade screen. 

In any case we have to ensure (for the MVP) that if e.g. a long position exists the short order can only get submitted with the same quantity as the long position (as we will not support partial closing of a position for MVP).

### Closing a Position on the coordinator

After the order is submitted to the coordinator, the coordinator looks for existing positions with the submitting counterparty. If a position is found matching the quantity and a "reversed" direction, the coordinator simply proposes to close the dlc_channel with the trader. 

_Note, I don't see the benefit of instructing the mobile app to initiate the closing of the dlc channel, but happy to be convinced otherwise._


### MVP Scope

If we do not support closing a position partially, then I guess we will also not support extending a position. That will basically mean, after you have an open position. The only action you can take as a trader is to close the position. Meaning that we can simplify things here in the implementation as we can assume that another order submitted by the trader can only be a request to close the position if a position already exists.

bonomat:
> In any case we have to ensure (for the MVP) that if e.g. a long position exists the short order can only get submitted with the same quantity as the long position (as we will not support partial closing of a position for MVP).

I think this is a fair limitation which we should be able to enforce through the frontend.



> ### Closing a Position on the coordinator
> 
> After the order is submitted to the coordinator, the coordinator looks for existing positions with the submitting counterparty. If a position is found matching the quantity and a "reversed" direction, the coordinator simply proposes to close the dlc_channel with the trader.

I'd like to keep the _normal_ order flow, i.e. 
1. the trader posts a `market order` to the `orderbook`[^1] , 
2. once there is a match, the `orderbook` tells the trader _how_ to trade
3. the trader then goes against the `coordinator`: since the only allowed action is a total close, the trader will end-up with no DLC channel.

Note: @2: for the MVP the only matched order will come from our maker

[...]

Issue #181, opened 2023-02-28
Create an ADR for the order + trade lifecycle within the systems


da-kami:
This ticket adds to https://github.com/get10101/meta/discussions/131

We should capture the lifecycle and flow of the order and the order execution in an ADR. 

---

After a discussion with @bonomat about the flow "from order to position" we agreed on:

- When an order is matched by the orderbook *both* trading parties (clients) get notified about the match by the orderbook
- Both clients then send an execution request to the coordinator
     - The order is what the parties agreed on, the order is what is used in the execution request.
     - The coordinator comes up with the `ContractInput` based on the order and returns the `ContractInput` to the client as response to the request; this is to be used by the client to trigger `propose_dlc_channel` on either client side
- Once the coordinator has both (matching) execution requests, i.e. both `propose_dlc_channel` recorded, the coordinator accepts both.
    - Timeout on coordinator: If either party does not execute the other party is notified with execution failure and the coordinator removes the pending execution request. 

Note: In the context of the coordinator we don't have to use "Maker" and "Taker", because for the coordinator it should not matter; it's just two trading parties / clients. But since we used these labels in other sequence diagrams I am re-using them here.

```mermaid
sequenceDiagram

participant m as Maker
participant o as Orderbook
participant c as Coordinator
participant t as Taker

o ->> m: matched {order_maker} with {order_taker, taker_pubkey}
o ->> t:  matched {order_taker} with {order_maker, maker_pubkey}
t ->> c: request execution: {mine: order_taker, counterparty: order_maker, pubkey_taker, pubkey_maker}
c ->> t: `ContractInfo` {contract_info_taker}
c ->> c: record pending execution: {party_a: order_taker, party_b: order_maker, pubkey_taker, pubkey_maker}: request_pending
m ->> c: request execution: {mine: order_maker, counterparty: order_taker, pubkey_taker, pubkey_maker}
c ->> m: `ContractInfo`  {contract_info_maker}
c ->> c: match pending execution: {party_a: order_taker, party_b: order_maker, pubkey_taker, pubkey_maker}: request_completed
t ->> c: propose_dlc_channel(contract_info_taker)
c ->> c: update pending execution: {party_a: order_taker, party_b: order_maker, pubkey_taker, pubkey_maker}: execution_pending
m ->> c: propose_dlc_channel(contract_info_maker)
c ->> c: update pending execution: {party_a: order_taker, party_b: order_maker, pubkey_taker, pubkey_maker}: in_execution
[...]

PR #228, opened 2023-03-07
feat: Close position

holzeis:
see also
 - https://github.com/get10101/10101/issues/181#issuecomment-1458458216
 - https://github.com/get10101/10101/issues/220#issuecomment-1458349496

This PR introduces closing a position using the rust-dlc implementation of a linear payout curve. I've created a ticket for the [inverse payout curve](https://github.com/get10101/10101/issues/250).

This PR also fixes an issue with starting the mobile app, which I had to fix here a795ec10c4891f4146413814a5a41706ef7f0e83 as I was blocked otherwise.

There are loads of todos left in the code, but for demo this should be good enough for merging. We can address the todos, in follow up PRs.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -79,14 +79,14 @@ impl Node {
         tracing::info!("total collatoral");
         let total_collateral = margin_coordinator + margin_trader;
         tracing::info!("get leverages");
-        let (leverage_short, leverage_long) = get_leverages(trade_params);
+        let (leverage_long, leverage_short) = get_leverages(trade_params);
 
         tracing::info!("build payout curve");
         let payout_function = build_payout_curve(
             total_collateral,
             trade_params.execution_price,
-            leverage_short,
             leverage_long,
+            leverage_short,
         )?;
 
         tracing::info!("starting payout function");
@@ -120,13 +120,13 @@ impl Node {
 
         let (margin_coordinator, margin_trader) = get_margins(trade_params);
         let total_collateral = margin_coordinator + margin_trader;
-        let (leverage_short, leverage_long) = get_leverages(trade_params);
+        let (leverage_long, leverage_short) = get_leverages(trade_params);
 
         let contract_descriptor = build_contract_descriptor(
             total_collateral,
             trade_params.execution_price,
-            leverage_short,
             leverage_long,
+            leverage_short,
         )?;
 
         let contract_symbol = trade_params.contract_symbol.label();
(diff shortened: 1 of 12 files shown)


coordinator/src/node.rs
```diff
@@ -11,38 +13,127 @@ use dlc_manager::payout_curve::PayoutPoint;
...
+        }
+    }
+
+    /// Checks if the existing position needs to be closed.
+    fn is_position_closing(&self, trade_params: &TradeParams) -> Result<bool> {
+        // FIXME: this needs to be replaced with a query from the database, matching the existing
+        // position on quantity and direction. If a "reversed" order was found we propose to
+        // collaboratively settle the dlc channel. If not the position either needs to be extended
+        // or reduced. e.g. if we have an existing long position of a quantity of 100 and we get a
+        // short position of a quantity of 75, we will simply reduce the existing dlc_channel to 75.
```
da-kami, line 61:
```suggestion
        // short position of a quantity of 75, we will simply reduce the existing dlc_channel to 25.
```

I think `100 - 75 = 25` 😄
Note that other parameters of the position will also represent the aggregate of the orders, e.g. average leverage, average entry price and the resulting collateral (margin).

crates/trade/src/lib.rs
```diff
@@ -66,11 +65,6 @@ pub struct TradeParams {
     /// A duration that defines how long the contract is meant to be valid.
     /// The coordinator calculates the maturity timestamp based on the current time and the expiry.
     pub expiry: Duration,
-
-    /// The public key of the oracle to be used
-    ///
-    /// The orderbook decides this when matching orders.
-    pub oracle_pk: XOnlyPublicKey,
```
da-kami, line 73: 🙃 I would keep it as part of the trade params and add the oracle URL. I think the orderbook is set to be the authority for deciding this. Not a big deal though, we can also add / change it later.

da-kami (approved):
LGTM!

coordinator/src/node.rs
```diff
@@ -51,90 +142,162 @@ impl Node {
...
+        trade_params.execution_price,
+        trade_params.quantity,
+        trade_params.leverage,
+    );
 
-        Ok(())
+    (margin_coordinator, margin_trader)
+}
+
+fn get_leverages(trade_params: &TradeParams) -> (f64, f64) {
```
da-kami, line 187: 🙃 To avoid confusion when reading the code it would be good if we could stick to always using the same order when returning tuples; in ItchySats we agreed on `(long, short)` (long was always first, short second).
Here we use `(short, long)` - we might want to turn this around.

da-kami: I adapted it to always be `long, short`

luckysori (approved):
Will tackle my comments and maybe other small things in a follow-up PR.

luckysori:
bors retry

Issue #223, opened 2023-03-07
update order in db when necessary


1010Tom:
https://github.com/get10101/10101/pull/208 set the foundation for storing the order in the DB. What is missing is to update the order whenever something happens, e.g. when it gets created or filled.

da-kami:
My understanding: This ticket is about implementing the *business logic* that triggers the creation  / update of the order in the database, i.e. use the database interface that was created. I will see to wire that together :)

da-kami:
I basically completed the (happy path) order lifecycle as part of this ticket and would go on with the position next; which requires the position handling in the DAL.

bonomat:
done

PR #234, opened 2023-03-08
Feat/database exposes crud for bl

da-kami:
Expose `Order` related CRUD on the db interface so we can use it in the business logic (BL).
This is pre-work for #223
I will go on with using this work in the BL, we can merge this PR separately already to keep the changeset smaller :)

### DIFF
--- a/mobile/native/src/db/mod.rs
+++ b/mobile/native/src/db/mod.rs
@@ -1,4 +1,6 @@
 use crate::api;
+use crate::db::models::Order;
+use crate::trade;
 use anyhow::anyhow;
 use anyhow::Context;
 use anyhow::Result;
@@ -13,6 +15,7 @@ use diesel_migrations::MigrationHarness;
 use state::Storage;
 use std::sync::Arc;
 use time::OffsetDateTime;
+use uuid::Uuid;
 
 mod custom_types;
 pub mod models;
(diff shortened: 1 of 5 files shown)


mobile/native/src/db/models.rs
```diff
@@ -108,16 +107,16 @@ impl Order {
         }
     }
 
-    pub fn get(order_id: Uuid, conn: &mut SqliteConnection) -> QueryResult<Order> {
+    pub fn get(order_id: String, conn: &mut SqliteConnection) -> QueryResult<Order> {
```
bonomat, line 151: I remember why I left this here as a Uuid: because I failed to store the Uuid in the DB and wanted to pick it up again. :D I think a string here is fine though .

mobile/native/src/db/models.rs
```diff
@@ -313,6 +427,25 @@ pub mod test {
         let loaded_order = Order::get(uuid.to_string(), &mut connection).unwrap();
         assert_eq!(order, loaded_order);
 
+        Order::update_state(
+            uuid.to_string(),
+            (crate::trade::order::OrderState::Filled {
+                execution_price: 100000.0,
```
bonomat, line 433: mhmm..... good price...

@remind_me_in_6_months

mobile/native/src/db/mod.rs
```diff
@@ -51,3 +54,32 @@ pub fn update_last_login() -> Result<api::LastLogin> {
     let last_login = models::LastLogin::update_last_login(now, &mut db)?.into();
     Ok(last_login)
 }
+
+pub fn insert_order(order: trade::order::Order) -> Result<trade::order::Order> {
+    let mut db = connection()?;
+    let order = Order::insert(order.into(), &mut db)?;
```
bonomat, line 60: In my PR #237 I decided to change this to free-floating functions move these functions into a separate module instead, e.g. `orders::insert`

bonomat (approved):
LGTM.

PR #235, opened 2023-03-08
Add Clippy lint workspace configuration

luckysori:
We've used a similar pattern in past projects.

One thing that has changed is that we use `[target.'cfg(feature = "cargo-clippy")']` instead of `[build]`. This allows us to enforce this lint configuration locally and not just on CI.

Using `rustflags` does mean that devs might encounter unexpected full rebuilds when switching between IDE and command line when using `cargo clippy`. This is a known [issue]. For emacs, I just solved this by setting a different Cargo `target-directory` to the one used by default. A similar approach should work for other IDEs.

[issue]: https://github.com/rust-lang/cargo/issues/8716.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,8 @@
+[target.'cfg(feature = "cargo-clippy")']
+rustflags = [
+  "-Wclippy::dbg_macro",
+  "-Wclippy::print_stderr",
+  "-Wclippy::print_stdout",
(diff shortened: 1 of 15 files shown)


holzeis (approved):
LGTM 👍

da-kami (approved):
lgtm

luckysori:
bors retry

PR #236, opened 2023-03-08
Remove unused `Cargo.lock` files

luckysori:
Since all crates are part of the same workspace, there should only be one `Cargo.lock` file.


bonomat (approved):
very nice :)

PR #237, opened 2023-03-08
feat: maker posts orders to orderbook

### LINKED ISSUE #216: Subscribe to BitMex pricefeed and post orders to orderbook
The maker should subscribe to the BitMEX pricefeed and post new orders to the orderbook

bonomat:
resolves #216

Some reasoning behind the restructoring of the orderbook: 
The orderbook is meant to become its own binary eventually, hence, I decided to go for a feature-based folder structure and moved the db into the orderbook module of the coordinator. This will allow us to simply _extract_ the orderbook later on. I've also went with a nice idea @da-kami came up with: hide the orderbook models completely and expose only free floating functions while passing in business-logic models. 
This again gives us a nice decoupled architecture allowing us to easily switch the database if wanted. 

Open todo: there are a few `expect`/`unwraps` all over the place. I plan to tackle this in a follow-up PR.


maker/src/cli.rs
```diff
@@ -17,6 +18,10 @@ pub struct Opts {
     /// Where to permanently store data, defaults to the current working directory.
     #[clap(long)]
     data_dir: Option<PathBuf>,
+
+    /// The HTTP addresss for the orderbook.
```
holzeis, line 22: nit: typo on **s** to much in addresss

maker/src/trading/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    };
+    let mut price_stream = bitmex_client::bitmex(network).await;
+
+    let mut last_bid = None;
+    let mut last_ask = None;
+
+    while let Some(quote) = price_stream.try_next().await? {
+        tracing::debug!("Received new quote {quote:?}");
+
+        last_bid = update_order(
```
holzeis, line 27: ❓ I guess we are not adding a spread yet?

maker/src/trading/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+            orderbook_url.clone(),
+            quote.ask(),
+            Direction::Long,
+            maker_id.clone(),
+            last_bid,
+            Default::default(),
```
holzeis, line 33: 🔧  Shouldn't we also create an order with some contracts? I guess this line will a 0 decimal.

maker/src/trading/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    orderbook_url: Url,
+    price: Decimal,
+    direction: Direction,
+    maker_id: String,
+    last_order: Option<OrderResponse>,
+    quantity: Decimal,
+) -> Option<OrderResponse> {
+    if let Some(last_order) = last_order {
+        let order_id = last_order.id;
+        if let Err(err) = orderbook_client::delete_order(orderbook_url.clone(), order_id).await {
```
holzeis, line 60: 🤔 Do we need to take care of some race conditions here? What if the coordinator is currently trying to setup a dlc_channel with a trader while this order is deleted?

holzeis (approved):
LGTM 👍

maker/src/trading/mod.rs
bonomat: Definitely 🙈

maker/src/trading/mod.rs
bonomat: Not yet :)

maker/src/trading/mod.rs
bonomat: If we are currently already setting up the DLC channel then all good. More interesting is what if the orderbook matches 1sec before that? 
I guess in this case the maker will need to reject the match request from the orderbook towards him.

PR #238, opened 2023-03-08
Feat/persist order

da-kami:
This drives the order forward.

The order lifecylcle for happy path scenarios should be complete with this PR.
Note that this also updates some parts of the position lifecycle, as we now have an order that we can retrieve data from for the position.

 We also handle order failure scenarios *before* the execution of the DLC protocol; failure scenarios (e.g. reject) after protocol execution are not handled yet. I would do that on top of this work, because it will touch how we query the dlc manager for dlcs.

---

Notes on the lifecylcle:

- We currently assume that there can only be one `Order` in execution in the app. If one `Order` is being executed the next `Order` has to wait until the first one finished.
- We don't retry anything. If we fail when triggering the trade setup we always just set the `Order` to `Failed` (with a failure reason) and the user would have to try again.
- A position is only created /update only once a (new) DLC exists; this is when the `Order` lifecycle ends in a happy path scenario (the order is in final state `Filled` then). Note that we don't create a position prior to having a DLC - this is something that always bugged me in ItchySats. I think we finally achieve that the position is *bound* to the existence of the DLC :)

--- 

The structure of "which handler calls which other handler" (i.e. function calls in `position::handler` and `order::handler`) and the responsibilities for handling errors might need another iteration. I basically just plugged this together to see it work and would gradually make it better now.

### DIFF
--- a/mobile/native/src/api.rs
+++ b/mobile/native/src/api.rs
@@ -103,12 +103,6 @@ pub async fn submit_order(order: NewOrder) -> Result<()> {
 }
 
 #[tokio::main(flavor = "current_thread")]
-pub async fn get_order(id: String) -> Result<Order> {
-    let order = order::handler::get_order(id).await?.into();
-    Ok(order)
-}
-
-#[tokio::main(flavor = "current_thread")]
 pub async fn get_orders() -> Result<Vec<Order>> {
     let orders = order::handler::get_orders_for_ui()
         .await?
--- a/mobile/native/src/api.rs
+++ b/mobile/native/src/api.rs
@@ -110,7 +110,7 @@ pub async fn get_order(id: String) -> Result<Order> {
 
 #[tokio::main(flavor = "current_thread")]
 pub async fn get_orders() -> Result<Vec<Order>> {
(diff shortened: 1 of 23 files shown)


da-kami:
Opening this PR because it's basically done - I will test by clicking through the UI after lunch, so there might be some minor changes pushed on top. The overall approach should not change anymore.

mobile/native/src/api.rs
```diff
@@ -110,7 +110,7 @@ pub async fn get_order(id: String) -> Result<Order> {
 
 #[tokio::main(flavor = "current_thread")]
 pub async fn get_orders() -> Result<Vec<Order>> {
-    let orders = order::handler::get_orders()
+    let orders = order::handler::get_orders_for_ui()
```
holzeis, line 107: nit: This naming convention is a bit strange. 😅 Could we place ui specifc requests in to a dedicated sub module so that it reads something like that?

```
let orders = order::handler::ui::get_orders();
```

or potentially only use some `.into()` magic?

holzeis (approved):
LGTM 👍

PR #240, opened 2023-03-08
feat: maker configurable network

holzeis:
Adds the network cli flag to the maker bin.

### DIFF
--- a/maker/src/cli.rs
+++ b/maker/src/cli.rs
@@ -19,15 +19,41 @@ pub struct Opts {
     #[clap(long)]
     data_dir: Option<PathBuf>,
 
+    #[clap(value_enum, default_value = "regtest")]
+    pub network: Network,
+
     /// The HTTP address for the orderbook.
     #[clap(long, default_value = "http://localhost:8000")]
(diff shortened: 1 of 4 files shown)


maker/src/cli.rs
```diff
@@ -17,6 +17,28 @@ pub struct Opts {
...
+    Signet,
+    Testnet,
+    Mainnet,
+}
+
+impl From<Network> for bitcoin::Network {
+    fn from(network: Network) -> Self {
+        match network {
+            Network::Regtest => bitcoin::Network::Regtest,
+            Network::Signet => bitcoin::Network::Signet,
```
luckysori, line 37: ❓ Should we add more information to the `Signet` and `Regtest` variants to identify the actual instance of `signet` or `regtest`?

luckysori (approved):

maker/src/cli.rs
holzeis: I guess you are referencing to e.g. what coordinator endpoint, pubkey, etc. to use? We can add this in a follow up PR.

PR #241, opened 2023-03-09
Fix LN wallet double balance

luckysori:
By updating a couple of dependencies.

The bug caused the LN wallet balance to appear to be doubled whilst a DLC channel was open.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,7 +800,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
-source = "git+https://github.com/get10101/rust-dlc?branch=feature/ln-dlc-channels#b3d31730874ffe422d27d4bbdbb508532a5ba8ee"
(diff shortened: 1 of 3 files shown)


luckysori:
> I am not sure if I understand. Do you mean there was a bug in `rust-dlc` and `rust-lightning` that let the ln wallet balance to be doubled? I never noticed while testing.
> 
> What bug are you referencing to?

Yep. The `available` balance was fine, but the `pending_close` balance was wrong. In tests we never look at pending close, but I did when testing adding the DLC channel and it was wrong. See [this matrix thread](https://matrix.to/#/!YOUZSBmuJuaGzmxzaA:matrix.org/$m2XEsrhu9uy9cumOoPVmQerD2w053d-rZnH8mMgA0ME?via=matrix.org) for more context.

holzeis (approved):
Nice 👍

PR #244, opened 2023-03-09
feat: Initiate the dlc channel setup from the coordinator

### LINKED ISSUE #243: Initiate the dlc channel setup from the coordinator
I've updated the diagram according to our discussion this morning. Note, due to the additional `Confirm` message that was originally not reflected in the diagram, the trade would have had the last word in the dlc channel creation. As a consequence the Coordinator needs to initiate the trade. This will also allow us to simplify things, as the coordinator is now really in control of coordinating the trade.

However, we need to provide the matched order details to the trader, simply open a dlc channel with out order context is not sufficient. This is not reflected in the diagram yet.

```mermaid
sequenceDiagram

participant m as Maker
participant c as Coordinator
participant o as Orderbook
participant t as Taker

m ->> o: create limit order

t ->> o: market order
o -->> o: match maker and taker

o ->> c: execute match

par Coordinator to Taker
    c ->> t: Propose DLC channel open
    t -->> c: Accept
and Coordinator to Maker
    c ->> m: Propose DLC channel open
    m -->> c: Accept 
end

par Coordinator to Taker
    c ->> t: Confirm
and Coordinator to Maker
    c ->> m: Confirm
end


Note over m,t: ...time flies when making money....

m ->> o: create limit order

t ->> o: "reversed" market order
o -->> o: match maker and taker

o ->> c: execute match

par Trader to Coordinator

c ->> t: Propose DLC channel close
t -->> c: Accept 
and Maker to Coordinator

c ->> m: Propose DLC channel close
m -->> c: Accept 
end

par Coordinator to Taker
    c ->> t: Confirm
and Coordinator to Maker
    c ->> m: Confirm
end
```

Side note, in the diagram, below we are automatically closing the position with the maker. This is fine for the MVP purpose, but post MVP we could find another match taking the original limit order from the maker so that we don't have to close the dlc channel with the maker.

_Originally posted by @holzeis in https://github.com/get10101/10101/issues/181#issuecomment-1461622475_

holzeis:
Checklist:

- [x] Submit order to orderbook
- [x] Refactor TradeParams to reflect a Match
- [x] Initiate dlc channel proposal from coordiantor
- [x] Accept dlc channel proposal from the taker
- [x] Accept dlc channel proposal from the maker
- [x] Manually broadcast confirm message once both taker and maker have accepted the dlc channel
- [ ] Publish order filling and order filled updates
- [ ] Testing

resolves #243


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -167,6 +167,42 @@ pub fn run(data_dir: String) -> Result<()> {
...
+                            tracing::error!(peer = %coordinator_pubkey.to_string(), "Unable to retrieve DLC channel offer: {e:#}");
+                            continue;
+                        }
+                    };
+
+                    tracing::info!(%coordinator_pubkey, "Found DLC channel offer");
+
+                    let channel_id = sub_channel.channel_id;
+
+                    // todo: the app should validate if the offered dlc channel matches it's submitted order.
```
da-kami, line 195: ❓How do we associate incoming DLC requests with orders? 

In the previous flow the client was in charge of triggering the `trade` based on the orderbook reporting a match. That match was for an order, and thus the client could identify the order.
The client then triggerd the execution, and only allowed one trade execution at the time - so if multiple match events from the orderbook would have come in, the client was in charge of scheduling the execution.

Now, the coordinator is in charge of scheduling the execution. If the client submitted multiple order to the orderbook, how would the client know which order is being executed here and update the order state accordingly?

By adapting this, the order / position lifecycle updates were removed, I think we should add them again so we can find an answer to the question above :)

Possibilities:

- The client asks the coordinator what is being executed
- The coordinator communicates the order id that is being executed to the client
- ~~The client asks the orderbook what was matched~~ -> I think this does not work, because the orderbook does not execute, and several oders could have been matched, so we run into the same problem

My conclusion is that the information what order is in execution has to come from the coordinator. We can easily validate that the trade parameters that the coordinator proposes match the order in question, so I think this is fine.

da-kami (commented):
Thanks for driving this forward, I left some comments

mobile/native/src/trade/order/handler.rs
```diff
@@ -1,46 +1,41 @@
...
-        order_id_counterparty: "the orderbook will know".to_string(),
-        contract_symbol: ContractSymbol::BtcUsd,
-        leverage: order.leverage,
-        leverage_counterparty: 2.0,
-        quantity: order.quantity,
-        execution_price: 55_000.0,
-        // in 24h
-        expiry: Duration::from_secs(60 * 60 * 24),
-        oracle_pk: ln_dlc::get_oracle_pubkey()?,
+    // todo: this model should include the contract_symbol
```
bonomat, line 18: Either this or you make the `contract_symbol` part of the route, e.g. 
`.../api/orderbook/orders/btcusd`

mobile/native/src/trade/order/handler.rs
```diff
@@ -1,46 +1,41 @@
...
+    let new_order = trade::NewOrder {
```
bonomat, line 19: We will need to know if this is a market order or a limit order

mobile/native/src/trade/order/handler.rs
```diff
@@ -1,46 +1,41 @@
...
+        price: dec!(22_000), // todo: replace with actual btcusd rate price
+        quantity: Decimal::try_from(order.quantity)?,
+        maker_id: "".to_string(), // todo: what should be provided on the maker_id?
```
bonomat, line 22: This was just a dummy field. Imho this should be the sender's pk, e.g. if maker posts the limit order, then this is the maker's pk, if it is a market order by the trader, then this is the trader's pk.

coordinator/src/node.rs
```diff
@@ -17,63 +17,108 @@ use std::sync::Arc;
...
+            let node = self.inner.clone();
+            async move {
+                let maturity_time = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs()
+                    + match_params.expiry.as_secs();
+
+                let contract_symbol = match_params.contract_symbol.label();
+
+                let margin_trader = calculate_margin(
+                    match_params.execution_price,
+                    match_params.quantity as f64, // todo: the quantity should not be f64
```
bonomat, line 90: What else should it be?

coordinator/src/node.rs
```diff
@@ -17,63 +17,108 @@ use std::sync::Arc;
...
+                    trade.leverage as f64,        // todo: the leverage should not be f64
```
bonomat, line 91: Unless we want to rule out fractional leverages this can be something else, otherwise I think this has to be a float.

crates/trade/src/lib.rs
```diff
@@ -1,60 +1,71 @@
...
-    /// Our leverage
+    /// The leverage of the trading party either maker or taker.
+    pub leverage: u8,
+
+    /// The direction of the trading party either long or short.
     ///
-    /// This has to correspond to our order's leverage.
-    pub leverage: f64,
+    /// Note, this needs to be reversed for the counter party.
+    pub direction: Direction,
```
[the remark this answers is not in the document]
bonomat: I think you are right, the `direction` might not be needed here.

crates/trade/src/lib.rs
```diff
@@ -42,44 +41,31 @@ pub struct MatchParams {
...
+    pub order_id: i32,
+}
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Match {
     /// The quantity to be used
     ///
     /// This quantity may be the complete amount of either order or a fraction.
-    pub quantity: f64,
+    pub quantity: u64,
```
bonomat, line 68: This quantity needs to support fractional amounts, see comment here: [`ab06660` (#244)](https://github.com/get10101/10101/pull/244/commits/ab06660ec958ee826ba9e018dcf7163c3ae939bd#r1131880986)

crates/trade/src/lib.rs
```diff
@@ -11,11 +11,36 @@ use uuid::Uuid;
...
+/// A match represents the matched orders of a taker and a maker and will be executed by the
+/// coordinator.
 ///
-/// Emitted by the orderbook when a match is found.
-/// Both trading parties will receive trade params and then request trade execution with said trade
-/// parameters from the coordinator.
+/// Emitted by the orderbook to the coordinator for execution when a match is found.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MatchParams {
+    /// The taker trade information
```
bonomat, line 27: I think we need to be able to identify which order was matched because the trader might have multiple orders pending.

maker/src/bin/maker.rs
```diff
@@ -14,9 +14,11 @@ use rand::thread_rng;
 use rand::RngCore;
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 use tracing::metadata::LevelFilter;
 
 const ELECTRS_ORIGIN: &str = "tcp://localhost:50000";
+const PROCESS_TRADE_REQUESTS_INTERVAL: Duration = Duration::from_secs(30);
```
bonomat, line 21: Can you add a comment why `30` seconds were chosen? it's also worth mentioning if it was a random choice

mobile/native/src/ln_dlc/mod.rs
bonomat: I still think that the orderbook should notify the trader as well and tell them: 
_hey, your order X,Y,Z have been matched, expect a DLC channel of size Q coming from coordinator c_.

bonomat (commented):
I think the flow is pretty much what we discussed. One critical question remains open though: [`2f2d435` (#244)](https://github.com/get10101/10101/pull/244/commits/2f2d435ad6bad05b0b4cc949e5f3a42f3af4489e#r1131821959)

mobile/native/src/ln_dlc/mod.rs
da-kami: > I still think that the orderbook should notify the trader as well and tell them:
> hey, your order X,Y,Z have been matched, expect a DLC channel of size Q coming from coordinator c.

We can do that, but I think we additionally need a way to get information from the coordinator about which execution (i.e. DLC) is for which order. I don't think the orderbook telling us is enough - especially for a maker that potentially has several takers trading with him at the same time. We have to be able to clearly associate one order with one DLC, and I don't think the orderbook has this information in this flow.

mobile/native/src/trade/order/handler.rs
da-kami: I'm in favor of adding it as field - this allows us a more data driven approach, where the context is clear from the data that was sent.

mobile/native/src/trade/order/handler.rs
da-kami: This is how the `NewOrder` is represented on the API for the app; I think this should be complete. Likely we will need a similar representation when sending an order to the orderbook.

```
#[frb]
#[derive(Debug, Clone)]
pub struct NewOrder {
    #[frb(non_final)]
    pub leverage: f64,
    #[frb(non_final)]
    pub quantity: f64,
    #[frb(non_final)]
    pub contract_symbol: ContractSymbol,
    #[frb(non_final)]
    pub direction: Direction,
    // Box needed for complex enum, otherwise generated Rust code complains about Default impl
    // missing
    #[frb(non_final)]
    pub order_type: Box<OrderType>,
}

```

coordinator/src/node.rs
da-kami: We *need* fractional leverage for a combined positions! This has to stay f64 from my point of view.

crates/trade/src/lib.rs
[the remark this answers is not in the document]
holzeis: Hmmm.. I've seen that it was not added in the TradeParams model, but figured it was simply forgotten.

If it's not needed, how will we be able to know if the trader wants to buy or sell?

mobile/native/src/trade/order/handler.rs
```diff
@@ -1,46 +1,41 @@
...
+        .post(format!(
+            "http://{}/api/orderbook/orders",
+            config::get_http_endpoint()
+        ))
+        .json(&new_order)
+        .send()
+        .await
+        .context("Failed submit order to order book")?;
+
+    // TODO: the order has not changed, why do we need to update the order here?
```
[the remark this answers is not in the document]
holzeis: Got it, thanks for the explanation. 👍

crates/trade/src/lib.rs
holzeis: Sorry, the match params ended up in the wrong commit, the trade model (introduced in the next commit) includes the order id.

mobile/native/src/trade/order/handler.rs
holzeis: @da-kami why is the quantity of type `f64`? 

and I guess the `order_type` is `limit` or `market`?

mobile/native/src/trade/order/handler.rs
```diff
@@ -1,46 +1,41 @@
...
     };
-    // TODO: Remove this call once we plug in the orderbook; we trigger trade upon being matched
-    // then
-    position::handler::trade(dummy_trade_params).await?;
+
+    let client = reqwest::Client::new();
+    client
+        .post(format!(
+            "http://{}/api/orderbook/orders",
+            config::get_http_endpoint()
```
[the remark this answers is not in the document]
holzeis: we don't have that config yet on the mobile app. I am going to add it later.

coordinator/src/node.rs
holzeis: fair enough, I wasn't aware of that requirement. In the past we used `u64` for the quantity.

do you think the leverage should also be `f64`?

mobile/native/src/trade/order/handler.rs
holzeis: I see.. Then we should probably name it `trader_id` or something.

coordinator/src/node.rs
holzeis: thanks for the explanation.

mobile/native/src/trade/order/handler.rs
holzeis: what about the `pub_key` of the trader?

maker/src/bin/maker.rs
```diff
@@ -76,6 +78,47 @@ async fn main() -> Result<()> {
         }
     });
 
+    tokio::spawn({
+        let node = node.clone();
+        async move {
+            loop {
+                tokio::time::sleep(PROCESS_TRADE_REQUESTS_INTERVAL).await;
```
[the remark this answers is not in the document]
holzeis: Having the sleep in the beginning has the advantage, that the continues below will not skip the sleep.

mobile/native/src/trade/order/handler.rs
da-kami: True, the pubkey of the trader should be added. On the app it's our own so we know it internally and don't send it to the API; for the orderbook we will have to add it.

da-kami:
I tried to sum up holistically what is the advantage and disadvantage of this new flow against the previous version where the clients were in charge of triggering the trade execution after a match. 
This does not mean we have to change anything, but I wanted to document this somewhere, so I started here. 
Eventually this should be documented in https://github.com/get10101/10101/issues/181.

Advantages:

- The coordinator can fail a trade setup faster if one of the parties is not reachable, because the coordinator reaches out to the parties for trade setup instead of waiting for both parties to reach out.
- If there is a bug in "how the trade is being triggered" we can fix it faster, because the coordinator is in control and we don't have to roll out a new version on the client(s). This is only an advantage for the first message; as soon as something from the client is involved this advantage becomes irrelevant.

Disadvantages:

- The orderbook coming up with a match is decoupled from the execution on the client side. If the orderbook contacts the client for a match, and the client triggers the execution it is clear that this execution is for that match. If the coordinator triggers the execution there is additional validation needed on the client, to ensure the execution is triggered according to the match of the orderbook. This can be overcome by e.g. the client querying the orderbook to retrieve match information. 
- Similarly, the client may not be able to match the `order_id` that is being executed with the client's orders. Further information from the coordinator is required (which has to be validated against the orderbook, see above).

---

Thinking about this made me come to this question: Who is in charge of optimizing the "trades" so that we don't execute everything sequentially - the orderbook or the coordinator?

If we assume that the orderbook, eventually, should be smart enough to batch execution together, i.e. the orderbook should ensure that not all orders are handled sequentially if there are multiple matches, this would mean that the coordinator should actually not question the authority of the orderbook. That could mean that the coordinator does not do any optimizations - in such a case we might as well let the client trigger the execution to overcome the disadvantages listed above. 
If we, however, expect the coordinator to do some form of trade execution optimization then the flow as proposed in this PR will make more sense.

coordinator/src/node.rs
luckysori: What precision do we want here though? I would avoid floats as much as possible.

coordinator/src/node.rs
luckysori: Can we use `Decimal` instead?

da-kami: > Can we use `Decimal` instead?

That might be a good idea!

Note: So far when using `Decimal`, we always use `float` as representation when serializing/deserializing it. It will still be better to use `Decimal` when performing calculations; but just for the sake of communicating information `f64` might actually be fine. We can however still use the `Decimal` type here, because then it becomes an implementation detail how it is serialized / desierialized.

bonomat: In the orderbook I've used `Decimal` in most places.

luckysori:
@holzeis, @da-kami: Shall we close this PR now that we have dealt with these changes in other PRs?

holzeis:
Yes let's close it. I kept it open so we could finish our discussions, but we have moved on already.

PR #247, opened 2023-03-10
Automate dev setup

klochowicz:
after chatting with @da-kami I have not implemented passing in the lightning invoice from the app, if we need to automate that stage, it can be added in a follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If you run into linking troubles when trying to build the coordinator you might 
 RUSTFLAGS='-L /path/to/libpq/lib' cargo build --bin coordinator
 ```
 
-Alternatively you can configure this flag in `.cargo/config.toml` so you don't have to configure it all the time:
+Alternatively you can configure this flag in `~/.cargo/config.toml` so you don't have to configure it all the time:
 
 ```toml
 [target.aarch64-apple-darwin]
@@ -170,24 +170,16 @@ Otherwise, you might have troubles starting 10101, due to port conflicts on cont
 
 1. Start the coordinator with `cargo run --bin coordinator` or `just coordinator`.
 
-   _Ensure that you are using your network ip address and not localhost. This is critical as the docker container will otherwise not be able to reach the coordinator._
-2. Open `http://localhost:8080/faucet/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
-3. Ensure you have enough balance on your bitcoin wallet. Hit the mine button a couple of times if not.
-4. Get a new address of your coordinator by running `curl http://localhost:8000/api/newaddress`
-5. Faucet some coins to your coordinator wallet. Hit the mine button afterwards so the transaction gets into a block.
-6. Open `http://localhost:8080/channel/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
-7. Copy the address of the lnd node and faucet that wallet as described in step 5.
-8. Open a channel with your coordinator (02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@[coordinator ip]:9045) and set a reasonable channel capacity.
-
-   _Note, if you're on mac or windows you can use `host.docker.internal` as coordinator ip._
-9. Mine a few blocks (at least 6) so that the channel gets announced.
+2. Fund and configure coordinator by running `just fund`
 
 #### Fauceting your lightning wallet
 
-10. Create an invoice in your 10101 app by navigating to the receive screen.
+3. Create an invoice in your 10101 app by navigating to the receive screen.
+   _Note, that you have to provide the coordinator host to the mobile app like that `just run`_
 
-    _Note, that you have to provide the coordinator host to the mobile app like that `just run`_
-11. Copy the invoice and enter it on the lightning faucet. Hit send and you will receive your funds momentarily.
+4. Open `http://localhost:8080/faucet/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
+
+5. Copy the invoice and enter it on the lightning faucet. Hit send and you will receive your funds momentarily.
 
 #### Resetting dev environment
 
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -18,6 +18,7 @@ use coordinator_commons::TradeParams;
 use diesel::r2d2::ConnectionManager;
 use diesel::r2d2::Pool;
 use diesel::PgConnection;
+use ln_dlc_node::node::NodeInfo;
 use orderbook_commons::OrderbookMsg;
 use serde::Deserialize;
 use serde::Serialize;
@@ -48,6 +49,7 @@ pub fn router(node: Node, pool: Pool<ConnectionManager<PgConnection>>) -> Router
         .route("/", get(index))
         .route("/api/fake_scid/:target_node", post(post_fake_scid))
         .route("/api/newaddress", get(get_new_address))
+        .route("/api/node", get(get_node_info))
         .route("/api/balance", get(get_balance))
         .route("/api/invoice", get(get_invoice))
         .route("/api/orderbook/orders", get(get_orders).post(post_order))
@@ -95,6 +97,13 @@ pub async fn get_new_address(
             AppError::InternalServerError(format!("Failed to get new address: {e:#}"))
         })?;
     Ok(Json(address.to_string()))
+}
+
+pub async fn get_node_info(
+    State(app_state): State<Arc<AppState>>,
+) -> Result<Json<NodeInfo>, AppError> {
+    let node_info = app_state.node.inner.info;
+    Ok(Json(node_info))
 }
 
 #[derive(Serialize, Deserialize)]
--- a/crates/ln-dlc-node/examples/fund.rs
+++ b/crates/ln-dlc-node/examples/fund.rs
@@ -1,0 +1,223 @@
+extern crate ln_dlc_node;
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+use bitcoin::Amount;
+use ln_dlc_node::node::NodeInfo;
+use local_ip_address::local_ip;
+use reqwest::Response;
+use serde::Deserialize;
+use std::time::Duration;
+use tracing::metadata::LevelFilter;
+use tracing_subscriber::filter::Directive;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+
+const RUST_LOG_ENV: &str = "RUST_LOG";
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ If you run into linking troubles when trying to build the coordinator you might
 RUSTFLAGS='-L /path/to/libpq/lib' cargo build --bin coordinator
 ```
 
-Alternatively you can configure this flag in `.cargo/config.toml` so you don't have to configure it all the time:
+Alternatively you can configure this flag in `~/.cargo/config.toml` so you don't have to configure it all the time:
```
bonomat, line 53: I prefer to use the config within the project folder

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    let lnd_addr: LndAddr = reqwest::get(&format!("{FAUCET_ORIGIN}/lnd/v1/newaddress"))
+        .await
+        .unwrap()
+        .json()
+        .await
+        .unwrap();
+    println!("lightning node: {}", node);
+
+    fund(&lnd_addr.address, Amount::ONE_BTC.checked_mul(2).unwrap())
```
bonomat, line 45: 1 BTC should be plenty 😅

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+        &lnd_addr.address,
+        Amount::ONE_BTC
+            .checked_mul(2)
+            .expect("small integers to multiply"),
+    )
+    .await?;
+    mine(20).await?;
+
+    let lnd_balance = get_text(&format!("{FAUCET}/lnd/v1/balance/blockchain")).await?;
+    tracing::info!("coordinator lightning balance: {}", lnd_balance);
```
1010Tom, line 56: Imho it would be better to do a busy wait until the balance > 0

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -419,11 +421,7 @@ impl Node {
             address,
         };
 
-        tracing::info!(
-            "Lightning node started with node ID {}@{}",
-            node_info.pubkey,
-            node_info.address
-        );
+        tracing::info!("Lightning node started with node ID {}", node_info);
```
1010Tom, line 424: Not sure if this compiles but imho this is the preferred way to print this info.
```suggestion
        tracing::info!(node_info, "Lightning node started");
```

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
+extern crate ln_dlc_node;
```
1010Tom, line 1: Why did you structure it as an `example` and not a standalone crate?

coordinator/src/routes.rs
```diff
@@ -97,6 +99,13 @@ pub async fn get_new_address(
     Ok(Json(address.to_string()))
 }
 
+pub async fn get_node_info(
+    State(app_state): State<Arc<AppState>>,
+) -> Result<Json<NodeInfo>, AppError> {
```
1010Tom, line 104: Does not have to return a result

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+    Ok(())
+}
+
+#[derive(Deserialize)]
+struct LndAddr {
+    address: String,
+}
+
+// Includes some bespoke text processing that ensures we can deserialise the response properly
+async fn get_coordinator_address() -> Result<String> {
```
1010Tom, line 77: I think the problem is that we return the address string as a json. We could fix that and either return a type holding the address or just the address as string.

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+    Ok(get_text(&format!("{COORDINATOR}/api/newaddress"))
+        .await?
+        .strip_prefix('"')
```
1010Tom, line 80: I think the problem is that we return the address string as a json. We could fix that and either return a type holding the address or just the address as string.

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+    post_query(
+        "lnd/v1/peers",
+        format!(
+            r#"{{"addr": {{ "pubkey": "{}", "host": "{host}" }}, "perm":false }}"]"#,
+            node_info.pubkey
+        ),
+    )
+    .await?;
+
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
1010Tom, line 163: Better would be to query lnd for the known peers: https://lightning.engineering/api-docs/api/lnd/lightning/list-peers#rest

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+        "lnd/v1/channels",
+        format!(
+            r#"{{"node_pubkey_string":"{}","local_funding_amount":"{}", "min_confs":1 }}"#,
+            node_info.pubkey,
+            amount.to_sat()
+        ),
+    )
+    .await?;
+
+    mine(10).await?;
```
1010Tom, line 176: We don't need 10 blocks. 
It would be better to only mine the required amount replace the remaining sleep with a busy wait which check for the channel being opened.

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+#[tokio::main]
+async fn main() {
+    init_tracing(LevelFilter::DEBUG).expect("tracing to initialise");
+    fund_everything().await.expect("to be able to fund");
+}
+
+async fn fund_everything() -> Result<()> {
+    let coord_addr = get_coordinator_address().await?;
+    fund(&coord_addr, Amount::ONE_BTC).await?;
+    mine(20).await?;
```
1010Tom, line 30: Why do you mine 20 blocks?  Isn't this waiting uneccesary long (20*5 seconds)?

crates/ln-dlc-node/examples/fund.rs
```diff
@@ -0,0 +1,223 @@
...
+        .await?;
+
+    fund(
+        &lnd_addr.address,
+        Amount::ONE_BTC
+            .checked_mul(2)
+            .expect("small integers to multiply"),
+    )
+    .await?;
+    mine(20).await?;
```
1010Tom, line 53: Why do you mine another 20 blocks here?

1010Tom (approved):
If it works, let's get it in. My comments can be addressed later on or not :)

README.md
```diff
@@ -170,21 +170,14 @@ Otherwise, you might have troubles starting 10101, due to port conflicts on cont
...
-3. Ensure you have enough balance on your bitcoin wallet. Hit the mine button a couple of times if not.
-4. Get a new address of your coordinator by running `curl http://localhost:8000/api/newaddress`
-5. Faucet some coins to your coordinator wallet. Hit the mine button afterwards so the transaction gets into a block.
-6. Open `http://localhost:8080/channel/` (note: ensure to add the trailing `/` as otherwise nginx will try to redirect the call)
-7. Copy the address of the lnd node and faucet that wallet as described in step 5.
-8. Open a channel with your coordinator (02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@[coordinator ip]:9045) and set a reasonable channel capacity.
-
-   _Note, if you're on mac or windows you can use `host.docker.internal` as coordinator ip._
-9. Mine a few blocks (at least 6) so that the channel gets announced.
+2. Fund and configure coordinator by running `just fund`
```
1010Tom, line 173: I would have loved a single command, i.e. `just fund <invoice>` :)

klochowicz: I'll try to add that in a follow-up ! I thought about having it too, but wanted to get something in first that takes the most of the pain away...

crates/ln-dlc-node/examples/fund.rs
klochowicz: it only waits for `5` seconds, it is equivalent to multiple clicks.

20 is an arbitrary large number, might reduce that later on.

da-kami (approved):
I'm happy to merge it, I'll see to test as well.

@klochowicz were you able to actually do a trade after running the setup, or did you always encounter the channel force close problems?

README.md
da-kami: I think we can add that as follow up fairly fast.
Combined with a command that starts everything (https://github.com/get10101/10101/pull/269) this will be easy. 

Then we first start everything, create the invoice in the app and then run `just fund <invoice>`

da-kami:
hm... I started the setup with #269 and I keep running into this error when I run `just fund`:

```
    Finished dev [unoptimized + debuginfo] target(s) in 1m 23s
     Running `target/debug/examples/fund`
2023-03-16T03:22:57.242412Z  INFO fund: Initialized logger
2023-03-16T03:23:02.499007Z  INFO fund: coordinator BTC balance: {"offchain":0,"onchain":100000000}
thread 'main' panicked at 'to be able to fund: error decoding response body: EOF while parsing a value at line 1 column 0

Caused by:
    EOF while parsing a value at line 1 column 0', crates/ln-dlc-node/examples/fund.rs:24:29
error: Recipe `fund` failed on line 55 with exit code 101

```

da-kami (changes requested):
It appears that `coordinator/api/node` changed and that command fails atm.

da-kami (approved):
Ok, after another

```
just wipe
just all
just fund
```

(with `run-everything-dev` rebased on this branch)

it worked - no idea what caused the problem before 😅

da-kami:
One more note:

What `margin` are we supposed to be able to use? I tried with a relatively small margin of `1000` sats, having a `50000` channel, but I ran into this error:

```
2023-03-16T03:49:53.363486Z ERROR native::ln_dlc::accept: Failed to accept subchannel: Invalid parameters were provided: Computed payout is greater than total collateral. channel_id=c49d90be5e086605d3e74fc167400818dd538009e84c64b228121929f30275bd
```

Is that a bug in our payout curve, or related to the setup?

klochowicz:
@1010Tom @da-kami I will merge this now to ensure speedy development time, but I will address the feedback in the follow up PR soon!

klochowicz:
> One more note:
> 
> What `margin` are we supposed to be able to use? I tried with a relatively small margin of `1000` sats, having a `50000` channel, but I ran into this error:
> 
> ```
> 2023-03-16T03:49:53.363486Z ERROR native::ln_dlc::accept: Failed to accept subchannel: Invalid parameters were provided: Computed payout is greater than total collateral. channel_id=c49d90be5e086605d3e74fc167400818dd538009e84c64b228121929f30275bd
> ```
> 
> Is that a bug in our payout curve, or related to the setup?

The number were chosen arbitrarily - they were not specified in the manual setup either - happy to hear suggestions for improvement!

PR #248, opened 2023-03-13
feat: add order matching function

1010Tom:
follow-up PR from #246

Next: 
- [x]  fill in correct types: needs types from https://github.com/get10101/10101/pull/244
- [ ] subscribe to match event in maker/taker/coordinator


coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,223 @@
...
+use std::cmp::Ordering;
+use trade::Direction;
+
+/// Matches a provided market order with limit orders from the DB
+///
+/// If the order is a long order, we return the orders with the highest price
+/// If the order is a short order, we return the orders with the lowest price
+pub fn match_order(
+    order: Order,
+    db_connection: &mut PooledConnection<ConnectionManager<PgConnection>>,
```
luckysori, line 19: 🔧 It would be better if this was written as a pure function without involving the database. Loading the orders from the DB should happen outside. Then you can write unit tests.

coordinator/src/schema.rs
```diff
@@ -4,18 +4,24 @@ pub mod sql_types {
     #[derive(diesel::sql_types::SqlType)]
     #[diesel(postgres_type(name = "Direction_Type"))]
     pub struct DirectionType;
+
+    #[derive(diesel::sql_types::SqlType)]
+    #[diesel(postgres_type(name = "OrderType_Type"))]
+    pub struct OrderTypeType;
```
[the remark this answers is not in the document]
bonomat: Yes, it's only a db type.

coordinator/src/orderbook/trading.rs
bonomat: Good idea :)

bonomat: [`94d65ee` (#248)](https://github.com/get10101/10101/pull/248/commits/94d65eed8af90a0ccd3fa10a24614f6f9f487e06)

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
     pub price: Decimal,
     pub quantity: Decimal,
-    pub maker_id: String,
+    pub trader_id: String,
     pub direction: Direction,
+    pub order_type: OrderType,
+}
+
+pub struct MatchParams {
+    // todo: waiting for https://github.com/get10101/10101/pull/244 to land
```
da-kami, line 90: Please note that https://github.com/get10101/10101/pull/244 will not be merged; parts of the refactoring might eventually be moved to a separate PR, but the model and flow changes will not be incorporated for now.

The `TradeParams` are defined on `main` already. The output of matching is expected to be one `TradeParams` for each trader.
Note that https://github.com/get10101/10101/pull/228 adds the `Direction` to the `TradeParams` and changes the `expiry` to be a timestamp (`OffsetDateTime`) instead of a `Duration`.

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
+            .parse()
+            .expect("To have a valid pub key");
+
+        match (
+            authenticated_users.get(&maker_public_key),
+            authenticated_users.get(&taker_public_key),
+        ) {
+            (Some(maker_sender), Some(taker_sender)) => {
+                maker_sender
+                    .send(PriceFeedResponseMsg::Match)
```
da-kami, line 137: @luckysori has already pointed this out somewhere I think.
I think we should rename the `PriceFeed` to be something like an `OrderbookFeed` (or similar).

As far as I could see we are not sending any price information on that "feed", but we are currently sending:

- Order(s)
- Matches
- Authentication related messages

I think eventually it would be nice to have an actual price feed, that returns the best bid/ask price (according to the orders); for now I think it's OK to send out the orders, but the name of the feed could reflect better what it is about :D

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
+                    .await
+                    .unwrap();
+
+                taker_sender
+                    .send(PriceFeedResponseMsg::Match)
+                    .await
+                    .unwrap();
+            }
+            (Some(_), None) => {
+                tracing::error!(
```
da-kami, line 147:
```suggestion
                tracing::debug!(
```

da-kami: Or is this *actually unexpected*? I think this is something that can and will happen, so I don't think it's an error.

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
+                    "Could not notify taker about the match - probably not online anymore"
+                )
+            }
+            (None, Some(_)) => {
+                tracing::error!(
```
da-kami, line 152:
```suggestion
                tracing::warn!(
```

da-kami: For the maker this *is* unexpected, but still not an error I think.

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
+                    "Could not notify maker about the match - probably not online anymore"
+                )
+            }
+            (None, None) => {
+                tracing::error!("Could not notify maker nor taker about the match - probably not online anymore")
```
da-kami, line 157:
```suggestion
                tracing::warn!("Could not notify maker nor taker about the match - probably not online anymore")
```

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::orderbook::routes::Order;
+use crate::orderbook::routes::OrderType;
+use anyhow::Result;
+use rust_decimal::Decimal;
+use std::cmp::Ordering;
+use trade::Direction;
+
+/// Matches a provided market order with limit orders from the DB
+///
+/// If the order is a long order, we return the orders with the highest price
```
da-kami, line 11:
```suggestion
/// If the order is a long order, we return the short orders sorted by price (highest first)
```

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,191 @@
...
+/// If the order is a short order, we return the orders with the lowest price
```
da-kami, line 12:
```suggestion
/// If the order is a short order, we return the long orders sorted by price (lowest first)
```

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,191 @@
...
+pub fn match_order(order: Order, all_orders: Vec<Order>) -> Result<Vec<MatchParams>> {
```
da-kami, line 13: 🤔 I am a bit confused - we are not matching the order by `Direction` as far as I can see. 

A long order can only be matched with a short order of the other party!
As far as I can see there is no filter step, but we match against all order.
If we would have only 2 long orders we would actually match them with each other...? Am I missing something? 😅

bonomat: The list of possible orders is passed in and already filtered outside.

1010Tom: >  If we would have only 2 long orders (one market, one limit) we would actually match them with each other...? Am I missing something? 😅

No, if we have 2 long orders they would not match with each other :)

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
+        .map_err(|e| AppError::BadRequest(format!("{e:#}")))?;
 
-    Json(order)
+    Ok(Json(order))
 }
 
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Clone)]
 pub struct NewOrder {
     pub price: Decimal,
```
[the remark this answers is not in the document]
1010Tom: On a `market` order this price does not matter. We can define it as `option` but than we have to check if it is set if we needed, better to just ignore it :)

coordinator/src/orderbook/routes.rs
1010Tom: Already on it

coordinator/src/orderbook/routes.rs
1010Tom: I know that clients can dissapear and come back online but I still think this is an error message here.

coordinator/src/orderbook/trading.rs
da-kami: OK, I followed it back to `all_by_direction_and_type` :) - adding a few comments on how we could potentially improve the readability.

da-kami, line 13:
```suggestion
pub fn match_order(order: Order, opposite_direction_orders: Vec<Order>) -> Result<Vec<MatchParams>> {
```

coordinator/src/orderbook/routes.rs
```diff
@@ -19,57 +27,139 @@ use tokio::sync::broadcast::Sender;
...
-    let sender = state.tx_pricefeed.clone();
-    update_pricefeed(PriceFeedResponseMsg::NewOrder(order.clone()), sender);
+    if new_order.order_type == OrderType::Limit {
+        // we only tell everyone about new limit orders
+        let sender = state.tx_pricefeed.clone();
+        update_pricefeed(PriceFeedResponseMsg::NewOrder(order.clone()), sender);
+    }
 
-    Json(order)
+    let all_orders =
```
da-kami, line 85:
```suggestion
    let opposite_direction_orders =
```

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let matched_orders = match_order(order, all_orders).unwrap();
+
+        assert_eq!(matched_orders.len(), 2);
+        let matched_order = matched_orders.get(0).unwrap();
+        assert_eq!(matched_order.maker_order.id, 1);
+        let matched_order = matched_orders.get(1).unwrap();
+        assert_eq!(matched_order.maker_order.id, 4);
+    }
+}
```
da-kami, line 196: It would be cool to have a test that showcases that we match with opposite direction only.

da-kami (approved):
Nice, it's coming together!

---

One random idea I had on a different model that could potentially simplify things:

We could think about representing `Short` with a minus quantity. Then we would not necessarily need the `Direction`. 
We had this idea in the past at some point - I thought that for the orderbook it could potentially make sense, but I'm not sure if it actually would have advantages over the `Direction`. We would have less decision logic I suppose, but it might not be as easy to understand the code.

coordinator/src/orderbook/trading.rs
1010Tom: [`a9cd5d9` (#248)](https://github.com/get10101/10101/pull/248/commits/a9cd5d9126caf2c5ac3e8effa5c118f4fbd27f2d)

coordinator/src/orderbook/trading.rs
1010Tom: [`a06f033` (#248)](https://github.com/get10101/10101/pull/248/commits/a06f033130d8a28ed2bad7bd209901abc6564df4)

coordinator/src/orderbook/trading.rs
```diff
@@ -16,6 +16,11 @@ pub fn match_order(order: Order, all_orders: Vec<Order>) -> Result<Vec<MatchPara
         return Ok(vec![]);
     }
 
+    let all_orders = all_orders
```
da-kami, line 19: I still think this should not be called `all_orders`, because it is not all orders.

coordinator/src/orderbook/routes.rs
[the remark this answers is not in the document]
da-kami: Fair enough, I thought about representing it like this in the app initially, but I came to the conclusion that it's not correct, and I would not be sure what to set the price to for a market order (or what would happen internally if I do so), so I decided to model the limit price explicitly. 
I'm OK with this simplification of the model here, but we have to be aware that in the app it's already represented in a different way.

coordinator/src/orderbook/trading.rs
1010Tom: happy to do so :)

coordinator/src/orderbook/db/orders.rs
```diff
@@ -88,6 +93,7 @@ struct NewOrder {
 impl From<OrderbookNewOrder> for NewOrder {
     fn from(value: OrderbookNewOrder) -> Self {
         NewOrder {
+            id: Uuid::new_v4(),
```
holzeis, line 96: 🔧 it's a bit unexpected that the transformation will generate a new id upon transforming. I would prefer that a transformation will always end up in the same result.

holzeis (approved):
I left a couple of remarks! None of them are blocking though. Let's get this in and address the remarks in follow up PRs if required.

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut orders = sort_orders(orders, is_long);
+
+    let mut remaining_quantity = order.quantity;
+    let mut matched_orders = vec![];
+    while !orders.is_empty() {
+        let matched_order = orders.remove(0);
+        remaining_quantity -= matched_order.quantity;
+        matched_orders.push(matched_order);
+
+        if remaining_quantity <= Decimal::ZERO {
```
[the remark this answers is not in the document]
1010Tom: Well, it means that the order provided by the maker was bigger than what the taker needs. This will very likely always be the case.

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,223 @@
...
+/// For matching market order and limit order we have to
+/// - take the highest rate if the market order is long
+/// - take the lowest rate if the market order is short
+/// hence, we sort the orders here accordingly
+/// - if long is needed: the resulting vec is ordered ascending.
+/// - if short is needed: the resulting vec is ordered descending.
+///
+/// Note: if two orders have the same rate, we give the earlier order (the one with the lower id)
+/// a higher ordering.
+fn sort_orders(mut orders: Vec<Order>, is_long: bool) -> Vec<Order> {
```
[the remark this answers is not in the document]
1010Tom: I thought about the economic side of things and how centralized exchanges work.

> e.g. filling a maker order of 100 with 100 takers is a bit much, don't you think?

It shouldn't matter given we are on lightning. The cost per trade is theoretical zero (despite the orderbook fee).

coordinator/src/orderbook/routes.rs
```diff
@@ -79,12 +79,53 @@ pub async fn post_order(
...
-    let _matched_orders = match_order(order.clone(), &mut conn).unwrap();
+    if new_order.order_type == OrderType::Limit {
+        // we only tell everyone about new limit orders
+        let sender = state.tx_pricefeed.clone();
+        update_pricefeed(PriceFeedResponseMsg::NewOrder(order.clone()), sender);
+    }
 
-    let sender = state.tx_pricefeed.clone();
-    update_pricefeed(PriceFeedResponseMsg::NewOrder(order.clone()), sender);
+    let matched_orders = match_order(order.clone(), &mut conn).unwrap();
```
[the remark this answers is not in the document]
1010Tom: Interesting idea, I guess this can be optimized later.

coordinator/src/orderbook/trading.rs
```diff
@@ -0,0 +1,223 @@
...
+        let matched_order = orders.remove(0);
+        remaining_quantity -= matched_order.quantity;
+        matched_orders.push(matched_order);
+
+        if remaining_quantity <= Decimal::ZERO {
+            break;
+        }
+    }
+
+    Ok(matched_orders
```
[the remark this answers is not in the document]
1010Tom: jap :D 
We should definitely do that.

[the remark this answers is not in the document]
1010Tom: I'll add this in a follow-up PR.

coordinator/src/orderbook/db/orders.rs
1010Tom: The type `OrderbookNewOrder` does not have an id yet, only once the order is inserted into the DB using the local type `NewOrder` we will get the ID. This is what is happening here. 
Ideally this ID is generated by the db but I wasn't able to get this working.

coordinator/src/orderbook/routes.rs
```diff
@@ -62,10 +64,22 @@ pub async fn get_order(
     Ok(Json(order))
 }
 
+#[derive(Clone)]
 pub struct MatchParams {
-    // todo: waiting for https://github.com/get10101/10101/pull/244 to land
-    pub taker_order: Order,
-    pub maker_order: Order,
+    pub taker_matches: TakerMatchParams,
```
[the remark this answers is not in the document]
1010Tom: In this function there is only one taker, hence, there is only on `TakerMatchParams`. Inside there is `matches` which are the multiple matches.

coordinator/src/orderbook/routes.rs
```diff
@@ -62,10 +64,22 @@ pub async fn get_order(
...
+}
+
+#[derive(Clone)]
+pub struct MakerMatchParams {
+    pub maker_id: PublicKey,
+    pub filled_with: FilledWith,
+}
+
+#[derive(Clone)]
+pub struct TakerMatchParams {
```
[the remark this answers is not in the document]
1010Tom: You are right, this is not necessary, I'll combine them in a follow-up PR.

PR #251, opened 2023-03-13
chore: minor clean-ups

1010Tom:
follow-up from #246

### DIFF
--- a/crates/orderbook-client/src/lib.rs
+++ b/crates/orderbook-client/src/lib.rs
@@ -7,8 +7,7 @@ use futures::Stream;
 use futures::StreamExt;
 use orderbook_commons::create_sign_message;
 use orderbook_commons::Signature;
-use secp256k1::SecretKey;
-use secp256k1::SECP256K1;
+use secp256k1::Message;
 use serde::Serialize;
 use serde_json::to_string;
 use tokio_tungstenite::tungstenite;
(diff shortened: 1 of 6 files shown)


luckysori (approved):
Awesome! Thanks for taking the time to work in my feedback :D

crates/orderbook-client/src/lib.rs
```diff
@@ -25,14 +24,15 @@ pub fn subscribe(url: String) -> impl Stream<Item = Result<String, Error>> + Unp
 /// It subscribes and yields all messages.
 pub fn subscribe_with_authentication(
     url: String,
-    credentials: Credentials,
+    authenticate: &dyn Fn(Message) -> Signature,
```
luckysori, line 27: ❓ Asking because I'm never quite sure: could this be `impl Fn(Message) -> Signature` instead? I might give it a go myself to check.

luckysori (approved):
Nice, I think this is better!

crates/orderbook-client/src/lib.rs
bonomat: The compiler was recommending this as it was complaining about the function not being Sized. 
Give it a try and let me know. I'll wait with merging this :)

luckysori: It just worked with `impl` for me, so I force-pushed the amended commit 🤷

bonomat: mhmmm ok :D  Thanks for giving it a try :)

PR #253, opened 2023-03-14
refactor: Start the lifecycle of the order after submission to orderbook

da-kami:
The previous flow was unnecessarily complicated. The lifecycle of an order should start once we got a reply from the orderbook.

If we succeed to submit the order we assign the `Uuid` returned from the orderbook to the order and safe it in the database. If the fail to submit the order we assign our own `Uuid` to the order and safe it in the database (in `Rejected` state, which is final and we won't query the orderbook with orders in this state). Given that the lifecycle now starts with the result of the submission to the orderbook we don't need an `Initial` state anymore.


luckysori (approved):
Concept ACK.

PR #254, opened 2023-03-14
refactor: Move trade params to `coordinator-commons`

da-kami:
The `TradeParams` are specific to the coordinator, it makes sense to move them to a `coordinator-commons` crate. This has the advantage that we will avoid cyclic dependencies, because we can use `orderbook-commons` to reference the `FilledWith` in `coordinator-commons` and are able to use the global common types of `trade` in other crates.

---

This is in preparation of adapting the `TradeParams` to use `FilledWith` of the `orderbook-commons` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -564,6 +564,7 @@ dependencies = [
  "bdk",
  "bitcoin",
  "clap",
+ "coordinator-commons",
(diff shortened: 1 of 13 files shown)


holzeis (approved):
LGTM 👍

PR #256, opened 2023-03-14
Improve upon close-position PR

luckysori:
Follow-up to https://github.com/get10101/10101/pull/228.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -22,6 +22,7 @@ use std::sync::Arc;
 use trade::cfd::calculate_long_liquidation_price;
 use trade::cfd::calculate_margin;
 use trade::cfd::calculate_short_liquidation_price;
+use trade::cfd::BTCUSD_MAX_PRICE;
 use trade::Direction;
 
 /// The leverage used by the coordinator for all trades.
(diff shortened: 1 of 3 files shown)


luckysori:
As discussed, feel free to take over and merge this, @holzeis.

crates/trade/src/cfd.rs
```diff
@@ -3,6 +3,8 @@ use bdk::bitcoin::Denomination;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 
+pub const BTCUSD_MAX_PRICE: u64 = 1_048_575;
```
holzeis, line 6: nit: I guess we should add a comment here, that this number is not chosen randomly, but is actually a requirement coming from the p2pdoracle we are using at the moment.

🤔 one place were we could put this constant could be on the oracle client, if we choose to wrap it.

coordinator/src/node.rs
```diff
@@ -259,17 +259,16 @@ fn build_contract_descriptor(
 /// function like we used to do in ItchySats.
 fn build_payout_curve(
     total_collateral: u64,
-    execution_price: f64,
+    initial_price: f64,
```
holzeis, line 263: 🤔 I don't think `initial_price` is completely right here, as this function will also be used with the `close_price`. I found `execution_price` to fitting best to reflect both kind of prices.

Note, I agree on the `build_contract_descriptor` function

holzeis (approved):
Thank you 👍

coordinator/src/node.rs
luckysori: It's true that we use it when closing the contract too, but you better use the `initial_price` to construct the payout function if you want it to be the same one again!

PR #260, opened 2023-03-15
Align clippy flags for all builds

klochowicz:
This reduces the chance of cache invalidation between the runs of rust-analyzer, cargo build/run and clippy.

On my machine (with no global Rust config atm), it made huge improvements, ie. no rebuilds from scratch anymore.

There might better ways, but I can't justify spending more time on this right now. Some possible further direction:
 - add a separate build profile for rust-analyzer & clippy


luckysori (approved):
Thanks!

PR #262, opened 2023-03-15
mark orders as taken in db

### LINKED ISSUE #261: Update order as taken in orderbook
After matching an order we should update the db so that the orders are not matched again. 

_Originally posted by @holzeis in https://github.com/get10101/10101/pull/248#discussion_r1135364628_

1010Tom:
resolves #261

### DIFF
--- a/coordinator/src/orderbook/routes.rs
+++ b/coordinator/src/orderbook/routes.rs
@@ -1,4 +1,5 @@
 use crate::orderbook;
+use crate::orderbook::db;
 use crate::orderbook::db::orders;
 use crate::orderbook::trading::match_order;
 use crate::orderbook::trading::notify_traders;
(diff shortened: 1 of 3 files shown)


coordinator/src/orderbook/routes.rs
```diff
@@ -92,17 +93,37 @@ pub async fn post_order(
...
+            .iter()
+            .map(|m| m.order_id)
+            .collect();
+
+        orders_to_set_taken.append(&mut order_ids);
+
         notify_traders(matched_orders, authenticated_users.clone()).await;
+
+        for order_id in orders_to_set_taken {
+            if let Err(err) = db::orders::update(&mut conn, order_id, true) {
```
holzeis, line 122: nit: the method name could be more clear to what this is doing. I would have expected an update method to update the whole order and not just the taken flag. How about `db::orders::taken(&mut conn, order_id, true)`?

holzeis (approved):
LGTM 👍

coordinator/src/orderbook/routes.rs
1010Tom: done

PR #264, opened 2023-03-15
Add capability to easily reset dev environment

klochowicz:
Simply run `just wipe` to reset the dev environment.

### DIFF
--- a/justfile
+++ b/justfile
@@ -57,6 +57,26 @@ clean:
     flutter clean
     cd native && cargo clean
 
+wipe:
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    docker-compose down -v
+    rm -rf data/coordinator/regtest
+    git checkout data/coordinator
+    # Array of possible app data directories (OS dependent)
+    # TODO: Add Linux locations
+    directories=( "$HOME/Library/Containers/finance.get10101.app/Data/Library/Application Support/finance.get10101.app")
+    # Remove all possible app data directories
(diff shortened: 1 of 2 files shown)


klochowicz:
if anyone with m1 Mac could test it, it would be awesome! 

also added a section how to wipe out Simulator data, but it's device dependent :)

justfile
```diff
@@ -57,6 +57,26 @@ clean:
...
 
+wipe:
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    docker-compose down -v
+    rm -rf data/coordinator/regtest
+    git checkout data/coordinator
+    # Array of possible app data directories (OS dependent)
+    # TODO: Add Linux locations
+    directories=( "$HOME/Library/Containers/finance.get10101.app/Data/Library/Application Support/finance.get10101.app")
```
holzeis, line 68: 🤔 This will also remove a directory which is not regtest, correct?

holzeis (approved):
Works on my M1 👍

bonomat (approved):
LGTM

justfile
klochowicz: `git checkout` will restore the seed (it's unfortunately in the regtest dir, the other way would be moving it out of the dir that gets nuked).

the app data will get removed for all, I don't think you can distinguish them in any way. (I would say if we want to do testnet/mainnet, then we should really stick to mobile anyway!).

PR #266, opened 2023-03-15
chore: Make the `PnL` of the `Position` a calculated value

da-kami:
We don't persist the `PnL`, but it is calculated in the UI using the latest price. The calculation is handled in Rust with a calculation endpoint.

@klochowicz this is relevant for wiring in the price events
In dart we will have to consume the price events in the `TradeValueChangeNotifier` and the `PositionChangeNotifier`.


da-kami:
Note on the pnl calculation: I pulled that in from ItchySats and simplified it *fast* to fit our current needs. This code really needs some cleanup and tests 😅

I am planning to finish the position lifecycle tomorrow (with the position in memory instead of DB) and am planning to do a small cleanup round after where I also add tests for this code :)

crates/trade/src/cfd.rs
```diff
@@ -54,3 +58,61 @@ pub fn calculate_short_liquidation_price(leverage: Decimal, price: Decimal) -> D
...
+/// curve determines.
+pub fn calculate_payout_at_price(
+    opening_price: f64,
+    closing_price: f64,
+    quantity: f64,
+    long_leverage: f64,
+    short_leverage: f64,
+    direction: Direction,
+) -> Result<u64> {
+    let opening_price = Decimal::try_from(opening_price).expect("price to fit into decimal");
```
holzeis, line 81: 🤔 Shouldn't we use the same payout function here as we use when closing the position?

crates/trade/src/cfd.rs
```diff
@@ -54,3 +58,61 @@ pub fn calculate_short_liquidation_price(leverage: Decimal, price: Decimal) -> D
...
+
+    // TODO: The fee offset is still missing, for this we need to track fees in the position first,
+    // see ItchySats
+
+    let payout = match direction {
+        Direction::Long => long_margin as i64 + uncapped_pnl_long,
+        Direction::Short => short_margin as i64 - uncapped_pnl_long,
+    };
+
+    let payout = if payout.is_negative() {
```
holzeis, line 110: nit: Wouldn't `cmp.min(0, payout)` work?

holzeis (approved):
LGTM 👍

crates/trade/src/cfd.rs
```diff
@@ -54,3 +58,61 @@ pub fn calculate_short_liquidation_price(leverage: Decimal, price: Decimal) -> D
 
     price * leverage / (leverage - Decimal::ONE)
 }
+
+// TODO: This was copied from ItchySats and adapted; we need tests for this!
+/// Compute the payout for the given CFD parameters at a particular `closing_price`.
```
bonomat, line 63: Can you update the doc please? It contains things which are not relevant anymore.

crates/trade/src/cfd.rs
bonomat: Good point: Either this function calculates the `PnL at price` or the `payout at price`. For the latter one we would need to use the payout funciton to determine the correct payout. 
I'd go for the former one

crates/trade/src/cfd.rs
```diff
@@ -54,3 +58,61 @@ pub fn calculate_short_liquidation_price(leverage: Decimal, price: Decimal) -> D
 
     price * leverage / (leverage - Decimal::ONE)
 }
+
+// TODO: This was copied from ItchySats and adapted; we need tests for this!
```
bonomat, line 62: Yes please, I did not check the functions and trust you to verify this later on.

mobile/native/src/api.rs
```diff
@@ -97,6 +97,21 @@ pub fn calculate_liquidation_price(
...
 }
 
+pub fn calculate_pnl(
+    opening_price: f64,
+    closing_price: f64,
+    quantity: f64,
+    leverage: f64,
+    direction: Direction,
+) -> SyncReturn<u64> {
+    // TODO: Handle the result and don't just return 0
```
klochowicz, line 107: perhaps you could return `Option<u64>` here, that should be easiest to deal with in Dart

mobile/native/src/calculations/mod.rs
```diff
@@ -15,6 +16,28 @@ pub fn calculate_quantity(opening_price: f64, margin: u64, leverage: f64) -> f64
     cfd::calculate_quantity(opening_price, margin, leverage)
 }
 
+pub fn calculate_pnl(
+    opening_price: f64,
+    closing_price: f64,
+    quantity: f64,
+    leverage: f64,
```
klochowicz, line 23: side note: it is pretty scary that we use f64 everywhere and hope we will always make the order right 😢

klochowicz (commented):
thank you for this part, I'll cherry-pick this locally into my price feed work!

mobile/lib/features/trade/position_change_notifier.dart
```diff
@@ -30,6 +36,9 @@ class PositionChangeNotifier extends ChangeNotifier implements Subscriber {
 
     if (event is bridge.Event_PositionUpdateNotification) {
       Position position = Position.fromApi(event.field0);
+
+      position.unrealizedPnl = Amount(_positionService.calculatePnl(position, _price));
```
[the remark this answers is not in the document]
da-kami: that's a good point, I think the price should really always be the current bid and ask and we should then use the potential closing price accordingly.

I would really like to achieve that the PnL that is displayed is also what the user would actually get if they close with said price. In ItchySats there could be discrepancies because we did not use the correct price in some cases.

crates/trade/src/cfd.rs
holzeis: ❓ hmm.. is that the way it's typically done? As a user I would like to see how much money I would get if I sell / buy right now. What's the difference between `PnL at price` and `payout at price`

side note: I know it can still not be guaranteed as there has to be an offer for the market order, but still that would be the closest approximation.

da-kami: `PnL` is the profit and loss, so it's on top of your collateral. 
I think it depends on the trader if they would like to see `PnL` or payout - but we can easily calculate the payout as `collateral + pnl` - if we assume that fees are already calculated in. 

I would display the pnl for no and see what users say

mobile/native/src/calculations/mod.rs
da-kami: indeed, we definitely should do refactoring rounds where we introduce our own types that should preferably wrap Decimal and not float 😬

da-kami:
Note: The test output was validated against our good old Excel sheet `CFD hedging model`

crates/trade/src/cfd.rs
holzeis: but still the underlying calculation is different. the way you describe it is that `payout at price` is actually part of the `pnl at price`. `PnL at price` = collateral + `payout at price`

PR #267, opened 2023-03-15
feat: add maker docker build

holzeis:
- [x] Publish latest maker docker image upon merge to main.
- [x] Replace `GITHUB_TOKEN` with a generated classic token. Note, `GITHUB_TOKEN` does only work when the workflow is triggered manually.

### DIFF
--- a/.github/workflows/publish_docker.yml
+++ b/.github/workflows/publish_docker.yml
@@ -61,3 +61,23 @@ jobs:
           push: true
           build-args: |
           tags: ${{ steps.meta.outputs.tags }}
+
+      - run: cargo build --release --bin maker --target-dir ./target
+
+      - uses: docker/metadata-action@v4
+        id: meta
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=tag
+
+      - name: Build docker image and push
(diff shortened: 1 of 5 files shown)


holzeis:
@bonomat How about building the docker images also for arm64? That way I could use these images on my raspberry pi cluster, without having to build them locally. 😅

.github/workflows/publish_docker.yml
```diff
@@ -61,3 +61,23 @@ jobs:
...
+          tags: |
+            type=ref,event=branch
+            type=ref,event=tag
+
+      - name: Build docker image and push
+        uses: docker/build-push-action@v3
+        with:
+          context: ./maker
+          secrets: |
+            GIT_AUTH_TOKEN=${{ secrets.GITHUB_TOKEN }}
```
bonomat, line 80: I've created a new token for this: 

```suggestion
            GIT_AUTH_TOKEN=${{ secrets.GH_ACTION_TOKEN }}
```

bonomat (approved):
Looks good

bonomat:
> @bonomat How about building the docker images also for arm64? That way I could use these images on my raspberry pi cluster, without having to build them locally. 😅

I think this is generally a good idea :)

bonomat:
@holzeis : I fixed the token for you.

PR #269, opened 2023-03-16
Add command to start everything at once

klochowicz:
added to be able to quickly test things. 
now one can say `just all` and everything will be started (I opted for native build, because the pace of iteration of rebuilding it is just way faster, typically doesn't require rebuilding at all).

side note: perhaps naming it `just start` would be better?


da-kami (approved):
I have `just wipe` and `just all` a try and it worked for me.
I'm about to test `just fund` on top of the started setup.

da-kami (changes requested):
When triggering a trade, I ran into the problem that my order did not get matched!

I think we also have to spin up the `maker` when starting all the systems, otherwise the order will never get matched!

CC @bonomat

klochowicz:
> When triggering a trade, I ran into the problem that my order did not get matched!
> 
> We also have to spin up the `maker` when starting all the systems, otherwise the order will never get matched!
> 
> I think we have to:
> 
> * add `maker` to `just all`
> * add cleaning up the maker data to `just wipe`

sure, I'll do that!

da-kami (approved):

Issue #275, opened 2023-03-16
Attempt to accept DLC only when offer is received


luckysori:
Currently we accept based on the state of the DLC, which can lead to attempting to accept multiple times.

PR #278, opened 2023-03-16
Reply to DLC open/close offers immediately (and only once)

luckysori:
Fixes https://github.com/get10101/10101/issues/275.


crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -75,6 +75,24 @@ impl Node {
...
+        tracing::info!(channel_id = %channel_id_hex, "Rejecting DLC channel offer");
+
+        let (node_id, accept_sub_channel) = self
+            .sub_channel_manager
+            .accept_sub_channel(channel_id)
+            .map_err(|e| anyhow!(e.to_string()))?;
+
+        self.dlc_message_handler.send_message(
+            node_id,
+            Message::SubChannel(SubChannelMessage::Accept(accept_sub_channel)),
```
holzeis, line 90: 🤔 This looks strange. Is `Accept` really the sub channel message that we are supposed to send here? 😅

mobile/native/src/ln_dlc/node.rs
```diff
@@ -36,78 +33,107 @@ impl Node {
...
+                        }
+                    }
+
+                    if let Some(reply_msg) = reply_msg {
+                        tracing::debug!(
+                            to = %node_id,
+                            msg = %sub_channel_message_as_str(&reply_msg),
+                            "Sending DLC channel message"
+                        );
+                        self.inner
```
holzeis, line 99: side note, we will have to delay the sending of the confirm messages here.

holzeis (approved):
Nice :) 👍

crates/ln-dlc-node/src/node/dlc_channel.rs
luckysori: Oh, good catch. I was investigating if there was a way to reject subchannel open/close offers (there is none currently) and left this hanging around by mistake. It's not used and it's obviously wrong, so I'll remove it.

mobile/native/src/ln_dlc/node.rs
luckysori: Confirm is sent by the coordinator though.

Issue #132, opened 2023-02-14
Price Feed and handling price updates in the UI


da-kami:
Scope: 

- [x] Price feed (from the orderbook) in Rust backend that streams prices to UI; it would be great if we can structure this as separate crate again so we can re-use it in different systems;  we can likely just pull in code from the poc or itchysats
- [x] Price events in rust that are being picked up by `FlutterSubscriber` to be streamed to the UI
- [x] Subscribe to price events in `TradeValuesChangeNotifier` (see how we subscribe and handle other events in `EventService` in dart)

Note: Eventually we might have *other* change notifiers that depend on the price, but for now subscribing to for `TradeValuesChangeNotifier` will be good enough.

da-kami:
@klochowicz I think this is a good first issue :)

bonomat:
Just to clarify the focus of this ticket: 
We will have two pricefeeds (for the time being)
1. BitMex
2. Orderbook

While the former one is only for showing a nice candle chart, the latter one for showing a rate to the user.

I'm understanding that this ticket focuses on 1.

bonomat:
This ticket is about streaming the orders from the orderbook to the app.

klochowicz:
This ticket has been done.
See PRs:  #326 #300

PR #281, opened 2023-03-16
feat: in memory position that can be closed in UI

da-kami:
This is in line with our current simplified flow:

When a position is created we store it in memory.
When the position is closed we remove it. Updating (i.e. extending / reducing) a position is not implemented. 

---

Note that for closing we remove the position in user interface for now; there is no particular trade history planned, but the opening and closing of the position will be visible in the transaction history of the wallet.

Note: The PnL will only be displayed (at the moment it's constantly loading) once we have the price feed #132

### DIFF
--- a/mobile/lib/features/trade/position_list_item.dart
+++ b/mobile/lib/features/trade/position_list_item.dart
@@ -129,7 +129,18 @@ class PositionListItem extends StatelessWidget {
                     : () async {
                         await onClose();
                       },
-                child: const Text("Close Position"),
+                child: notNullPosition.positionState == PositionState.closing
+                    ? Row(
+                        children: const [
+                          SizedBox(
+                            width: 10,
+                            height: 10,
+                            child: CircularProgressIndicator(),
+                          ),
+                          Text("Closing ...")
+                        ],
+                      )
+                    : const Text("Close Position"),
               ),
             ],
           ),
--- a/mobile/lib/features/trade/position_list_item.dart
+++ b/mobile/lib/features/trade/position_list_item.dart
@@ -7,9 +7,10 @@ import 'package:get_10101/features/trade/trade_theme.dart';
 import 'contract_symbol_icon.dart';
 
 class PositionListItem extends StatelessWidget {
-  const PositionListItem({super.key, required this.position});
+  const PositionListItem({super.key, required this.position, required this.onClose});
 
   final Position? position;
+  final Function onClose;
 
   @override
   Widget build(BuildContext context) {
@@ -123,8 +124,8 @@ class PositionListItem extends StatelessWidget {
                 width: 10,
               ),
               ElevatedButton(
-                onPressed: () {
-                  // TODO: bottom sheet to close position
+                onPressed: () async {
+                  await onClose();
                 },
                 child: const Text("Close Position"),
               ),
--- a/mobile/lib/features/trade/position_list_item.dart
+++ b/mobile/lib/features/trade/position_list_item.dart
@@ -124,9 +124,11 @@ class PositionListItem extends StatelessWidget {
                 width: 10,
               ),
               ElevatedButton(
-                onPressed: () async {
-                  await onClose();
-                },
+                onPressed: notNullPosition.positionState == PositionState.closing
+                    ? null
+                    : () async {
+                        await onClose();
+                      },
                 child: const Text("Close Position"),
               ),
             ],
--- a/mobile/native/src/db/models.rs
+++ b/mobile/native/src/db/models.rs
@@ -294,7 +294,7 @@ impl Position {
             .execute(conn)?;
 
         if effected_rows == 0 {
-            bail!("Could not update order")
+            bail!("Could not update position")
         }
 
         Ok(())
@@ -302,7 +302,6 @@ impl Position {
 
     // TODO: This is obviously only for the MVP :)
     /// deletes all positions in the database
-    /// Deletes given order from DB, in case of success, returns > 0, else 0 or Err
     pub fn delete_all(conn: &mut SqliteConnection) -> QueryResult<usize> {
         diesel::delete(positions::table).execute(conn)
     }
--- a/mobile/native/src/db/models.rs
+++ b/mobile/native/src/db/models.rs
@@ -282,6 +282,24 @@ impl Position {
         positions::table.load(conn)
     }
 
+    /// updates the status of the given order in the db
+    pub fn update_state(
+        contract_symbol: ContractSymbol,
+        state: PositionState,
+        conn: &mut SqliteConnection,
+    ) -> Result<()> {
+        let effected_rows = diesel::update(positions::table)
+            .filter(schema::positions::contract_symbol.eq(contract_symbol))
+            .set(schema::positions::state.eq(state))
+            .execute(conn)?;
+
+        if effected_rows == 0 {
+            bail!("Could not update order")
+        }
+
+        Ok(())
+    }
+
     // TODO: This is obviously only for the MVP :)
     /// deletes all positions in the database
     /// Deletes given order from DB, in case of success, returns > 0, else 0 or Err
@@ -491,6 +509,7 @@ pub enum FailureReason {
     NodeAccess,
     NoUsableChannel,
     ProposeDlcChannel,
+    CannotExtendOrReduce,
 }
 
 impl From<FailureReason> for crate::trade::order::FailureReason {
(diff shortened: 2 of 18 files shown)


klochowicz:
@da-kami looking into the issue now

da-kami:
Note: I came to the conclusion that I can also wire the `Close` button of the position in with this change; so that we literally trigger a close and not an extend / reduce of the position.
We could even go as far as to restrict the user interface so that the user can only `sell` once they have a `long` position, and only `buy` once they have a `short` position. This could be added as part of adding validation for now, so we restrict the behavior of the user interface to what the system is currently capable of :)

da-kami:
Note: Using the `Storage` as I intended in this PR does actually *not* work, because one cannot set the value after it was initialized by set (setting it again will just return `false` and not change the value).

So, I implemented saving the position into the database quickly :)

da-kami:
Note: Tests with the previous version using `Storage` already worked until the point where we tried to re-set the storage (which did not work).
Will re-test with the database in place and then autosquash once that is done :)

mobile/native/src/trade/order/handler.rs
```diff
@@ -16,6 +17,11 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
```
luckysori, line 18: 🙃 Would be good to move this next to where it is used.

mobile/native/src/trade/position/handler.rs
```diff
@@ -49,50 +50,69 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
-        contract_symbol: ContractSymbol::BtcUsd,
-        direction: Direction::Long,
-        average_entry_price: 20000.0,
-        liquidation_price: 14000.0,
-        position_state: PositionState::Open,
-        collateral: 2000,
-    };
+/// Update position once an order was filled
+///
+/// This crates or updates the position.
```
luckysori, line 58: 🔧 Typo.

```suggestion
/// This creates or updates the position.
```

mobile/native/src/trade/position/handler.rs
```diff
@@ -49,50 +50,69 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
-        leverage: 2.0,
-        quantity: 10000.0,
-        contract_symbol: ContractSymbol::BtcUsd,
-        direction: Direction::Long,
-        average_entry_price: 20000.0,
-        liquidation_price: 14000.0,
-        position_state: PositionState::Open,
-        collateral: 2000,
-    };
+/// Update position once an order was filled
```
luckysori, line 56: 🔧 Looks like a copy-paste error.

```suggestion
/// Update position once an order was submitted.
```

mobile/native/src/trade/position/handler.rs
```diff
@@ -49,50 +50,69 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
+    Ok(())
 }
 
-pub fn order_filled(filled_order: Order, collateral: u64) -> Result<()> {
-    // TODO: Persist the position
-    // TODO: Decide if we have to create or update the position:
-    //  Probably best to have a "insert or update" function for the db that returns the position
+/// Update position once an order was filled
+///
+/// This crates or updates the position.
```
luckysori, line 78: 🔧 Typo.

```suggestion
/// This creates or updates the position.
```

mobile/native/src/trade/position/handler.rs
```diff
@@ -49,50 +50,69 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
-        liquidation_price: 14000.0,
-        position_state: PositionState::Open,
-        collateral: 2000,
-    };
+/// Update position once an order was filled
+///
+/// This crates or updates the position.
+/// If the position was closed we set it to `Closed` state.
+pub fn update_position_after_order_submitted(submitted_order: Order) -> Result<()> {
+    if let Some(position) = db::get_positions()?.first() {
```
luckysori, line 61: 🙃 It would be a bit more correct if we `find` the first position with the same `contract_symbol`.

mobile/native/src/trade/position/handler.rs
```diff
@@ -49,50 +50,69 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
-    // TODO: Decide if we have to create or update the position:
-    //  Probably best to have a "insert or update" function for the db that returns the position
+/// Update position once an order was filled
+///
+/// This crates or updates the position.
+/// If the position was closed we set it to `Closed` state.
+pub fn update_position_after_order_filled(filled_order: Order, collateral: u64) -> Result<()> {
+    // We don't have a position yet
+    if db::get_positions()?.is_empty() {
+        tracing::debug!("We don't have a position at the moment, creating it");
```
luckysori, line 83: 🔧 Would be good to log the `filled_order` and the collateral.

luckysori (approved):
LGTM!

mobile/native/src/event/api.rs
```diff
@@ -35,10 +37,23 @@ impl From<EventInternal> for Event {
...
+            EventInternal::PositionCloseNotification(contract_symbol) => {
+                Event::PositionClosedNotification(PositionClosed { contract_symbol })
+            }
         }
     }
 }
 
+/// Wrapper struct needed by frb
+///
+/// The mirrored `ContractSymbol` does not get picked up correctly when using it directly as
```
holzeis, line 49: ❓hmm, what error do you get when you try to mirror the `ContractSymbol` enum?

mobile/native/src/trade/position/handler.rs
```diff
@@ -65,7 +69,21 @@ pub async fn get_positions() -> Result<Vec<Position>> {
...
 
-pub fn order_filled(filled_order: Order, collateral: u64) -> Result<()> {
+// TODO: Remove this temporary in-memory storage of the position and safe it in the database
+static POSITION: Storage<Arc<Mutex<Option<Position>>>> = Storage::new();
+
+pub(crate) fn get() -> MutexGuard<'static, Option<Position>> {
+    POSITION
+        .get()
+        .lock()
+        .expect("failed to get lock on event hub")
```
holzeis: nit: typo "failed to get lock on the positions"

holzeis (approved):
LGTM 👍 

Note, I was a little bit confused with the title of the PR and the actual implementation 😅

da-kami:
> LGTM 👍
> 
> Note, I was a little bit confused with the title of the PR and the actual implementation 😅

It changed quite a bit on the way - sorry, forgot to update the title and description in the end 😅

mobile/native/src/db/models.rs
```diff
@@ -242,6 +243,102 @@ impl TryFrom<Order> for crate::trade::order::Order {
     }
 }
 
+#[derive(Queryable, QueryableByName, Insertable, Debug, Clone, PartialEq)]
+#[diesel(table_name = positions)]
+pub(crate) struct Position {
```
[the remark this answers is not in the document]
da-kami: Yeah, that file is getting quite big. I would refactor this separately once we agreed on what we would like to see.
At the moment the database is not part of the "feature first" structure.
We could also move each `DAO` into a `db` module in the respective feature module.

mobile/native/src/event/api.rs
da-kami: I reported this separate when I initially opened the PR and pinged you there as well :)
Eventually I found a way to overcome it which is wrapping the struct. I don't have the exact error anymore and would not invest time into solving this at this point :)

If you are keen to see the error you can:

- Remove the wrapper struct, use `trade::ContractSymbol` directly in the enum varient directly.
- run `just gen`
- run clippy, the problem will be reported in the generated code

mobile/lib/features/trade/position_list_item.dart
```diff
@@ -123,9 +124,11 @@ class PositionListItem extends StatelessWidget {
...
               ),
               ElevatedButton(
-                onPressed: () {
-                  // TODO: bottom sheet to close position
-                },
+                onPressed: notNullPosition.positionState == PositionState.closing
+                    ? null
+                    : () async {
+                        await onClose();
+                      },
```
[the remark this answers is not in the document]
da-kami: I tested it again and it does work, but it might not happen immediately.
To make this better we would have to e.g. track the state in widget using widget state independent of the state of the position. Upon pressing the button we would set it to disabled internally - this could be released by e.g. a timer. The position's state would disable the button independently. 

I did not add this logic yet.

PR #283, opened 2023-03-17
make things more configurable so that they run on our testnet vm

bonomat:


coordinator/src/bin/coordinator.rs
luckysori, line 1: Great file!

luckysori (approved):

PR #284, opened 2023-03-17
chore: move all dev setup to dev readme

da-kami:
We started having information split up in multiple files; the main README and the DEVELOPMENT readme. This may be quite confusing, because the information in neither file is complete.

In an attempt to consolidate information I opted for:

- The main README describes the project at a higher level; eventually we can also add some more user friendly information on how to run it.
- The DEVELOPMENT readme describes what is relevant for the developer.

---

Note: I am not 100% sure we need separate DEVELOPMENT readme - but I'm not against it. 
In the past we said that the project on GitHub is for devs, and our users should usually not go there, but they should be served through the website / app-store. So, the README on GitHub always reflected the dev setup. For `comit-rs` we had an additional section on top of the README that described the project more high level.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,116 +4,125 @@ This document specifies how to start up `10101` as a developer and how to run th
 For getting a better understanding of the software architecture decisions you can refer to our [architecture decision records (ADRs)](/docs/readme.adoc).
 Aimed primarily for developers.
 
-## 10101 app
+## Getting started
 
-### Flutter Rust Bridge
+10101 consists of sevaral systems:
 
-10101 uses Rust and Flutter, and leverages [flutter-rust-bridge](https://github.com/fzyzcjy/flutter_rust_bridge) to generate code.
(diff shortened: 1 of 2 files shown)


klochowicz:
hm... I'm not sure whether this is in line with my intention when I tried to split the files. 

My idea was splitting them as follows:
- `README.md` - enough info to actually try out the project as a dev (ie. give it a spin).
- `DEVELOPMENT.md` - detailed info on things when I want to dig deeper into it, or contribute.

This was dictated by the realisation that users have a limited attention span, and putting *everything* there makes it harder to actually pick the relevant bits for a user.

Perhaps we should unify them again @da-kami ? maybe just reorg would suffice.

da-kami:
Converted to deaft because I don't have the time to work on this atm.
I am happy with combining it into one README again and making that README better.

klochowicz:
@da-kami I'm closing this for now as the last activity in this ticket was over 2 months ago and it looks stale.... feel free to reopen when you'd like to push it over the line.

PR #285, opened 2023-03-20
Fixing bugs appearing during environment setup

luckysori:


docker-compose.yml
```diff
@@ -156,13 +156,19 @@ services:
...
 
   adminer:
     container_name: adminer
     image: adminer
     restart: always
     ports:
       - 8888:8080
+    networks:
+      vtto:
+        ipv4_address: 10.5.0.9
```
bonomat, line 171: If noone is using this container we can actually remove it. I've only used it during the initial work at the db :)

bonomat (approved):
Thanks for the fixes

docker-compose.yml
luckysori: Are you referring to the `adminer`? I'll remove it if it's not being used at all.

bonomat: yes

luckysori: Done: https://github.com/get10101/10101/pull/285/commits/bc64ddc82ebdbb9a0ce8365a32078cfbb36066ec.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -116,9 +116,18 @@ pub fn run(data_dir: String) -> Result<()> {
         let node = Arc::new(Node { inner: node });
 
         // todo: should the library really be responsible for managing the task?
```
holzeis, line 118: nit: I guess it should not, but then we can remove the comment 🙂

PR #292, opened 2023-03-20
fix: Mount correct paths to volumes

holzeis:
Looks like we did not mount the bitcoin and lnd folders correctly, which should have lead to issues when restarting these services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       - "28332:28332"
       - "28333:28333"
     volumes:
-      - bitcoin:/config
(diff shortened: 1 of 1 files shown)


bonomat (approved):
good find.

docker-compose.yml
```diff
@@ -58,8 +58,6 @@ services:
     ports:
       - "50000:50000"
       - "30000:30000"
-    volumes:
```
bonomat, line 61: Not critical: electrs could use a data volume as well so that it does not have to reindex the whole chain on a restart.

luckysori (approved):

PR #294, opened 2023-03-21
feat: release to testflight description

1010Tom:
While @bonomat seems to be the only one being able to publish to testflight at the moment I still think it has value to have this checked in.


justfile
```diff
@@ -210,4 +211,12 @@ wait-for-electrs-to-be-ready:
         fi
     done
 
+build-ipa:
+    cd mobile && flutter build ipa --dart-define="ELECTRS_ENDPOINT=${ELECTRS_ENDPOINT}" --dart-define="COORDINATOR_P2P_ENDPOINT=${COORDINATOR_P2P_ENDPOINT}"
+
+publish-testflight:
+    cd mobile && xcrun altool --upload-app --type ios --file ./build/ios/ipa/10101.ipa --apiKey ${ALTOOL_API_KEY} --apiIssuer ${ALTOOL_API_ISSUER}
+
+release-testfligth: gen ios build-ipa publish-testflight
```
klochowicz, line 220: nit: typo `flight` instead of `fligth` (here and in readme).

you can probably shorten this to `TestFlight` and add a comment in the line above `# build an iOS binary and release to TestFlight` (when you query `just --list`, then the comments appear too!)

RELEASE.md
```diff
@@ -0,0 +1,51 @@
+# How to release
+
+## Docker
+
+Docker images are being released up on push to `main` at the moment.
+
+As of now, there is no other way to release a different versio!
```
klochowicz, line 7: nit: typo

```suggestion
As of now, there is no other way to release a different version!
```

RELEASE.md
```diff
@@ -0,0 +1,51 @@
...
+
+```bash
+just publish-testflight
+```
+
+For a single command to execute all this in one go, you can use:
+
+```bash
+just release-testfligth
+```
```
klochowicz, line 49: I would probably describe that at the top, otherwise everyone will follow the longer process 😄

klochowicz (approved):
nice & clear instructions!

PR #296, opened 2023-03-21
Harmonize order state handling

da-kami:
Just some minor cleanup

Harmonize how the order state is updated.
Initially we opted for not exposing `Rejected` because we would exit with an error in the order submission dialog; however, it is not guaranteed that the user will see this error. To keep things simple we map the internal `Rejected` state to `Failed` in the user interface, and update the order accordingly. This guarantees that the user will know that the order failed.

### DIFF
--- a/mobile/native/src/trade/order/api.rs
+++ b/mobile/native/src/trade/order/api.rs
@@ -102,9 +102,9 @@ impl From<order::OrderState> for OrderState {
             order::OrderState::Initial => unimplemented!(
                 "don't expose orders that were not submitted into the orderbook to the frontend!"
             ),
-            order::OrderState::Rejected => unimplemented!(
-                "don't expose orders that were rejected by the orderbook to the frontend!"
-            ),
+            // TODO: At the moment the UI does not depict Rejected, we map it to Failed; for better
+            // feedback we should change that eventually
+            order::OrderState::Rejected => OrderState::Failed,
             // We don't expose this state, but treat it as Open in the UI
             order::OrderState::Filling { .. } => OrderState::Open,
         }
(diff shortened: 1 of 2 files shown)


mobile/native/src/trade/order/api.rs
```diff
@@ -102,9 +102,8 @@ impl From<order::OrderState> for OrderState {
             order::OrderState::Initial => unimplemented!(
                 "don't expose orders that were not submitted into the orderbook to the frontend!"
             ),
-            order::OrderState::Rejected => unimplemented!(
-                "don't expose orders that were rejected by the orderbook to the frontend!"
-            ),
+            // TODO: At the moment the UI does not depict Failed, we could change that
```
luckysori, line 105: ❓ Did you mean that it doesn't depict `Rejected`?

mobile/native/src/trade/order/handler.rs
```diff
@@ -27,16 +27,11 @@ pub async fn submit_order(order: Order) -> Result<()> {
...
+        update_order_state(order.id, OrderState::Rejected)?;
         bail!("Could not post order to orderbook");
     }
-    db::update_order_state(order.id, OrderState::Open)?;
 
-    let order = Order {
-        state: OrderState::Open,
-        ..order
-    };
-    ui_update(order);
```
luckysori, line 39: ❓ Why do we no longer update the UI with the `order` here? I don't think the commit message talks about this change.

da-kami: We do! `update_order_state` handles the UI update, that's why I used it here.

da-kami: I added another commit that might help :)

luckysori: Ah, nice. Thank you. I clearly didn't understand the first version 😛

luckysori (approved):

klochowicz (approved):
👍

PR #297, opened 2023-03-21
feat: show app info in settings

bonomat:
I found myself wondering if the environment variables are working as expected and quickly added some app info into the settings. 

Design could need some love but it's functional: 

image

### DIFF
--- a/mobile/lib/common/settings_screen.dart
+++ b/mobile/lib/common/settings_screen.dart
@@ -1,6 +1,8 @@
 import 'package:flutter/material.dart';
 import 'package:get_10101/features/trade/settings_screen.dart';
 import 'package:get_10101/features/wallet/settings_screen.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
+import 'package:provider/provider.dart';
 
 class SettingsScreen extends StatelessWidget {
   const SettingsScreen({required this.fromRoute, super.key});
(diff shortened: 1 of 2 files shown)


mobile/lib/common/settings_screen.dart
```diff
@@ -1,9 +1,11 @@
 import 'package:flutter/material.dart';
 import 'package:get_10101/features/trade/settings_screen.dart';
 import 'package:get_10101/features/wallet/settings_screen.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
 
 class SettingsScreen extends StatelessWidget {
-  const SettingsScreen({required this.fromRoute, super.key});
+  final bridge.Config config;
+  const SettingsScreen({required this.fromRoute, super.key, required this.config});
```
holzeis, line 8: 🔧 The config is registered as provider so instead of passing it through as an argument you could have simple used it like this.

```
final config = context.read<bridge.Config>();
```

holzeis (approved):

mobile/lib/common/settings_screen.dart
bonomat: Nice, thanks for this :)

da-kami (approved):
I'm happy to adapt the design - I would use `ValueDataRow` for displaying the data.
Would extend it with adding `ValueType` `text` and defining a flag that defines if the text should be `selectable`. That way we have it re-usable. 

I'm happy to do that in a follow-up :)

PR #300, opened 2023-03-21
Amend model to provide the prices from the orderbook

klochowicz:
The orderbook can query the orders to provide the "best" orders for a given contract symbol.

I had to deal with the situation when there's no price (either bid or ask).

For the simplicity of wiring things together, I'm only piping through the BTCUSD
    best prices into Flutter for now.

PR for Flutter is incoming soon (excluded from this commit, as I'm still working on that part) - in the meantime it would be good to get some feedback on this backend code.

There was a bit of complexity that I had to tackle because:
- there might be no orders for particular direction (which means there's no bid or ask price from orderbook),
- initial misunderstandings on how orderbooks work (thanks @da-kami for clarification on that).

### DIFF
--- a/crates/orderbook-commons/src/price.rs
+++ b/crates/orderbook-commons/src/price.rs
@@ -1,0 +1,137 @@
+use crate::Order;
+use crate::ToPrimitive;
+use rust_decimal::Decimal;
+use serde::Deserialize;
+use serde::Serialize;
+use std::collections::HashMap;
+use trade::ContractSymbol;
+use trade::Direction;
+
+#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq)]
+pub struct Price {
+    pub bid: Option<Decimal>,
+    pub ask: Option<Decimal>,
+}
+
+pub type Prices = HashMap<ContractSymbol, Price>;
+
+/// Best prices across all current orders for given ContractSymbol in the orderbook
+/// Taken orders are not included in the average
+pub fn best_current_price(current_orders: &[Order]) -> Prices {
+    let mut prices = HashMap::new();
+    let mut add_price_for_symbol = |symbol| {
+        prices.insert(
+            symbol,
+            Price {
+                bid: best_bid_price(current_orders, symbol),
+                ask: best_ask_price(current_orders, symbol),
+            },
+        );
+    };
+    add_price_for_symbol(ContractSymbol::BtcUsd);
+    prices
+}
+
+/// Best price (highest) of all long (buy) orders in the orderbook
+fn best_bid_price(current_orders: &[Order], symbol: ContractSymbol) -> Option<Decimal> {
+    best_price_for(current_orders, Direction::Long, symbol)
+}
+
+/// Best price (lowest) of all short (sell) orders in the orderbook
+fn best_ask_price(current_orders: &[Order], symbol: ContractSymbol) -> Option<Decimal> {
+    best_price_for(current_orders, Direction::Short, symbol)
+}
+
+fn best_price_for(
+    current_orders: &[Order],
+    direction: Direction,
+    symbol: ContractSymbol,
+) -> Option<Decimal> {
+    assert_eq!(
+        symbol,
+        ContractSymbol::BtcUsd,
+        "only btcusd supported for now"
+    );
+    let use_max = direction == Direction::Long;
+    current_orders
+        .iter()
+        .filter(|order| !order.taken && order.direction == direction)
+        .map(|order| order.price.to_f64().expect("to represent decimal as f64"))
+        // get the best price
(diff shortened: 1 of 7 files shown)


crates/orderbook-commons/src/price.rs
```diff
@@ -0,0 +1,137 @@
...
+pub struct Price {
+    pub bid: Option<Decimal>,
+    pub ask: Option<Decimal>,
+}
+
+pub type Prices = Box<HashMap<ContractSymbol, Price>>;
+
+/// Best prices across all current orders for given ContractSymbol in the orderbook
+/// Taken orders are not included in the average
+pub fn best_current_price(current_orders: &[Order]) -> Prices {
```
holzeis, line 20: nit: It feels odd that this function is also enforcing that taken orders are not included. It feels like this should rather be a responsibility of the calling component. e.g. only fetch orders that are not taken and pass them to the `best_current_price`

mobile/native/src/orderbook.rs
```diff
@@ -65,6 +66,13 @@ pub fn subscribe(secret_key: SecretKey) -> Result<()> {
                                     tracing::error!("Trade request sent to coordinator failed. Error: {e:#}");
                                 }
                             },
+                            OrderbookMsg::AllOrders(orders) => {
+                                // TODO: How to trigger this more often? this comes only initially
```
holzeis, line 70: I think the idea is that any subsequent order data will be updated through delta events.

- `OrderbookMsg::NewOrder(Order)`
- `OrderbookMsg::DeleteOrder(Uuid)`
- `OrderbookMsg::Update(Order)`

The client will then have to update her local list correspondingly.

holzeis (approved):
LGTM 👍

mobile/native/src/orderbook.rs
klochowicz: ouch, that's a tough one, because having that being piped through might be really heavy.

might be best to just ask for the best price in the context of all these events after all...

crates/orderbook-commons/src/price.rs
klochowicz: after quite a bit of thinking, I realised the taken orders by definition cannot be part of the price feed (because one *cannot* use them). 

is there any situation where this might be useful to present it as part of the price feed? this API call specifically filters the orders to get the best price, one can still access all orders if they want to.

crates/orderbook-commons/src/price.rs
```diff
@@ -0,0 +1,137 @@
...
+        prices.insert(
+            symbol,
+            Price {
+                bid: best_bid_price(current_orders, symbol),
+                ask: best_ask_price(current_orders, symbol),
+            },
+        );
+    };
+    add_price_for_symbol(ContractSymbol::BtcUsd);
+    Box::new(prices)
```
bonomat, line 32: Why does it have to be box?

mobile/native/src/orderbook.rs
bonomat: As Richie said: this event is only fired on first opening the websocket connection, afterwards you only get the diff with `NewOrder`, `DeleteOrder` and `UpdateOrder`. 

Wouldn't it be easy to just maintain a list of orders in this file and whenever there is an update you check if it is worth calling `position::handler::price_update...` or not?

klochowicz: great idea. I will maintain the the list of orders here - assembling the full picture out of a delta would be too laborious in flutter.

crates/orderbook-commons/src/price.rs
klochowicz: I thought I would be smart by ensuring it's on the heap, but after reading the docs, `HashMap` is *always* on the heap regardless. 🙈 
removed.

mobile/native/src/orderbook.rs
```diff
@@ -38,6 +39,9 @@ pub fn subscribe(secret_key: SecretKey) -> Result<()> {
             Signature { pubkey, signature }
         };
 
+        // Consider using a HashMap instead to optimize the lookup for removal/update
```
da-kami, line 42: 🙃 I'd action that TODO now - we can still easily extract the `values` of the hashmap as `Vec` if that's more convenient for `best_current_price`. Don't mind much though.

da-kami (approved):

mobile/native/src/orderbook.rs
klochowicz: I will amend the `best_current_price` to also take the hashmap, because it's easier - but after I have it end-to-end working, otherwise it's just an optimisation :-)

note: even `AllOrders` would be best as HashMap, because then we can specify stricter contract (ie. no duplicates!).

PR #305, opened 2023-03-21
chore: Use same price for pnl and collab settlement

### LINKED ISSUE #301: Don't use payout curve when settling collaboratively
We should use the price for displaying PnL and for settling collaboratively.

holzeis:
resolves #301 

This uses the pnl calculation to derive the `accept_settlement_amount`. However, since we do not have the opening price at hand, the calculation will still not be correct. For this we need to load the position from the database (which should store the opening price).

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -25,6 +25,7 @@ use ln_dlc_node::PeerManager;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 use std::sync::Arc;
+use trade::cfd;
 use trade::cfd::calculate_long_liquidation_price;
 use trade::cfd::calculate_margin;
 use trade::cfd::calculate_short_liquidation_price;
@@ -110,34 +111,24 @@ impl Node {
             "Closing position"
         );
 
-        let margin_trader = margin_trader(trade_params);
-        let margin_coordinator = margin_coordinator(trade_params);
-
         let leverage_long = leverage_long(trade_params);
         let leverage_short = leverage_short(trade_params);
 
-        let total_collateral = margin_coordinator + margin_trader;
+        let closing_price = trade_params.average_execution_price();
 
-        // FIXME: This is wrong as we cannot use the closing price to
-        // rebuild the payout function. We must save the initial price
-        // when creating the position and use it here again for
-        // closing.
-        let initial_price = trade_params.average_execution_price();
+        // FIXME: This is wrong as we cannot use the closing price to calculated the
+        // `accept_settlement_amount`. We must save the initial price when creating the position and
+        // use it here again for closing.
+        let opening_price = closing_price;
 
-        let payout_function = build_payout_function(
-            total_collateral,
-            initial_price,
+        let accept_settlement_amount = calculate_accept_settlement_amount(
+            opening_price,
+            closing_price,
+            trade_params.quantity,
             leverage_long,
             leverage_short,
+            trade_params.direction,
         )?;
-
-        let accept_settlement_amount = payout_function
-            .to_range_payouts(total_collateral, &get_rounding_intervals())
-            .map_err(|e| anyhow!("{e:#}"))?
-            .iter()
-            .find(|p| trade_params.weighted_execution_price() < Decimal::from(p.start + p.count))
-            .map(|p| p.payout.accept)
-            .context("Failed to find payout.")?;
 
         tracing::debug!(
             "Settling position of {accept_settlement_amount} with {}",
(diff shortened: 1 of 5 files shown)


coordinator/src/node.rs
```diff
@@ -201,6 +192,34 @@ enum TradeAction {
...
+    let pnl = cfd::calculate_pnl(
+        opening_price,
+        closing_price,
+        quantity,
+        long_leverage,
+        short_leverage,
+        direction,
+    )?;
+
+    let margin_trader = calculate_margin(opening_price, quantity, long_leverage);
```
da-kami, line 215: 🛠️ I think it is incorrect to assume that the `margin_trader` is always using the `long_leverage`, we would have to determine the leverage to be used based on the direction I suppose.

coordinator/src/node.rs
```diff
@@ -110,35 +111,25 @@ impl Node {
...
 
-        // FIXME: This is wrong as we cannot use the closing price to
-        // rebuild the payout function. We must save the initial price
-        // when creating the position and use it here again for
-        // closing.
-        let initial_price = trade_params.average_execution_price();
+        // FIXME: This is wrong as we cannot use the closing price to calculated the
+        // `accept_settlement_amount`. We must save the initial price when creating the position and
+        // use it here again for closing.
+        let closing_price = opening_price;
```
da-kami, line 122: 🛠️ We have that in the position, it's the `average_entry_price` - because the position can be filled by multiple orders with different prices; there is no `opening_price`, but multiple of them, so we called it `average_entry_price`.

coordinator/src/node.rs
```diff
@@ -32,14 +32,7 @@ use trade::cfd::BTCUSD_MAX_PRICE;
...
 /// The leverage used by the coordinator for all trades.
-///
-/// TODO: In case of leverage 1.0 the liquidation will be set to
-/// 21_000_000. I guess this will result in issues when the overall
-/// upper boundary is smaller than that. It looks like the payout
-/// function will never finish in that case. Setting this value
-/// temporarily to 2.0 so that the liquidation price is below that
-/// total upper boundary.
-const COORDINATOR_LEVERAGE: f64 = 2.0;
+const COORDINATOR_LEVERAGE: f64 = 1.0;
```
luckysori, line 36: 👍 Ah, thanks. I thought I changed this but I didn't!

coordinator/src/node.rs
```diff
@@ -208,17 +192,45 @@ enum TradeAction {
     // Reduce,
 }
 
+/// Calculates the accept settlement amount based on the pnl.
+///
+/// The smallest possible result is 0.
```
luckysori, line 197: The type already encodes this information 😛

coordinator/src/node.rs
```diff
@@ -208,17 +192,45 @@ enum TradeAction {
...
+        closing_price,
+        quantity,
+        long_leverage,
+        short_leverage,
+        direction,
+    )?;
+
+    let margin_trader = calculate_margin(opening_price, quantity, long_leverage);
+
+    let accept_settlement_amount = (margin_trader as i64).saturating_add(pnl);
```
luckysori, line 217: I'm unsure if this saturating add is needed at this stage. It's slightly safer, but are we really gonna hit `i64::MAX`? I would just add the PNL to make it more readable.

---

For reference:

```rust
i64::MAX == 9_223_372_036_854_775_807
```

coordinator/src/node.rs
```diff
@@ -208,17 +192,45 @@ enum TradeAction {
...
+    // the amount can only be positive, adding a safeguard here with the max comparison to
+    // ensure the i64 fits into u64
+    Ok(accept_settlement_amount.max(0) as u64)
```
luckysori, line 220: 👍 As a small nit, I would prefer if an important operation was on a line other than the return position. Otherwise it's too easy to ignore.

coordinator/src/node.rs
```diff
@@ -403,3 +415,90 @@ pub fn process_incoming_messages_internal(
 
     Ok(())
 }
+
+#[cfg(test)]
+pub mod tests {
```
luckysori, line 420: 👍 You might want to look into `proptest` or `quickcheck` to generate enough test cases to give you more confidence in the properties you are checking.

luckysori (approved):
LGTM apart for the 2 things that Daniel already pointed out.

coordinator/src/node.rs
holzeis: indeed 🙈 good find 👍

coordinator/src/node.rs
holzeis: but we don't store the position on the coordinator yet. so we can't get this price during closing yet.

coordinator/src/node.rs
holzeis: fair enough 🙈

coordinator/src/node.rs
da-kami: hm... that's unfortunate - can we just store them in memory in a hashmap for now as a quick fix? If that's too much work feel free to merge as is :)

da-kami (approved):

coordinator/src/node.rs
holzeis: I'll merge as is and address your comment in a follow up PR.

PR #307, opened 2023-03-22
Feat/wallet history

da-kami:
It does work - if nothing goes "wrong" :D


crates/ln-dlc-node/src/ln_dlc_wallet.rs
```diff
@@ -52,10 +55,33 @@ impl LnDlcWallet {
...
+            .context("Can't acquire lock on bdk_ldk wallet")?
+            .list_transactions(false)
+            .context("bar")?;
+
+        txs.sort_by(|a, b| {
+            b.confirmation_time
+                .as_ref()
+                .map(|t| t.height)
+                .cmp(&a.confirmation_time.as_ref().map(|t| t.height))
+        });
```
luckysori, line 81: This sorting is a bit superfluous for the task at hand because we later combine all the transactions, payments and trades and sort them again. But I thought it might be good to sort this regardless in case we ever need to consume the list of on-chain transactions separately.

crates/ln-dlc-node/src/node/wallet.rs
```diff
@@ -104,4 +108,57 @@ impl Node {
...
+            flow: PaymentFlow::Outbound,
+            amount_msat: info.amt_msat.0,
+            timestamp: info.timestamp,
+        });
+
+        let mut payments = inbound_payments
+            .chain(outbound_payments)
+            .collect::<Vec<_>>();
+
+        payments.sort_by(|a, b| b.timestamp.cmp(&a.timestamp));
```
luckysori, line 147: Same as for on-chain transactions: This sorting is a bit superfluous for the task at hand because we later combine all the transactions, payments and trades and sort them again. But I thought it might be good to sort this regardless in case we ever need to consume the list of off-chain payments separately.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -291,20 +295,20 @@ impl Node {
 
         // TODO: Persist inbound payment info to disk
         let inbound_payments: PaymentInfoStorage = Arc::new(Mutex::new(HashMap::new()));
+        // TODO: Persist outbound payment info to disk
+        let outbound_payments: PaymentInfoStorage = Arc::new(Mutex::new(HashMap::new()));
```
luckysori, line 299: Persisting this is now covered in https://github.com/get10101/10101/issues/314.

mobile/native/src/api.rs
```diff
@@ -37,25 +37,28 @@ pub struct Balances {
...
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub enum WalletType {
-    OnChain,
-    #[default]
-    Lightning,
+    OnChain { txid: String },
+    Lightning { payment_hash: String },
```
luckysori, line 60: I couldn't figure out if there was a way to identify the sender or the receiver of the payment, so it was finally omitted.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -23,17 +26,22 @@ use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
+use time::OffsetDateTime;
 use tokio::runtime::Runtime;
```
luckysori, line 31: 🔧 Unwanted newline.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -162,6 +169,134 @@ pub fn run(data_dir: String) -> Result<()> {
...
+    let trades = {
+        let orders = crate::db::get_filled_orders()
+            .context("Failed to get filled orders; skipping update")?;
+
+        orders.into_iter().enumerate().map(|(i, order)| {
+            let flow = if i % 2 == 0 {
+                api::PaymentFlow::Outbound
+            } else {
+                api::PaymentFlow::Inbound
+            };
```
luckysori, line 264: 🔧 I should have probably added a TODO about how insanely hacky this is. Although it works very well for the time being since we only support an order history that goes `[open, close, open, close, open, ...]`.

mobile/native/src/trade/order/mod.rs
```diff
@@ -123,6 +124,18 @@ impl Order {
             }
         }
     }
+
+    /// This returns the trader's margin once known (based on the execution price).
+    pub fn trader_margin(&self) -> Option<u64> {
+        let opening_price = self.execution_price()?;
+
+        // QUESTION: Is the leverage always the trader's leverage or the self_leverage?
```
luckysori, line 132: 🔧 Can remove this if someone answers haha.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -162,6 +169,134 @@ pub fn run(data_dir: String) -> Result<()> {
...
+
+            let amount_sats = order
+                .trader_margin()
+                .expect("Filled order to have a margin");
```
luckysori, line 268: 🤔 The `amount_sats` of a trade varies whether we are closing a position or not:

- When opening a position we "pay" for the margin, so this calculated value would be correct.
- When closing a position we "earn" the payout (margin plus or minus PNL), so this calculated value would be _wrong_.

We've captured this bug in #311.

luckysori (approved):
I only commented on my Rust code to point out some shortcomings.

Since I've witnessed the thing working on the UI I'm content with the Dart changes, but I haven't reviewed them in depth.

mobile/native/src/ln_dlc/mod.rs
da-kami: 😄 I guess we should capture these hacks in tickets that are attached to feature epics eventually.

The feature epic here is: Support extend/reduce the position - once we do that this hack has to be fixed.

mobile/native/src/trade/order/mod.rs
da-kami: I am confused by the question. `self` in this context is owned by the `trader`for me:
Since the order is by the trader (owned by the trader) that order's leverage is always the trader's leverage.

luckysori: I think I worded this poorly. Let me try again:

Does the `Order` always represent the view of the party holding it? Or is it always fixed to the trader's perspective?

From your answer I understand that it depends on who holds it.

In conclusion, let's remove this question if it's still there.

PR #316, opened 2023-03-22
Add a refresh indicator so that we can update the wallet manually

da-kami:
Needs testing on the phone

### DIFF
--- a/mobile/lib/features/wallet/wallet_screen.dart
+++ b/mobile/lib/features/wallet/wallet_screen.dart
@@ -35,82 +35,87 @@ class _WalletScreenState extends State<WalletScreen> {
     );
 
     return Scaffold(
-      body: Padding(
-        padding: const EdgeInsets.symmetric(horizontal: 20),
-        child: Column(
-          crossAxisAlignment: CrossAxisAlignment.stretch,
-          children: [
-            ExpansionPanelList(
-              children: [
-                ExpansionPanel(
-                  headerBuilder: (BuildContext context, bool isExpanded) {
-                    return Row(
-                      children: [
-                        // https://stackoverflow.com/a/70192038 - do not know if this is principled
-                        const SizedBox(
-                            width:
-                                64), // ExpansionPanelList IconContainer size: end margin 8 + padding 16*2 + size 24),
-                        Expanded(
-                          child: Center(
-                              child: AmountText(
-                                  amount: walletChangeNotifier.total(),
-                                  textStyle: const TextStyle(
-                                      fontSize: 20.0, fontWeight: FontWeight.bold))),
-                        )
-                      ],
+      body: RefreshIndicator(
+        onRefresh: () async {
+          await walletChangeNotifier.refreshWalletInfo();
+        },
+        child: Padding(
+          padding: const EdgeInsets.symmetric(horizontal: 20),
+          child: Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            children: [
+              ExpansionPanelList(
+                children: [
+                  ExpansionPanel(
+                    headerBuilder: (BuildContext context, bool isExpanded) {
+                      return Row(
+                        children: [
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Concept ACK.

mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -35,82 +35,87 @@ class _WalletScreenState extends State<WalletScreen> {
...
-                          child: BalanceRow(walletType: WalletHistoryItemDataType.lightning),
-                        ),
-                        Padding(
-                          padding: EdgeInsets.symmetric(horizontal: 8.0),
-                          child: BalanceRow(walletType: WalletHistoryItemDataType.onChain),
-                        )
-                      ],
+      body: RefreshIndicator(
+        onRefresh: () async {
+          await walletChangeNotifier.refreshWalletInfo();
```
holzeis, line 40: 🔧 `refresh_wallet_info` is not doing much! It is only reconstructing the wallet info from the existing information. You will unlikely get new results by doing that, unless we also run a wallet sync on the refresh wallet info. 

Note, after a wallet sync we automatically push a refreshed wallet info to the client. so refreshing it manually will most likely never yield any new results until the sync happened.

holzeis (approved):
LGTM 👍

mobile/lib/features/wallet/wallet_screen.dart
holzeis: Oh, I missed that this was a PR to the transaction history which added the sync to the wallet refresh. I only looked at your changes. 😊

crates/ln-dlc-node/src/node/wallet.rs
```diff
@@ -104,4 +108,57 @@ impl Node {
...
+        self.wallet
+            .on_chain_transactions()
+            .context("Failed to retrieve on-chain transaction history")
+    }
+
+    pub fn get_off_chain_history(&self) -> Vec<PaymentDetails> {
+        let inbound_payments = self
+            .inbound_payments
+            .lock()
+            .expect("to be able to acquire lock");
```
bonomat, line 122: Let's try to avoid expects from now on if possible

luckysori: I think `unwrap`ping a mutex lock is the normal way to handle this stuff. If a mutex is poisoned there is not much else to do and crashing the application is best. Here's [some context](https://www.reddit.com/r/rust/comments/10f6328/is_this_a_reasonable_way_to_handle_unwrapping/).

luckysori: Obviously a thread panicking and the rest of the application running as if nothing bad has happened is not ideal either. We could try something like [this](https://stackoverflow.com/a/74854542/6185443).

Eventually we should probably go down the route of supervising any long-running spawned threads and respawn them in case of panic, like we used to do in ItchySats with supervised actors.

bonomat: Fair enough :)

klochowicz: another way of dealing with poisoned mutex is to evaluate whether we could ignore the poisoning...

https://users.rust-lang.org/t/mutex-poisoning-why-and-how-to-recover/72192/4
https://stackoverflow.com/questions/68426260/what-is-the-easiest-way-to-ignore-a-poison-error-on-mutex

mobile/lib/features/wallet/domain/wallet_history.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  // on-chain
+  final String? txid;
+
+  // lightning
+  final String? paymentHash;
+
+  // trade
+  final String? orderId;
+
```
klochowicz, line 24: this shows rather well that these should rather be distinct types on the domain level 🙃

klochowicz (approved):
concept ACK. let's get this in, so we can keep building upon it!

PR #317, opened 2023-03-22
feat: add a button to share logs

1010Tom:
Share the logs as file (Even though the logs are not useful yet)

IMG_81A54E326C29-1

### DIFF
--- a/mobile/lib/common/settings_screen.dart
+++ b/mobile/lib/common/settings_screen.dart
@@ -1,8 +1,15 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
 import 'package:get_10101/features/trade/settings_screen.dart';
 import 'package:get_10101/features/wallet/settings_screen.dart';
 import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
+import 'package:intl/intl.dart';
+import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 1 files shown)


mobile/lib/common/settings_screen.dart
```diff
@@ -66,6 +73,19 @@ class SettingsScreen extends StatelessWidget {
...
+        ElevatedButton(
+            onPressed: () async {
+              var file = await FLog.exportLogs();
+              var logsAsString = await file.readAsString();
+              final List<int> bytes = utf8.encode(logsAsString);
+              final Directory tempDir = await getTemporaryDirectory();
+              String now = DateFormat('yyyy-MM-dd_HH:mm:ss').format(DateTime.now());
+              final String filePath = '${tempDir.path}/$now.logs';
+              await File(filePath).writeAsBytes(bytes);
+              final XFile logFile = XFile(filePath);
```
holzeis, line 85: 🤔  `Flog.exportLogs()` is already writing the file to the disk, why do read it and write it again? Shouldn't it be suffcient to simply do the following?

```
final file = await.FLog.exportLogs();
find XFile logFile = XFile(file.path);
String now = DateFormat('yyyy-MM-dd_HH:mm:ss').format(DateTime.now());
Share.shareXFiles([logFile], text: 'Logs from $now');
```

holzeis (approved):

mobile/lib/common/settings_screen.dart
bonomat: I tried this but it did not work.

bonomat:
bors retry

PR #318, opened 2023-03-22
fix: `just android` to not fail

da-kami:
Add necessary dependencies to bundle the rust application in android native code.

look what helped... https://github.com/comit-network/comit-rs/blob/dev/cnd/Cargo.toml#L32

bundling diesel db drivers... CC @thomaseizinger


thomaseizinger:
> look what helped... [comit-network/comit-rs@`dev`/cnd/Cargo.toml#L32](https://github.com/comit-network/comit-rs/blob/dev/cnd/Cargo.toml?rgh-link-date=2023-03-22T09%3A24%3A11Z#L32)
> 
> bundling diesel db drivers... CC @thomaseizinger

You might also want to take a look at `rustls`, it is a pure Rust impl of TLS and thus doesn't have any native code dependencies. SQLite is one of the most portable C codebases out there and widely deployed on Android so I bundling that (which will compile it using your machine's C compiler for the specified `target` as part of the build script) makes sense :)

holzeis (approved):
After testing the correct branch it worked without issues 😅

PR #319, opened 2023-03-22
fix: Compute payout is greater than total collateral

### LINKED ISSUE #289: App is unable to accept DLC channel offer with certain parameters
In this instance of this error we had 10k sats outbound liquidity on the app and we attempted to create a position long position with 2k sats of margin at 2x leverage. I was able to reproduce this error multiple times with the same parameters.

```
Failed to accept DLC channel offer: Invalid parameters were provided: Computed payout is greater than total collateral. channel_id=ba320523272cae7d300c1ebd991320d422f46a8eee69cc5722cd65b4751864f1
```

We think this has to do with the fact that we are using a dummy linear payout curve instead of a proper inverse payout curve.

### LINKED ISSUE #306: Computed payout is greater than total collateral
At the moment we are using a `rounding_mod` of 500 for our rounding interval. That makes the process of opening a position much faster. However, it also means that the generated payouts only increase by 500 steps (0, 500, 1_000, 1_500, etc). These 500 steps are currently not considered in our margin calculation, which results sometimes into the error.

```
Failed to accept DLC channel offer: Invalid parameters were provided: Computed payout is greater than total collateral.
```

We need to choose a sensible value for the `rounding_mod`. It might not have an impact on a collaborative settlement as the traders can agree on an settlement amount, however on a force close it also means only 500 steps payout can be broadcast.

Note 1, during debugging I managed to prevent that error by simply ensuring that the margins used in the `payout curve` are a multiple of 500. However, going that path would also require us to change the general margin calculation (displayed to the user).

Note 2, I also tested with a `rounding_mod` of 1 as we had some time ago, but for me the coordinator got so occupied that it got unavailable to the mobile app, resulting in a loss of connection and weird errors afterwards. (The position did not open successfully)

holzeis:
Resolves #289. 
Resolves #306.

The `rust-dlc` [branch](https://github.com/p2pderivatives/rust-dlc/pull/97) that we depend on has already been rebased onto `master`, so we can fix the bug like this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -825,7 +825,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
-source = "git+https://github.com/get10101/rust-dlc?rev=ef3fe72c5e0f8930010fb13b6ea5dc3d91457c0f#ef3fe72c5e0f8930010fb13b6ea5dc3d91457c0f"
+source = "git+https://github.com/get10101/rust-dlc?rev=2194d6bce901b85684998bff57d8a2ccd9d35a4c#2194d6bce901b85684998bff57d8a2ccd9d35a4c"
 dependencies = [
  "bitcoin",
  "miniscript",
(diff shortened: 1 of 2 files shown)


luckysori (approved):
I approve my own changes.

luckysori:
bors retry

holzeis:
@luckysori why don't we just take the one commit that fixes the issue referenced in this PR instead of adding up multiple things here?

PR #320, opened 2023-03-22
chore: Add release workflow for android apk

holzeis:
Mostly copied from our poc workflow and adapted to the new folder structure, just and required versions.

Note, it seems the cache is not really working. Should be looked at in a follow up PR.

Tested on my fork: https://github.com/holzeis/10101/releases/tag/0.0.1

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -1,0 +1,144 @@
+name: "Create release"
+
+on:
+  pull_request:
+    types:
+      - closed
+
+env:
+  FLUTTER_VERSION: "3.7.7"
+  RUST_VERSION: "1.68.0"
+
+jobs:
+  build:
+    runs-on: macos-latest
+    if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/')
+    outputs:
+      ANDROID_APK_NAME: ${{ steps.build.outputs.ANDROID_APK_NAME }}
+    steps:
+      - uses: actions/checkout@v2
+
+      # #499, https://github.com/actions/virtual-environments/issues/5595
+      - name: Configure ndk
+        run: |
+          ANDROID_HOME=$HOME/Library/Android/sdk
+          SDKMANAGER=$ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager
+
+          echo y | $SDKMANAGER "ndk;21.4.7075529"
+
+          ln -sfn $ANDROID_HOME/ndk/21.4.7075529 $ANDROID_HOME/ndk-bundle
+
+      - name: Setup | Rust
+        uses: ATiltedTree/setup-rust@v1
+        with:
+          rust-version: ${{ env.RUST_VERSION }}
+          components: rustfmt
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "11.x"
+          cache: "gradle"
+
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: "stable"
+          flutter-version: ${{ env.FLUTTER_VERSION }}
+          architecture: x64
+
+      - uses: actions/cache@v3
+        id: cache-deps
+        with:
+          path: |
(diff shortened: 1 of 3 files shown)


.github/workflows/create_release.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Setup | Rust
+        uses: ATiltedTree/setup-rust@v1
+        with:
+          rust-version: ${{ env.RUST_VERSION }}
+          components: rustfmt
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "11.x" # "betterprogramming.pub" says must be java "8"
```
bonomat, line 40: the comment says it should be java 8 but here it is java 11?

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,144 @@
...
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ./rust/target
+          key: ${{ runner.os }}-cargo-integrate-android-${{ hashFiles('**/Cargo.lock') }}-${{ steps.checkout.outputs.rustc_hash }}
+
+      - name: Install FFI bindings
+        if: steps.cache-deps.outputs.cache-hit != 'true'
+        run: cargo install flutter_rust_bridge_codegen
```
bonomat, line 62: We should use `just deps-gen` to ensure we use the same version everywhere.

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,144 @@
+name: "Create release"
+
+on:
+  pull_request:
+    types:
+      - closed
```
bonomat, line 6: Are you sure this is the right trigger? 
This will publish a new version whenever any PR is closed.
I think it should be this to release on every push to main - provided we want this
```
on:
  push:
    branches:
      - main
```

.github/workflows/create_release.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Build Rust lib
+        working-directory: ./mobile/native
+        run: cargo ndk -o ../android/app/src/main/jniLibs build
+
+      - name: Parse version from pubspec.yaml
+        id: version
+        uses: jbutcher5/read-yaml@1.6
+        with:
+          file: "mobile/pubspec.yaml"
+          key-path: '["version"]'
```
bonomat, line 87: Below we are extracting the version from the branch. I think we need to be careful to have the same version everywhere, otherwise things might get weird.

.github/workflows/draft_new_release.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Initialize mandatory git config
+        run: |
+          git config user.name "${{ secrets.GET10101_NAME }}"
+          git config user.email ${{ secrets.GET10101_EMAIL }}
+
+      - name: Update changelog
+        uses: thomaseizinger/keep-a-changelog-new-release@v1
+        with:
+          version: ${{ github.event.inputs.version }}
```
bonomat, line 35: I love our changelog 😅

bonomat (approved):
Looks good, only minor comments.

.github/workflows/create_release.yml
holzeis: removed the comment 😝

.github/workflows/create_release.yml
holzeis: I think it doesn't matter as we are setting the pubspec version to the input version already during the `draft_new_release` workflow.

```
  - name: Bump pubspec version
    uses: fjogeleit/yaml-update-action@v0.12.2
    with:
      valueFile: "mobile/pubspec.yaml"
      propertyPath: "version"
      value: ${{ github.event.inputs.version }}
      commitChange: false
      updateFile: true
```

It's different sources but it should be the same everywhere 😅. As mentioned, I just copied it over from the poc. I did change anything except what was required to make it run 😉

PR #322, opened 2023-03-23
chore: listen on broadcast address

bonomat:
The problem on our VM was that we could not listen on our public ip address from inside the container. Instead, we should listen on 0.0.0.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,7 @@ jobs:
           flutter-version: "3.7.7"
           channel: "stable"
       - name: Install FFI bindings
-        uses: baptiste0928/cargo-install@v1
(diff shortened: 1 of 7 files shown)


luckysori (approved):
Nice! Let's roll with this.

da-kami (approved):

bonomat:
bors retry

holzeis:
> The problem on our VM was that we could not listen on our public ip address from inside the container. Instead, we should listen on 0.0.0.0

I guess that's a change that wasn't transferred to the the cloud-conf. I also noticed that we can't bind on our public ip and changed to `0.0.0.0`. Your fix is still better though, as it will allow us now to separate the node announcement from how to listen. 👍

PR #324, opened 2023-03-23
fix: on restart a node is not registered anymore

bonomat:
hence we need to fail the intercepted htlc backwards

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -402,13 +402,6 @@ impl EventHandler {
                 inbound_amount_msat,
                 expected_outbound_amount_msat,
             } => {
-                let fake_channel_payments = self.fake_channel_payments.clone();
-                let result = fake_channel_payments.lock();
-                let guard = result.unwrap();
-                let target_node_id = guard
-                    .get(&requested_next_hop_scid)
-                    .expect("To have a target node id stored");
-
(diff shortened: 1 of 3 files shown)


luckysori (approved):

bonomat:
bors retry

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -445,6 +438,21 @@ impl EventHandler {
                     return;
                 }
 
+                let fake_channel_payments = self.fake_channel_payments.clone();
+                let result = fake_channel_payments.lock();
+                let guard = result.unwrap();
+
+                let target_node_id = match guard.get(&requested_next_hop_scid) {
```
holzeis, line 445: 🤔 not sure if this is 100% correct. What if the `requested_next_hop_scid` isn't a fake one, because the user wants to pay to an known node?

cc: @bonomat @luckysori

luckysori: Then we should not trigger this event. From the docs:


>Used to indicate that we've intercepted an HTLC forward. This event will only be generated if you've encoded an intercept scid in the receiver's invoice route hints using `ChannelManager::get_intercept_scid`

holzeis: i see, I though any HTLC will be intercepted by this event. 👍

Issue #328, opened 2023-03-23
Outbound lightning payments do not appear in the transaction history


holzeis:
Currently we do not track any outbound payments on the ln_dlc node. Consequently the history will not show if an outbound payment happened.

### How to test
1. initialize development setup with a funded wallet.
2. start a second 10101 wallet. e.g. by using two different simulators.
3. Create an invoice with one wallet and pay it with the funded one.

### Expected behavior
The outbound transaction is shown in the transaction history

### Observed behavior
The balance is updated but the transaction history does not show the outbound transaction.

PR #329, opened 2023-03-24
fix: Failed to create just in time channel due to disconnected peer

### LINKED ISSUE #330: Failed to create just-in-time channel due to disconnected node
If the coordinator is not connected to the payee upon creating a just-in-time channel it will panic with the following error message.

```
thread '<unnamed>' panicked at 'To open channel: Could not create channel with 02c8280854b03ab289fa0a76264b599413418092e242de26dcd1ec95679d8a4d4c due to Channel unavailable: Not connected to node: 02c8280854b03ab289fa0a76264b599413418092e242de26dcd1ec95679d8a4d4c', crates/ln-dlc-node/src/ln/event_handler.rs:492:22
```

After this error occurs the coordinator is in a faulty state and has to be restarted.

holzeis:
The change also includes error handling in case the payment could not be forwarded and sets the HTLC to failed.

resolves #330

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -12,6 +12,8 @@ use crate::PaymentInfo;
 use crate::PaymentInfoStorage;
 use crate::PendingInterceptedHtlcs;
 use anyhow::anyhow;
+use anyhow::Context;
+use anyhow::Result;
 use bitcoin::secp256k1::Secp256k1;
 use lightning::chain::chaininterface::BroadcasterInterface;
 use lightning::chain::chaininterface::ConfirmationTarget;
@@ -66,7 +68,7 @@ impl EventHandler {
         }
     }
 
-    fn match_event(&self, event: Event) {
+    fn match_event(&self, event: Event) -> Result<()> {
         match event {
             Event::FundingGenerationReady {
                 temporary_channel_id,
@@ -94,8 +96,9 @@ impl EventHandler {
                         );
                         self.channel_manager
                             .close_channel(&temporary_channel_id, &counterparty_node_id)
-                            .expect("To be able to close a channel we cannot open");
-                        return;
+                            .map_err(|e| anyhow!("{e:?}"))?;
+
+                        return Ok(());
                     }
                 };
 
@@ -195,7 +198,8 @@ impl EventHandler {
                         &counterparty_node_id,
                         0,
                     )
-                    .expect("To be able to accept a 0-conf channel");
+                    .map_err(|e| anyhow!("{e:?}"))
+                    .context("To be able to accept a 0-conf channel")?;
             }
             Event::PaymentPathSuccessful { .. } => {}
             Event::PaymentPathFailed { .. } => {}
@@ -287,21 +291,18 @@ impl EventHandler {
                 });
             }
             Event::SpendableOutputs { outputs } => {
-                let destination_address = self.wallet.inner().get_unused_address().unwrap();
+                let destination_address = self.wallet.inner().get_unused_address()?;
                 let output_descriptors = &outputs.iter().collect::<Vec<_>>();
                 let tx_feerate = self
                     .wallet
                     .get_est_sat_per_1000_weight(ConfirmationTarget::Normal);
-                let spending_tx = self
-                    .keys_manager
-                    .spend_spendable_outputs(
-                        output_descriptors,
-                        Vec::new(),
-                        destination_address.script_pubkey(),
-                        tx_feerate,
-                        &Secp256k1::new(),
-                    )
-                    .unwrap();
+                let spending_tx = self.keys_manager.spend_spendable_outputs(
+                    output_descriptors,
+                    Vec::new(),
+                    destination_address.script_pubkey(),
+                    tx_feerate,
+                    &Secp256k1::new(),
+                )?;
                 self.wallet.broadcast_transaction(&spending_tx);
             }
             Event::ChannelClosed {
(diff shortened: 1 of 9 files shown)


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -175,22 +174,38 @@ impl EventHandler {
                 fee_paid_msat,
                 ..
             } => {
```
holzeis, line 176: @luckysori: It appears that the outbound payment is never added to the the `self.outbound_payments`(At least I did not find any code for it). Hence the tests were not able to assert on a failed payment. I've adapted the code the code analogue to how we handle the Claimed Payment and insert the payment if `Vacant`. 

I suspected that this will also fix #328, but after a quick manual tests, it appears the issue still persists. Might be still something missing there. However, the tests are able to find the `outbound_payments` now.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -480,16 +525,25 @@ impl EventHandler {
...
                             "Could not create channel with {} due to {e:?}",
                             target_node_id
                         )
-                    })
-                    .expect("To open channel");
+                    }) {
+                    Ok(temp_channel_id) => temp_channel_id,
+                    Err(_e) => {
+                        // todo: check why this tracing fails here.
+                        // tracing::warn!("Failed to open just in time channel. Error: {e:?}");
```
holzeis, line 540: @luckysori: Do you know why this is failing? The event_handler panics at `backtrace::trace_unsynchronized`.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -226,10 +242,26 @@ impl EventHandler {
...
+                    Entry::Occupied(mut e) => {
+                        let payment = e.get_mut();
+                        payment.status = HTLCStatus::Failed;
+                    }
+                    Entry::Vacant(e) => {
+                        e.insert(PaymentInfo {
+                            preimage: None,
+                            secret: None,
+                            status: HTLCStatus::Failed,
+                            // todo: decode amount from payment hash?
```
bonomat, line 260: I don't think you can decode it from a hash? You'd need to retrieve it from somewhere.

crates/ln-dlc-node/src/ln/event_handler.rs
bonomat: commenting to follow this...

bonomat (approved):
Nice, this should hopefully get rid of the poison errors.

crates/ln-dlc-node/src/ln/event_handler.rs
holzeis: I think you are right.. otherwise that hash wouldn't be a good hash 😅

crates/ln-dlc-node/src/ln/event_handler.rs
luckysori: Good catch! I think we only ever tested manually with inbound payments 😨

crates/ln-dlc-node/src/ln/event_handler.rs
luckysori: Haha I've never come across this, I'll take a closer look.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -113,13 +113,14 @@ impl Node {
...
+    pub fn create_intercept_scid(&self, target_node: PublicKey) -> Result<u64> {
         let intercept_scid = self.channel_manager.get_intercept_scid();
-        self.fake_channel_payments
+        let mut fake_channel_payment = self
+            .fake_channel_payments
             .lock()
-            .unwrap()
-            .insert(intercept_scid, target_node);
-        intercept_scid
+            .map_err(|e| anyhow!("{e:#}"))?;
```
luckysori, line 121: 🤔 I think we should just `unwrap` or `expect` when dealing with locks. The reason why this panics is because the mutex was poisoned in the event handler that has all those _other_ unwraps. A mutex is poisoned if the guard is held and a panic occurs.

luckysori (commented):
I will take it over so that we can merge it asap.

crates/ln-dlc-node/src/ln/event_handler.rs
luckysori: Could not reproduce in a test (where I caused a disconnect an saw the log being emitted).

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -69,7 +71,7 @@ impl EventHandler {
         }
     }
 
-    fn match_event(&self, event: Event) {
+    fn match_event(&self, event: Event) -> Result<()> {
```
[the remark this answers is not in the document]
luckysori: I've split into multiple patches now.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -175,22 +174,38 @@ impl EventHandler {
                 fee_paid_msat,
                 ..
             } => {
-                let mut payments = self.outbound_payments.lock().unwrap();
-                for (hash, payment) in payments.iter_mut() {
-                    if *hash == payment_hash {
+                let mut payments = self
+                    .outbound_payments
+                    .lock()
+                    .map_err(|e| anyhow!("{e:#}"))?;
```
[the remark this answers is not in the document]
klochowicz: just chiming in: IMHO the best option would be to override the panic handler from Tokio to take down the whole application - if we're resilient to such restarts, we can *appear* really stable and will also spot problems much easier during development.
When an essential task crashes, there's nothing really we can do. Crashing and starting again seems much better than having a "zombie" app that appears to work, but doesn't do what it supposed to.

PR #337, opened 2023-03-27
fix: Switch offer to accept collateral

holzeis:
The app is accepting the dlc hence the apps collateral can be found in the `accept_collateral` attribute.


luckysori (approved):
Good find!

bonomat (approved):

PR #339, opened 2023-03-27
feat: Use opening price when closing position

### LINKED ISSUE #331: Use opening price when calculating collab settlement amount
Since we do not store the position yet on the coordinator, we could not load the opening price from the to be closed position. As a quick fix the execution (closing) price is currently used. This obviously incorrect and should be fixed once we have position persistency on the coordinator.

holzeis:
This is a temporary fix until we store the position into the database.

@da-kami: Addressing your comment on #305. This will store and load the position in memory.

resolves #331


coordinator/src/node.rs
```diff
@@ -24,19 +24,32 @@ use ln_dlc_node::node::SubChannelManager;
...
+    pub positions: Mutex<HashMap<String, Position>>,
+}
+
+pub struct Position {
+    pub contract_symbol: ContractSymbol,
+    pub leverage: f64,
+    pub quantity: f64,
+    pub direction: Direction,
+    pub trader: PublicKey,
+    pub average_entry_price: f64,
```
luckysori, line 52: 🙃 Could just store the `average_entry_price` since that all we care about and this is just a band-aid fix.

coordinator/src/node.rs
```diff
@@ -116,10 +146,21 @@ impl Node {
...
+            .remove(&trade_params.pubkey.to_string())
+        {
+            Some(position) => Decimal::try_from(position.average_entry_price)?,
+            None => {
+                tracing::warn!("Did not find position in memory, thus we do not have the opening price to calculate a correct accept settlement amount. Using the closing price.");
+                // FIXME: This is wrong as we cannot use the closing price to calculated the
+                // `accept_settlement_amount`. We must save the initial price when creating the
+                // position and use it here again for closing.
+                closing_price
+            }
```
luckysori, line 162: 👍 And we know this will happen if the coordinator is restarted, since the position info won't be persisted.

luckysori (approved):
LGTM to make the situation better if no restarts take place.

PR #341, opened 2023-03-29
Ensure separate sqlite DB is used for different networks

### LINKED ISSUE #277: App database file location is shared between all networks
In particular, if a dev first runs the app on testnet and then on regtest, the same database file (`trader.sqlite`) will be used for both. This could result in unexpected behaviour or simply overwriting data by mistake.

klochowicz:
Closes #277

### DIFF
--- a/mobile/native/src/db/mod.rs
+++ b/mobile/native/src/db/mod.rs
@@ -7,6 +7,7 @@ use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
+use bdk::bitcoin;
 use diesel::connection::SimpleConnection;
 use diesel::r2d2;
 use diesel::r2d2::ConnectionManager;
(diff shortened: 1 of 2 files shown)


mobile/native/src/db/mod.rs
```diff
@@ -61,8 +62,8 @@ impl r2d2::CustomizeConnection<SqliteConnection, r2d2::Error> for ConnectionOpti
     }
 }
 
-pub fn init_db(db_dir: String) -> Result<()> {
-    let database_url = format!("sqlite://{db_dir}/trader.sqlite");
+pub fn init_db(db_dir: &str, network: bitcoin::Network) -> Result<()> {
+    let database_url = format!("sqlite://{db_dir}/{network}.sqlite");
```
da-kami, line 66:
```suggestion
    let database_url = format!("sqlite://{db_dir}/{network}-trades.sqlite");
```

I would still indicate that this is the database that stores the trades / 10101 data and not just "Bitcoin related".
It might be confusing to people that we use the network as database name.

We could also consider moving the data into the network folder, but I'm not sure that's less confusing. 
Not super important though.

da-kami (approved):

mobile/native/src/db/mod.rs
klochowicz: how about `trader-$NETWORK`?

luckysori (approved):

mobile/native/src/db/mod.rs
klochowicz: I'll go with `trades-{network}.sqlite` - good suggestion! :)

PR #342, opened 2023-03-29
Downgrade log message when receiving list of all orders

### LINKED ISSUE #323: App received all orders from orderbook even though it already had some orders
I ran into this on iOS:

```
2023-03-23 14:12:49.016 | LogLevel.WARNING | Received all orders from orderbook, but we already have some orders. This should not happen 
```

klochowicz:
This would have come if the websocket stream got broken & reestablished.

After we get this message we swap all orders we had previously with new list of initial orders - and then proceed to accumulate deltas.

Closes #323


luckysori (approved):
Perfect, thank you!

PR #343, opened 2023-03-29
chore: just command to build apk

da-kami:
Can be improved - I think we can also add commands to e.g. start the app against the testnet setup


justfile
```diff
@@ -3,6 +3,9 @@ set dotenv-load
 line_length := "100"
 coordinator_log_file := "$PWD/data/coordinator/regtest.log"
 maker_log_file := "$PWD/data/maker/regtest.log"
+pubspec := "$PWD/mobile/pubspec.yaml"
+testnet_coordinator := "COORDINATOR_P2P_ENDPOINT=038fb0eb9d9deeb4f0aa885d6c64abb0a26002871d2f7501349cd8d00f1f130045@195.201.81.94:9045"
+testnet_electrs := "ELECTRS_ENDPOINT=195.201.81.94:50000"
```
klochowicz, line 8: I think you have these in your `.env`, so you could refer to them by `${VARIABLE}` notation instead of duplicating these

justfile
```diff
@@ -189,6 +192,14 @@ all: services gen native run
 # Run everything at once, tailored for iOS development (rebuilds iOS)
 all-ios: services gen ios run
 
+testnet-apk:
+    #!/usr/bin/env bash
+    BUILD_NAME=$(yq -r .version {{pubspec}})
+    BUILD_NUMBER=$(git rev-list HEAD --count)
+    echo $BUILD_NAME
```
klochowicz, line 199: nit: whenever you are quoting bash variables, it's a good practice to use `${BUILD_NAME}` - otherwise things might be parsed improperly (e.g. if there is a space or something).

klochowicz (approved):
nice! just some nits :)

justfile
da-kami: I don't think the testnet configuration should be in the `.env` file - I think it can just be hardcoded for everybody.

klochowicz: fair enough!

justfile
```diff
@@ -1,11 +1,14 @@
...
+set export
+
 line_length := "100"
 coordinator_log_file := "$PWD/data/coordinator/regtest.log"
 maker_log_file := "$PWD/data/maker/regtest.log"
 pubspec := "$PWD/mobile/pubspec.yaml"
-testnet_coordinator := "COORDINATOR_P2P_ENDPOINT=038fb0eb9d9deeb4f0aa885d6c64abb0a26002871d2f7501349cd8d00f1f130045@195.201.81.94:9045"
-testnet_electrs := "ELECTRS_ENDPOINT=195.201.81.94:50000"
+TESTNET_COORDINATOR := "038fb0eb9d9deeb4f0aa885d6c64abb0a26002871d2f7501349cd8d00f1f130045@195.201.81.94:9045"
+TESTNET_ELECTRS := "195.201.81.94:50000"
```
klochowicz, line 11: why are these uppercase? I'd suggest to stick to the same convention and keep them lowercase...

.env.sample
```diff
@@ -1,6 +1,2 @@
 ALTOOL_API_KEY=TBA
 ALTOOL_API_ISSUER=TBA
-# test vm electrs endpoint
-ELECTRS_ENDPOINT=35.189.57.114:50000
-# test vm coordinator
-COORDINATOR_P2P_ENDPOINT=026e1de0e76902d71d2c27bfdda9d48c19c4c550f08bb4189d7208246e0e3af97e@35.189.57.114:9045
```
klochowicz, line 6: why are we removing these? are these not useful anymore? I suspect that you might need to change some documentation to reflect that, because if I remember correctly they were described earlier

justfile
```diff
@@ -1,11 +1,14 @@
 # To use this file, install Just: cargo install just
 set dotenv-load
+# eports all justfile variables as environment variables
```
klochowicz, line 3: nit: typo `exports`

klochowicz: I'm not sure whether I'm in favour of these, variables such as line length are exported with this, and there's no point for that?
IMHO it's better to stick to the previously established convention with the substitutions.

justfile
```diff
@@ -223,6 +230,11 @@ wait-for-electrs-to-be-ready:
     done
 
 build-ipa:
+    #!/usr/bin/env bash
+    ELECTRS_ENDPOINT="${ELECTRS_ENDPOINT:-$TESTNET_ELECTRS}"
+    COORDINATOR_P2P_ENDPOINT="${COORDINATOR_P2P_ENDPOINT:-$TESTNET_COORDINATOR}"
```
klochowicz, line 235: are you sure this does what you wanted it to do?
maybe something more like:
`ELECTRS_ENDPOINT="${ELECTRS_ENDPOINT}:-${TESTNET_ELECTRS}" ? (quoting the individual variables?)

klochowicz (commented):
I'm not sure whether update was a step in the right direction? 
leaking env values from justfile seems of no use ;/

klochowicz:
can we avoid dependency on yq? what are you using it for?

.env.sample
da-kami: They are not useful - if they change they change globally. 
If this is in your local `.env` file everybody has to change the values if the values change. (the file that is checked in is just `.env.sample` - it's just a sample...)

The `.env` file is meant to be for things that are specific to each machine (e.g. signing keys), not the global settings.

justfile
da-kami: Am I misunderstanding how to set a default if the variable is not set? 
I followed this: https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash

.env.sample
klochowicz: I was referring to the fact that they were in the sample and were meant to be used / overriden. as such, adding capability to release on Android has nothing to do with removing these?

justfile
da-kami: that is indeed quite random, I can change that :)

justfile
klochowicz: please maybe add a comment what's happening here - I've been doing Bash for years and thought I'm quite decent, and you totally blew my mind with this! 😂

thank you for enlightening me!

da-kami:
@klochowicz I noticed that some of the things I did are an overkill - we don't have to export all the variables from the justfile, I can just read the variables with `{{}}` instead and assign them - I will adapt that

klochowicz:
> @klochowicz I noticed that some of the things I did are an overkill - we don't have to export all the variables from the justfile, I can just read the variables with `{{}}` instead and assign them - I will adapt that

yes! that was my main gripe, then we don't need to use `set export` :)
ready whenever you are, thank you for your patience @da-kami

justfile
```diff
@@ -189,6 +195,18 @@ all: services gen native run
...
+    #!/usr/bin/env bash
+    BUILD_NAME=$(yq -r .version {{pubspec}})
+    BUILD_NUMBER=$(git rev-list HEAD --count)
+    echo "build name: ${BUILD_NAME}"
+    echo "build number: ${BUILD_NUMBER}"
+    ELECTRS_ENDPOINT="${ELECTRS_ENDPOINT:-$TESTNET_ELECTRS}"
+    COORDINATOR_P2P_ENDPOINT="${COORDINATOR_P2P_ENDPOINT:-$TESTNET_COORDINATOR}"
+    echo "electrs: ${ELECTRS_ENDPOINT}"
+    echo "coordinator: ${COORDINATOR_P2P_ENDPOINT}"
+    cd mobile && flutter build apk --build-name=${BUILD_NAME} --build-number=${BUILD_NUMBER} --release --dart-define="ELECTRS_ENDPOINT=${ELECTRS_ENDPOINT}" --dart-define="COORDINATOR_P2P_ENDPOINT=${COORDINATOR_P2P_ENDPOINT}"
```
holzeis, line 208: fyi: we have a github workflow for that. It's the same workflow copied from our poc.

1. draft new release: https://github.com/get10101/10101/actions/workflows/draft_new_release.yml
2. create release: once the draft release gets merged the release is created. https://github.com/get10101/10101/actions/workflows/create_release.yml

Ideally we can move the ios build to the workflow as well, eventually 🙂

klochowicz:
@da-kami are you updating this? :) I wanted to re-review it, but it seems nothing changed yet?

da-kami:
> @da-kami are you updating this? :) I wanted to re-review it, but it seems nothing changed yet?

Does not have prio at the moment, will update once we target android

PR #344, opened 2023-03-29
Fix rust_flutter_bridge not being a singleton issue

### LINKED ISSUE #315: FlutterRustBridgeBase should be singleton
I've noticed the following line in our logs: 

```
Error: Exception: Subclasses of `FlutterRustBridgeBase` should be singletons - there should not be two instances (runtimeType=NativePlatform) 
```

Looks like a bug to me.

klochowicz:
Fixes #315 

frb template already had a fix for iOS, but not for macOS:

https://github.com/fzyzcjy/flutter_rust_bridge/issues/1025

### DIFF
--- a/mobile/macos/Runner/AppDelegate.swift
+++ b/mobile/macos/Runner/AppDelegate.swift
@@ -4,7 +4,8 @@ import FlutterMacOS
 @NSApplicationMain
 class AppDelegate: FlutterAppDelegate {
   override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
-    dummy_method_to_enforce_bundling()
+    let dummy = dummy_method_to_enforce_bundling()
+    print(dummy)
     return true
   }
 }


mobile/macos/Runner/AppDelegate.swift
```diff
@@ -4,7 +4,8 @@ import FlutterMacOS
 @NSApplicationMain
 class AppDelegate: FlutterAppDelegate {
   override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
-    dummy_method_to_enforce_bundling()
+    let dummy = dummy_method_to_enforce_bundling()
+    print(dummy)
```
holzeis, line 8: wait, what? 😅 this is fixing the issue 😂

klochowicz: yup, dummy methods can be optimised away and therefore stripped in builds if they're not doing anything :-)

klochowicz: I just copied the solution referenced in the linked issue

holzeis: leaving this for reference why this fixes the issue: https://github.com/fzyzcjy/flutter_rust_bridge/issues/496#issuecomment-1150038776

holzeis (approved):

PR #348, opened 2023-03-30
chore: orderbook returns an error if no match

bonomat:
An order which cannot be matched right now should transition into a failed state for now. While in the future we want to be able to notify the trader later on once there is enough liquidity.

This will prevent us from having stale orders in the frontend.


holzeis (approved):
👍 Makes sense!

PR #349, opened 2023-03-31
Expose Electrum blockchain in LnDlcWallet

klochowicz:
At times, we need to query Electrum; having it exposed is much more convenient instead of exposing the API part by part.


luckysori (approved):
Clippy is unsure, but LGTM.

Issue #352, opened 2023-03-31
Closing channel due to close-required error: Invalid commitment tx signature from peer


holzeis:
If the app loses the connection to the coordinator in the middle of a dlc setup the channel can gets closed.

More specifically, if the `app` reconnects after accepting the dlc channel offer the `coordinator` will force-close the apps channel. It seems the dlc state does not get affected by that and will still show that the contract has been signed.

```
2023-03-31T07:26:58.154706Z DEBUG lightning::ln::channelmanager: Received channel_update for channel 4855ed5ad051db406116c88769ec8aef8ec77f10392293ab7c40f401643035a6.    
2023-03-31T07:26:58.155806Z ERROR lightning::ln::channelmanager: Closing channel 4855ed5ad051db406116c88769ec8aef8ec77f10392293ab7c40f401643035a6 due to close-required error: Invalid commitment tx signature from peer    
2023-03-31T07:26:58.155837Z DEBUG lightning::ln::channelmanager: Finishing force-closure of channel with 0 HTLCs to fail    
2023-03-31T07:26:58.155850Z  INFO lightning::chain::channelmonitor: Applying update to monitor 4855ed5ad051db406116c88769ec8aef8ec77f10392293ab7c40f401643035a6, bringing update_id from 0 to 1 with 1 changes.    
2023-03-31T07:26:58.155877Z DEBUG lightning::chain::channelmonitor: Getting signed latest holder commitment transaction!    
2023-03-31T07:26:58.156305Z  INFO lightning::chain::channelmonitor: Broadcasting local commitment tx with txid f614546b5ca0a32412a41d948e3e238a6dab4b81767e3a7156e56e8f3342e0db    
2023-03-31T07:26:58.174399Z DEBUG lightning::chain::chainmonitor: Persistence of ChannelMonitorUpdate for channel 4855ed5ad051db406116c88769ec8aef8ec77f10392293ab7c40f401643035a6 completed    
2023-03-31T07:26:58.174433Z ERROR lightning::ln::channelmanager: Invalid commitment tx signature from peer    
2023-03-31T07:26:58.174838Z DEBUG lightning::ln::channelmanager: Received channel_update for channel 4855ed5ad051db406116c88769ec8aef8ec77f10392293ab7c40f401643035a6.    
[...]

PR #353, opened 2023-03-31
test: Add test reproducing the invalid commitment tx signature error

holzeis:
Reproduces the error described in https://github.com/get10101/10101/issues/352

TODO: This test will be executed on our slow running tests, but should not as it is not fixed yet. However, it's valuable to a test case already written to reproduce the error.

- [x] Refactor the way slow running tests are run

Suggestion from @luckysori 

- Not ignore any ln-dlc-node tests by default.
- Only run ln-dlc-node tests on the special job which runs with bors (not just the slow ones).
- Ignore tests which we actually want to ignore for the time being.

Latest suggestion from @luckysori reflected in https://github.com/get10101/10101/pull/353/commits/1472ad8cc7120a92df761e54adc7c45f5bd48175:

- Expect the failure case in the test assertion.
- Run the test like any other ignored test.
- If it ever fails it means the bug has been fixed!

### DIFF
--- a/crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
+++ b/crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
@@ -1,0 +1,91 @@
+use crate::node::Node;
+use crate::tests::dummy_contract_input;
+use crate::tests::init_tracing;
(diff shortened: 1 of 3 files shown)


luckysori (approved):

holzeis:
bors retry

PR #354, opened 2023-03-31
chore: Remove unwrap from node sync

holzeis:


luckysori (approved):
Noice!

luckysori:
bors retry.

Triggering CI again in case #355 was transient.

PR #356, opened 2023-04-03
fix(dlc_channel): Remove unwrap from offer sub channel

### LINKED ISSUE #309: Fix an unwrap that blows up the coordinator
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidState("Received sub channel offer but a non closed sub channel already exists")', /Users/dakami/CoBloX/repos/10101/crates/ln-dlc-node/src/node/dlc_channel.rs:46:14
```

once this happend the coordinator was not reachable anymore (http endpoints down)

holzeis:
resolves #309 

However, that fix does not address what to do with offers that remain in an pending state. We could either (given that the mobile app is aware of these offers)

1. Reject any pending dlc channel offer, prior to creating a new trade 
2. Continue the dlc setup flow if one is already there.

I guess some kind of timeout would be in anyway sensible.


luckysori (approved):

da-kami (approved):
> 1. Reject any pending dlc channel offer, prior to creating a new trade 
> 2. Continue the dlc setup flow if one is already there.
>
> I guess some kind of timeout would be in anyway sensible.

For now it's fine to just fail it and let the user try again. We should call `handler::order_failed` for now for pretty much anything that does not go as expected. I would only add retry later.

PR #357, opened 2023-04-03
Stop all spawned tasks on `Node` drop

luckysori:
At least the ones that are spawned in the constructor. Otherwise they could keep running indefinitely unexpectedly in situations where we remake a `Node` e.g. in tests.


da-kami (approved):

luckysori:
bors retry

PR #358, opened 2023-04-03
Watch LN channels using original funding TXO

### LINKED ISSUE #299: `ChannelMonitor was stored in the wrong file`
```
flutter: | 2023-03-21 15:42:23.872 | LogLevel.ERROR | Failed to initialise: Error: ChannelMonitor was stored in the wrong file
```

Once the we run into this the app won't start anymore - will always run into this error. 
I ran into this error after repeatedly starting and stopping the app - there is no clear pattern to reproduce this though, it happend quite randomly. 

Daniel's guess: We don't close *some* database when we exit the app gracefully, so it can happen that a lock is still there or some file gets corrupted.

### LINKED ISSUE #308: Unable to close position after a restart
App logs; failed to accept:

```
2023-03-22T02:50:06.401533Z DEBUG native::ln_dlc::node: Processing DLC channel message from=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9 msg=CloseOffer
2023-03-22T02:50:06.409307Z  INFO ln_dlc_node::node::dlc_channel: Accepting DLC channel collaborative settlement channel_id=f925052ec6e71ea170eeabf59dee467a7fa8893b807a3e01843e0ef433fe9697
2023-03-22T02:50:06.411071Z ERROR lightning::chain::chainmonitor: Failed to update channel monitor funding txo: no such monitor registered    
2023-03-22T02:50:06.412608Z ERROR native::ln_dlc::node: Failed to accept DLC channel close offer: Invalid parameters were provided: Misuse error: Could not update channel funding transaction. channel_id=f925052ec6e71ea170eeabf59dee467a7fa8893b807a3e01843e0ef433fe9697
2023-03-22T02:50:11.36406Z DEBUG ln_dlc_node::node: Announcing node on [IPv4 { addr: [0, 0, 0, 0], port: 62184 }]
```

Weirdest all:

I think the position was closed somehow because my balance changed afterwards - but I don't know what happened exactly.

luckysori:
Fixes #299.
Fixes #308.

With the introduction of `rust-dlc` LN-DLC channels, `rust-lightning` was forked to introduce some minor changes. Among those, `rust-lightning` must now be able to consider the `glue_transaction` as the actual funding transaction in certain scenarios, whilst still considering the _original_ `funding_transaction` as the funding transaction for other scenarios.

One scenario where `rust-lightning` still expects the original `funding_transaction` is when setting up the `ChainMonitor` to watch channels on startup.

Our code was partially ported from `10101-poc`, where we did not use LN-DLC channels provided by `rust-dlc`. As such, we kept a version of the `ChainMonitor` setup which still called `get_funding_txo`, even though we are now expected to call `get_original_funding_txo` there. The correct setup was already exemplified [here](https://github.com/p2pderivatives/ldk-sample/blob/fc8bac24a66c9a8fcf817b9c2f76780b89480d5d/src/main.rs#L561), but we missed this detail.

---

As you can see this PR does not include a test, which is a shame. @holzeis and I were able to reproduce the issues listed above and subsequently verify that this code change fixes them.

We weren't able to write a Rust `ln-dlc-node` test with a `Node` restart because it's [seemingly impossible to open a `sled` DB soon after dropping it](https://github.com/spacejam/sled/issues/1234#issuecomment-754769425). Instead we modified some test utils and wrote a _couple_ of tests which shared state between them. Here's the patch for posterity: https://github.com/get10101/10101/commit/9f46fcd8e558a9e2716efea48765d2f953b93b98.

### DIFF
--- a/crates/ln-dlc-node/src/node/channel_manager.rs
+++ b/crates/ln-dlc-node/src/node/channel_manager.rs
@@ -120,7 +120,14 @@ pub(crate) fn build(
     // Give ChannelMonitors to ChainMonitor to watch
     for confirmable_monitor in confirmable_monitors.drain(..) {
         let channel_monitor = confirmable_monitor.0;
-        let funding_txo = channel_monitor.get_funding_txo().0;
+
+        // ATTENTION: This must be `get_original_funding_txo` and _not_ `get_funding_txo`, because
+        // we are using LN-DLC channels. `rust-dlc` is manipulating the funding TXO so that LDK
+        // considers the `glue_transaction` as the `funding_transaction` for certain purposes.
+        //
+        // For other purposes, LDK must still refer back to the original `funding_transaction`. This
+        // is one such case.
+        let funding_txo = channel_monitor.get_original_funding_txo().0;
         assert_eq!(
             chain_monitor.watch_channel(funding_txo, channel_monitor),
             ChannelMonitorUpdateStatus::Completed


holzeis (approved):
Very nice 👌

da-kami (approved):
omg, a one-liner fixing two major problems 😄

klochowicz (approved):
great work!!!

Issue #359, opened 2023-04-03
Unable to close position (settle DLC) on current `main`: `Unable to process incoming messages: No confirmed DLC found`


da-kami:
app logs:

```
2023-04-03T22:49:17.184387Z DEBUG native::event::event_hub: Publishing event WalletInfoUpdateNotification(WalletInfo { balances: Balances { on_chain: 0, lightning: 98167 }, history: [WalletHistoryItem { flow: Outbound, amount_sats: 1000, timestamp: 1680562095, status: Confirmed, wallet_type: Trade { order_id: "45d45442-82df-46de-a8fb-9cf4d843c61e" } }, WalletHistoryItem { flow: Inbound, amount_sats: 100000, timestamp: 1680562078, status: Confirmed, wallet_type: Lightning { payment_hash: "0bc1ec91a9b8b759cdb50da58b2ade82d662943a987cd150a781709d76f0beb4" } }] })
2023-04-03T22:49:27.457137Z DEBUG native::event::event_hub: Publishing event WalletInfoUpdateNotification(WalletInfo { balances: Balances { on_chain: 0, lightning: 98167 }, history: [WalletHistoryItem { flow: Outbound, amount_sats: 1000, timestamp: 1680562095, status: Confirmed, wallet_type: Trade { order_id: "45d45442-82df-46de-a8fb-9cf4d843c61e" } }, WalletHistoryItem { flow: Inbound, amount_sats: 100000, timestamp: 1680562078, status: Confirmed, wallet_type: Lightning { payment_hash: "0bc1ec91a9b8b759cdb50da58b2ade82d662943a987cd150a781709d76f0beb4" } }] })
2023-04-03T22:49:36.371358Z DEBUG native::event::event_hub: Publishing event PositionUpdateNotification(Position { leverage: 2.0, quantity: 0.55622, contract_symbol: BtcUsd, direction: Long, average_entry_price: 27811.0, liquidation_price: 18540.666666666668, position_state: Open, collateral: 1000 })
[...]

PR #360, opened 2023-04-04
fix(mobile): Handle opening and closing DLC channels separately

luckysori:
Fixes https://github.com/get10101/10101/issues/359.

A closed DLC channel results in a `Contract::Closed` state, as opposed to a `Contract::Confirmed` state. As such, we were unable to find a `Confirmed` DLC after handling the `CloseFinalize` sub-channel message, which led to a [bug](https://github.com/get10101/10101/issues/359).

But it turns out that if we are reacting to a DLC channel being closed, we already know that the position should change in a different manner (it should simply be deleted). We therefore separate the handling of opening and closing DLC channels on the way to fixing the bug.


holzeis:
@luckysori verified on my local machine, opening and closing a position works without any errors.

PR #361, opened 2023-04-04
Fix/forward payment on existing channel

holzeis:
fix: Forward payment on channel if existing

The current logic was flawed due to the following reasons.

1. The `requested_next_hop_scid` does not match the `short_channel_id`, is the fake scid generated by the coordinator. Also note, that this fake scid, is always different.
2. Matching on the `short_channel_id` is generally not ideal as the `short_channel_id` is only set after the channel has sufficient confirmations.

The fix matches on the `target_node_id` instead, assuming that we will currently only ever have exactly one channel with the user.

Eventually we should not use interceptable invoices once we have an open channel. However, I decided to stick with only interceptable invoices for now as mixing regular invoices with interceptable invoices seems to be failure prune upon the first invoice(s). e.g. Assume a second invoice is created before the first invoice has been paid.

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -432,20 +432,37 @@ impl EventHandler {
                     "Intercepted HTLC"
                 );
 
-                // if we have already a channel with them, we try to forward the payment.
-                // TODO: here we would need to increase the channel size if the channel is too small
-                if let Some(channel) =
-                    self.channel_manager
-                        .list_channels()
-                        .iter()
-                        .find(|channel_details| {
-                            if let Some(scid) = channel_details.short_channel_id {
-                                scid == requested_next_hop_scid
-                            } else {
-                                false
+                let target_node_id = {
+                    let fake_channel_payments = self.fake_channel_payments.lock().unwrap();
+                    match fake_channel_payments.get(&requested_next_hop_scid) {
+                        None => {
+                            tracing::warn!(fake_scid = requested_next_hop_scid, "Could not forward the intercepted HTLC because we didn't have a node registered with said fake scid");
+
+                            if let Err(err) =
+                                self.channel_manager.fail_intercepted_htlc(intercept_id)
+                            {
+                                tracing::error!("Could not fail intercepted htlc {err:?}")
                             }
-                        })
+
+                            return Ok(());
+                        }
+                        Some(target_node_id) => *target_node_id,
+                    }
+                };
+
+                // if we have already a channel with them, we try to forward the payment.
+                if let Some(channel) = self
+                    .channel_manager
+                    .list_channels()
+                    .iter()
+                    // The coordinator can only have one channel with each app. Hence, if we find a
+                    // channel with the target of the intercepted HTLC, we know
+                    // that it is the only channel between coordinator and
+                    // target app and we can forward the intercepted HTLC through it.
+                    .find(|channel_details| channel_details.counterparty.node_id == target_node_id)
                 {
+                    // Note, the forward intercepted htlc might fail due to insufficient balance,
+                    // since we do not check yet if the channel outbound capacity is sufficient.
                     if let Err(error) = self.channel_manager.forward_intercepted_htlc(
                         intercept_id,
                         &channel.channel_id,
@@ -461,21 +478,6 @@ impl EventHandler {
 
                     return Ok(());
                 }
-
-                let fake_channel_payments = self.fake_channel_payments.lock().unwrap();
-
-                let target_node_id = match fake_channel_payments.get(&requested_next_hop_scid) {
-                    None => {
-                        tracing::warn!(fake_scid = requested_next_hop_scid, "Could not forward the intercepted HTLC because we didn't have a node registered with said fake scid");
-
-                        if let Err(err) = self.channel_manager.fail_intercepted_htlc(intercept_id) {
-                            tracing::error!("Could not fail intercepted htlc {err:?}")
-                        }
-
-                        return Ok(());
-                    }
-                    Some(target_node_id) => target_node_id,
-                };
 
                 // FIXME: This will set the channel capacity to twice the amount that is
                 // transferred or the `JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT` ensuring there
@@ -503,7 +505,7 @@ impl EventHandler {
                 // `max_inbound_htlc_value_in_flight_percent_of_channel`
                 // configuration value
                 let temp_channel_id = match self.channel_manager.create_channel(
-                    *target_node_id,
+                    target_node_id,
                     channel_value,
                     0,
                     0,
@@ -534,7 +536,7 @@ impl EventHandler {
                 let pending_intercepted_htlcs = self.pending_intercepted_htlcs.clone();
                 let mut pending_intercepted_htlcs = pending_intercepted_htlcs.lock().unwrap();
                 pending_intercepted_htlcs.insert(
-                    *target_node_id,
+                    target_node_id,
                     (intercept_id, expected_outbound_amount_msat),
                 );
             }
--- a/crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
+++ b/crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
@@ -28,7 +28,7 @@ async fn just_in_time_channel() {
     let coordinator_outbound_liquidity_sat =
         min_outbound_liquidity_channel_creator(&payer, payer_outbound_liquidity_sat);
 
-    let payer_coordinator_channel_details = coordinator
+    coordinator
         .open_channel(
             &payer,
             coordinator_outbound_liquidity_sat,
@@ -36,14 +36,6 @@ async fn just_in_time_channel() {
         )
         .await
         .unwrap();
-
-    let payer_balance_before = payer.get_ldk_balance();
-    let coordinator_balance_before = coordinator.get_ldk_balance();
-    let payee_balance_before = payee.get_ldk_balance();
-
-    // Act
-
-    let intercept_scid = coordinator.create_intercept_scid(payee.info.pubkey);
 
     // This comment should be removed once the implementation of
     // `Node` is improved.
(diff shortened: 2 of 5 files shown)


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -432,20 +432,34 @@ impl EventHandler {
...
+                        tracing::warn!(fake_scid = requested_next_hop_scid, "Could not forward the intercepted HTLC because we didn't have a node registered with said fake scid");
+
+                        if let Err(err) = self.channel_manager.fail_intercepted_htlc(intercept_id) {
+                            tracing::error!("Could not fail intercepted htlc {err:?}")
+                        }
+
+                        return Ok(());
+                    }
+                    Some(target_node_id) => target_node_id,
+                };
```
luckysori, line 448: 🔧 It's generally good practice to drop a `MutexGuard` as soon as you're done with it, so that some other thread can access the resource immediately.

```suggestion
                let target_node_id = {
                    let fake_channel_payments = self.fake_channel_payments.lock().unwrap();
                    match fake_channel_payments.get(&requested_next_hop_scid) {
                        None => {
                            tracing::warn!(fake_scid = requested_next_hop_scid, "Could not forward the intercepted HTLC because we didn't have a node registered with said fake scid");

                            if let Err(err) =
                                self.channel_manager.fail_intercepted_htlc(intercept_id)
                            {
                                tracing::error!("Could not fail intercepted htlc {err:?}")
                            }

                            return Ok(());
                        }
                        Some(target_node_id) => *target_node_id,
                    }
                };
```

Definitely not a problem with your code as you've just moved code around, but it's something to keep in mind in general.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -432,20 +432,34 @@ impl EventHandler {
...
-                                scid == requested_next_hop_scid
-                            } else {
-                                false
-                            }
-                        })
+                if let Some(channel) = self
+                    .channel_manager
+                    .list_channels()
+                    .iter()
+                    .find(|channel_details| channel_details.counterparty.node_id == *target_node_id)
```
luckysori, line 455: 🔧 It would help to leave a comment about why this should work e.g.
```suggestion
                    // The coordinator can only have one channel with each app. Hence, if we find a channel
                    // with the target of the intercepted HTLC, we know that it is the only channel between
                    // coordinator and target app and we can forward the intercepted HTLC through it.
                    .find(|channel_details| channel_details.counterparty.node_id == *target_node_id)
```

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -432,20 +432,34 @@ impl EventHandler {
...
                 {
+                    if channel.outbound_capacity_msat < expected_outbound_amount_msat {
+                        // TODO: here we would need to increase the channel size if the channel is
+                        // too small
+                        tracing::warn!("Cannot forward payment on channel, due to insufficient outbound capacity.")
+                    }
```
luckysori, line 461: 🤔 Looking back on this, I still don't like checking the pre-condition here. This is the job of the `forward_intercepted_htlc` API we call a few lines below.

I think this just leads to:

- Redundant error/warning messages if we emit this log and subsequently fail below.
- Confusing logs, where we log a warning but still succeed below for some unforeseen reason.

Even keeping the TODO is debatable for me, because it's not the only way that we can solve the problem of not having enough liquidity to route a payment to an app. For example, we could alternatively fail during invoice creation, since we know the inbound liquidity of our channel with the coordinator, and show the user an option to explicitly increase the liquidity on the UI.

Given all this, I'd be happy if this bit of code was removed.

crates/ln-dlc-node/src/tests/just_in_time_channel_with_multiple_payments.rs
luckysori, line 1: 🔧 Can we create a `just_in_time_channel` module with `just_in_time_channel/create` and `just_in_time_channel/multiple_payments` sub-modules?

crates/ln-dlc-node/src/tests/just_in_time_channel_with_multiple_payments.rs
```diff
@@ -0,0 +1,59 @@
...
+use bitcoin::Amount;
+
+#[tokio::test]
+#[ignore]
+async fn just_in_time_channel_with_multiple_payments() {
+    init_tracing();
+
+    // Arrange
+
+    let user_a = Node::start_test_app("payer").await.unwrap();
```
luckysori, line 15:
```suggestion
    let user_a = Node::start_test_app("user_a").await.unwrap();
```

crates/ln-dlc-node/src/tests/just_in_time_channel_with_multiple_payments.rs
```diff
@@ -0,0 +1,59 @@
...
+    let coordinator = Node::start_test_coordinator("coordinator").await.unwrap();
+    let user_b = Node::start_test_app("payee").await.unwrap();
```
luckysori, line 17:
```suggestion
    let user_b = Node::start_test_app("user_b").await.unwrap();
```

crates/ln-dlc-node/src/tests/just_in_time_channel.rs
```diff
@@ -70,44 +62,108 @@ async fn just_in_time_channel() {
...
+        .iter()
+        .find_map(|c| (&c.channel_id == channel_id).then_some(c.channel_value_satoshis))
+        .context("No matching channel")?;
+    let channel_size_sat = Decimal::from(channel_size_sat);
+
+    let max_inbound_htlc_sat =
+        channel_size_sat * (max_inbound_htlc_as_percent_of_channel / Decimal::ONE_HUNDRED);
+
+    Ok(htlc_amount_sat <= max_inbound_htlc_sat)
+}
```
luckysori, line 106: 🔧 Since this is less important let's move it to the bottom of the file.

crates/ln-dlc-node/src/tests/just_in_time_channel.rs
```diff
@@ -120,43 +176,16 @@ async fn just_in_time_channel() {
...
 
-    let channel_size_sat = receiving_node
-        .channel_manager
-        .list_channels()
-        .iter()
-        .find_map(|c| (&c.channel_id == channel_id).then_some(c.channel_value_satoshis))
-        .context("No matching channel")?;
-    let channel_size_sat = Decimal::from(channel_size_sat);
+    // no additional just-in-time channel should be created.
+    assert_eq!(coordinator.channel_manager.list_channels().len(), 2);
```
luckysori, line 188: 🙃 We could move this to the test since we don't actually know if the caller's `coordinator` should only have 2 channels.

crates/ln-dlc-node/src/tests/just_in_time_channel.rs
```diff
@@ -70,44 +62,108 @@ async fn just_in_time_channel() {
...
+
+    Ok(htlc_amount_sat <= max_inbound_htlc_sat)
+}
+
+pub(crate) async fn send_interceptable_payment(
+    payer: &Node,
+    payee: &Node,
+    coordinator: &Node,
+    invoice_amount: u64,
+    coordinator_balance_before_payment: u64,
```
luckysori, line 113: 🔧 I led you astray when I suggested this name yesterday. This is actually the `coordinator_just_in_time_channel_creation_outbound_liquidity`.

It might be clearer to pass in an `Option<u64>`, because `None` might more clearly indicate that the caller knows that no JIT channel will be created with that interceptable payment. Then the function can use `unwrap_or_default` to get either the `value` in `Some(value)` or `0`.

luckysori (approved):
LGTM!

crates/ln-dlc-node/src/ln/event_handler.rs
holzeis: I like the idea with moving the check to the invoice. However, this approach is also not 100% bullet proof as the user might create multiple invoices.

holzeis: I removed the if condition and added a comment to the `forward_intercepted_htlc` function call.

luckysori: > I like the idea with moving the check to the invoice. However, this approach is also not 100% bullet proof as the user might create multiple invoices.

Yeah, you're right! I was primarily trying to argue that the TODO is a bit too opinionated for me at this stage.

crates/ln-dlc-node/src/tests/just_in_time_channel_with_multiple_payments.rs
holzeis: sure 🙂

holzeis:
bors retry

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -432,20 +432,37 @@ impl EventHandler {
...
-                        .list_channels()
-                        .iter()
-                        .find(|channel_details| {
-                            if let Some(scid) = channel_details.short_channel_id {
-                                scid == requested_next_hop_scid
-                            } else {
-                                false
+                let target_node_id = {
+                    let fake_channel_payments = self.fake_channel_payments.lock().unwrap();
+                    match fake_channel_payments.get(&requested_next_hop_scid) {
```
bonomat, line 437: Do I understand it correctly that `intercept_htlc` is really just being called for `fake_channel_ids` and for `known_channel_ids` it is not being called?

bonomat: > But even if we did support "regular" invoices, the matching HTLC should not trigger this event and should be handled differently (I think automatically by LDK).

Cool, I did not know this. 

> The app currently only supports interceptable invoices

I thought I had added logic to decided wether to go for intercept able invoices or _normal_ invoices.

luckysori: It looks like the app only supports interceptable invoices currently:

https://github.com/get10101/10101/blob/e2552992fb3ad2bdbd9bd954ef9c002b6099bdf0/mobile/native/src/ln_dlc/mod.rs#L313-L346

We could change that in the future, although I think @holzeis had reservations.

holzeis: I think we need both implementations eventually. I don't think we can go without that fix as it is not always possible for the app to know what invoice to create. e.g. assume the user creates an interceptable invoice (since no channel exists yet). Before that invoice has been paid, the user creates a second invoice. At this point the app has not enough information to create either a regular invoice (assuming the interceptable invoice will be paid first) or another interceptable invoice (assuming the first invoice does not get paid).

Eventually we should go with a fix that does not create a interceptable invoice if a channel already exists, but this is not critical for now. That's why this fix on focused on ensuring that multiple interceptable invoices will not lead to multiple channels.

PR #365, opened 2023-04-04
feat: The orderbook rejects orders with zero price

da-kami:
An order with zero price does not make any sense and should be rejected.

### DIFF
--- a/coordinator/src/orderbook/routes.rs
+++ b/coordinator/src/orderbook/routes.rs
@@ -28,6 +28,7 @@ use orderbook_commons::OrderType;
 use orderbook_commons::OrderbookMsg;
 use orderbook_commons::OrderbookRequest;
 use orderbook_commons::Signature;
+use rust_decimal::Decimal;
 use serde::Deserialize;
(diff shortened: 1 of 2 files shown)


klochowicz (approved):

coordinator/src/orderbook/routes.rs
```diff
@@ -82,6 +83,12 @@ pub async fn post_order(
     State(state): State<Arc<AppState>>,
     Json(new_order): Json<NewOrder>,
 ) -> Result<Json<Order>, AppError> {
+    if new_order.price == Decimal::ZERO {
```
bonomat, line 86: We could do `<=0` to also not allow negative prices.

bonomat (approved):

holzeis (approved):
Only downside to this fix is, that it is fixing a symptom rather than the issue. 🙂

bonomat:
> Only downside to this fix is, that it is fixing a symptom rather than the issue. 🙂

What issue do you see?

PR #367, opened 2023-04-05
Use `get_offered_sub_channels` after fix in `rust-dlc`

luckysori:
The bug mentioned in the diff was fixed a while ago. We are now on a version of `rust-dlc` where `get_offered_sub_channels` is safe to call.

### DIFF
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -116,11 +116,13 @@ impl Node {
     }
 
     pub fn get_dlc_channel_offer(&self, pubkey: &PublicKey) -> Result<Option<SubChannel>> {
-        let matcher = |dlc_channel: &&SubChannel| {
-            dlc_channel.counter_party == *pubkey
-                && matches!(&dlc_channel.state, SubChannelState::Offered(_))
-        };
-        let dlc_channel = self.get_dlc_channel(&matcher)?; // `get_offered_sub_channels` appears to have a bug
+        let dlc_channel = self
+            .dlc_manager
+            .get_store()
+            .get_offered_sub_channels()
+            .map_err(|e| anyhow!(e.to_string()))?
+            .into_iter()
+            .find(|dlc_channel| dlc_channel.counter_party == *pubkey);
 
         Ok(dlc_channel)
     }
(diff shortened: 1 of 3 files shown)


holzeis (approved):
LGTM, but are we risking here fixing code that did not have any issues? I guess I wouldn't have touched this as long as it is working 😅

luckysori:
> LGTM, but are we risking here fixing code that did not have any issues? I guess I wouldn't have touched this as long as it is working 😅

We have tests that call the API, so if the tests pass I think we are in the clear.

crates/ln-dlc-node/src/node/connection.rs
```diff
@@ -40,16 +48,20 @@ impl Node {
     }
 
     pub async fn keep_connected(&self, peer: NodeInfo) -> Result<()> {
-        // TODO: Let this time out
-        let connection_closed_future = loop {
-            tracing::debug!(%peer, "Attempting to establish initial connection");
+        #[allow(clippy::async_yields_async)] // we want to poll the returned future in a loop
+        let connection_closed_future = tokio::time::timeout(Duration::from_secs(30), async {
```
holzeis, line 52: 🔧 I am not in favor of this, as I think it makes sense for the app to indefinitely try to reconnect to the coordinator. If I read this correctly the behavior would change towards the app only tries 30 seconds after loosing the connection and would then end up in a failed state.

luckysori: This is only with respect to the initial connection. It was designed to only return after establishing the initial connection, which is why it can block execution indefinitely.

In practice, the feature of waiting for the initial connection does not matter for the only production user (the app) because we spawn a task to run `keep_connected` anyway.

That means the feature is only useful for the tests. But it is debatable if it is even necessary, as we could just call `connect_to_peer`, since we _shouldn't_ need to reconnect in the tests.

luckysori: Your comment does make me think that maybe the `keep_connected` code should be in the `app`. For now I think I will revert this part of the change and add another patch which:

- Replaces `keep_connected` with `connect_to_peer` in the tests.
- Rewrites `keep_connected` to not do anything special with the initial connection and simply keep the connection alive.

luckysori: Moved these unrelated changes to https://github.com/get10101/10101/pull/370.

luckysori:
bors retry

PR #369, opened 2023-04-05
fix: Reset lnd mission control on send payment

### LINKED ISSUE #333: 10101 mobile app can reach unfundable state
### Repro

1. Fresh install of 10101 app on TestFlight (version 1.0.0, build 46).
2. Generate an invoice to fund the app.
3. _Close the app_.
4. Attempt to pay invoice via public faucet.
5. Payment fails because the app is closed (`no_route`).
6. Open the app and generate a new invoice.
7. Keep the app open.
8. Attempt to pay invoice via public faucet.
9. Payment fails due to `no_route`.

The app never recovers from this. The cause is unknown. The only workaround is to reinstall the app, generating a new seed.

---

A note on step 5: we know that the app has to be open in order to receive payments. Ideally we would leave the HTLC pending until the user opens the app (via a push notification perhaps). For now, we don't need to fix the error in step 5, as the user is only ever funding their app once and we can instruct them to keep it open.

holzeis:
lnd is keeping a heuristic for routing attempts based on past payment attempts. This information is kept in the mission control and seems to block any subsequent payment to a node that has previously failed. If it is the very first payment that fails I guess this heuristic is even heavier opinionated.

https://docs.lightning.engineering/lightning-network-tools/lnd/payments#mission-control

In order to bypass this mission control, this change simply resets the mission control on every payment. resolves #333 

----

Note, I also tried setting setting the maximum number of payment results held on disk by mission control (`routerrpc.maxmchistory`) to 0, but without success. Other than that I found no meaningful configuration to disable lnds mission control. see also https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf

### DIFF
--- a/faucet/faucet.html
+++ b/faucet/faucet.html
@@ -109,8 +109,22 @@
         xhr.send();
       }
 
+      function clearMissionControl() {
+        const xhr = new XMLHttpRequest();
+        xhr.open('post', '/lnd/v2/router/mc/reset', false);
+        xhr.setRequestHeader('Content-Type', 'application/json');
+        xhr.send();
+        if (xhr.status === 200) {
+          console.log("Successfully cleared mission control.");
+        } else {
+          console.warn("Failed to clear mission control.")
+        }
+      }
+
       function sendPayment() {
         const encoded_invoice = document.getElementById("inputInvoice").value;
+
+        clearMissionControl();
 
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Wooo, amazing find!

bonomat (approved):

PR #370, opened 2023-04-05
Improve LN connection handling

luckysori:
I have some very small hope that these changes might fix our intermittent CI issues.


holzeis (approved):
LGTM 👍

Issue #355, opened 2023-04-03
CI often hangs on `run-tests-ln-dlc-node` job


luckysori:
We started seeing this on and off since [this run](https://github.com/get10101/10101/actions/runs/4444630040). We cannot associate it with a particular code change because it doesn't always happen.

It could be related to GitHub's infrastructure but I wouldn't know why.

I am not able to reproduce this problem locally.

luckysori:
Interestingly, the previous PR triggered exactly the same tests (2 hours before this issue started affecting us) and they passed: https://github.com/get10101/10101/actions/runs/4570833858.

luckysori:
Turns out it was just a transient problem with CI.

luckysori:
We're actually still seeing this, but not necessarily with the same test: https://github.com/get10101/10101/actions/runs/4612585983/jobs/8153652603. I'm not sure what to do about this issue if it only happens sometimes.

luckysori:
Re-opening as it keeps happening.

luckysori:
[This CI run](https://github.com/get10101/10101/actions/runs/4619199701/jobs/8167629311) was automatically cancelled after 30 minutes given the changes in #371. We can see that the last test to reach completion was actually running until the very end, so there is a chance that we can debug this problem.

klochowicz:
I wonder whether this would reproducible with https://github.com/nektos/act ?

klochowicz:
I'll try to migrate our old setup for partitioned tests from ItchySats to see whether it will make a difference - it should be just copy-paste; it would certainly distribute test load, might as well give it a go.

bonomat:
This keeps happening here as well: https://github.com/get10101/10101/pull/414#issuecomment-1508958660

luckysori:
Definitely not fixed with f00ec25 unfortunately.

luckysori:
I may have a new lead!

I randomly checked this run that failed because it was hanging and it's interesting because it hanged on _the very first test_ it tried to run. This means that we should be able to definitively determine where the test is hanging on CI (as far as the logs can provide context).

Here's the last interesting log message AFAICT: https://github.com/get10101/10101/actions/runs/5005062691/jobs/8968486859#step:8:249.

Here's the line where I think it hangs in the test (although it could possibly be the next time we call `process_incoming_messages`): 

[...]

PR #371, opened 2023-04-05
Time out `run-tests-ln-dlc-node` job after 30 minutes

luckysori:
We have seen these tests hang for hours, which is wasteful and annoying as we don't get feedback fast enough unless we actively check the Actions tab.

Choosing 30 minutes as a timeout seems sensible as no successful run has ever gone over 15 minutes and most of them are much faster than that.

---

Related to #355.


holzeis (approved):
seems reasonable, thanks 👍

luckysori:
bors retry

luckysori:
bors retry

PR #384, opened 2023-04-10
Fix/add pnl calc to tx history

### LINKED ISSUE #311: Use payout calculation when calculating close position trade amount in TX history
At the moment we use the margin calculation, which is incorrect because it doesn't include the PNL.

See: https://github.com/get10101/10101/pull/307#discussion_r1144213133.

holzeis:
resolves #311


holzeis:
bors r+

@luckysori Thanks for the review, I created a Technical Debt ticket from your remarks. Going ahead with the merge though as it is functional what we need right now.

PR #387, opened 2023-04-11
fix: the called workflow needs access to our secrets

bonomat:
Let's try this to fix our testflight pipeline

Taken from: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsecretsinherit


holzeis (approved):
Nice 👍

PR #389, opened 2023-04-11
Constrain max leverage to 5

### LINKED ISSUE #380: Limit maximum leverage to 5
At this stage we're not planning to go with higher leverages - we don't want for people to get rekt. 5x should be plenty for most users.

klochowicz:
Make it easier to adjust these values in the future.

Fixes #380


holzeis (approved):
LGTM 👍

PR #390, opened 2023-04-11
add logo in svg

bonomat:
Replaced all logos with the latest one from figma. I've also added all kinds of different versions to the repository :)


klochowicz (approved):
looks good!

nit: formatting failed on some json, we can either exclude the xcode asset dirs or re-run `dprint` before merging.

PR #391, opened 2023-04-11
Feat/seed backup

holzeis:
This feature has been mainly copied over from our poc and does not adhere to our current figma design. It is the very minimum we need for mainnet.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -48,6 +48,11 @@ pub async fn refresh_wallet_info() -> Result<()> {
     Ok(())
 }
 
+pub fn get_seed_phrase() -> Result<Vec<String>> {
+    let node = NODE.try_get().context("failed to get ln dlc node")?;
+    Ok(node.get_seed_phrase())
+}
+
 pub fn get_node_key() -> Result<SecretKey> {
     NODE.try_get()
         .context("failed to get ln dlc node")?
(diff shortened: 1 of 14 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -48,6 +48,11 @@ pub async fn refresh_wallet_info() -> Result<()> {
     Ok(())
 }
 
+pub fn get_seed_phrase() -> Result<Vec<String>> {
+    let node = NODE.try_get().context("failed to get ln dlc node")?;
```
klochowicz, line 52: I would argue that if we never manipulate / change it, we should just ignore `PoisonError` and be always reaching into the actual Seed. That will slightly simplify this code as there's no `Result<>`.

klochowicz (approved):
It does the job! 

feel free to ignore the feedback for now, it can be addressed later and it's not crucial :)

mobile/native/src/ln_dlc/mod.rs
holzeis: Agreed! But then we should also adapt this on all other functions where we are accessing the node like that. (and I didn't want to do this now ;)

luckysori: Not sure if I get you, @klochowicz, but I would propose to always `expect` when accessing the `NODE` since it must be set at all times after startup (and if it isn't crashing the app is best).

PR #394, opened 2023-04-11
WIP: charge just in time channel creation fees

### LINKED ISSUE #393: Only charge fees on opening channel
Currently we charge 2% on every routing transaction. Instead we want to charge the user for the on-chain transaction costs when opening an just-in-time channel. Routing fees are set 1 sat + 1% (this is the default config on the lnd) for the beta phase.

// edit: changed routing fees from 0 to 1 sat + 1%.

holzeis:
resolves #393

### DIFF
--- a/crates/ln-dlc-node/src/node/mod.rs
+++ b/crates/ln-dlc-node/src/node/mod.rs
@@ -112,7 +112,12 @@ pub struct NodeInfo {
 
 /// Liquidity-based routing fee in millionths of a routed amount. In
 /// other words, 10000 is 1%.
-pub(crate) const LIQUIDITY_ROUTING_FEE_MILLIONTHS: u32 = 20_000;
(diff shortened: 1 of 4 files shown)


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -112,7 +112,12 @@ pub struct NodeInfo {
 
 /// Liquidity-based routing fee in millionths of a routed amount. In
 /// other words, 10000 is 1%.
-pub(crate) const LIQUIDITY_ROUTING_FEE_MILLIONTHS: u32 = 20_000;
+pub(crate) const LIQUIDITY_ROUTING_FEE_MILLIONTHS: u32 = 10_000;
+
+/// We have to cover our costs for broadcasting the funding transactions on chain. In order to
+/// simplify things we simply charge a fee of 1000 sats instead of substracting the actual
```
luckysori, line 118: 🙃 It's risky to document the value as it could diverge from the actual value.

PR #396, opened 2023-04-12
Persist payments in app's DB

luckysori:
The `PaymentPersister` trait is introduced so that different consumers of `ln-dlc-node` can bring their own implementations. Alternatively, we could start moving away from `ln-dlc-node` altogether, since most of the important code differs between consumers, but that's too much work at the moment.

The app brings an implementation based on its SQLite DB. The coordinator and the maker use `ln_dlc_node::PaymentMap`, so their behaviour shouldn't change w.r.t. the previous patch.

---

~Small caveat: doesn't quite work yet, as I can't see the payment history at all. But the code shouldn't change much after the fix~. This was fixed. The problem was that we were not updating the `amount_msat` correctly, so the payments were being ignored for display purposes.

### DIFF
--- a/crates/ln-dlc-node/src/node/mod.rs
+++ b/crates/ln-dlc-node/src/node/mod.rs
@@ -11,7 +11,6 @@ use crate::util;
 use crate::ChainMonitor;
 use crate::FakeChannelPaymentRequests;
 use crate::InvoicePayer;
-use crate::PaymentInfoStorage;
 use crate::PeerManager;
 use anyhow::ensure;
 use anyhow::Context;
@@ -60,6 +59,7 @@ mod dlc_manager;
 pub(crate) mod invoice;
 mod ln_channel;
 mod oracle_client;
+mod payment_persister;
 mod sub_channel_manager;
 mod wallet;
 
@@ -68,20 +68,21 @@ pub use ::dlc_manager as rust_dlc_manager;
 pub use channel_manager::ChannelManager;
 pub use dlc_channel::sub_channel_message_as_str;
 pub use invoice::HTLCStatus;
+pub use payment_persister::PaymentMap;
+pub use payment_persister::PaymentPersister;
 pub use sub_channel_manager::SubChannelManager;
 pub use wallet::PaymentDetails;
-pub use wallet::PaymentFlow;
 
 // TODO: These intervals are quite arbitrary at the moment, come up with more sensible values
 const BROADCAST_NODE_ANNOUNCEMENT_INTERVAL: Duration = Duration::from_secs(60);
 
 /// An LN-DLC node.
-pub struct Node {
+pub struct Node<P> {
     network: Network,
 
     pub(crate) wallet: Arc<LnDlcWallet>,
     pub peer_manager: Arc<PeerManager>,
-    invoice_payer: Arc<InvoicePayer<EventHandler>>,
+    invoice_payer: Arc<InvoicePayer<EventHandler<P>>>,
     pub(crate) channel_manager: Arc<ChannelManager>,
     chain_monitor: Arc<ChainMonitor>,
     keys_manager: Arc<CustomKeysManager>,
@@ -98,8 +99,7 @@ pub struct Node {
     pub sub_channel_manager: Arc<SubChannelManager>,
     oracle: Arc<P2PDOracleClient>,
     pub dlc_message_handler: Arc<DlcMessageHandler>,
-    inbound_payments: PaymentInfoStorage,
-    outbound_payments: PaymentInfoStorage,
+    payment_persister: Arc<P>,
 
     pub(crate) user_config: UserConfig,
 }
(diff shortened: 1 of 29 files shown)


crates/ln-dlc-node/src/node/payment_persister.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn insert(&self, payment_hash: PaymentHash, info: PaymentInfo) -> Result<()>;
+    fn upsert(
+        &self,
+        payment_hash: &PaymentHash,
+        flow: PaymentFlow,
+        htlc_status: HTLCStatus,
+        preimage: Option<PaymentPreimage>,
+        secret: Option<PaymentSecret>,
+    ) -> Result<()>;
+    fn get(&self, payment_hash: &PaymentHash) -> Result<Option<(PaymentHash, PaymentInfo)>>;
```
klochowicz, line 24: perhaps some comments would be good? e.g. why it can return an Result<Option<T>>

crates/ln-dlc-node/src/node/payment_persister.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    fn upsert(
+        &self,
+        payment_hash: &PaymentHash,
+        flow: PaymentFlow,
+        htlc_status: HTLCStatus,
+        preimage: Option<PaymentPreimage>,
+        secret: Option<PaymentSecret>,
+    ) -> Result<()> {
+        let mut payments = self.0.lock().unwrap();
```
klochowicz, line 46: did you mean to keep this as `unwrap()`? this seems to be the only opportunity to get an error in this call

crates/ln-dlc-node/src/node/payment_persister.rs
```diff
@@ -0,0 +1,92 @@
...
+                    },
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    fn get(&self, payment_hash: &PaymentHash) -> Result<Option<(PaymentHash, PaymentInfo)>> {
+        let payments = self.0.lock().unwrap();
```
klochowicz, line 78: same as above - can we return an `Err`?

crates/ln-dlc-node/src/node/payment_persister.rs
```diff
@@ -0,0 +1,92 @@
...
+        let info = payments.get(payment_hash);
+
+        let payment = info.map(|info| (*payment_hash, *info));
+
+        Ok(payment)
+    }
+
+    fn all(&self) -> Result<Vec<(PaymentHash, PaymentInfo)>> {
+        let payments = self.0.lock().unwrap();
```
klochowicz, line 87: same as above - can we return an Err?

mobile/native/src/db/custom_types.rs
```diff
@@ -168,6 +170,54 @@ impl FromSql<Text, Sqlite> for PositionState {
...
+        let string = <String as FromSql<Text, Sqlite>>::from_sql(bytes)?;
+
+        return match string.as_str() {
+            "Inbound" => Ok(Flow::Inbound),
+            "Outbound" => Ok(Flow::Outbound),
+            _ => Err("Unrecognized enum variant".into()),
+        };
+    }
+}
+
```
klochowicz, line 220: (unrelated to this particular change, just a meta observation) : it's sad that we need to do that manually with diesel :( can't `serde` deal with this?

mobile/native/src/db/models.rs
```diff
@@ -551,11 +555,265 @@ impl From<crate::trade::order::FailureReason> for FailureReason {
...
+
+    /// updates the status of the given order in the db
+    pub fn update(
+        payment_hash: String,
+        htlc_status: HtlcStatus,
+        preimage: Option<String>,
+        secret: Option<String>,
+        conn: &mut SqliteConnection,
+    ) -> Result<i64> {
+        let updated_at = OffsetDateTime::now_utc().unix_timestamp();
```
klochowicz, line 609: I like the fact that we don't need to pass in this update time manually, and it got hidden deep :)

klochowicz (approved):
great job!!!  🥇

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -11,7 +11,7 @@ use crate::util;
 use crate::ChainMonitor;
 use crate::FakeChannelPaymentRequests;
 use crate::InvoicePayer;
-use crate::PaymentInfoStorage;
+
```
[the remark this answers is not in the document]
luckysori: I blame my new lsp client: `lsp-bridge`.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -98,8 +101,7 @@ pub struct Node {
     pub sub_channel_manager: Arc<SubChannelManager>,
     oracle: Arc<P2PDOracleClient>,
     pub dlc_message_handler: Arc<DlcMessageHandler>,
-    inbound_payments: PaymentInfoStorage,
-    outbound_payments: PaymentInfoStorage,
+    payment_persister: P,
```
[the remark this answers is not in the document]
luckysori: Yeah, I think what you propose makes more sense!

crates/ln-dlc-node/src/node/payment_persister.rs
luckysori: I did mean to keep it, because I think `unwrap`ping `LockResult`s is usually best.

The reason why we have to return an error is because we are implementing a trait method which returns a `Result`. For a `HashMap` it's irrelevant, but when accessing a database it's a different story.

mobile/native/src/db/custom_types.rs
luckysori: This boilerplate is annoying, I agree.

luckysori:
bors retry

PR #400, opened 2023-04-12
Flip ask/bid once for all

bonomat:
The second commit might have been the reason why we saw weird prices and quick P/L


holzeis (approved):
I guess 😅

PR #407, opened 2023-04-13
fix: Accept settlement amount calculation

holzeis:
Settling a position must calculate the accept settlement amount from the viewpoint of the position at the time it was opened. The reversed order in terms of direction has to be flipped. Accordingly, we need to also flip the leverage as we would otherwise use the coordinators leverage.


1010Tom (approved):
Sounds reasonable to me :)

PR #408, opened 2023-04-13
Feat/show expiry on position

### LINKED ISSUE #385: Show the expiry in the position and when creating an order
e.g. 00:00 every day

holzeis:
resolves #385

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -66,7 +66,7 @@ impl Node {
         let margin_trader = margin_trader(trade_params);
         let margin_coordinator = margin_coordinator(trade_params);
 
-        let tomorrow = OffsetDateTime::now_utc().date() + Duration::days(1);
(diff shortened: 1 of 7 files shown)


coordinator/src/node.rs
```diff
@@ -66,7 +66,7 @@ impl Node {
         let margin_trader = margin_trader(trade_params);
         let margin_coordinator = margin_coordinator(trade_params);
 
-        let tomorrow = OffsetDateTime::now_utc().date() + Duration::days(1);
+        let tomorrow = OffsetDateTime::now_utc().date() + Duration::days(2);
```
1010Tom, line 69: Why +2 days?

holzeis: so it will be open for at least 24h, up to 47 hours and 59 mins.

1010Tom (approved):
LGTM

PR #415, opened 2023-04-13
Placeholders for empty position / orders tabs

da-kami:
- If there is no position / orders then show placeholder message
- If there is an order currently being filled then show an according message in position tab

image

image


holzeis (approved):
LGTM 👍

bonomat (approved):
Functional :)

PR #417, opened 2023-04-14
Prevent order state from going back in time for app

### LINKED ISSUE #336: We have a DLC but were unable to set the order to filled
When opening a dlc from the phone, the filling order can not be found in the database. No other error messages were seen in the logs until the point where the order is fetched by state.

Note, this is only happening on the App released on TestFlight build 46 and is not reproducible on the simulator.

```
| 2023-03-27 13:59:57.455 | LogLevel.DEBUG | New message from orderbook 
| 2023-03-27 13:59:57.456 | LogLevel.INFO | Received match from orderbook 
| 2023-03-27 13:59:59.156 | LogLevel.INFO | Sent trade request to coordinator successfully 
| 2023-03-27 14:00:13.512 | LogLevel.DEBUG | Processing DLC channel message 
| 2023-03-27 14:00:13.534 | LogLevel.INFO | Accepting DLC channel offer 
| 2023-03-27 14:00:34.095 | LogLevel.DEBUG | Processing DLC channel message 
| 2023-03-27 14:00:34.256 | LogLevel.DEBUG | Sending DLC channel message 
| 2023-03-27 14:00:34.257 | LogLevel.ERROR | Critical Error! We have a DLC but were unable to set the order to filled: There is no order in state filled in the database 
```

luckysori:
Fixes #336.

Tested it at least once and it works™️.

### DIFF
--- a/mobile/native/src/db/mod.rs
+++ b/mobile/native/src/db/mod.rs
@@ -114,6 +114,23 @@ pub fn insert_order(order: trade::order::Order) -> Result<trade::order::Order> {
 
 pub fn update_order_state(order_id: Uuid, order_state: trade::order::OrderState) -> Result<()> {
     let mut db = connection()?;
+
+    // We should not walk back the state of orders in general. Here we specifically want to prevent
+    // `Filling` orders from going back to `Open`.
+    if let trade::order::OrderState::Open = order_state {
+        let order = get_order(order_id)?;
+        let current_state = order.state;
+
+        use crate::trade::order::OrderState::*;
+        if let Rejected | Open | Filling { .. } | Failed { .. } | Filled { .. } = current_state {
+            tracing::debug!(
+                ?current_state,
+                "Ignoring order update; Open state is old news"
+            );
+            return Ok(());
+        };
+    }
+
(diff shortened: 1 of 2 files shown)


klochowicz (approved):

mobile/native/src/db/mod.rs
```diff
@@ -114,6 +114,23 @@ pub fn insert_order(order: trade::order::Order) -> Result<trade::order::Order> {
 
 pub fn update_order_state(order_id: Uuid, order_state: trade::order::OrderState) -> Result<()> {
     let mut db = connection()?;
+
```
klochowicz, line 117: Shall we also prevent going from a final state back to an open state?

luckysori: Maybe the comment is confusing, because what you propose is what we are doing.

With the comment I just wanted to highlight that we are preventing a specific scenario that has happened to us: filling to open. But we are _also_ preventing final state to open.

klochowicz: yup, got confused by the comment!

luckysori: Do adapt it if you think it's important 😃 But I need to 💤 now.

PR #418, opened 2023-04-14
chore: electrum is a url and not a socket

1010Tom:
this allows us to use domains and not just IPs.

This way we can pass in `ELECTRS_ENDPOINT=tcp://blockstream.info:110"`


holzeis (approved):
Makes sense! Thanks for fixing.

bonomat:
Tests run locally, f*ck bors 😈

PR #420, opened 2023-04-14
Feat/collab close position expired positions

### LINKED ISSUE #386: Collaboratively close position at expiry time
when the user opens the app, we should collaboratively close the position at the price at expiry time

holzeis:
resolves #386 

TODO
- [x] Set the execution price on the fake order at the time the position got closed

### DIFF
--- a/crates/ln-dlc-node/Cargo.toml
+++ b/crates/ln-dlc-node/Cargo.toml
@@ -32,7 +32,7 @@ log = "0.4.17"
 p2pd-oracle-client = { version = "0.1.0" }
 rand = "0.8.5"
 reqwest = { version = "0.11", default-features = false, features = ["json"] }
-secp256k1-zkp = { version = "0.7.0" }
+secp256k1-zkp = { version = "0.7.0", features = ["global-context"] }
 serde = "1.0.147"
 sha2 = "0.10"
 simple-wallet = "0.1.0"


coordinator/src/bin/coordinator.rs
```diff
@@ -121,13 +127,75 @@ async fn main() -> Result<()> {
...
+                        let positions = positions
+                            .into_iter()
+                            .filter(|p| {
+                                p.position_state == PositionState::Open
+                                    && OffsetDateTime::now_utc().ge(&p.expiry_timestamp)
+                            })
+                            .collect::<Vec<Position>>();
+
+                        for position in positions.iter() {
+                            tracing::debug!(%position.expiry_timestamp, "Attempting to closed expired position with {}", position.trader);
```
1010Tom: Nit: (not sure if this compiles) 

```suggestion
tracing::debug!(trader_pk=position.trader, %position.expiry_timestamp, "Attempting to close expired position");
```

coordinator/src/bin/coordinator.rs
```diff
@@ -121,13 +127,75 @@ async fn main() -> Result<()> {
...
+                                    position.trader
+                                ),
+                            }
+                        }
+                    }
+                    Err(e) => {
+                        tracing::error!("Failed to get positions. Error: {e:?}");
+                    }
+                }
+                tokio::time::sleep(Duration::from_secs(600)).await;
```
1010Tom: Any particular reason why you went with every 1 minutes?

coordinator/src/node.rs
```diff
@@ -164,26 +169,25 @@ impl Node {
...
+            position.quantity,
             leverage_long,
             leverage_short,
             position.direction,
         )?;
 
         tracing::debug!(
             "Settling position of {accept_settlement_amount} with {}",
-            trade_params.pubkey
+            position.trader.to_string()
```
1010Tom, line 190: Would be cool if we make it a habbit to move variables out of the message. This will allow us to filter the logs more easily.

coordinator/src/node.rs
```diff
@@ -53,6 +53,21 @@ pub struct Node {
 }
 
 impl Node {
+    /// Returns true or false, whether we can find an useable channel with the provider trader.
```
1010Tom, line 54: typo: 
```suggestion
     /// Returns true or false, whether we can find an useable channel with the provided trader.
```

1010Tom (approved):
Very nice. If it works, let's go

coordinator/src/bin/coordinator.rs
holzeis: well maybe we need to revisit that. We don't want that interval to be too small as not to interfere with an ongoing closure nor too big to risk not reaching the user when she is online.

holzeis:
bors retry

PR #431, opened 2023-04-16
Add welcome screen

klochowicz:
image

TODO
- [x] sending the registration info into the coordinator
- [x] storing registration info in coordinator

### DIFF
--- a/coordinator/migrations/2023-04-17-185626_create_users/up.sql
+++ b/coordinator/migrations/2023-04-17-185626_create_users/up.sql
@@ -1,0 +1,6 @@
+CREATE TABLE "users" (
+    id SERIAL PRIMARY KEY NOT NULL,
+    pubkey TEXT NOT NULL,
+    email TEXT NOT NULL,
+    nostr TEXT NOT NULL
+);
--- a/coordinator/src/db/user.rs
+++ b/coordinator/src/db/user.rs
@@ -1,0 +1,38 @@
+use coordinator_commons::RegisterParams;
+use diesel::prelude::*;
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::schema::users;
+
+#[derive(Insertable, Queryable, Identifiable, Debug, Clone, Serialize, Deserialize)]
+#[diesel(primary_key(id))]
+pub struct User {
+    #[diesel(deserialize_as = i32)]
+    pub id: Option<i32>,
+    pub pubkey: String,
+    pub email: String,
+    pub nostr: String,
+}
+
+impl From<RegisterParams> for User {
+    fn from(value: RegisterParams) -> Self {
+        User {
+            id: None,
+            pubkey: value.pubkey.to_string(),
+            email: value.email.unwrap_or("".to_owned()),
+            nostr: value.nostr.unwrap_or("".to_owned()),
+        }
+    }
+}
+
+pub fn all(conn: &mut PgConnection) -> QueryResult<Vec<User>> {
+    users::dsl::users.load(conn)
+}
+
+pub fn insert(conn: &mut PgConnection, user: User) -> QueryResult<User> {
+    let user: User = diesel::insert_into(users::table)
+        .values(&user)
+        .get_result(conn)?;
+    Ok(user)
+}
--- a/mobile/lib/features/welcome/welcome_screen.dart
+++ b/mobile/lib/features/welcome/welcome_screen.dart
@@ -1,0 +1,91 @@
+import 'package:f_logs/model/flog/flog.dart';
+import 'package:flutter/material.dart';
+import 'package:get_10101/features/wallet/wallet_screen.dart';
+import 'package:get_10101/util/preferences.dart';
+import 'package:go_router/go_router.dart';
+import 'package:get_10101/ffi.dart';
+
+class WelcomeScreen extends StatefulWidget {
+  static const route = "/welcome";
+  static const label = "Welcome";
+
+  const WelcomeScreen({Key? key}) : super(key: key);
+
+  @override
+  State<WelcomeScreen> createState() => _WelcomeScreenState();
+}
+
+class _WelcomeScreenState extends State<WelcomeScreen> {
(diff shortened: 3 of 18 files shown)


mobile/lib/main.dart
```diff
@@ -83,92 +85,116 @@ class TenTenOneApp extends StatefulWidget {
...
-                  return const SendScreen();
-                },
-              ),
-              GoRoute(
-                path: SeedScreen.subRouteName,
-                // Use root navigator so the screen overlays the application shell
-                parentNavigatorKey: _rootNavigatorKey,
+      navigatorKey: _rootNavigatorKey,
+      initialLocation: WelcomeScreen.route,
+      routes: <RouteBase>[
```
klochowicz, line 93: I have no idea why suddenly it formatted with 4 spaces instead of 2.... @bonomat is this the problem you had too recently?

mobile/lib/features/welcome/welcome_screen.dart
```diff
@@ -0,0 +1,96 @@
...
+              SizedBox(height: 10),
+              Text("Please be patient with us as we work out the rough edges."),
+              Text("Any feedback is welcome!"),
+              SizedBox(height: 10),
+              Text("Please enter your email address to continue:"),
+            ],
+          ),
+          const SizedBox(height: 20),
+          TextFormField(
+            keyboardType: TextInputType.emailAddress,
```
bonomat, line 54: No NPub support?

mobile/lib/main.dart
bonomat: No, I haven't had this problem yet 😬

mobile/lib/features/welcome/welcome_screen.dart
```diff
@@ -0,0 +1,96 @@
...
+      ),
+    );
+  }
+
+  @override
+  void initState() {
+    super.initState();
+
+    Preferences.instance.getEmailAddress().then((value) => setState(() {
+          FLog.info(text: "retrieved stored email from the preferences: $_email.");
```
bonomat, line 92: I think you want to print `value` here as `_email` is only set after the log message. 
Or move the log message down.

mobile/lib/main.dart
```diff
@@ -83,92 +85,116 @@ class TenTenOneApp extends StatefulWidget {
 }
 
 class _TenTenOneAppState extends State<TenTenOneApp> {
+  bool showOnboarding = false;
```
bonomat, line 88: 🚧
This field doesn't seem to be use anywhere?

mobile/native/src/api.rs
```diff
@@ -217,3 +217,8 @@ pub fn get_seed_phrase() -> Result<SyncReturn<Vec<String>>> {
     let seed_phrase = ln_dlc::get_seed_phrase()?;
     Ok(SyncReturn(seed_phrase))
 }
+
+pub fn register_beta(email: String) -> Result<()> {
+    tracing::debug!("Registering beta user: {}", email);
```
bonomat, line 222: Nothing is happening in here?

bonomat (changes requested):
Screenshot would have been nice :) 

> SharedPreferences package doesn't seem to be able to retrieve the email at
    startup (bug?)

On my simulator it looks like it is able to read from preferences just fine.

🚧 Am I missing something or is the email only stored on the user's phone? That won't be very useful for us. 
// edit // looks like you wanted to only post the flutter part first. Let's merge this all together please.

Here is the screenshot for others to see: 

image

mobile/lib/features/welcome/welcome_screen.dart
klochowicz: > I think you want to print `value` here as `_email` is only set after the log message.
> Or move the log message down.

omg, thank you. nothing was wrong, I just read the wrong field. 🙈

mobile/native/src/api.rs
klochowicz: yup, the goal was to put the frontend part, till the API is this PR (as it was described). 
could have left a TODO to make it more clear in code though.

klochowicz:
> Screenshot would have been nice :)
> 
> > SharedPreferences package doesn't seem to be able to retrieve the email at
> > startup (bug?)
> 
> On my simulator it looks like it is able to read from preferences just fine.
> 
> 🚧 Am I missing something or is the email only stored on the user's phone? That won't be very useful for us. // edit // looks like you wanted to only post the flutter part first. Let's merge this all together please.
> 
> Here is the screenshot for others to see:
> 
> image

I have updated review to include the backend in the same PR. Hidden this conversation as the screenshot is outdated (will keep this in mind to always send a screenshot!)

coordinator/src/db/user.rs
```diff
@@ -0,0 +1,38 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::schema::users;
+
+#[derive(Insertable, Queryable, Identifiable, Debug, Clone, Serialize, Deserialize)]
+#[diesel(primary_key(id))]
+pub struct User {
+    #[diesel(deserialize_as = i32)]
+    pub id: Option<i32>,
```
bonomat, line 12: 👍
Nice that this works . I'll need to remember this :)

bonomat (approved):
Thanks for this. 
When do you plan to update the frontend to also support nostr pubkeys?

I think we can roll with this already though :)

coordinator/migrations/2023-04-17-185626_create_users/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "users" (
+    id SERIAL PRIMARY KEY NOT NULL,
+    pubkey TEXT NOT NULL,
```
bonomat, line 3: We potentially could have gone with making the pubkey the id, but I guess it doesn't matter that much. :)

klochowicz:
> Thanks for this. When do you plan to update the frontend to also support nostr pubkeys?
> 
> I think we can roll with this already though :)

working on it now, just didn't want to block the release by fiddling with Flutter!

PR #436, opened 2023-04-17
Feat/add reverse proxy

holzeis:
see also https://github.com/get10101/cloud-conf/pull/177

ssl-offloading will come with a follow up PR.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -1,0 +1,33 @@
+worker_processes auto;
+
+events {
+    worker_connections 1024;
+}
+
+http {
+
+    server {
+        listen 80;
+        server_name api.test.10101.finance 35.189.57.114 api.10101.finance 34.30.20.202;
+
+        # reject any attempt to access the admin endpoint from the outside.
+        location /api/admin {
+            deny all;
+            return 403;
+        }
+
+        # Reverse proxy any other endpoint to the ordernator
+        location / {
+            proxy_pass http://10.5.0.100:8000;
+            proxy_http_version  1.1;
+            proxy_cache_bypass  $http_upgrade;
+
+            proxy_set_header Host              $host;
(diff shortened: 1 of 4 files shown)


nginx/nginx.conf
```diff
@@ -0,0 +1,33 @@
...
+
+            proxy_set_header Host              $host;
+            proxy_set_header X-Real-IP         $remote_addr;
+            proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_set_header X-Forwarded-Host  $host;
+
+        }
+    }
+}
```
bonomat, line 33: nit: new line

bonomat (approved):
Good idea with moving sensitive endpoints in a separate file!

.github/workflows/build-release-binaries.yml
```diff
@@ -25,9 +25,11 @@ jobs:
     uses: ./.github/workflows/ios.yml
     secrets: inherit
     with:
-      electrs_endpoint: tcp://blockstream.info:110
-      coordinator_p2p_endpoint: 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156@34.30.20.202:9045
-      network: mainnet
+      ELECTRS_ENDPOINT: tcp://blockstream.info:110
+      COORDINATOR_PK: 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156
+      COORDINATOR_HOST: api.10101.finance
```
holzeis, line 30: not sure if this will work as we are also using the host for announcing our node. We might have to go with the actual ip address here.

holzeis: I ended up adding the ip addresses to the `nginx.conf` as well, so that we have to change a little as possible.

PR #437, opened 2023-04-17
Do not call `peer_manager.process_events` manually

luckysori:
I left this note a while back and running the `ln-dlc-node` tests now confirms that the snippet is not needed.


holzeis (approved):
LGTM 👍

bonomat (approved):
nice

Issue #438, opened 2023-04-17
Fail to update position after DLC is closed


luckysori:
I am tripping on this error _every time_ I try to close a position. The DLC is removed correctly, but the `position` cannot be updated.

```
2023-04-17T07:23:23.411906567Z DEBUG native::ln_dlc::node: Sending DLC channel message to=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9 msg=CloseFinalize
2023-04-17T07:23:23.412159894Z  WARN native::ln_dlc::node: Could not find a filling position in the database. This might be, because the coordinator closed an expired position. Error: There is no order in state filling in the database
2023-04-17T07:23:23.412715278Z ERROR native::ln_dlc::node: Critical Error! We have a DLC but were unable to set the order to filled. Error: Exactly one positions should be open when trying to close a position
```

---

This is on https://github.com/get10101/10101/commit/13bcef6d320f35db0c10f2c2f4778edfca2c30d6 on Linux.

Issue #410, opened 2023-04-13
"Clear pending DLC" cleanup button in app


da-kami:
This will help the user to overcome "stuck" states where trading in the app becomes unusable

holzeis:
I am adding the MVP flag again as I think this will be super important especially in the first weeks.

luckysori:
Should not do this.

PR #439, opened 2023-04-17
Some tweaks to our DLC message handling

luckysori:
I noticed this when I was looking into #410.

Sorry about the last two patches effectively being fixups of the first one. I chose to leave it like this because rebasing was annoying.

I'm pretty sure it works, but I wasn't able to test it in full because of #438.

### DIFF
--- a/mobile/native/src/ln_dlc/node.rs
+++ b/mobile/native/src/ln_dlc/node.rs
@@ -2,8 +2,10 @@ use crate::db;
 use crate::trade::order;
 use crate::trade::position;
 use anyhow::anyhow;
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
+use bdk::bitcoin::secp256k1::PublicKey;
 use bdk::TransactionDetails;
 use dlc_messages::sub_channel::SubChannelFinalize;
 use dlc_messages::Message;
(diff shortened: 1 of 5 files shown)


mobile/native/src/ln_dlc/node.rs
```diff
@@ -82,151 +85,156 @@ impl Node {
...
+                    // TODO: We should probably verify that: (1) the counterparty is the
+                    // coordinator and (2) the DLC channel close offer is expected and correct.
+                    self.inner
+                        .accept_dlc_channel_collaborative_settlement(&channel_id)
+                        .with_context(|| {
+                            format!(
+                                "Failed to accept DLC channel close offer for channel {}",
+                                hex::encode(channel_id)
+                            )
+                        })?;
```
1010Tom: Shouldn't this be a match block?

mobile/native/src/ln_dlc/node.rs
```diff
@@ -174,28 +173,32 @@ impl Node {
 
                     match reply_msg {
                         SubChannelMessage::Finalize(SubChannelFinalize { channel_id, .. }) => {
-                            let contract = self
+                            let contracts = self
                                 .inner
                                 .dlc_manager
                                 .get_store()
-                                .get_contract(&channel_id)
```
klochowicz: too bad these were only type aliases, because such a API misuse seems to be quite common :(

klochowicz (approved):
LGTM!

mobile/native/src/ln_dlc/node.rs
luckysori: Definitely an insidious source of bugs. We could PRs on the relevant repos to introduce new types. I reckon `rust-dlc` would accept it.

mobile/native/src/ln_dlc/node.rs
luckysori: For reference: the `if let`s were already there, I just moved them around.

I do agree that a `match` is a bit clearer as one can more easily see that the cases are independent. On the other hand, it does mean that we have to explicitly handle the catch-all pattern explicitly, which is a bit noisy.

luckysori: Improved with https://github.com/get10101/10101/pull/439/commits/412605a2af653877ac1a64510aeae1d8a95c2a4e.

PR #442, opened 2023-04-17
Ignore testnet testflight for releases

1010Tom:
When triggering a release, we get a merge commit in main which will trigger the release to production and testnet. This can fail, because each build gets a unique number, and when building at the same time, they might get the same which makes one of the two fail. To overcome this we simply ignore certain commits with the message starting with


holzeis (approved):
Nice 👍

klochowicz (approved):

PR #443, opened 2023-04-17
fix: signing a message using our nodes key

bonomat:
this seems to be the correct function to use.

It works, i.e. you can sign, but I did not manage to log into amboss or lightningswap yet :)

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -4,7 +4,6 @@ use axum::extract::Path;
 use axum::extract::Query;
 use axum::extract::State;
 use axum::Json;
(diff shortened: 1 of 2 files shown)


bonomat:
@holzeis : tested locally with correct key and it works

bonomat:
- https://amboss.space/node/022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156
- https://lightningnetwork.plus/nodes/022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156

holzeis (approved):
Awesome 👍

PR #444, opened 2023-04-17
fix max sats/vbyte for predictability

bonomat:
We want to be able to predict how much money we use for tx fee, hence, we set a upper limit of 20 sats


holzeis (approved):
👍

Issue #383, opened 2023-04-10
Cap the channel size


klochowicz:
- the limit should be shown on the invoice view.
- the max amount of sats that can be received applies whether a channel already exist or not. (As we don't support extending channels yet)
- if a payment is received with an amount exceeding the channel cap, the htlc should transition to failed. (note, this is already the case if a channel already exists)

holzeis:
We could show a similar limit on the send payment screen.

da-kami:
Notes:

- after #397 we already have the `200k` in the `TradeService` but since this value concerns the wallet as well it should be moved into a separate service.
- I would add Form validation to the `Receive` screen to validate the value
- In the trade screen, when the user runs low on capacity we recommend to balance the channel 50/50 - I would probably do the same in the Receive screen for now to keep it simple.

PR #446, opened 2023-04-17
Feat/cap channel capacity

### LINKED ISSUE #402: Fix the flutter exception that is thrown in the trade bottom sheets when editing values
After making the `TradeBottomSheetTab` stateful in #397  we keep seeing this exception in the logs (when modifying the `margin`; we also see it when modifying the `quantity`, but slightly different because it is always triggered by the other text field):

```
══╡ EXCEPTION CAUGHT BY FOUNDATION LIBRARY ╞════════════════════════════════════════════════════════
The following assertion was thrown while dispatching notifications for TextEditingController:
setState() or markNeedsBuild() called during build.
This Form widget cannot be marked as needing to build because the framework is already in the
process of building widgets. A widget can be marked as needing to be built during the build phase
only if one of its ancestors is currently building. This exception is allowed because the framework
builds parent widgets before children, which means a dirty descendant will always be built.
Otherwise, the framework might not visit this widget during this build phase.
The widget on which setState() or markNeedsBuild() was called was:
  Form-[LabeledGlobalKey<FormState>#3fa9e]
The widget which was currently being built when the offending call was made was:
  DoubleTextInputFormField

When the exception was thrown, this was the stack:

...
```
When using `TextEdigintController`s and a `Form` it is recommended to have a `Stateful` widget.

Part of the problem comes from the complexity of having two text fields that update each other - sounds easy, but is actually quite complex. But even without editing both fields we can run into the exception of triggering multiple builds. 

I added `print` statements to see how often certain code pathes (onChange, and build) are triggered and they are triggered too often - but I just don't understand how. In many cases I did not even interact with the UI and somehow it triggered a rebuild. I am not sure what in our setup triggers this, but something is and I don't have the time to look into this now.

---

Daniel's conclusion:

The problem must be related to using the providor's `Selector` in combination with the `FormTextFields` and validation. I stopped investigating because it's too time intense. 

Recommendation:

1. Take a step back
2. Keep the requirements for the trade bottom sheet screens (how the text fields should behave, ...)
3. Re-designing the implementation from the ground up keeping the requirements in mind

da-kami:
resolves https://github.com/get10101/10101/issues/383

Note: I have not added form validation for the `Send` screen, am planning to do that on top of this work.

---

Sidenote: I tried to fix #402 on the way - calling `setState` in the `validator` function is *not* good - but removing that part still did not help with resolving the problem, and brought other problems, so I decided to keep the logic as if for now to not fall into a rabbit hole.

---

Here's a GIF that shows validation for the receive screen:

### DIFF
--- a/mobile/lib/features/wallet/create_invoice_screen.dart
+++ b/mobile/lib/features/wallet/create_invoice_screen.dart
@@ -9,8 +9,7 @@ import 'package:get_10101/features/wallet/wallet_change_notifier.dart';
 import 'package:get_10101/features/wallet/wallet_screen.dart';
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
-
-import '../../common/modal_bottom_sheet_info.dart';
+import 'package:get_10101/common/modal_bottom_sheet_info.dart';
 import 'application/wallet_service.dart';
 
 class CreateInvoiceScreen extends StatefulWidget {
@@ -42,9 +41,10 @@ class _CreateInvoiceScreenState extends State<CreateInvoiceScreen> {
   @override
   Widget build(BuildContext context) {
     int channelCapacity = channelConstraintsService.getLightningChannelCapacity();
-    int totalMaxAmount = (channelCapacity / 2).floor();
+    int usableChannelCapacity = channelConstraintsService.getUsableChannelCapacity();
     int balance = context.watch<WalletChangeNotifier>().walletInfo.balances.lightning.sats;
-    int maxAmount = max(totalMaxAmount - balance, 0);
+    // it can go below 0 if the user has an unbalanced channel
+    int maxAmount = max(usableChannelCapacity - balance, 0);
 
     int minAmount = channelConstraintsService.getChannelReserve() +
         channelConstraintsService.getFeeReserve() +
@@ -90,7 +90,7 @@ class _CreateInvoiceScreenState extends State<CreateInvoiceScreen> {
                           try {
                             int amount = int.parse(value);
 
-                            if (balance > totalMaxAmount) {
+                            if (balance > usableChannelCapacity) {
                               return "Maximum beta balance exceeded";
                             }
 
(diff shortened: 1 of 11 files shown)


mobile/lib/features/wallet/create_invoice_screen.dart
```diff
@@ -1,23 +1,23 @@
...
+import 'package:get_10101/common/application/channel_constraints_service.dart';
 import 'package:get_10101/common/domain/model.dart';
-import 'package:get_10101/features/wallet/domain/wallet_info.dart';
 import 'package:get_10101/features/wallet/share_invoice_screen.dart';
 import 'package:get_10101/features/wallet/wallet_change_notifier.dart';
 import 'package:get_10101/features/wallet/wallet_screen.dart';
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 
+import '../../common/modal_bottom_sheet_info.dart';
```
klochowicz, line 13: nit: use the `package` import method

mobile/lib/features/wallet/create_invoice_screen.dart
```diff
@@ -27,78 +27,140 @@ class _CreateInvoiceScreenState extends State<CreateInvoiceScreen> {
...
+    _amountController.dispose();
+    super.dispose();
+  }
+
   @override
   Widget build(BuildContext context) {
-    WalletInfo info = context.watch<WalletChangeNotifier>().walletInfo;
-    log("Refresh receive screen: ${info.balances.onChain}");
+    int channelCapacity = channelConstraintsService.getLightningChannelCapacity();
+    int totalMaxAmount = (channelCapacity / 2).floor();
```
klochowicz, line 45:
```suggestion
    int totalMaxAmount = channelConstraintsService.getTotalMaxAmount();
```

I'd suggest instead of doing math in widget, abstract it away into the same channel contstraint service

mobile/lib/features/wallet/create_invoice_screen.dart
```diff
@@ -27,78 +27,140 @@ class _CreateInvoiceScreenState extends State<CreateInvoiceScreen> {
...
+    int balance = context.watch<WalletChangeNotifier>().walletInfo.balances.lightning.sats;
+    int maxAmount = max(totalMaxAmount - balance, 0);
```
klochowicz, line 47: can this actually be a negative number??? if this happened, then lower in the file we would display `0` in few places, where this would really confuse the user. e.g. "you can receive up to 0 sats".

(if this is a shortcut, then it's fine, but please add a comment about it ;D )

klochowicz (approved):

holzeis (approved):
LGTM 👍

holzeis:
resolves #383

mobile/lib/features/wallet/create_invoice_screen.dart
da-kami: You can receive up to `0` sats is correct in this context. I understand it can be confusing, but I would come back to it once we have more user feedback. 
This goes towards "explaining the channel boundaries to the user" - which is not easy.

klochowicz: can we please add a comment in the code to reflect that? e.g.
`// it can go below 0 if the user has an unbalanced channel`

thanks 🙏

klochowicz:
bors retry

holzeis:
bors retry

PR #449, opened 2023-04-17
chore: reset a failed position to open

### LINKED ISSUE #435: Set position state to open if close order failed
When we close a position and it gets to failed, the position remains in the state Closing. Resulting into the position being stuck in that state. We should as well reset the state to Open if the closing failed.

bonomat:
in case we are closing a position but fail at matching, we need to set the position to open again so that we can try again later.

resolves #435


holzeis (approved):
👍

PR #451, opened 2023-04-17
Send screen overhaul

da-kami:
- Decode invoice
- Form validation

Note: we are not displaying the decoded values yet.

TODO:

- [x] Display decoded values once available

---

### DIFF
--- a/mobile/lib/features/wallet/application/wallet_service.dart
+++ b/mobile/lib/features/wallet/application/wallet_service.dart
@@ -1,6 +1,7 @@
 import 'package:f_logs/f_logs.dart';
 import 'package:get_10101/common/domain/model.dart';
 import 'package:get_10101/ffi.dart' as rust;
+import 'package:get_10101/features/wallet/domain/lightning_invoice.dart';
 
 class WalletService {
   const WalletService();
@@ -37,6 +38,22 @@ class WalletService {
       FLog.error(text: "Error: $error", exception: error);
       return null;
     }
+  }
+
+  Future<LightningInvoice?> decodeInvoice(String invoice) async {
+    try {
+      FLog.debug(text: "Decoding invoice $invoice");
+      rust.LightningInvoice lightningInvoice = await rust.api.decodeInvoice(invoice: invoice);
+      FLog.debug(text: "Successfully decoded invoice.");
+      return LightningInvoice.fromApi(lightningInvoice);
+    } catch (error) {
+      FLog.debug(text: "Failed to decode invoice: $error", exception: error);
+      return null;
+    }
+  }
+
+  Future<void> payInvoice(String invoice) async {
+    await rust.api.sendPayment(invoice: invoice);
   }
 
   String getNewAddress() {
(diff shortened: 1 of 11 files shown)


mobile/lib/features/wallet/application/wallet_service.dart
```diff
@@ -39,6 +40,17 @@ class WalletService {
     }
   }
 
+  Future<LightningInvoice?> decodeInvoice(String invoice) async {
+    try {
+      LightningInvoice lightningInvoice = await rust.api.decodeInvoice(invoice: invoice);
+      FLog.info(text: "Successfully decoded invoice.");
```
bonomat, line 46: Nit: I guess debug would be enough for this message.

mobile/native/src/api.rs
```diff
@@ -217,3 +224,38 @@ pub fn get_seed_phrase() -> Result<SyncReturn<Vec<String>>> {
...
+
+pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
+    anyhow::ensure!(!invoice.is_empty(), "received empty invoice");
+    let invoice = &Invoice::from_str(&invoice).context("Could not parse invoice string")?;
+    let description = match invoice.description() {
+        InvoiceDescription::Direct(direct) => direct.to_string(),
+        InvoiceDescription::Hash(_) => "".to_string(),
+    };
+    // can't use as_millis as the frb does not support u128
+    let timestamp = invoice
```
bonomat, line 244: What's the timestamp needed for?

bonomat (approved):
Looks good :)

mobile/native/src/api.rs
da-kami: I just blindly copied this from the POC - will have another look :)

mobile/lib/features/wallet/send_screen.dart
```diff
@@ -14,33 +21,118 @@ class SendScreen extends StatefulWidget {
...
+
+                    return null;
+                  },
+                ),
+              ),
+              Center(
+                  child: Padding(
+                padding: const EdgeInsets.only(bottom: 10.0, left: 32.0, right: 32.0),
+                child: Text(
+                  "During the beta a minimum of $channelReserve sats have to remain in the wallet."
```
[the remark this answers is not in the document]
da-kami: I like the suggestions - we should do this for all those popups I suppose, but it's quite some work and I would not give it prio

mobile/native/src/api.rs
da-kami: In the poc we just displayed it - no other use. 
As far as I understand it's the expiry timestamp of the invoice. 
I would keep it for completion.

bonomat: > As far as I understand it's the expiry timestamp of the invoice.

I'm confused :) The expiry is 4 lines below.

holzeis: The timestamp is when the invoice has been created and the expiry defines the number of seconds until it is invalid.

The following example should make it clear. 
```
"timestamp": "1496314658",
"expiry": "3600",
```

da-kami:
Note: Fixed up the expiry after taking the gif record :)

bonomat (approved):
LGTM

PR #455, opened 2023-04-18
Add LN close buttons to app's settings page

luckysori:
Related to #410.

This is extremely dumb at the moment, but it worked nicely for testing as it allowed me to keep going without having to wipe any data.

### DIFF
--- a/mobile/lib/common/settings_screen.dart
+++ b/mobile/lib/common/settings_screen.dart
@@ -64,6 +64,16 @@ class _SettingsScreenState extends State<SettingsScreen> {
                   ? FontWeight.bold
                   : FontWeight.normal),
         ),
+        ElevatedButton(
+            onPressed: () {
+              rust.api.closeChannel();
+            },
+            child: const Text("Close channel")),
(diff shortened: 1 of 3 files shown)


klochowicz (approved):
I like it. Does the job 👍

bonomat:
Let's put this on hold for now and force users to reach out to us if they get stuck. This way we have a chance to learn more then if they try to forc close and we don't hear anything from them.

klochowicz:
as @luckysori pointed out in the PR, making testing easier is big plus. 

@bonomat @holzeis would you guys be ok if we only showed these buttons on regtest?

bonomat:
> @bonomat @holzeis would you guys be ok if we only showed these buttons on regtest for now?

Yes, we could do that .

klochowicz:
@bonomat @da-kami can I merge this? I added a flag that only shows it on regtests builds for now, and at least it will not need rebasing again...

klochowicz:
going ahead, as it does not affect production in any way.

bors r+

PR #457, opened 2023-04-18
fix: if condition release latest main

1010Tom:
The problem seems to have been that we mixed syntax: `{}` and without brackets.


holzeis (approved):
👍

PR #464, opened 2023-04-19
Ensure that the counterparty is able to fulfil

da-kami:
Make sure that the balance of the counterparty can fulfil the calculated counterparty margin.


holzeis (approved):
looks good to me. lets give it a go, if the manual testing verified that its working

PR #468, opened 2023-04-19
feat: show build number and node id in settings

### LINKED ISSUE #466: Add more info into settings for debug purposes
We want to see the build number and the node id in the settings in the app so that we can debug better.

bonomat:
resolves #466

image


mobile/lib/common/settings_screen.dart
```diff
@@ -10,6 +10,8 @@ import 'package:intl/intl.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:package_info_plus/package_info_plus.dart';
+import '../ffi.dart' as rust;
```
holzeis: nit: don't use relative imports

holzeis (approved):

PR #469, opened 2023-04-19
chore: create new order then delete old

bonomat:
this way we should always have an order in the orderbook and don't run into no-match found errors


luckysori (approved):
LGTM!

PR #471, opened 2023-04-19
chore: Export 10101 app logs with '*.log' extension

klochowicz:
"*.logs" is non-standard, making it harder to open it in IDE.

image

### DIFF
--- a/mobile/lib/common/settings_screen.dart
+++ b/mobile/lib/common/settings_screen.dart
@@ -164,8 +164,8 @@ class _SettingsScreenState extends State<SettingsScreen> {
               var logsAsString = await file.readAsString();
               final List<int> bytes = utf8.encode(logsAsString);
               final Directory tempDir = await getTemporaryDirectory();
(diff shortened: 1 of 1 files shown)


mobile/lib/common/settings_screen.dart
```diff
@@ -165,7 +165,7 @@ class _SettingsScreenState extends State<SettingsScreen> {
               final List<int> bytes = utf8.encode(logsAsString);
               final Directory tempDir = await getTemporaryDirectory();
               String now = DateFormat('yyyy-MM-dd_HH:mm:ss').format(DateTime.now());
```
bonomat, line 167: Since we are on it :) 
```suggestion
String now = DateFormat('yyyy-MM-dd_HHmmss').format(DateTime.now());
```

`:` are weirdly expressed as `/` which is super annoying.

bonomat (approved):
Thanks, this annoyed me as well :D

holzeis (approved):

klochowicz:
bors retry

PR #473, opened 2023-04-20
Get rid of all the bad `unwrap`s, and keep the good ones

luckysori:
I was still a bit worried about the number of `unwrap`s that were lying around unattended in the codebase, so I decided to ask Clippy for help.

Now this is not definitive as we might still have a few dubious `expect`s.


holzeis (approved):
LGTM 👍

luckysori:
Had no problems locally, the failure is just another instance of #355. Merging.

Issue #470, opened 2023-04-19
Failed to close position after app restart: Invalid state: Expected Signed state but got Accepted


holzeis:
When restarting the app with an open position the position can't be closed anymore. see also error message below.

### How to reproduce:

1. startup test environment with a funded app.
2. Open a position and wait for the position to get filled.
3. Kill the app (do not just put the app into the background)
4. Open app again
5. Close the position.

```
 | 2023-04-19 16:23:18.620 | LogLevel.ERROR | Failed to process incoming message: Invalid state: Expected Signed state but got Accepted(AcceptedSubChannel { offer_per_split_point: PublicKey(f24e13345df34251200d392b0f424090282b598510d5098a98892e9a2e075304c40f3783714cdbc28172368f7305db615ead3e0cc61398925ae38c155abbd6f9), accept_per_split_point: PublicKey(f366a0de4617eb3e987884b3af58b90295a239b280c61b70e56d120c99b78e9f63e440e6c25fd0798dc4a5d33ac435edba5af8eb8a092e8d1fff36a0b414ea17), accept_split_adaptor_signature: EcdsaAdaptorSignature(03db0f25b2cb518e5036d7dd8d7856d025dd2f3ea4a0bdecc89556cd62288e0c890304ac642e9ea6681d993d4c1c4a33f0f35117076cc6abd3a5abcbb676f39fab28f822dae21c91f4c15c3410146398603a0cc64d6a1b5c063498acb7a324181a09346dc85a278f0d2c494f740b51b2434671052580c0334d56a0fb96285c3940d4637095386cd3255e95bce21293edd4ad98bfba84f300d09931eefeadeaa413a2), split_tx: SplitTx { transaction: Transaction { version: 2, lock_time: PackedLockTime(0), input: [TxIn { previous_output: OutPoint { tx<…>
```

klochowicz:
happened to me too.

holzeis:
I presume that the issue is caused by a state not being written to disk after the dlc was signed. Restarting will read an old state from disk leading to that error when trying to close the dlc.

PR #474, opened 2023-04-21
Update `rust-dlc` to fix restart bug

luckysori:
Fixes https://github.com/get10101/10101/issues/470.


luckysori:
Not much to review and it was already tested by @holzeis. Merging.

PR #481, opened 2023-04-21
feat: Delay payment for the user to open 10101

### LINKED ISSUE #478: Wait for target peer to become online before completing HTLCIntercepted event
Currently the app needs to be running at the time of sending the payment. This is an unlikely scenarios for our users, as most of them will onboard from other mobile lightning wallet on the same phone.

We should wait a couple of seconds for the app to come online again to allow the user to switch back to 10101 to receive the funds.

holzeis:
In cases the user will onboard to 10101 using another mobile app. 10101 will be closed at the time the payment is sent. Without delaying payment processing before the user opens 10101 the payment will always fail.

This is a quick fix to solve that issue, by simply waiting for 10 seconds for the 10101 receiver to come online.

Eventually we should implement some logic that is keeping 10101 online in the background for some time to cater for that scenario more nicely.

Also note, that this fix assumes that the sending party will remain online while the payment is sent.

resolves #478

### DIFF
--- a/crates/ln-dlc-node/src/ln/mod.rs
+++ b/crates/ln-dlc-node/src/ln/mod.rs
@@ -22,3 +22,10 @@ pub(crate) use logger::TracingLogger;
 ///
 /// This constant only applies to the coordinator.
 pub(crate) const JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT: u64 = 200_000;
+
+/// When handling the [`Event::HTLCIntercepted`], the user might not be online right away. This
+/// could be because she is funding the wallet through another wallet. In order to give the user
+/// some time to open 10101 again we wait for a bit to see if we can establish a connection.
+///
+/// This constant specifies the amount of time (in seconds) we are willing to delay a payment.
+pub(crate) const HTLC_INTERCEPTED_CONNECTION_TIMEOUT: u64 = 30;
--- a/crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
+++ b/crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
@@ -3,11 +3,13 @@ use crate::node::Node;
 use crate::node::PaymentMap;
 use crate::node::LIQUIDITY_ROUTING_FEE_MILLIONTHS;
 use crate::tests::init_tracing;
+use crate::tests::just_in_time_channel::TestPath;
 use crate::tests::min_outbound_liquidity_channel_creator;
 use anyhow::Context;
 use anyhow::Result;
 use bitcoin::Amount;
 use rust_decimal::Decimal;
+use std::time::Duration;
 
 #[tokio::test]
 #[ignore]
@@ -64,6 +66,7 @@ async fn just_in_time_channel() {
     let invoice_amount = 1_000;
 
     send_interceptable_payment(
+        TestPath::OnlineFunding,
         &payer,
         &payee,
         &coordinator,
@@ -75,6 +78,7 @@ async fn just_in_time_channel() {
 }
 
 pub(crate) async fn send_interceptable_payment(
+    test_path: TestPath,
     payer: &Node<PaymentMap>,
     payee: &Node<PaymentMap>,
     coordinator: &Node<PaymentMap>,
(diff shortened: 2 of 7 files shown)


crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
```diff
@@ -123,6 +127,17 @@ pub(crate) async fn send_interceptable_payment(
 
     payer.send_payment(&invoice)?;
 
+    if TestPath::MobileFunding == test_path {
+        // simulate the user switching from another app to 10101
+
+        // Note, hopefully Breez, Phoenix or any other non-custodial wallet is able to run in the
+        // background when sending a payment as otherwise a disconnect on their side would happen
+        // when switching to 10101 resulting into a failed payment.
+        // payer.disconnect(coordinator.info);
```
klochowicz, line 136: did you mean to disconnect the payer here? 

the line seems relevant, but it's commented out.

crates/ln-dlc-node/src/tests/just_in_time_channel/offline_receiver.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::node::Node;
+use crate::tests::init_tracing;
+use crate::tests::just_in_time_channel::create::send_interceptable_payment;
+use crate::tests::just_in_time_channel::TestPath;
+use crate::tests::min_outbound_liquidity_channel_creator;
+use bitcoin::Amount;
+
+#[tokio::test]
+#[ignore]
+async fn offline_receiver() {
```
klochowicz, line 11: could we come up with a better name for this test?
`able_to_receive_an_offline_payment`? or something? it would be immensely helpful to have a name that's obvious when it fails.

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
holzeis: I guess the comment above could be more clear. I did comment it that because it wouldn't work otherwise.

klochowicz: perhaps:
`// line below is commented out on purpose`

or actually remove the commented out line of code and just keep the comment?

crates/ln-dlc-node/src/ln/mod.rs
```diff
@@ -22,3 +22,10 @@ pub(crate) use logger::TracingLogger;
 ///
 /// This constant only applies to the coordinator.
 pub(crate) const JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT: u64 = 200_000;
+
+/// When handling the [`Event::HTLCIntercepted`], the user might not be online right away. This
+/// could be because she is funding the wallet through another wallet. In order to give the user
+/// some time to open 10101 again we wait for a bit to see if we can establish a connection.
+///
+/// This constant specifies the amount of time (in seconds) we are willing to delay a payment.
+pub(crate) const HTLC_INTERCEPTED_CONNECTION_TIMEOUT: u64 = 10;
```
klochowicz, line 31: I would go with 30 seconds, 10s is pretty tricky to make it on time when you're trying to open another app / website.

klochowicz (approved):
Happy to give this a try, if it buys us more time to solve things properly :)

crates/ln-dlc-node/src/ln/mod.rs
holzeis: I can increase it, but I am not sure if the other hops will simply timeout before us then.

crates/ln-dlc-node/src/tests/just_in_time_channel/offline_receiver.rs
holzeis: The nameing could really be better, but it's also not really an offline payment. I am going to merge it as is and we can change the name in a follow up PR.

holzeis:
slow running tests timed out on the ci. local run succeeded. merging by bypassing branch protections.

crates/ln-dlc-node/src/ln/mod.rs
bonomat: They won't timeout that fast. Afaik the HTLC will be added already to your node and the timeout is related to the CTLV timout, so actually quiet high.

PR #482, opened 2023-04-21
fix: Add sign tx input implementation

holzeis:
While debugging, I noticed that the `sign_tx_input` was not implemented. I copied over the implementation from the simple wallet.


klochowicz (approved):

holzeis:
bors retry

PR #489, opened 2023-04-23
feat: Add connect and is connected endpoints

### LINKED ISSUE #479: Coordinator: API to connect to peer by id@ip
It looks like people have troubles routing through us and their peers do not show up when doing
```
curl localhost:8000/api/admin/peers | jq .
```

This is suboptimal because users cannot route through us. 

I'd like to test if we can connect to other nodes from our side before trying to find out if there is a problem in our code regarding network loss. 

For that, I'd like to have a simple http endpoint which allows us to connect to a peer by it's `id` and `ip`. 

https://lightningnetwork.plus/nodes/02abd50ef667e05d26a1132eaac0521512605444e95bf12c474a114c7407994eb7

holzeis:
If the coordinator restarts it does not attempt to reconnect to previously connected peers. This is a quick fix to add an admin endpoint to connect to any given peer.

### Tested with maker and taker.

1. Open channel from taker to maker.
```
curl -v -X POST -H "Content-Type: application/json" -d '{"target": { "pubkey": "025c578b8e05cd203e2a486388ed4de4e480456923192a0f288ef96c863940ce6b", "address": "127.0.0.1:19045"}, "local_balance": 300000 }' http://localhost:8000/api/admin/channels
```
2. Check if peer is connected
```
# will list all connected peers.
curl -X GET http://localhost:8000/api/admin/peers

# returns true if given peer is connected or false otherwise.
curl -X GET http://localhost:8000/api/admin/is_connected/025c578b8e05cd203e2a486388ed4de4e480456923192a0f288ef96c863940ce6b
```
3. Stop the maker
4. Verify that coordinator is not connected to maker
```
# will return an empty list
curl -X GET http://localhost:8000/api/admin/peers

# returns false
curl -X GET http://localhost:8000/api/admin/is_connected/025c578b8e05cd203e2a486388ed4de4e480456923192a0f288ef96c863940ce6b
```
5. Start maker again. At this point the maker and coordinator remain disconnected as neither tries to reconnect to a peer they have a channel with. This should be fixed for public channels.
6. Connect coordinator to maker
```
curl -X POST -H "Content-Type: application/json" -d '{"pubkey": "025c578b8e05cd203e2a486388ed4de4e480456923192a0f288ef96c863940ce6b", "address": "127.0.0.1:19045"}' http://localhost:8000/api/admin/connect
```
7. Verify that maker and coordinator are connected (see step 4)

resolves #479

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -175,3 +175,24 @@ pub async fn sign_message(
 
     Ok(Json(signature))
 }
+
+pub async fn connect_to_peer(
+    State(state): State<Arc<AppState>>,
+    target: Json<NodeInfo>,
+) -> Result<(), AppError> {
+    let target = target.0;
+    state.node.inner.connect(target).await.map_err(|err| {
+        AppError::InternalServerError(format!("Could not connect to {target}. Error: {err}"))
+    })?;
+    Ok(())
+}
+
+pub async fn is_connected(
+    State(state): State<Arc<AppState>>,
+    Path(target_pubkey): Path<String>,
+) -> Result<Json<bool>, AppError> {
+    let target = target_pubkey.parse().map_err(|err| {
+        AppError::BadRequest(format!("Invalid public key {target_pubkey}. Error: {err}"))
+    })?;
+    Ok(Json(state.node.is_connected(&target)))
+}
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -26,7 +26,9 @@ use ln_dlc_node::node::NodeInfo;
 use orderbook_commons::OrderbookMsg;
 
 use crate::admin::close_channel;
+use crate::admin::connect_to_peer;
 use crate::admin::get_balance;
+use crate::admin::is_connected;
 use crate::admin::list_channels;
 use crate::admin::list_dlc_channels;
 use crate::admin::list_peers;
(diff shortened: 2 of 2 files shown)


coordinator/src/admin.rs
```diff
@@ -175,3 +175,24 @@ pub async fn sign_message(
...
     Ok(Json(signature))
 }
+
+pub async fn connect_to_peer(
+    State(state): State<Arc<AppState>>,
+    target: Json<NodeInfo>,
+) -> Result<(), AppError> {
+    let target = target.0;
+    state.node.inner.connect(target).await.map_err(|err| {
+        AppError::InternalServerError(format!("Could not connection to {target}. Error: {err}"))
```
bonomat, line 185: Nit: small typo : could not connect

coordinator/src/admin.rs
```diff
@@ -175,3 +175,24 @@ pub async fn sign_message(
...
+    })?;
+    Ok(())
+}
+
+pub async fn is_connected(
+    State(state): State<Arc<AppState>>,
+    Path(target_pubkey): Path<String>,
+) -> Result<Json<bool>, AppError> {
+    let target = target_pubkey.parse().map_err(|err| {
+        AppError::InternalServerError(format!("Invalid public key {target_pubkey}. Error: {err}"))
```
bonomat, line 195: Nit: since it's an invalid user Input it should be a clientSide error, e.g. 4xx something

bonomat (approved):
Very nice. Thanks for that .

PR #493, opened 2023-04-24
Keep `rust-dlc` dependency up-to-date

luckysori:
No notable functional changes other than ensuring that our fork of `rust-dlc` is as similar as possible to upstream.


holzeis (approved):
Nice 👍

Issue #480, opened 2023-04-21
Order failed in the coordinator


klochowicz:
Could not open a new position

```
| 2023-04-21 09:43:42.522 | LogLevel.INFO | Updated order state new_state: Filling,order_id: 0522710a-8842-492b-a665-cdaa35af9a85 
| 2023-04-21 09:43:42.523 | LogLevel.DEBUG | Publishing event OrderUpdateNotification(Order { id: 0522710a-8842-492b-a665-cdaa35af9a85, leverage: 5.0, quantity: 1.495325, contract_symbol: BtcUsd, direction: Long, order_type: Market, state: Filling { execution_price: 29906.5 }, creation_timestamp: 2023-04-21 13:43:41.0 +00:00:00 })  
| 2023-04-21 09:43:43.539 | LogLevel.ERROR | Failed to execute trade for order Some(0522710a-8842-492b-a665-cdaa35af9a85): TradeResponse: Could not post trade to coordinator: {"error":"Could not handle trade request: Storage error Nonsense bytes didn't map to the type they were interpreted as"}  
| 2023-04-21 09:43:43.541 | LogLevel.INFO | Updated order state new_state: Failed,order_id: 0522710a-8842-492b-a665-cdaa35af9a85 
```

Investigating the coordinator side, more details coming

luckysori:
Is this on `mainnet`?

The `mainnet` coordinator reports the same error when I try to list the DLC channels:

```
lucas@get10101-production ~ $ curl localhost:8000/api/admin/dlc_channels | jq .

{
  "error": "Failed to list DLC channels: Storage error Nonsense bytes didn't map to the type they were interpreted as"
}
```

klochowicz:
Had the same issue on regrest. Can reproduce locally with the db I copied before wiping coordinator.

luckysori:
There was at least [one breaking change in the serialisation of `SubChannel`s](https://github.com/p2pderivatives/rust-dlc/commit/adc7cef5650cbba56f869a275b7b0a273c414ef1#diff-98c1a2b50758c12662586ffc088a6113bf2e4635402d799164c78fdea0129628R175-R176) in the past few days and it's the likely cause of this error.

I think we:

- Created a DLC channel before the patch.
- Upgraded `rust-dlc` to a version that included the patch.
- Deployed a new version of the coordinator.
- Tried to load the DLC channel from the `sled` DB and encountered the error on deserialising the data.

---

[...]

PR #494, opened 2023-04-24
Use temporary version of `rust-dlc` to circumvent bug

luckysori:
This is an attempt at dealing with the consequences of https://github.com/get10101/10101/issues/480.

I don't think `rust-dlc#ln-dlc-channels` intends to be backwards-compatible at this stage, so this is not something that we want to upstream.

---

The idea here is to use this branch to start up the mainnet coordinator and close a few DLC channels that are outdated. We can use the logs to identify which ones are old, or simply close all of them.

This is not guaranteed to "fix" #480, but it's the leading theory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -825,7 +825,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
-source = "git+https://github.com/get10101/rust-dlc?rev=87c3f8b69884dbdfa4e0404de0bb6ec5aea3ff79#87c3f8b69884dbdfa4e0404de0bb6ec5aea3ff79"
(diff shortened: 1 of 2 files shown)


klochowicz:
@luckysori I'll have a look whether this fixes the coordinator state I ran into on regtest.

bonomat:
> The idea here is to use this branch to start up the mainnet coordinator and close a few DLC channels that are outdated. We can use the logs to identify which ones are old, or simply close all of them.

It looks like our mainnet is effected by this problem. 

> Storage error Nonsense bytes didn't map to the type they were interpreted as

We know that we do have a "broken" channel on mainnet. 
Can we close said channel somehow?

What's the intention once the channel is closed?

klochowicz:
@luckysori I believe we can now close this PR, right?

Issue #498, opened 2023-04-24
App cannot receive payment with open position


holzeis:
### how to reproduce

1. Fund the wallet 
2. Open position
3. Receive money on the wallet

The transaction is shown in the history but not in the balance.

![image](https://user-images.githubusercontent.com/382048/234048083-fb3931c2-9fe6-48ea-bb9f-d390f3d6a171.png)

holzeis:
Note, I waited long enough to exclude that the wallet has simply not been synchronised.

bonomat:
Can confirm this bug. It happens on my iPhone as well. 
Also after a restart the money does not show up. 

The problem is, you also can't use the funds because the wallet thinks that the money is not there and sending is disabled because of this: 


![image](https://user-images.githubusercontent.com/224613/234051904-4c3f0d87-579c-45ab-9077-bd14e71436d6.png)

bonomat:
We decided to not allow creating invoices while a position is open.

klochowicz:
Removed the 'critical' label in light of our decision to not allow new invoices whilst the position is open.

holzeis:
Just tested this bug on the latest version and it seems it is fixed. 🚀🎉

PR #500, opened 2023-04-24
# Prevent creating invoices when a position is open

klochowicz:
This is a temporary stop-gap for https://github.com/get10101/10101/issues/498


bonomat (approved):
Nice. Thanks .

PR #503, opened 2023-04-24
chore: exponentially retry

bonomat:
It is better to retry not too often as we might annoy poor nodes by retrying to often. Additionally, the retry mechanism should happen less often than the check on the outside.

It's an attempt to make our connection handling a bit smarter. 
I'm not sure it will change much though.


holzeis (approved):
LGTM 👍

klochowicz (approved):
LGTM!

luckysori:
Can we try something like [`backoff`](https://docs.rs/backoff/0.4.0)? I can try to do this later.

klochowicz:
> Can we try something like [`backoff`](https://docs.rs/backoff/0.4.0)? I can try to do this later.

I'd still merge it though, it's a step forward compared to how we're doing it now....
can always make it pretty later ;)

klochowicz:
bors retry

klochowicz:
bors retry

PR #505, opened 2023-04-25
fix: Announce node only on channel creation and startup

holzeis:
We noticed that the coordinator is very chatty, which might be a reason that counterparty do not respond as fast as possible on pings.

Worst, this only reducing the network noise.


luckysori (approved):
Tests pass on my machine.

PR #506, opened 2023-04-25
Fix/open public channels through api

holzeis:


luckysori (approved):

holzeis:
integration tests have been executed locally successfully.

PR #509, opened 2023-04-25
Reduce node sync interval in coordinator

klochowicz:


bonomat (approved):
Let's give this a try 👍

PR #511, opened 2023-04-25
feat: Add json logging option for coordinator and maker

klochowicz:
As usual, use `--json` to enable JSON output.

Turns out that we copied the functionality from ItchySats long time ago, we just
never exposed it.


bonomat (approved):
Too easy

PR #513, opened 2023-04-25
Add even more logs

klochowicz:


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -93,6 +94,11 @@ where
                 output_script,
                 ..
             } => {
+                tracing::info!(
+                    "Funding generation ready for channel with counterparty {}",
```
bonomat, line 98: would be good to have the `counterparty_node_id` as a field

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -244,7 +251,7 @@ where
                     counterparty,
                     funding_satoshis,
                     push_msat,
-                    "Accepting 0-conf channel request"
```
holzeis, line 247: I believe 0-conf was correct here. Although we are just implying this code is only executed for the mobile app. See also configs for the mobile app where we manually accept channel requests.

holzeis (approved):

PR #516, opened 2023-04-26
Re-enable list-DLC-channels endpoint on coordinator

luckysori:
We cannot query the endpoint at the moment. The server automatically reports a 404.


holzeis (approved):
Thanks 👍

PR #517, opened 2023-04-26
chore: add auto mining script

holzeis:
A simply script to automatically mine every 60 seconds a new block. I am using it for testing locally and thought it might be useful to others as well.


klochowicz (approved):
awesome! I'll add `just mine` command on top of this :)

PR #520, opened 2023-04-26
chore: restore deleted coordinator regtest seed file

holzeis:


klochowicz (approved):
thanks for bringing it back, it "might" come in handy :)

PR #526, opened 2023-04-26
fix: Properly print out errors converted into `thiserror`

klochowicz:
Override Display trait of newly constructed `thiserror` errors to print out the
underlying error by default.

### DIFF
--- a/crates/bdk-ldk/src/lib.rs
+++ b/crates/bdk-ldk/src/lib.rs
@@ -46,9 +46,9 @@ const MAX_SATS_PER_V_BYTE: u32 = 20;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("BDK wallet error")]
+    #[error("BDK wallet error: {0}")]
     Bdk(#[from] bdk::Error),
-    #[error("Other")]
+    #[error("Other: {0}")]
     Other(#[from] anyhow::Error),
 }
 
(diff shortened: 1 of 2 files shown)


bonomat (approved):

klochowicz:
bors retry

luckysori (commented):
Given that we use `anyhow` everywhere, I think this is just redundant given that the `from` attribute implies `source` which will then be included when we debug print (`{e:?}`) or even print using the alternate selector (`{e:#}`).
[
Link to Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=be1f8cb365e876c65cf5b5969a3484cd).

klochowicz:
> Given that we use `anyhow` everywhere, I think this is just redundant given that the `from` attribute implies `source` which will then be included when we debug print (`{e:?}`) or even print using the alternate selector (`{e:#}`).
> [ Link to Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=be1f8cb365e876c65cf5b5969a3484cd).

yup, however this will not be present if one does use the `Display` trait to print. Small fix, gets rid of headaches (and we missed important info already due to that). it makes no difference if you use `Debug` trait to print, so no errors there.

given the above, I'm not sure whether it's redundant - this just plugs a hole if one e.g. prints `err` field in `tracing` log.

Issue #189, opened 2023-03-01
The lndlc node should not panic to start if the oracle is not available


holzeis:
```
2023-03-01 07:36:10 DEBUG reqwest::connect: starting new connection: https://oracle.holzeis.me/    
thread 'tokio-runtime-worker' panicked at 'to be able to create the p2pd oracle: IOError(Custom { kind: Other, error: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("oracle.holzeis.me")), port: None, path: "/oracle/publickey", query: None, fragment: None }, source: TimedOut } })', /Users/holzeis/workspace/coblox/10101/crates/ln-dlc-node/src/node.rs:470:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: task 11 panicked
```

holzeis:
Fixed with #527

PR #527, opened 2023-04-26
Retry connecting to the oracle a few times before bailing

klochowicz:
Current setting - retry every 1s up to 10 times.
I created easy consts to adjust the settings if needed.

Tested by unplugging internet temporarily on desktop.

### DIFF
--- a/crates/ln-dlc-node/src/node/oracle_client.rs
+++ b/crates/ln-dlc-node/src/node/oracle_client.rs
@@ -4,11 +4,35 @@ use anyhow::Result;
 use bitcoin::secp256k1::XOnlyPublicKey;
 use dlc_manager::Oracle;
 use p2pd_oracle_client::P2PDOracleClient;
+use std::thread::sleep;
+use std::time::Duration;
+
+// TODO: This should come from the configuration.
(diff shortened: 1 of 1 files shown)


crates/ln-dlc-node/src/node/oracle_client.rs
```diff
@@ -4,11 +4,35 @@ use anyhow::Result;
 use bitcoin::secp256k1::XOnlyPublicKey;
 use dlc_manager::Oracle;
 use p2pd_oracle_client::P2PDOracleClient;
+use std::thread::sleep;
```
holzeis, line 7: 🤔 do we have issues using std::thread instead of tokio?

holzeis (approved):
Nice 👍

holzeis:
I guess this will also resolve https://github.com/get10101/10101/issues/189

crates/ln-dlc-node/src/node/oracle_client.rs
luckysori: It's used in a blocking thread. Where we are using we actually cannot use `tokio::time::sleep`, because it's just an `FnOnce` closure, not a future.

klochowicz:
bors retry

PR #528, opened 2023-04-26
Add message that mentions `2%` fee when receiving funds

da-kami:
Note:
Adding this 2% fee is already an experiment - we will have to closely evaluate (by getting into contact with users) if the fee is a blocker for some users.

@bonomat @klochowicz I this fee is an initial blocker for users we should definitely experiment to charge at a later point (e.g. when trading, or if the channel was not used for some time, ...). Please think through if we can derive the fee being a blocker or not from the data we have available. 
I think we should be able to correlate the app being opened by a users and the user sending funds. It would be great to also have stats on which users succeeded to get funds in and which users did not.


da-kami:
Let's rather wait for them to complain.

PR #534, opened 2023-04-27
chore: Change data directory on phone

holzeis:
On iOS the ApplicationSupportDirectory is never exposed to the user and can not be backed up. Contrary the ApplicationDocumentsDirectory is user generated data, that can be backed up. I think this still not perfect as we have both kinds of data, but being able to access the data dir on the phone has IMHO several benefits. Especially during the beta phase.

This however, will be a breaking change, as the data dir folder will change. If we want that, we should get this in before the first beta users get to test the app.

Screenshot 2023-04-27 at 15 40 55

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -2,6 +2,8 @@ PODS:
   - Flutter (1.0.0)
   - flutter_native_splash (0.0.1):
     - Flutter
+  - package_info_plus (0.4.5):
+    - Flutter
   - path_provider_foundation (0.0.1):
     - Flutter
     - FlutterMacOS
@@ -14,6 +16,7 @@ PODS:
 DEPENDENCIES:
   - Flutter (from `Flutter`)
   - flutter_native_splash (from `.symlinks/plugins/flutter_native_splash/ios`)
+  - package_info_plus (from `.symlinks/plugins/package_info_plus/ios`)
   - path_provider_foundation (from `.symlinks/plugins/path_provider_foundation/ios`)
   - share_plus (from `.symlinks/plugins/share_plus/ios`)
   - shared_preferences_foundation (from `.symlinks/plugins/shared_preferences_foundation/ios`)
@@ -23,6 +26,8 @@ EXTERNAL SOURCES:
     :path: Flutter
   flutter_native_splash:
     :path: ".symlinks/plugins/flutter_native_splash/ios"
+  package_info_plus:
+    :path: ".symlinks/plugins/package_info_plus/ios"
   path_provider_foundation:
     :path: ".symlinks/plugins/path_provider_foundation/ios"
   share_plus:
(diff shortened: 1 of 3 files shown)


bonomat (approved):
Could we not write a migration for this and roll out later? 
Otherwise we need to wait again for Apple to review it before we can reach out to users. ...

holzeis:
> Could we not write a migration for this and roll out later? Otherwise we need to wait again for Apple to review it before we can reach out to users. ...

Yes of course.. but that will mean more implementation efforts. Also I am not sure how easy the migration script will be, considering that we have to move data that is currently accessed..

mobile/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: get_10101
 description: 10101 combines the power of a self-custodial on-chain and off-chain wallet with the vast world of trading.
 publish_to: none
-version: 1.0.12
+version: 0.0.1
```
da-kami, line 4: Why is this commit needed and part of this PR?

I read the commit message `This changes the pubspec version to run a test build on TestFlight that will not be shown on the top but on version 0.0.1` but I don't understand *why* this was necessary?

da-kami (commented):
I don't think we should block releasing the beta on this as it is not strictly necessary. 
Yes, there will be a file migration needed but that's fine.
I would even go as far as telling users backwards compatibility is not ensured in the first few versions. We can migrate, but if it does not work I would not bother too much and just tell the users.

mobile/pubspec.yaml
holzeis: read the commit message 😜 : [Remove this commit before merging](https://github.com/get10101/10101/pull/534/commits/c4c2886f3c4fa538ab17e5f43b5b68ad89a1e200)

klochowicz:
@holzeis asked an important question in a different PR: https://github.com/get10101/10101/pull/538#pullrequestreview-1404510512

in essence: This might be a breaking change - how we should deal with that? (do we have a policy?)

in my view, if we use #538 mechanism, the seed will stay in the same location, but any open trades or lightning channels might get lost, so at the very least recovering funds will be delayed. 
TL;DR: it will be not good, but not *bad* (no money lost, just delayed).

Ideally, we should avoid even that situation. Maybe we could *copy* the data over in this case? What do you think @holzeis ? 

basically, if the `app` dir in the app support dir was not empty (contained `bitcoin` dir), then copy it over. I think that should be graceful enough?

klochowicz (approved):

klochowicz:
@holzeis I have updated the PR to keep the seed location as it was, with the change to app support directory. 

I'm not sure how to go about the breaking change caused by this (a feature flag for new users? ) a migration?

holzeis:
> @holzeis I have updated the PR to keep the seed location as it was, with the change to app support directory.
> 
> I'm not sure how to go about the breaking change caused by this (a feature flag for new users? ) a migration?

I've added a simple check to use the `seedDir` in case it has already been initialised with data. 0cc815a773841213a544c97725515cfad61d2a24

Now its not a breaking change anymore 🙂 .

klochowicz (approved):
thank you for adding the backwards compatibility part.  LGTM! 👍

PR #538, opened 2023-04-27
feat(mobile): Decouple application data and seed locations

klochowicz:
This allows for more flexibility, e.g. store app data in a different location
from the app seed.

This PR does not immediately use this feature, defaulting to the same directory.

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -260,7 +260,7 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
       final appSupportDir = await getApplicationSupportDirectory();
       FLog.info(text: "App data will be stored in: $appSupportDir");
 
-      await rust.api.run(config: config, appDir: appSupportDir.path);
+      await rust.api.run(config: config, appDir: appSupportDir.path, seedDir: appSupportDir.path);
 
       await orderChangeNotifier.initialize();
       await positionChangeNotifier.initialize();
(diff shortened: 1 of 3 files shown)


holzeis (approved):
LGTM!

Note, this is a breaking change. We don't have a policy yet for how to deal with that.

I suggest not being downward compatible and simply inform the users e.g. to drain there wallet before the next version.

klochowicz:
> LGTM!
> 
> Note, this is a breaking change. We don't have a policy yet for how to deal with that.
> 
> I suggest not being downward compatible and simply inform the users e.g. to drain there wallet before the next version.

how is this a breaking change? I resisted the urge to remove the `network` dir from it, and thought I avoided breaking anything! (breaking would be only introduced in #534

holzeis:
> > LGTM!
> 
> > 
> 
> > Note, this is a breaking change. We don't have a policy yet for how to deal with that.
> 
> > 
> 
> > I suggest not being downward compatible and simply inform the users e.g. to drain there wallet before the next version.
> 
> 
> 
> how is this a breaking change? I resisted the urge to remove the `network` dir from it, and thought I avoided breaking anything! (breaking would be only introduced in #534 

Fair enough. The question should better be addressed on #534, without which this change wouldn't be needed.

klochowicz:
I will ask this question it the other place.

PR #540, opened 2023-04-27
Cargo lock update

da-kami:
The toml changed but it appears the lock was not updated?

I ran into this problem when running clippy locally.


klochowicz (approved):

klochowicz:
thanks for spotting this!

PR #542, opened 2023-04-28
WIP: add esplora client

bonomat:
this is a wip to replace electrum with an async esplora client. 

Last commit does not compile. I stopped half-way.


da-kami:
Taking this over :)

PR #548, opened 2023-04-28
fix(mobile): Only store email address in preferences after successful registration

klochowicz:
Don't store email address in preferences if we somehow could not register (e.g.
backend was not started properly).

### DIFF
--- a/mobile/lib/features/welcome/welcome_screen.dart
+++ b/mobile/lib/features/welcome/welcome_screen.dart
@@ -74,9 +74,9 @@ class _WelcomeScreenState extends State<WelcomeScreen> {
                   onPressed: () {
                     if (_formKey.currentState != null && _formKey.currentState!.validate()) {
                       _formKey.currentState?.save();
+                      api.registerBeta(email: _email);
                       Preferences.instance.setEmailAddress(_email);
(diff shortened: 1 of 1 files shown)


mobile/lib/features/welcome/welcome_screen.dart
```diff
@@ -74,9 +74,9 @@ class _WelcomeScreenState extends State<WelcomeScreen> {
                   onPressed: () {
                     if (_formKey.currentState != null && _formKey.currentState!.validate()) {
                       _formKey.currentState?.save();
+                      api.registerBeta(email: _email);
```
da-kami, line 77: 🙃 It appears I missed this in a previous review: 
We should avoid `api` calls in the user interface!

It would be great if you could clean this up and wrap it in a service.
I think the service call should be wrapped in a `try catch` here so we are exiting gracefully.

da-kami (approved):

mobile/lib/features/welcome/welcome_screen.dart
klochowicz: I'll create a refactor ticket out of this, right now it's not the priority

PR #552, opened 2023-04-28
Do not start the UI in case initialisation fails

klochowicz:
I have added the naive approach of just closing the app.

Hopefully exit(1) will result in submitting a TestFlight reports (exit with
error).

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -272,11 +272,24 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
       await walletChangeNotifier.refreshWalletInfo();
     } on FfiException catch (error) {
       FLog.error(text: "Failed to initialise: Error: ${error.message}", exception: error);
+      exitAppAfterErrors();
     } catch (error) {
       FLog.error(text: "Failed to initialise: $error", exception: error);
+      exitAppAfterErrors();
     } finally {
       FlutterNativeSplash.remove();
     }
+  }
+
+  exitAppAfterErrors() {
+    FLog.info(text: "Exiting app due to startup errors");
+
+    // Add arbitrary amount of sleep to allow accepting permission dialogs if
+    // lack of permissions caused the app to crash, so that the user can start the app next time
+    var duration = const Duration(seconds: 10);
+    sleep(duration);
+
+    exit(1); // Hopefully this triggers TestFlight crash report
   }
 
   setupRustLogging() {


klochowicz:
note: eventhough it seems like a trivial change, I'm a bit worried that this could cause unintended consequences (what if user needs to explicitly enable local network permissions? like @holzeis showed on his screen once - I tried catering for that, but I'm not sure whether they would pop up over the splashscreen).

Ideally I would have guarded it with a feature flag (managed as per feature flag / A/B testing discussion), especially given our TestFlight release cycle.

klochowicz:
the alternative approach could involve storing a flag in preferences `initialised = false` at startup, and setting to true at the end of init, and then adjusting welcome screen based on that.

might be more elegant approach, I'll give it a quick try.

mobile/lib/main.dart
```diff
@@ -272,13 +272,26 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
...
 
+  exitAppAfterErrors() {
+    FLog.info(text: "Exiting app due to startup errors");
+
+    // Add arbitrary amount of sleep to allow accepting permission dialogs if
+    // lack of permissions caused the app to crash, so that the user can start the app next time
+    var duration = const Duration(seconds: 10);
+    sleep(duration);
+
+    exit(1); // Hopefully this triggers TestFlight crash report
```
holzeis, line 292: I guess it will crash the app, but I guess we will only get a crash report in TestFlight with the StackTrace pointing to this exit. Why not remove the try catch from above?

holzeis (commented):
I guess this change will get the job done in terms of crashing the app if the initialisation fails, but will not result in a crash report thats helpful to us.

holzeis:
> but I'm not sure whether they would pop up over the splashscreen

For me it did pop up over the splash screen.

> the alternative approach could involve storing a flag in preferences initialised = false at startup, and setting to true at the end of init, and then adjusting welcome screen based on that.

I think we are fine with the splash screen only being removed once the app is initialised, if we ensure the app crashes in case the initialisation failed.

mobile/lib/main.dart
klochowicz: > > but I'm not sure whether they would pop up over the splashscreen
> 
> For me it did pop up over the splash screen.

great, thanks for testing.

> 
> > Ideally I would have guarded it with a feature flag (managed as per feature flag / A/B testing discussion), especially given our TestFlight release cycle.
> 
> I don't think that this fix justifies a feature flag. If we would use feature flags for little things like that I'd fear we would quickly end up with loads of flags, which I guess will get hell to maintain. I think the best course of action here is to write a simple test, that fails the intitialisation and assert that the app crashes.

the feature flag idea is for *temporary* adding things that are potentially risky, 
Such feature flag can be removed in the upcoming release after it stabilises.

Sometimes a test would do, it's not a panacea though.

> 
> > the alternative approach could involve storing a flag in preferences initialised = false at startup, and setting to true at the end of init, and then adjusting welcome screen based on that.
> 
> I think we are fine with the splash screen only being removed once the app is initialised, if we ensure the app crashes in case the initialisation failed.

klochowicz:
> I guess this change will get the job done in terms of crashing the app if the initialisation fails, but will not result in a crash report thats helpful to us.

it's still a step in the right direction, yes? At least the user cannot due silly things anymore in the app.... we're tightening the constraints.

As a further enhancement, we could either:
- send the logs automatically (e.g. via email)
OR
- open the Settings screen and disable wallet and trading functionality (so that the user can show / send the logs to help us debug.

I kinda like the second approach listed above.

As such, I consider these out of scope of these PR.

holzeis:
> > I guess this change will get the job done in terms of crashing the app if the initialisation fails, but will not result in a crash report thats helpful to us.
> 
> it's still a step in the right direction, yes? At least the user cannot due silly things anymore in the app.... we're tightening the constraints.

@klochowicz Why not remove the try catch from above? instead of using the `exit(1)`. That way we would also get the correct StackTrace in the Apple Crash Report.

klochowicz:
> > > I guess this change will get the job done in terms of crashing the app if the initialisation fails, but will not result in a crash report thats helpful to us.
> > 
> > 
> > it's still a step in the right direction, yes? At least the user cannot due silly things anymore in the app.... we're tightening the constraints.
> 
> @klochowicz Why not simply remove the try catch from above? instead of using the `exit(1)`. That way we would also get the correct StackTrace in the Apple Crash Report.

I have tested this on native and iOS Simulatar. In both cases the Dart VM displays the backtrace but does *not* exit the app (crash). 
The thing with Dart VM is that it is resilient in face of exceptions, so I'm not sure how we would get the correct StackTrace then? (At least with the `exit(1)` we catch and display the logs.

am I missing anything here @holzeis ?

holzeis:
> > > > I guess this change will get the job done in terms of crashing the app if the initialisation fails, but will not result in a crash report thats helpful to us.
> 
> > > 
> 
> > > 
> 
> > > it's still a step in the right direction, yes? At least the user cannot due silly things anymore in the app.... we're tightening the constraints.
> 
> > 
> 
> > @klochowicz Why not simply remove the try catch from above? instead of using the `exit(1)`. That way we would also get the correct StackTrace in the Apple Crash Report.
> 
> 
> 
> I have tested this on native and iOS Simulatar. In both cases the Dart VM displays the backtrace but does *not* exit the app (crash) - we hang on the splashscreen forever.
> 
> The thing with Dart VM is that it is resilient in face of exceptions, so I'm not sure how we would get the correct StackTrace then? (At least with the `exit(1)` we catch and display the logs.
> 
> 
> 
> am I missing anything here @holzeis ? 

Thanks for the explanation. I hoped a dart exception would have crashed the app.

mobile/lib/main.dart
klochowicz: closing this conversation, as removing the try/catch does not close the app, and the app just hangs.

da-kami:
I don't really understand the improvement in this PR.

Here is my understanding:

Prior to this PR the app would just crash when we run into an error during initialization.
Now the app exits instead of crashing...? How does this make a difference for the user?
Does the app still hang after this PR? (I assume not, it would just close) - Is the fact that the app does not hang anymore the improvement of this PR?

klochowicz:
> I don't really understand the improvement in this PR.
> 
> Here is my understanding:
> 
> Prior to this PR the app would just crash when we run into an error during initialization. Now the app exits instead of crashing...? How does this make a difference for the user? Does the app still hang after this PR? (I assume not, it would just close) - Is the fact that the app does not hang anymore the improvement of this PR?

the app would *not* crash previously. 
Now it will crash and exit.

previously, just like in corporate Java-land, we had a giant catch-all exceptions and go-on init 🙈
now we're actually exiting with error code - hopefully trigerring a crash report in the Apple (to be tested in release).

holzeis:
> hopefully trigerring a crash report in the Apple (to be tested in release).

We could actually test it. Simply build a TestFlight release using the GitHub action. I would advise to temporarily change the version in pubspec before though, so we do not block a build number on the current release version.

da-kami (approved):

klochowicz:
> > hopefully trigerring a crash report in the Apple (to be tested in release).
> 
> We could actually test it. Simply build a TestFlight release using the GitHub action. I would advise to temporarily change the version in pubspec before though, so we do not block a build number on the current release version.

I'll test it when the internal TestFlight release gets deployed.

klochowicz:
as we don't want to hold off a main TestFlight release, I've opted to merge it only after the new release, so that there's time to revert it easily if it doesn't work as expected without causing any friction.

holzeis:
@klochowicz I looked a little bit into error handling in flutter and found the following.

https://docs.flutter.dev/testing/errors#errors-caught-by-flutter

I think we should do the following.

1. Add a general handling for any unhandled exception and exit the app. 
`main.dart`
```
FlutterError.onError = (details) {
    FlutterError.presentError(details);
    if (kReleaseMode) exit(1);
  };
```
2. Build our iOS app with the release flag. As otherwise errors will be swallowed by flutter.
`ios.yml`
```
      - name: Build iOS Archive
        id: build-ios
        run: |
          BUILD_NAME=${{ steps.version.outputs.data }}
          BUILD_NUMBER=$(git rev-list HEAD --count)
          cd mobile
          flutter build ipa --export-options-plist=ios/exportOptions.plist \
                            --dart-define="ELECTRS_ENDPOINT=$ELECTRS_ENDPOINT" \
                            --dart-define="COORDINATOR_P2P_ENDPOINT=$COORDINATOR_P2P_ENDPOINT" \
                            --dart-define="NETWORK=$NETWORK" \
                            --dart-define="COMMIT=$(git rev-parse HEAD)" \
                            --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
                            --dart-define="COORDINATOR_PORT_HTTP=$COORDINATOR_PORT_HTTP" \
                            --build-name=$BUILD_NAME \
                            --build-number=$BUILD_NUMBER \
                            --release
```
3. Remove the big try catch on the main function.
4. Test by running the `Build and release ios app to TestFlight` workflow (ios.yml) manually. Throw a simple exception in the startup and check if the app crashes. (We don't have to wait for a release to do that.)
Note, we should just make sure that we adapt the pubspec version for that test. e.g. to 0.0.1 so that we do not block a build number.

klochowicz:
> 1. Add a general handling for any unhandled exception and exit the app.

I don't think this is a good idea at this stage (eventually, yes). Sometimes exceptions happen just in the UI layer, e.g. the trading screen -  #402 - we absolutely don't want to exit the app on this exception.

> FlutterError.onError = (details) {
    FlutterError.presentError(details);
    if (kReleaseMode) exit(1);
  };

I have no idea why wouldn't we exit in debug mode - IMHO we should be even more strict in exiting in debug mode.

I don't mind going eventually in that direction (especially exiting in *debug* first and foremost, to notice any exception).

anyway, maybe we can spin this discussion off to a separate issue @holzeis? It goes outside of scope of this PR and is a step further.

holzeis:
> I have no idea why wouldn't we exit in debug mode - IMHO we should be even more strict in exiting in debug mode.

I copied that from the flutter example. Not sure why they've added it only for release mode, but I think it's because in debug mode error are never unhandled?

klochowicz:
> > I have no idea why wouldn't we exit in debug mode - IMHO we should be even more strict in exiting in debug mode.
> 
> I copied that from the flutter example. Not sure why they've added it only for release mode, but I think it's because in debug mode error are never unhandled?

they are unhandled in debug mode - I test locally on debug mode.

klochowicz:
I've decided to go with an alternative to this PR that will ensure that we get a crash report #595 .

Basically, it avoids the cunundrum with Flutter exceptions by not emitting one at all  😊(we 'expect' it to initialise).

Issue #544, opened 2023-04-28
Replace electrs with esplora async


bonomat:
It is very likely that our connection dropouts are due to our blocking electrs sync. 
We want to replace this with an async client. 

I've started with this here already : https://github.com/get10101/10101/pull/542

bonomat:
Hopefully this solves https://github.com/get10101/10101/issues/492

klochowicz:
Done in #576

PR #557, opened 2023-04-29
fix: Set default fee rate to max

### LINKED ISSUE #555: Fee rate much too low: force closure
We have another channel force closure

```
{"timestamp":"2023-04-28T19:25:25.452196137Z","level":"INFO","fields":{"message":"\nChannel closed","channel":"cc61caf56546cc48cd1039b06609b14d55aebc3e5f38f79890ccee3d95cd8c44","reason":"CounterpartyForceClosed { peer_msg: \"Peer's feerate much too low. Actual: 253. Our expected lower limit: 5060 (- 250)\" }"},"target":"ln_dlc_node::ln::event_handler"}
```

and this one: 

```
2023-04-25 17:44:51 TRACE lightning::ln::peer_handler: Handling SendErrorMessage HandleError event in peer_handler for node 03fe8461ebc025880b58021c540e0b7782bb2bcdc99da9822f5c6d2184a59b8f69 with message Peer's feerate much too low. Actual: 2081. Our expected lower limit: 3036 (- 250)    
```

This is critical because our money is locked up for 2 weeks

holzeis:
In case we can't fetch a fee estimate through the electrum client, we used to return the default fee rate which was just 1 sats per v byte. This was much too low resulting into channel force closure upon channel fee updates.

resolves #555 

Note, this is more like a quick patch to prevent further force closures and should become obsolete once #544 is done.

### DIFF
--- a/crates/bdk-ldk/src/lib.rs
+++ b/crates/bdk-ldk/src/lib.rs
@@ -13,6 +13,7 @@ use bdk::database::BatchDatabase;
 use bdk::wallet::AddressIndex;
 use bdk::wallet::Wallet;
 use bdk::Balance;
+use bdk::FeeRate;
 use bdk::SignOptions;
 use bdk::SyncOptions;
 use std::cmp::min;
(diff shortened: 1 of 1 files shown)


crates/bdk-ldk/src/lib.rs
```diff
@@ -406,7 +407,18 @@ where
             ConfirmationTarget::HighPriority => 1,
         };
 
-        let estimate = self.client.estimate_fee(target_blocks).unwrap_or_default();
+        let estimate = match self.client.estimate_fee(target_blocks) {
+            Ok(estimate) => estimate,
+            Err(e) => {
+                tracing::warn!("Failed to get fee estimate. Using max sats per v byte {MAX_SATS_PER_V_BYTE} instead. Error: {e:?}");
+
+                // FIXME: Setting the fee rate to the maximum is a safe guard to prevent the channel
```
bonomat, line 415: I thought the problem of the force closure was because dour feetate was set to too high, compared to the counterpart's?

holzeis: I think it depends on the perspective..? I have to look into it to be completely sure.

bonomat:
If fetching the fee rate is the source of the problem then I have a different solution. PR is up in a minute. :)

holzeis:
Closed in favour of #559

PR #545, opened 2023-04-28
 WIP: add esplora client 

bonomat:
this is a wip to replace electrum with an async esplora client.

Last commit does not compile. I stopped half-way.

Replaces #542

PR #559, opened 2023-04-29
feat: add mempool space client

bonomat:
Mempool.space provices the most accurate fee estimate today.

I failed at using the async function because all the caller functions are sync. Hence I just added a second sync function using reqwest blocking. Imho this should be fine because the call is quite fast. 

Maybe that's a better solution than #557 ?

### DIFF
--- a/crates/bdk-ldk/src/lib.rs
+++ b/crates/bdk-ldk/src/lib.rs
@@ -481,7 +481,8 @@ fn estimate_fee_rate(target_blocks: u32, fee_estimator: &MempoolFeeEstimator) ->
 mod tests {
     #[test]
     fn it_works() {
-        let result = 2 + 2;
-        assert_eq!(result, 4);
+        let ten_ten_one = "10101";
+        let ten_ten_one = isize::from_str_radix(ten_ten_one, 2).unwrap();
+        assert_eq!(ten_ten_one, 21);
     }
 }
--- a/crates/bdk-ldk/src/lib.rs
+++ b/crates/bdk-ldk/src/lib.rs
@@ -13,6 +13,7 @@ use bdk::database::BatchDatabase;
 use bdk::wallet::AddressIndex;
 use bdk::wallet::Wallet;
 use bdk::Balance;
+use bdk::FeeRate;
 use bdk::SignOptions;
 use bdk::SyncOptions;
 use std::cmp::min;
@@ -25,6 +26,7 @@ use lightning::chain::chaininterface::FeeEstimator;
 use lightning::chain::Confirm;
 use lightning::chain::Filter;
 use lightning::chain::WatchedOutput;
+use mempool_space::MempoolFeeEstimator;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::Mutex;
@@ -43,6 +45,12 @@ mod indexed_chain;
 /// We pick 20 sats/vbyte because at the time of writing this was the requirement to get into the
 /// next block.
 const MAX_SATS_PER_V_BYTE: u32 = 20;
+
+/// Min TX fee for all transactions
+///
+/// If our fee estimator throws an error we want to return a min fee rate. At the time of writing,
+/// the min relay fee was 1 sat/vbyte which generally is the lower limit for transactions fees.
+const MIN_SATS_PER_V_BYTE: u32 = 1;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
@@ -86,6 +94,7 @@ impl Default for TxFilter {
 /// needed to use lightning with LDK.  Note: The bdk::Blockchain you use
 /// must implement the IndexedChain trait.
 pub struct LightningWallet<B, D> {
+    fee_estimator: MempoolFeeEstimator,
     client: Arc<B>,
     wallet: Mutex<Wallet<D>>,
     filter: Mutex<TxFilter>,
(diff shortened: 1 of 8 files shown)


crates/bdk-ldk/src/lib.rs
```diff
@@ -366,8 +374,10 @@ where
             ConfirmationTarget::HighPriority => 1,
         };
 
-        let estimate = self.client.estimate_fee(target_blocks).unwrap_or_default();
-        let sats_per_vbyte = estimate.as_sat_per_vb() as u32;
+        let sats_per_vbyte = self
+            .fee_estimator
+            .estimate_fee_blocking(target_blocks)
+            .unwrap_or_default();
```
holzeis, line 380: This would still end up with the default value of u32 `0` if the mempool call fails. I guess thats kind of the same issue we had using the electrum_client as fee estimator. At least I would not use `.unwrap_or_default()`, but rather handle the error and print that we failed to fetch the fee rate from mempool. I also wouldn't return 0 as fee rate as this would always end up with a fee rate of 253 sats.

Instead I'd suggest to return something like `MAX_SATS_PER_V_BYTE`. This is of course not perfect, but I guess a better value than `0`, as judging from the ldk code I think we have better chances of preventing the channel to get force closed. But its definitely not a guarantee.

----

LDK will check on the upper limit like this.

```
let upper_limit = cmp::max(250 * 25,
			fee_estimator.bounded_sat_per_1000_weight(ConfirmationTarget::HighPriority) as u64 * 
```

and for the lower limit like this, with the `feerate_per_kw` being the counterparties proposed feerate.

```
if feerate_per_kw + 250 < lower_limit {
  if let Some(cur_feerate) = cur_feerate_per_kw {
    if feerate_per_kw > cur_feerate {
      log_warn!(logger, "Accepting feerate that may prevent us from closing this channel because it's higher than what we have now. Had {} s/kW, now {} s/kW.", cur_feerate, feerate_per_kw);
      return Ok(());
    }
  }
  return Err(ChannelError::Close(format!("Peer's feerate much too low. Actual: {}. Our expected lower limit: {} (- 250)", feerate_per_kw, lower_limit)));
}
```

bonomat: Thanks for the explanation. I'll see to make this better.

bonomat: > Instead I'd suggest to return something like `MAX_SATS_PER_V_BYTE`. This is of course not perfect, but I guess a better value than `0`, as judging from the ldk code I think we have better chances of preventing the channel to get force closed. But its definitely not a guarantee.

Why do you want to return `MAX_SATS_PER_V_BYTE`? Wouldn't `1 sat/vbyte` be better?

holzeis: > > Instead I'd suggest to return something like `MAX_SATS_PER_V_BYTE`. This is of course not perfect, but I guess a better value than `0`, as judging from the ldk code I think we have better chances of preventing the channel to get force closed. But its definitely not a guarantee.
> 
> Why do you want to return `MAX_SATS_PER_V_BYTE`? Wouldn't `1 sat/vbyte` be better?

If I read the ldk code correctly, the 
- upper limit is at least `250 * 25 = 6_250` , so with a `MAX_SATS_PER_V_BYTE` of `20`, `253 * 20 = 5_060` we should not surpass that threshold.
- lower limit is whatever electrs or now mempool would say. Now if one side would `unwrap_or_default()` to the minimum fee rate of `1 sat/vbyte` we would end up with `253 sats/vbyte`. It's likely that the current fee rate is higher than that.

Hence in order to reduce the risk of a force closure due to a `peer's feerate much too low`, I would pick a larger feerate as we will not reach the upper limit, but have more chances on being larger than the lower limit. But of course its just a "better guess" and not a guarantee, that we will not run into this error again.

crates/bdk-ldk/src/lib.rs
```diff
@@ -453,6 +464,19 @@ pub enum ScriptStatus {
...
 
+fn estimate_fee_rate(target_blocks: u32, fee_estimator: &MempoolFeeEstimator) -> u32 {
+    match fee_estimator.estimate_fee_blocking(target_blocks) {
+        Ok(sats_per_vbyte) => sats_per_vbyte,
+        Err(err) => {
+            tracing::error!(
+                default_fee_rate = MIN_SATS_PER_V_BYTE,
+                "Could not retrieve fee rate from mempool {err}. Falling back to default rate"
+            );
+            1
```
holzeis, line 475: 🔧 We should make use of the constant here 🙂

crates/bdk-ldk/src/lib.rs
```diff
@@ -481,7 +481,8 @@ fn estimate_fee_rate(target_blocks: u32, fee_estimator: &MempoolFeeEstimator) ->
 mod tests {
     #[test]
     fn it_works() {
-        let result = 2 + 2;
-        assert_eq!(result, 4);
+        let ten_ten_one = "10101";
+        let ten_ten_one = isize::from_str_radix(ten_ten_one, 2).unwrap();
+        assert_eq!(ten_ten_one, 21);
```
holzeis, line 486: 🚀

bonomat:
While I think we should use mempool.space for the fee estimate I think we should hold this back until https://github.com/get10101/10101/pull/545 lands to avoid unnecessary merge conflicts.

klochowicz:
> While I think we should use mempool.space for the fee estimate I think we should hold this back until #545 lands to avoid unnecessary merge conflicts.

shall I pick it up and rebase this on master as #545 already landed on main?

PR #561, opened 2023-05-01
chore: Add test for claiming funds of a closed channel

holzeis:
Turns out the relevant configuration is `our_to_self_delay` and not `their_to_self_delay`. The later only specifies the amount of blocks we are at max willing to wait for claim funds after a force closed channel. `our_to_self_delay` is eventually the configuration we are using in case we are opening the channel. We used the default value of 144 blocks which is about 24 hours. I think this value is reasonable, but could be decreased even more for our just in time channels.

This PR pulls this config to 10101 and sets it to the default value of 144. So nothing changes here. I also added a test to ensure that force closing is actually working. Additionally I added a collaborative channel closure tests, which shows that the user has to wait for 6 blocks to receive the `[Event::SpendableOutput]` to broadcast the transaction to claim the on-chain funds. This will require the user to wait an additional block, before the funds are back to her wallet.

TIL: the app needs to be online after 
 - 144 blocks in case of a force closed channel
 - 6 blocks in case of a collaborative closed channel
to broadcast a transaction to be able to claim the funds. LDK does not do that automatically.

### DIFF
--- a/crates/bdk-ldk/src/lib.rs
+++ b/crates/bdk-ldk/src/lib.rs
@@ -421,7 +421,7 @@ where
     fn broadcast_transaction(&self, tx: &Transaction) {
         let tx_hex = serialize_hex(tx);
         let txid = tx.txid();
-        tracing::info!(%tx_hex, %txid, "Broadcasting transaction");
+        tracing::info!(%txid, %tx_hex, "Broadcasting transaction");
         if let Err(e) = self.client.broadcast(tx) {
             tracing::error!("Error broadcasting transaction: {e:#}");
         }
--- a/crates/ln-dlc-node/src/ln/config.rs
+++ b/crates/ln-dlc-node/src/ln/config.rs
@@ -14,6 +14,8 @@ pub fn app_config() -> UserConfig {
             minimum_depth: 1,
             // There is no risk in the leaf channel to receive 100% of the channel capacity.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
+            // the delay before the fund of a force closed channel can be claimed again
+            our_to_self_delay: 144,
             ..Default::default()
         },
         channel_handshake_limits: ChannelHandshakeLimits {
@@ -50,6 +52,8 @@ pub fn coordinator_config() -> UserConfig {
             minimum_depth: 1,
             // We set this 100% as the coordinator is online 24/7 and can take the risk.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
+            // the delay before the fund of a force closed channel can be claimed again
+            our_to_self_delay: 144,
             ..Default::default()
         },
         channel_handshake_limits: ChannelHandshakeLimits {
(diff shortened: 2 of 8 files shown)


holzeis:
This insight should remove the critical flag on the ticket #555

crates/ln-dlc-node/src/tests/just_in_time_channel/force_close.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    // this sync triggers the `[Event::SpendableOutputs]` broadcasting the transaction to claim the
+    // payees coins.
+    payee.sync().unwrap();
+
+    // mine a single block to claim the spendable output after waiting for the force close delay.
+    bitcoind::mine(1).await.unwrap();
+    payee.sync().unwrap();
+
+    // 1_000 - 122 fees = 878 sats
```
luckysori: :upside_down_face: It's preferable if we do the calculation in the code, like we do in other tests.

crates/ln-dlc-node/src/ln/config.rs
```diff
@@ -14,6 +14,8 @@ pub fn app_config() -> UserConfig {
             minimum_depth: 1,
             // There is no risk in the leaf channel to receive 100% of the channel capacity.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
+            // the delay before the fund of a force closed channel can be claimed again
+            our_to_self_delay: 144,
```
luckysori, line 18: :wrench: This patch is not just "adding a test". Can you either rename the patch or split it in 2?

crates/bdk-ldk/src/lib.rs
```diff
@@ -421,7 +421,7 @@ where
     fn broadcast_transaction(&self, tx: &Transaction) {
         let tx_hex = serialize_hex(tx);
         let txid = tx.txid();
-        tracing::info!(%tx_hex, %txid, "Broadcasting transaction");
+        tracing::info!(%txid, %tx_hex, "Broadcasting transaction");
```
luckysori, line 424: Nice!

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
```diff
@@ -0,0 +1,169 @@
...
+        .list_usable_channels()
+        .first()
+        .unwrap()
+        .channel_id;
+
+    assert_eq!(payee.get_on_chain_balance()?.confirmed, 0);
+    assert_eq!(payee.get_ldk_balance().available, 1_000);
+    assert_eq!(payee.get_ldk_balance().pending_close, 0);
+
+    if TestPath::ForceClose == test_path {
```
luckysori, line 125: :wrench: `match`ing is more idiomatic and generally safer if we end up adding more variants.

crates/ln-dlc-node/src/tests/just_in_time_channel/mod.rs
```diff
@@ -9,4 +9,6 @@ pub enum TestPath {
     OnlineFunding,
     // funding through a mobile lightning node (on the same phone)
     MobileFunding,
+    ForceClose,
+    CollabClose,
```
luckysori, line 13: :wrench: I would prefer it if we had a separate enum for this. Since the new variants only apply to `channel_close`. Otherwise you always have to handle all variants, or work with `if let else` which kind of defeats the point of using enums.

luckysori (approved):
Awesome, love the tests.

holzeis:
bors retry

PR #565, opened 2023-05-01
Store trades in db

### LINKED ISSUE #560: Add timestamp to users table
That makes it easier to track when we got new users.

### LINKED ISSUE #564: Metric for successful trades
Afaik we don't have an easy way to track successful trades from our `coordinator` db: 

1. The `order` table in `coordinator` is actually the `orderbook`, i.e. it includes `orders` which might not have been filled yet
2. The `position` table shows positions with states `open` and `closing`. 
	- Afaik `closing` means that a position has been opened and closed, i.e. these were _two_ trades. 

We should add an extra table which tracks successful trades per trader

bonomat:
Resolves #564
Resolves #560


coordinator/src/node.rs
```diff
@@ -194,6 +209,20 @@ impl Node {
             .propose_dlc_channel_collaborative_settlement(&channel_id, accept_settlement_amount)?;
 
         let mut connection = self.pool.get()?;
+        db::trades::insert(
```
bonomat, line 212: I know that at this point it is not guaranteed that the trade will be successful, but I guess this is good for now. We will see based on the `position_id` if a trade was successful or not.

luckysori (approved):

Issue #558, opened 2023-04-29
Position stuck in `Closing` forever in app


da-kami:
I had a position open for about 20 hours, when closing everything looked good, but the order remained in `Open` forever, and the position forever in `Closing`. 

I did not find any problem - neither in the app nor the coordinator - here's the [relevant timeframe Grafana link](https://itchysats.grafana.net/goto/-fkdFvy4R?orgId=1)

App logs:

```
| 2023-04-29 09:05:01.848 | LogLevel.DEBUG | New message from orderbook msg: {"Match":{"order_id":"7418d745-2e99-4124-84a1-4c2652a146a6","expiry_timestamp":[2023,121,0,0,0,0,0,0,0],"oracle_pk":"16f88cf7d21e6c0f46bcbc983a4e3b19726c6c98858cc31c83551a88fde171c0","matches":[{"order_id":"40885ca4-3a6d-4228-aff0-59408abf3273","quantity":0.59,"pubkey":"03f75f318471d32d39be3c86c622e2c51bd5731bf95f98aaa3ed5d6e1c0025927f","execution_price":29292.5}]}} 
| 2023-04-29 09:05:01.850 | LogLevel.INFO | Received match from orderbook order_id: 7418d745-2e99-4124-84a1-4c2652a146a6 
[...]

PR #567, opened 2023-05-01
Only update position state after order got submitted

klochowicz:
.

### DIFF
--- a/mobile/native/src/trade/order/handler.rs
+++ b/mobile/native/src/trade/order/handler.rs
@@ -17,9 +17,12 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
-        order_failed(Some(order.id), FailureReason::OrderNotAcceptable, e)?;
-        bail!("Could not submit order because extending/reducing the position is not part of the MVP scope");
+    let position = position::handler::get_current_position()?;
+
+    if !position::handler::is_order_acceptable_for_position(&position, &order)? {
+        let err = anyhow::anyhow!("Could not submit order because extending/reducing the position is not part of the MVP scope");
+        order_failed(Some(order.id), FailureReason::OrderNotAcceptable, &err)?;
+        return Err(err);
     }
 
     db::insert_order(order)?;
@@ -35,6 +38,7 @@ pub async fn submit_order(order: Order) -> Result<()> {
     }
 
     update_order_state_in_db_and_ui(order.id, OrderState::Open)?;
+    position::handler::update_position_after_order_submitted(order)?;
 
     Ok(())
 }
@@ -45,7 +49,7 @@ pub(crate) fn order_filling(order_id: Uuid, execution_price: f32) -> Result<()> 
 
     if let Err(e) = update_order_state_in_db_and_ui(order_id, state) {
         let e_string = format!("{e:#}");
-        match order_failed(Some(order_id), FailureReason::FailedToSetToFilling, e) {
+        match order_failed(Some(order_id), FailureReason::FailedToSetToFilling, &e) {
             Ok(()) => {
                 tracing::debug!(%order_id, "Set order to failed, after failing to set it to filling");
             }
@@ -81,7 +85,7 @@ pub(crate) fn order_filled() -> Result<Order> {
 pub(crate) fn order_failed(
     order_id: Option<Uuid>,
     reason: FailureReason,
-    error: anyhow::Error,
+    error: &anyhow::Error,
 ) -> Result<()> {
     tracing::error!("Failed to execute trade for order {order_id:?}: {reason:?}: {error:#}");
 
--- a/mobile/native/src/trade/position/handler.rs
+++ b/mobile/native/src/trade/position/handler.rs
@@ -50,7 +50,7 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
         .context("Could not update order to filling")?;
 
     if let Err((reason, e)) = ln_dlc::trade(trade_params).await {
-        order::handler::order_failed(Some(order.id), reason, e)
+        order::handler::order_failed(Some(order.id), reason, &e)
             .context("Could not set order to failed")?;
     }
 
--- a/mobile/native/src/trade/position/handler.rs
+++ b/mobile/native/src/trade/position/handler.rs
@@ -62,25 +62,29 @@ pub async fn get_positions() -> Result<Vec<Position>> {
     db::get_positions()
 }
 
+pub fn is_order_acceptable_for_position(
+    position: &Position,
+    submitted_order: &Order,
+) -> Result<bool> {
+    Ok(position.direction == submitted_order.direction.opposite()
+        && position.quantity == submitted_order.quantity)
+}
+
 /// Update the position once an order was submitted
 ///
 /// If the new order submitted is an order that closes the current position, then the position will
 /// be updated to `Closing` state.
 pub fn update_position_after_order_submitted(submitted_order: Order) -> Result<()> {
-    if let Some(position) = db::get_positions()?.first() {
-        // closing the position
-        if position.direction == submitted_order.direction.opposite()
-            && position.quantity == submitted_order.quantity
-        {
-            db::update_position_state(position.contract_symbol, PositionState::Closing)?;
-            let mut position = position.clone();
-            position.position_state = PositionState::Closing;
-            event::publish(&EventInternal::PositionUpdateNotification(position));
-        } else {
-            bail!("Currently not possible to extend or reduce a position, you can only close the position with a counter-order");
-        }
-    }
+    let position = get_current_position()?;
 
+    if is_order_acceptable_for_position(&position, &submitted_order)? {
+        db::update_position_state(position.contract_symbol, PositionState::Closing)?;
+        let mut position = position;
+        position.position_state = PositionState::Closing;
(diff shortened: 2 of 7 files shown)


mobile/native/src/trade/order/handler.rs
```diff
@@ -17,9 +17,12 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
-        order_failed(Some(order.id), FailureReason::OrderNotAcceptable, e)?;
-        bail!("Could not submit order because extending/reducing the position is not part of the MVP scope");
+    let position = position::handler::get_current_position()?;
+
+    if !position::handler::is_order_acceptable_for_position(&position, &order)? {
+        let err = anyhow::anyhow!("Could not submit order because extending/reducing the position is not part of the MVP scope");
```
luckysori, line 23: :upside_down_face: Prefer `anyhow!` to `anyhow::anyhow!`.

luckysori (approved):
LGTM, thanks!

mobile/native/src/trade/order/handler.rs
```diff
@@ -17,9 +17,12 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
-        order_failed(Some(order.id), FailureReason::OrderNotAcceptable, e)?;
-        bail!("Could not submit order because extending/reducing the position is not part of the MVP scope");
+    let position = position::handler::get_current_position()?;
+
+    if !position::handler::is_order_acceptable_for_position(&position, &order)? {
```
holzeis, line 22: 🤔 We are handling that scenario also on the coordinator. Handling this in the app definitely makes sense, but will also mean that we will not see this trade attempt in our unsuccessful trades statistics. So maybe its better to not check on the app side? as this should anyways not happen?

mobile/native/src/trade/position/handler.rs
```diff
@@ -206,3 +210,14 @@ pub async fn close_position() -> Result<()> {
     db::insert_order(order)?;
     update_position_after_dlc_closure(order)
 }
+
+/// Get the current position
+///
+/// FIXME: We are assuming that there is only one position open. This will
+/// change when we support multiple trading pairs.
+pub fn get_current_position() -> Result<Position> {
+    Ok(db::get_positions()?
```
holzeis, line 218: 🔧 It is very likely that we don't have position. e.g. First startup. I suggest we return an Option `None` if no position was found, instead of failing.

mobile/native/src/trade/order/handler.rs
```diff
@@ -17,9 +17,12 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
-        order_failed(Some(order.id), FailureReason::OrderNotAcceptable, e)?;
-        bail!("Could not submit order because extending/reducing the position is not part of the MVP scope");
+    let position = position::handler::get_current_position()?;
```
holzeis, line 20: 🤔 Can we be sure that there is always a current position? I think we are still deleting a position from the app once it gets closed, don't we? If yes, that will fail whenever we are opening a position?

mobile/native/src/trade/order/handler.rs
```diff
@@ -35,6 +38,7 @@ pub async fn submit_order(order: Order) -> Result<()> {
     }
 
     update_order_state_in_db_and_ui(order.id, OrderState::Open)?;
+    position::handler::update_position_after_order_submitted(order)?;
```
holzeis, line 41: 🤔 Since our close button state is dependent on the Closing state of the position, isn't it risky to wait so long to set the state of the position to `Closing`. e.g. It sometimes takes quite some time for the `post_order` http call to return. In that that time the user could click close position again? leading to other potential issues?

holzeis (commented):
I don't think this will resolve #558 - or at least not all scenarios, where the position is stuck in closing.

The issue in the ticket was due to a lost connection while closing he position. This fix will not recover the position from that closing state. However, I don't think we should set the position state back to `Open` in that case either, because the dlc is truly stuck in some closing state. 

In order to fix that properly we need the fix from tibo. see https://matrix.to/#/!YOUZSBmuJuaGzmxzaA:matrix.org/$DH8KXYw_90J8QZDzohruFEv1y3IALBNmyNMhn30IgXw?via=matrix.org

mobile/native/src/trade/position/handler.rs
klochowicz: this is just a refactor (it was exactly like this in the code already), and I had to stop somewhere. follow-up PRs welcome.

klochowicz:
> I don't think this will resolve #558 - or at least not all scenarios, where the position is stuck in closing.
> 
> The issue in the ticket was due to a lost connection while closing he position. This fix will not recover the position from that closing state. However, I don't think we should set the position state back to `Open` in that case either, because the dlc is truly stuck in some closing state.
> 
> In order to fix that properly we need the fix from tibo. see https://matrix.to/#/!YOUZSBmuJuaGzmxzaA:matrix.org/$DH8KXYw_90J8QZDzohruFEv1y3IALBNmyNMhn30IgXw?via=matrix.org

This PR will deal exactly with the issue you mention - it will *prevent* them, because the HTTP request will fail, thus not allowing to close a position. Preemptive update without recovery was not a good idea.

It is true that it will not fix ones that got already stuck, as mentioned in the ticket at hand - I have created another ticket for that and started #568 - happy to collaborate and discuss the best solutions there!

holzeis:
> This PR will deal exactly with the issue you mention - it will prevent them, because the HTTP request will fail, thus not allowing to close a position. Preemptive update without recovery was not a good idea.

There are still scenarios where the position can get stuck even if the HTTP request was successful.

mobile/native/src/trade/position/handler.rs
holzeis: well isn't this now behaving differently? I might be reading it wrong, but wouldn't this logic now return an error if the position is not found?

Before your refactoring we were doing something like that.

```
if let Some(position) = db::get_positions()?.first() {
```

mobile/native/src/trade/order/handler.rs
holzeis: was it? in the previous call we were handling if the was a current position. This is simply failing if not position is found. Or am I reading this wrong?

```
if let Some(position) = db::get_positions()?.first() {
```

klochowicz:
> > This PR will deal exactly with the issue you mention - it will prevent them, because the HTTP request will fail, thus not allowing to close a position. Preemptive update without recovery was not a good idea.
> 
> There are still scenarios where the position can get stuck even if the HTTP request was successful. Like the scenario reported in #558.

OK. Shall I remove 'closing' from the PR? Happy to do so.
I acknowledge that this doesn't fix *all* the things that can go wrong, but it gets rid of one source of errors.

klochowicz:
> > OK. Shall I remove 'closing' from the PR? Happy to do so. I acknowledge that this doesn't fix _all_ the things that can go wrong, but it gets rid of one source of errors.
> 
> Probably just remove that this PR will fix #558.

I can't unlink it 😥 I tried removing description, but that ticket is grayed out. I'll make sure to reopen it if this gets in.

mobile/native/src/trade/order/handler.rs
da-kami: If we are unable to submit the order to the orderbook we should set it to `Failed`. 
This logic has been removed now - I don't think this is intended.

mobile/native/src/trade/position/handler.rs
```diff
@@ -62,25 +62,29 @@ pub async fn get_positions() -> Result<Vec<Position>> {
...
-        {
-            db::update_position_state(position.contract_symbol, PositionState::Closing)?;
-            let mut position = position.clone();
-            position.position_state = PositionState::Closing;
-            event::publish(&EventInternal::PositionUpdateNotification(position));
-        } else {
-            bail!("Currently not possible to extend or reduce a position, you can only close the position with a counter-order");
-        }
-    }
+    let position = get_current_position()?;
```
da-kami, line 78: 🛠️ This call should not result in an error if there is no position! - if can happen that we don't have a position - so it should not result in an error here.

mobile/native/src/trade/position/handler.rs
```diff
@@ -62,25 +62,29 @@ pub async fn get_positions() -> Result<Vec<Position>> {
...
 
+    if is_order_acceptable_for_position(&position, &submitted_order)? {
```
da-kami, line 80: 🛠️ Calling `is_order_acceptable_for_position` in `update_position_after_order_submitted` does not make sense for me.  I don't think we should have submitted the order if we have a position *and* the order is not acceptable for the position.

Maybe there is a problem in the original code where we submit an order into the orderbook even though it is not allowed because we have a position and the order does not fit? 
I'm happy to investigate this.

da-kami (changes requested):

da-kami:
Thank you for pointing out the problems :)

I think what we have to do it:

1. Check if we are having a position and we are allowed to create this order (only allowed if closing the position). 
    1. If have a position and we are not closing it with the order we should fail the order and not update the position (it remains open). in this case we exit early
2. Submit the order to the orderbook
3. If submitted successfully we updated order and position as needed

Issue #543, opened 2023-04-28
Collaboratively closing the ln channel lost the apps funds


holzeis:
After ending up in a stuck state on my 10101 app (mainnet) Build 1.0.12 (985) I initiated a collaborative channel close through the coordinator api.

```
curl -X DELETE http://localhost:8000/api/admin/channels/f0bc467b35a4b6a72d0107d3a03556af4cb2aba3780cc7398c26b9bc9d42fc80
```

Both sides reported a successful channel closure and the commit transaction has been published https://mempool.space/tx/b5b3efb670887bf0dfa30f708835e5c3555796231d392da28dfe5a82cd4bc574 and confirmed.

Logs from the app
```
| 2023-04-28 08:33:50.451 | LogLevel.DEBUG | lightning::ln::peer_handler: Handling SendClosingSigned event in peer_handler for node 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156 for channel f0bc467b35a4b6a72d0107d3a03556af4cb2aba3780cc7398c26b9bc9d42fc80  
| 2023-04-28 08:33:50.658 | LogLevel.INFO | lightning::ln::channelmanager: Broadcasting closing tx with txid b5b3efb670887bf0dfa30f708835e5c3555796231d392da28dfe5a82cd4bc574  
| 2023-04-28 08:33:50.705 | LogLevel.INFO | Received event event: ChannelClosed { channel_id: [240, 188, 70, 123, 53, 164, 182, 167, 45, 1, 7, 211, 160, 53, 86, 175, 76, 178, 171, 163, 120, 12, 199, 57, 140, 38, 185, 188, 157, 66, 252, 128], user_channel_id: 0, reason: CooperativeClosure } 
| 2023-04-28 08:33:50.706 | LogLevel.INFO | 
Channel closed channel: f0bc467b35a4b6a72d0107d3a03556af4cb2aba3780cc7398c26b9bc9d42fc80,reason: CooperativeClosure 
| 2023-04-28 08:33:50.707 | LogLevel.DEBUG | Successfully handled event event: "ChannelClosed { channel_id: [240, 188, 70, 123, 53, 164, 182, 167, 45, 1, 7, 211, 160, 53, 86, 175, 76, 178, 171, 163, 120, 12, 199, 57, 140, 38, 185, 188, 157, 66, 252, 128], user_channel_id: 0, reason: CooperativeClosure }"
```

[...]

PR #563, opened 2023-05-01
WIP: chore: Add test to close channel with dlc

holzeis:
The test is reproducing the bug reported in #543.

PR #569, opened 2023-05-02
Use on-chain address for shutdown and destination

luckysori:
#### Collaborative close of LN channel

Before, the collab close transaction paid to an address owned by LDK, which had to be confirmed and then spent to the on-chain wallet.

After overriding `get_shutdown_scriptpubkey`, the close transaction pays directly to the on-chain wallet. This means that there is no risk of losing funds as soon as the close transaction is on-chain if we remember the seed, whereas before we could lose funds whilst waiting for the "claim transaction".

#### Force-close of LN channel

The effect of overriding `get_destination_script` to point to an on-chain wallet address is less clear to me, because the commitment transaction must still be processed by LDK to produce a `SpendableOutputs` event with a `SpendableOutputDescriptor::DelayedPaymentOutput` which can then be spent to an on-chain wallet address (after waiting for the corresponding delay).

#### Other changes

The tests were fixed to pass after the new changes were introduced. At the same time we decided to refactor them with the goal of making them clearer and simpler. These two tests diverged enough that it made sense to remove the common helper function.

#### Caveat

The trait method implementations modified include a few `expect`s because the trait methods are defined as infallible. We copied the code from `ldk-node`, so at least that suggests that they are okay with these `expect`s.

### DIFF
--- a/crates/ln-dlc-node/src/dlc_custom_signer.rs
+++ b/crates/ln-dlc-node/src/dlc_custom_signer.rs
@@ -1,6 +1,7 @@
 //! This file has temporarily been copied from `https://github.com/p2pderivatives/rust-dlc/pull/97`.
 //! We should reimplement some of these traits for production.
 
+use crate::ln_dlc_wallet::LnDlcWallet;
 use anyhow::anyhow;
 use anyhow::Result;
 use bitcoin::Script;
@@ -250,11 +251,15 @@ impl Sign for CustomSigner {}
 
 pub struct CustomKeysManager {
     keys_manager: KeysManager,
+    wallet: Arc<LnDlcWallet>,
 }
 
 impl CustomKeysManager {
-    pub fn new(keys_manager: KeysManager) -> Self {
-        Self { keys_manager }
+    pub fn new(keys_manager: KeysManager, wallet: Arc<LnDlcWallet>) -> Self {
+        Self {
+            keys_manager,
+            wallet,
+        }
     }
 }
 
@@ -292,11 +297,25 @@ impl KeysInterface for CustomKeysManager {
     }
 
     fn get_destination_script(&self) -> Script {
-        self.keys_manager.get_destination_script()
+        let address = self
+            .wallet
+            .get_last_unused_address()
+            .expect("Failed to retrieve new address from wallet.");
+        address.script_pubkey()
     }
 
     fn get_shutdown_scriptpubkey(&self) -> ShutdownScript {
-        self.keys_manager.get_shutdown_scriptpubkey()
+        let address = self
+            .wallet
+            .get_last_unused_address()
+            .expect("Failed to retrieve new address from wallet.");
+        match address.payload {
+            bitcoin::util::address::Payload::WitnessProgram { version, program } => {
+                ShutdownScript::new_witness_program(version, &program)
+                    .expect("Invalid shutdown script.")
+            }
+            _ => panic!("Tried to use a non-witness address. This must not ever happen."),
+        }
     }
 
     fn get_secure_random_bytes(&self) -> [u8; 32] {
--- a/crates/ln-dlc-node/src/ln/config.rs
+++ b/crates/ln-dlc-node/src/ln/config.rs
@@ -14,7 +14,8 @@ pub fn app_config() -> UserConfig {
             minimum_depth: 1,
             // There is no risk in the leaf channel to receive 100% of the channel capacity.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
-            // the delay before the fund of a force closed channel can be claimed again
+            // We want the coordinator to recover force-close funds as soon as possible. We choose
+            // 144 because we can't go any lower according to LDK.
             our_to_self_delay: 144,
             ..Default::default()
         },
@@ -23,8 +24,9 @@ pub fn app_config() -> UserConfig {
             trust_own_funding_0conf: true,
             // Enforces that incoming channels will be private.
             force_announced_channel_preference: true,
-            // lnd's max to_self_delay is 2016, so we want to be compatible.
-            their_to_self_delay: 2016,
+            // We want app users to only have to wait ~24 hours in case of a force-close. We choose
+            // 144 because we can't go any lower according to LDK.
+            their_to_self_delay: 144,
             ..Default::default()
         },
         channel_config: ChannelConfig {
@@ -52,7 +54,8 @@ pub fn coordinator_config() -> UserConfig {
             minimum_depth: 1,
             // We set this 100% as the coordinator is online 24/7 and can take the risk.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
-            // the delay before the fund of a force closed channel can be claimed again
+            // Our channel peers are allowed to get back their funds ~24 hours after a
+            // force-closure.
             our_to_self_delay: 144,
             ..Default::default()
         },
@@ -64,7 +67,7 @@ pub fn coordinator_config() -> UserConfig {
             // Enforces incoming channels to the coordinator to be public! We
             // only want to open private channels to our 10101 app.
             force_announced_channel_preference: true,
-            // lnd's max to_self_delay is 2016, so we want to be compatible.
+            // LND's max to_self_delay is 2016, so we want to be compatible.
             their_to_self_delay: 2016,
             ..Default::default()
         },
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -20,16 +20,17 @@ use anyhow::Context;
 use anyhow::Result;
 use bitcoin::consensus::encode::serialize_hex;
 use bitcoin::secp256k1::PublicKey;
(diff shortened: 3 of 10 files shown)


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -364,21 +356,8 @@ where
...
-                let spending_tx = self.keys_manager.spend_spendable_outputs(
-                    output_descriptors,
-                    Vec::new(),
-                    destination_address.script_pubkey(),
-                    tx_feerate,
-                    &Secp256k1::new(),
-                )?;
-                self.wallet.broadcast_transaction(&spending_tx);
+            Event::SpendableOutputs { .. } => {
+                // Spendable outputs are sent to the node's on-chain wallet directly
```
bonomat, line 360: You could log here that this should never happen :)

bonomat (approved):
Looks good. Nice and easy fix :)

crates/ln-dlc-node/src/ln/event_handler.rs
luckysori: Not quite, LDK still generates the event regardless of how we configure the channel or how we generate the final addresses.

bonomat: oh right. Thanks for clarification

crates/ln-dlc-node/src/dlc_custom_signer.rs
```diff
@@ -292,11 +297,25 @@ impl KeysInterface for CustomKeysManager {
...
+            .get_new_address()
+            .expect("Failed to retrieve new address from wallet.");
+        address.script_pubkey()
     }
 
     fn get_shutdown_scriptpubkey(&self) -> ShutdownScript {
-        self.keys_manager.get_shutdown_scriptpubkey()
+        let address = self
+            .wallet
+            .get_new_address()
```
holzeis, line 310: 🤔 should we use last unused address here?

crates/ln-dlc-node/src/ln/config.rs
```diff
@@ -14,6 +14,7 @@ pub fn app_config() -> UserConfig {
             minimum_depth: 1,
             // There is no risk in the leaf channel to receive 100% of the channel capacity.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
+            commit_upfront_shutdown_pubkey: false,
```
holzeis, line 17:
```suggestion
            // todo: check why we have to set this to false for our test setup to work
            commit_upfront_shutdown_pubkey: false,
```

crates/ln-dlc-node/src/ln/config.rs
```diff
@@ -50,6 +51,7 @@ pub fn coordinator_config() -> UserConfig {
             minimum_depth: 1,
             // We set this 100% as the coordinator is online 24/7 and can take the risk.
             max_inbound_htlc_value_in_flight_percent_of_channel: 100,
+            commit_upfront_shutdown_pubkey: false,
```
holzeis, line 54:
```suggestion
            // todo: check why we have to set this to false for our test setup to work
            commit_upfront_shutdown_pubkey: false,
```

crates/ln-dlc-node/src/ln_dlc_wallet.rs
```diff
@@ -91,6 +91,12 @@ impl LnDlcWallet {
 
         Ok(txs)
     }
+
+    pub fn get_new_address(&self) -> Result<Address> {
```
holzeis, line 95: 🔧 It's a bit confusing that the method is named `get_new_address` when it actually returns an unused address.

holzeis (approved):
Nice 👍 

I do wonder though why the funds did not endup in my wallet with the `p2wsh` script. After 6 blocks the `Event::SpendableOutputs` should have been thrown and the claim transaction should have been publised. After 1 confirmation I should have seen the funds in my wallet.

This change is definitely an improvement, as it makes it easier to get your funds back, but I am not sure how it fixes #543. Note, I've waited for 8 confirmations, before deleting the 10101 wallet. It might have been a failed broadcast transaction, after 6 blocks, that did not get retried? 🤔

holzeis:
> I was able to get my funds back after collab close of the LN channel via the settings page. I followed https://github.com/get10101/10101/issues/543#issuecomment-1527328044.

I just followed the same steps testing your fix, mined several block and it did not work for me. I tested on the iOS Simulator (if that makes any difference). The funds are still not back to my onchain wallet.

However, I wasn't able to reproduce that behavior in the tests.. which might indicate that its only a phone issue? https://github.com/get10101/10101/pull/563

crates/ln-dlc-node/src/dlc_custom_signer.rs
bonomat: Yes please, let's use `last_unused` here. Otherwise we get gaps in the derivation paths.

luckysori: It's getting a `last_unused` under the hood :D

luckysori:
As @holzeis correctly pointed out, this does not fix #543. This patch focuses on ensuring that the funds are safe in case there is a node restart between closing or force-closing an LN channel and getting enough confirmations to trigger the `SpendableOutputs` event. This was the theme of https://github.com/get10101/10101/issues/543#issuecomment-1529680489, which is what I focused on.

When trying to apply the same fix to a collaborative closure of the _LN-DLC_ channel, we realise that directly closing the LN channel cannot possibly work, because LDK has no insight on the state of the DLC channel. As such, we need to first settle the DLC channel collaboratively before triggering the LN channel collaborative close. Therefore this PR is currently irrelevant for that.

crates/ln-dlc-node/src/ln/config.rs
luckysori: It's not needed actually!

luckysori:
I've reworked this PR considerably after rebasing on top of your tests, @holzeis. So please take another look.

crates/ln-dlc-node/src/ln/config.rs
```diff
@@ -64,7 +67,7 @@ pub fn coordinator_config() -> UserConfig {
             // Enforces incoming channels to the coordinator to be public! We
             // only want to open private channels to our 10101 app.
             force_announced_channel_preference: true,
-            // lnd's max to_self_delay is 2016, so we want to be compatible.
+            // LND's max to_self_delay is 2016, so we want to be compatible.
```
holzeis, line 70: 🔧 I think we also don't want to accept channels with `2016` blocks delay opened to us. It might be safe to reduce that value to less? but not exceed 2016. We could easily test that with our faucet setup.

crates/ln-dlc-node/src/ln/config.rs
```diff
@@ -23,8 +24,9 @@ pub fn app_config() -> UserConfig {
             trust_own_funding_0conf: true,
             // Enforces that incoming channels will be private.
             force_announced_channel_preference: true,
-            // lnd's max to_self_delay is 2016, so we want to be compatible.
-            their_to_self_delay: 2016,
+            // We want app users to only have to wait ~24 hours in case of a force-close. We choose
+            // 144 because we can't go any lower according to LDK.
+            their_to_self_delay: 144,
```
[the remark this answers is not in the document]
luckysori: LDK won't let you go any lower. See this discussion: https://github.com/orgs/lightningdevkit/discussions/2260.

crates/ln-dlc-node/src/ln/config.rs
luckysori: Yeah, we might want to play around with this number, although I didn't do so with this PR.

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
```diff
@@ -55,19 +50,53 @@ async fn collab_close() {
...
+    payee.sync().unwrap();
+
+    // Assert
+
+    let ln_balance = payee.get_ldk_balance();
+    assert_eq!(ln_balance.available, 0);
+    assert_eq!(ln_balance.pending_close, 0);
+
+    assert_eq!(
+        payee.get_on_chain_balance().unwrap().confirmed,
```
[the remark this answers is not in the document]
luckysori: I don't think so. I think it just means that there is one less transaction involved when getting the coins back into the wallet, making it cheaper overall.

This test should certainly assert on the coordinator funds too, but I just kept it the same way it was before for now.

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
```diff
@@ -107,87 +136,59 @@ async fn force_close() {
...
+    assert_eq!(payee.get_ldk_balance().available, 0);
+    assert_eq!(payee.get_ldk_balance().pending_close, invoice_amount);
+
+    // Mine enough blocks so that the payee's revocable output in the commitment transaction
+    // is spendable
+    bitcoind::mine(
+        payee
+            .user_config
+            .channel_handshake_limits
+            .their_to_self_delay,
```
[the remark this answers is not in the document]
luckysori: Fixed it up onto the patch.

PR #578, opened 2023-05-02
Bump rust-dlc

klochowicz:
This update attempts to deal with losing connection between parties during
closing of the DLC.


klochowicz:
This hopefully deals with #558 - or at least reduces the frequency.

During @holzeis testing it turned out that it sometimes still fails, but it is still an improvement.

I propose to merge it in and test it in the new release in prod.

luckysori (approved):

Issue #577, opened 2023-05-02
JIT - Channel Fee


bonomat:
We know that our on-boarding procedure causes friction because the user has to make changes in their wallet they are sending from (i.e. change the max acceptable routing fee). 
The reason for this is that we are charging a 2% routing fee. This fee is hardcoded in the invoice created by the app. 

We don't want to offer channel opening for free because (a) it is meant to be a business model in the future and (b) we have on-chain expenses which we need to cover. We don't need to be profitable from day 1 but we should charge something from day 1.

I see three steps forward: 

- 1. Reduce the channel routing fee 0.5% (for now to) which seems to be acceptable in most apps. #579 
- 2. Use different flow: (1) LSP creates channel, (2) node pays invoice. This is based on trust but given we control both sides (coordinator and app) we can just send a payment from the node to the coordinator up-on channel creation. We can then increase the channel opening fee again to whatever amount we want. Payment routing will not be affected by this anymore. 
- https://github.com/get10101/10101/issues/843

I propose to tackle this in 3 steps: 

1. do 1. now
2. work on 2 next - imho this is a great approach as the fee will be shown as a payment in the app
3. work on 3 once the spec are finalized: see issue https://github.com/get10101/10101/issues/843

I see 1. as a quick fix and not a solution so I would like to avoid a prolonged time of having 1. done but not 2.

klochowicz:
@bonomat do we need this for the MVP, ie. next 2 weeks? or can I remove the label for this issue?

bonomat:
> @bonomat do we need this for the MVP, ie. next 2 weeks? or can I remove the label for this issue?

Yes, i think at least point 2 should be done. 
We might need to do https://github.com/get10101/10101/issues/604 as well.

bonomat:
Ticket is up for grabs, I don't have the capacity to work on this at the moment :(

PR #579, opened 2023-05-03
Set JIT funding rate to 0.5% and make it configurable from the coordinator

bonomat:
So far we had it at 2% which was a hurdle for users. 
Pushing it down to 0.5% will be expensive for us in the long run, hence, this is just a temporarily solution until we go for point 2 in #577 .

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -23,6 +23,7 @@ use diesel::r2d2::ConnectionManager;
 use diesel::r2d2::Pool;
 use diesel::PgConnection;
 use ln_dlc_node::node::NodeInfo;
+use orderbook_commons::FakeScidResponse;
 use orderbook_commons::OrderbookMsg;
 
 use crate::admin::close_channel;
@@ -63,6 +64,10 @@ pub fn router(node: Node, pool: Pool<ConnectionManager<PgConnection>>) -> Router
     Router::new()
         .route("/", get(index))
         .route("/api/fake_scid/:target_node", post(post_fake_scid))
+        .route(
+            "/api/fake_scid/v2/:target_node",
+            post(post_fake_scid_with_fee_hint),
+        )
         .route("/api/newaddress", get(get_new_address))
         .route("/api/node", get(get_node_info))
         .route("/api/invoice", get(get_invoice))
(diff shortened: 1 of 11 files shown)


crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
```diff
@@ -95,11 +100,21 @@ pub(crate) async fn send_interceptable_payment(
 
     // Act
 
-    let intercept_scid = coordinator.create_intercept_scid(payee.info.pubkey);
+    let (intercept_scid, fee_millionth) = coordinator.create_intercept_scid(payee.info.pubkey);
 
     let flat_routing_fee = 1; // according to the default `ChannelConfig`
-    let liquidity_routing_fee =
-        (invoice_amount * LIQUIDITY_ROUTING_FEE_MILLIONTHS as u64) / 1_000_000;
+    let liquidity_rounding_fee = Decimal::from_u64(invoice_amount)
```
luckysori, line 106: :upside_down_face: Doesn't `Decimal::from` work without an `unwrap` for `u64`?

luckysori: :wrench: Should be `liquidity_routing_fee`.

coordinator/src/routes.rs
```diff
@@ -112,9 +117,26 @@ pub async fn post_fake_scid(
...
     })?;
 
-    Ok(Json(
-        app_state.node.inner.create_intercept_scid(target_node),
-    ))
+    let (fscid, _) = app_state.node.inner.create_intercept_scid(target_node);
+    Ok(Json(fscid))
+}
+
+pub async fn post_fake_scid_with_fee_hint(
```
luckysori, line 124: :wrench: I would either split these two, so that you can just introduce a new API that only fetches the fee hint, or rename the API and the route to reflect that it's not just the fake SCID anymore.

We could go for `register_interceptable_invoice` or something like that to connect it to the fact that the call is used to inform the coordinator that we are gonna route an interceptable payment through it (with the fake SCID and fee that it gives us).

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -116,6 +116,10 @@ pub struct Node<P> {
     pub(crate) alias: [u8; 32],
     #[cfg(test)]
     pub(crate) announcement_addresses: Vec<NetAddress>,
+
+    /// Liquidity-based routing fee in millionths of a routed amount. In
+    /// other words, 10000 is 1% and 5_000 is 0.5%
+    jit_funding_rate_millionth: Option<u32>,
```
luckysori, line 122: :wrench: Do we need this to be an `Option`? I think we can just make it mandatory.

Even if `new_app` doesn't care about it for now (because it doesn't make any sense since the app will not route payments anyway), it's not worth introducing an annoying `Option`. 

We just need to split the app and coordinator code at some point, which is the proper fix.

luckysori (approved):

PR #582, opened 2023-05-03
Run `ln-dlc-node` tests sequentially on CI

luckysori:
I think we can remove the `-j 1` flag, because I introduced it to achieve the same thing and it clearly wasn't working.

We must go down to `--test-threads=1` if we want to ensure that no two tests are run in parallel. Running tests in parallel at all seems to pretty consistently cause either test failures or non-terminating tests. Perhaps this change might fix
https://github.com/get10101/10101/issues/355.

---

This change will obviously slow down CI, but I don't have any better ideas at this stage.


klochowicz:
> Build failed:
> 
> * [tests-ln-dlc-node](https://github.com/get10101/10101/actions/runs/4874175062/jobs/8695369772)

doesn't seem to have fixed the issue on CI :( still worth merging though.

klochowicz:
bors retry

PR #585, opened 2023-05-03
Fix/only update position if order succeeded

klochowicz:
Improved version of #567 . Made sure I did the least amount of changes needed, and addressed issues raised by @da-kami and @holzeis in previous review.

### DIFF
--- a/mobile/native/src/trade/order/handler.rs
+++ b/mobile/native/src/trade/order/handler.rs
@@ -7,6 +7,7 @@ use crate::trade::order::FailureReason;
 use crate::trade::order::Order;
 use crate::trade::order::OrderState;
 use crate::trade::position;
+use crate::trade::position::handler::update_position_after_order_submitted;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
@@ -17,7 +18,7 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
+    if let Err(e) = position::handler::get_position_matching_order(&order) {
         order_failed(Some(order.id), FailureReason::OrderNotAcceptable, e)?;
         bail!("Could not submit order because extending/reducing the position is not part of the MVP scope");
     }
@@ -35,6 +36,7 @@ pub async fn submit_order(order: Order) -> Result<()> {
     }
 
     update_order_state_in_db_and_ui(order.id, OrderState::Open)?;
+    update_position_after_order_submitted(&order)?;
 
     Ok(())
 }
--- a/mobile/native/src/trade/position/handler.rs
+++ b/mobile/native/src/trade/position/handler.rs
@@ -9,7 +9,6 @@ use crate::trade::order::OrderState;
 use crate::trade::order::OrderType;
 use crate::trade::position::Position;
 use crate::trade::position::PositionState;
-use anyhow::bail;
 use anyhow::ensure;
 use anyhow::Context;
 use anyhow::Result;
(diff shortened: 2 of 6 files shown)


mobile/native/src/trade/position/handler.rs
```diff
@@ -66,24 +66,28 @@ pub fn get_positions() -> Result<Vec<Position>> {
...
-
     Ok(())
 }
 
+/// Returns the position that would be closed by submitted, if there is any
+pub fn get_position_matching_order(order: &Order) -> Result<Option<Position>> {
+    Ok(if let Some(position) = db::get_positions()?.first() {
+        // closing the position
+        ensure!(position.direction == order.direction.opposite()
+                && position.quantity == order.quantity, "Currently not possible to extend or reduce a position, you can only close the position with a counter-order");
```
luckysori, line 84: I understand that we want to check this, but for me it's unexpected that the check is happening implicitly when calling `get_position_matching_order`.

I think it would be clearer if we extracted this into a separate function `does_order_close_position` and just call it if we do have an order in the database. It's two calls rather than one, but the current behaviour is surprising to me.

mobile/native/src/trade/order/handler.rs
```diff
@@ -17,7 +18,7 @@ pub async fn submit_order(order: Order) -> Result<()> {
     let url = format!("http://{}", config::get_http_endpoint());
     let orderbook_client = OrderbookClient::new(Url::parse(&url)?);
 
-    if let Err(e) = position::handler::update_position_after_order_submitted(order) {
+    if let Err(e) = position::handler::get_position_matching_order(&order) {
```
da-kami, line 21:
```suggestion
    if let Err(e) = position::handler::does_order_close_position(&order) {
```

I would love to have a name that includes the purpose of this call - "check if there is a position and the order is closing it". If you don't think that helps feel free to ignore this comment.

da-kami, line 21: I remembered another thing 😅

It was on purpose that we updated the position immediately after submitting the order in case we have a closing scenarios - this update is basically what sets the button state to `Closing` in the UI so the user does not click it again. 

I think the best way would be updating the position to `Closing` immediately in case we are in a closing scenarios, but resetting it to open in case we fail to submit the order. I think resetting the state to open was missing, not necessarily moving the update to a later point.

Sorry, I did not think about this detail yesterday.

mobile/native/src/trade/position/handler.rs
luckysori: Fair enough. I guess now that you extracted this function it just feels weird that this _new_ function can error like this, whereas originally it didn't bother me. It's a minor detail, specially since as you say it was already effectively the same before your changes.

da-kami: I agree with @luckysori that `does_order_close_position` is a better name than `get_position_matching_order` - the function was newly introduced. I think it'd be great to rename it.

luckysori (approved):

da-kami (approved):

mobile/native/src/trade/order/handler.rs
da-kami: Best would be to change the state in the UI to set it to `Closing` upon clock until we get an update from the backend. But that might be a bit tricky as well :)

klochowicz: IMHO the problem was there beforehand, and I just uncovered it.

klochowicz: I perceive the risk from this being small:  the action is swiping not just clicking and we get the success dialog in between.

(e.g. one can't fat-finger twice accidentally the same way as with buttons).

This is a shortcoming though, and I'll work to solve it!

mobile/native/src/trade/position/handler.rs
klochowicz: @da-kami : @luckysori didn't want to rename this, we wanted to extract one more function, probably with `bool` return. I opted to not do this for now, out of fear of last-minute in-adverted changes to the logic.

klochowicz: Let's fix that up when we're removing `PositionState` from being stored in the db.

PR #596, opened 2023-05-04
fix: Do not use `getApplicationDocumentsDirectory`

klochowicz:
On desktop platforms, using `getApplicationDocumentsDirectory` caused issues:
- placed the contents of DB straight into ~/Documents directory (not even in
10101 subdir)
- this in turn broke `just wipe` behaviour.

As it impedes our testing, I have decided to bringe back previous behaviour for
the time being - we can get back to it when resolution to this problem is
found; it's not our highest priority atm though.

The DB will be placed in the same spot as the seed is.


holzeis (approved):
Thanks 👍

luckysori (approved):
Nice, makes sense!

PR #597, opened 2023-05-04
Add timeout to all HTTP requests

klochowicz:
Reqwest doesn't add request timeout by default.

### DIFF
--- a/maker/src/trading/orderbook_client.rs
+++ b/maker/src/trading/orderbook_client.rs
@@ -12,7 +12,10 @@ pub struct OrderbookClient {
 impl OrderbookClient {
     pub fn new() -> Self {
         Self {
-            client: reqwest::Client::new(),
+            client: reqwest::Client::builder()
+                .timeout(std::time::Duration::from_secs(30))
+                .build()
+                .expect("Could not build reqwest client"),
         }
     }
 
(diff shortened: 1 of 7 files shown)


maker/src/trading/orderbook_client.rs
```diff
@@ -12,7 +12,10 @@ pub struct OrderbookClient {
 impl OrderbookClient {
     pub fn new() -> Self {
         Self {
-            client: reqwest::Client::new(),
+            client: reqwest::Client::builder()
+                .timeout(std::time::Duration::from_secs(10))
```
holzeis, line 16: 🤔 is 10 seconds conservative enough? e.g. the generation of the fake scid takes sometimes more than 10 seconds. how about 30 seconds to be on the safe side?

holzeis (approved):
LGTM 👍

maker/src/trading/orderbook_client.rs
klochowicz: > 🤔 is 10 seconds conservative enough? e.g. I haven't stopped the time but the generation of the fake scid takes sometimes very long. how about 30 seconds to be on the safe side?

Sure, 10s was arbitrary! Shall we put 30s everywhere to be safe? or were you just talking about this place?

holzeis: Yes, let's go for 30 seconds to be safe. 😊

PR #598, opened 2023-05-04
chore: Use brand colours on trading screen

klochowicz:
Continue roll-out of our brand colours for added consistency.

Note: Info button uses a lighter shade of the same colour.

Screenshot 2023-05-04 at 4 41 28 pm


bonomat (approved):
LGTM

PR #603, opened 2023-05-05
feat: Keep backup button always visible

### LINKED ISSUE #602: Keep backup seed button visible at all times
A user reached out via twitter that he funded the wallet and checked the "I have made a backup of my seed" checkbox when creating a backup.

It seems he did not write the seed phrase down though and now can't create a backup anymore. I suggest we remove the "feature" of hiding the backup button once the backup has been confirmed once.

holzeis:
Feedback from a user, who confirmed the backup, but didn't create a backup. An easy quick fix that will help our users.

resolves #602

### DIFF
--- a/mobile/lib/util/preferences.dart
+++ b/mobile/lib/util/preferences.dart
@@ -14,8 +14,10 @@ class Preferences {
   }
 
   Future<bool> isUserSeedBackupConfirmed() async {
-    SharedPreferences preferences = await SharedPreferences.getInstance();
-    return preferences.getBool(userSeedBackupConfirmed) ?? false;
+    // FIXME: disabling the user seed backup confirmed preference so that the backup button is always visible. Eventually, we should think about how we want to make the seed backup accessible to the user at all times.
+    // SharedPreferences preferences = await SharedPreferences.getInstance();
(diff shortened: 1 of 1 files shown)


mobile/lib/util/preferences.dart
```diff
@@ -14,8 +14,10 @@ class Preferences {
   }
 
   Future<bool> isUserSeedBackupConfirmed() async {
-    SharedPreferences preferences = await SharedPreferences.getInstance();
-    return preferences.getBool(userSeedBackupConfirmed) ?? false;
+    // FIXME: disabling the user seed backup confirmed preference so that the backup button is always visible. Eventually, we should think about how we want to make the seed backup accessible to the user at all times.
+    // SharedPreferences preferences = await SharedPreferences.getInstance();
+    // return preferences.getBool(userSeedBackupConfirmed) ?? false;
+    return false;
```
klochowicz, line 20: instead of changing the logic, it might be more obvious to remove the conditional display from the widget itself

klochowicz (approved):
solves the problem for now

in longer term: maybe we could keep it always on in wallet settings?

PR #608, opened 2023-05-09
fix: use unified boolean logic

bonomat:
in most examples I could find the logic is like this. Let's see if this helps


holzeis (approved):

bonomat:
bors retry

Issue #591, opened 2023-05-04
Force-closing an LN-DLC channel fails


holzeis:
### Context

A force-closure of the channel can happen in 2 ways:

- _Automatically_: e.g. LDK will do so if the parties cannot agree on a channel fee rate.
- _Manually_: e.g. We instruct our coordinator to force-close all LN-DLC channels so that we can introduce breaking changes.

### Manual force-closure

We know how to handle the manual case. We just need to call the right APIs provided by `rust-dlc`. We must also write a test to verify this. 

- [x] Implement manual force-closure of LN-DLC channel (see https://github.com/get10101/10101/pull/612).

### Automatic force-closure

It appears to be more challenging to handle the automatic case, as it seems like LDK will eagerly publish the commitment transaction to force-close the channel in a lot of cases, even if the split transaction and the glue transaction are missing.

In our experience, when LDK fails on the commitment transaction publication it still removes the channel from the `ChannelManager`, so we're left in a state where it's hard (perhaps impossible) to recover the funds. In theory we should be able to recover the funds with the information stored in the `ChannelMonitor`, but we haven't verified this yet.

- [ ] #779

---

Related to #51, which focuses on the effect of force-closure of the LN-DLC channel on the DLC channel branch i.e. buffer transaction and CET/refund transaction.

holzeis:
blocks #580

holzeis:
It's fair to say that the manual force-closing is also incomplete, as only the coordinator is able to force-close. Furthermore, we only asserted that the coordinator gets their funds back. Tests are still missing for the app.

luckysori:
> It's fair to say that the manual force-closing is also incomplete, as only the coordinator is able to force-close. Furthermore, we only asserted that the coordinator gets their funds back. Tests / Assertions are still missing for the app.

This bit should be fixed with #771.

luckysori:
Closing as #771 has covered half of this and https://github.com/get10101/10101/issues/779 covers the rest.

PR #612, opened 2023-05-09
Implement force-close of LN-DLC channel

luckysori:
Fixes part of https://github.com/get10101/10101/issues/591.

This patch finally adds a `force_close_ln_dlc_channel` test, which verifies that the party effecting the force-close of the LN-DLC
channel can get their coins back if they:

- Manually call `force_close_channel`, publishing the split transaction.

- Wait for 288 confirmations on the split transaction.

- Manually call `finalize_force_close_ln_dlc_channel`, publishing the buffer, glue and commitment transactions.

- Wait for 144 confirmations on the commitment transaction, so that the delayed output on it becomes spendable.

- Wait for a further 144 confirmations, so that the CET can spend from the buffer transaction.

For the time being only the coordinator is able to force-close a channel by first using the DELETE /api/admin/channels/:channel_id
endpoint and later the new DELETE /api/admin/channels/finalize_force_close/:channel_id endpoint.

#### Limitations

- This work does not yet verify that the party who did not trigger the force-close gets all their coins back. Specifically we are aware that there is a bug with the `StaticPaymentOutput` on the commitment transaction not being spendable.

- The updates to the `rust-dlc` and `rust-lightning` dependencies represent very hacky changes that "just work".

- The `force_close_ln_dlc_channel` test should not assert on a magic number, as this is hard to understand and brittle.

---

- Relevant changes on `rust-dlc`: https://github.com/p2pderivatives/rust-dlc/commit/f4953a4aa54b298fe9da2ecff478c6794c3aee92.

- Relevant changes on `rust-lightning`: https://github.com/lightningdevkit/rust-lightning/compare/a97b1e0219ea73f0fcdc2a02ab77cf4bdaa2c5f7...get10101:rust-lightning:split-tx-experiment-114.

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -132,6 +132,37 @@ pub async fn close_channel(
     Ok(())
 }
 
+#[instrument(skip_all, err(Debug))]
+pub async fn finalize_force_close_ln_dlc_channel(
+    Path(channel_id): Path<String>,
+    State(state): State<Arc<AppState>>,
+) -> Result<(), AppError> {
+    let byte_array =
+        hex::decode(channel_id.clone()).map_err(|err| AppError::BadRequest(err.to_string()))?;
+
+    if byte_array.len() > 32 {
+        return Err(AppError::BadRequest(
+            "Provided channel id was invalid".to_string(),
+        ));
+    }
+    // Create a fixed-length byte array of size 8
+    let mut fixed_length_array = [0u8; 32];
+
+    // Copy the decoded bytes to the fixed-length array
+    let length = std::cmp::min(byte_array.len(), fixed_length_array.len());
+    fixed_length_array[..length].copy_from_slice(&byte_array[..length]);
+
+    tracing::info!(%channel_id, "Attempting to finalize channel force-close");
+
+    state
+        .node
+        .inner
+        .finalize_force_close_ln_dlc_channel(fixed_length_array)
+        .map_err(|e| AppError::InternalServerError(format!("{e:#}")))?;
+
+    Ok(())
+}
+
 pub async fn delete_subchannel(
     Path(channel_id): Path<String>,
     State(state): State<Arc<AppState>>,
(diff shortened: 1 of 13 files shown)


coordinator/src/admin.rs
```diff
@@ -132,6 +132,37 @@ pub async fn close_channel(
...
+    // Copy the decoded bytes to the fixed-length array
+    let length = std::cmp::min(byte_array.len(), fixed_length_array.len());
+    fixed_length_array[..length].copy_from_slice(&byte_array[..length]);
+
+    tracing::info!(%channel_id, "Attempting to finalize channel force-close");
+
+    state
+        .node
+        .inner
+        .finalize_force_close_ln_dlc_channel(fixed_length_array)
```
holzeis, line 160: nit: `fixed_length_array` is a bit ambiguous. How about `channel_id_fixed_length_array`

crates/ln-dlc-node/src/node/ln_channel.rs
```diff
@@ -60,53 +57,90 @@ impl<P> Node<P> {
...
+    fn collab_close_channel(&self, channel: &ChannelDetails) -> Result<()> {
+        let channel_id = channel.channel_id;
+        let channel_id_str = hex::encode(channel_id);
+        let peer = channel.counterparty.node_id;
+
+        tracing::info!(channel_id = %hex::encode(channel_id), %peer, "Collaboratively closing channel");
+
+        self.is_safe_to_close_ln_channel_collaboratively(&channel_id)
+            .with_context(|| {
+                format!("Could not collaboratively close LN channel {channel_id_str}")
```
holzeis, line 86: 🤔 Probably add that the existing DLC should be closed (settled) before the channel can be collaboratively closed? IMHO that would be more helpful as an error message.

holzeis (approved):
LGTM 👍

luckysori:
Merging because the tests passed locally (see https://github.com/get10101/10101/issues/355).

PR #613, opened 2023-05-09
Adapt fund binary to be able to fund public regtest instance

klochowicz:
Add a new just command `just fund-regtest` (run from local machine).


luckysori (approved):
Cool, thanks!

PR #614, opened 2023-05-10
Release test app version with dedicated bundle

### LINKED ISSUE #609: Release test version of 10101

holzeis:
Adds a justfile target to manually release the app to our test bundle id (finance.get10101.app.test) on TestFlight.

- Works on my machine ™️: https://appstoreconnect.apple.com/apps/6448948293/testflight/ios
- Pipeline to release to our main bundle id still works: https://github.com/get10101/10101/actions/runs/4934136628/jobs/8818814627

resolves #609 

Todo:

- [ ]  #616


justfile
```diff
@@ -246,9 +246,21 @@ wait-for-electrs-to-be-ready:
...
+    fi
+
+    cd mobile && flutter build ipa "${args[@]}" \
+           --dart-define="ESPLORA_ENDPOINT=${ESPLORA_ENDPOINT}" \
+           --dart-define="COORDINATOR_P2P_ENDPOINT=${COORDINATOR_P2P_ENDPOINT}" \
+           --dart-define="NETWORK=${NETWORK}" \
+           --dart-define="COMMIT=$(git rev-parse HEAD)" \
+           --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
+           --dart-define="COORDINATOR_PORT_HTTP=${COORDINATOR_PORT_HTTP}" \
+           --build-number=${BUILD_NUMBER}
```
klochowicz, line 263: I wonder whether it wouldn't be simpler to ditch `.env.sample` and just have a separate `build-ipa` and build-ipa-regtest` targets, but that might be an overkill if it works for you as it is 🙃

klochowicz (approved):
thank you for adding this.

I must say, after initial surprise about the new icon, it's growing on me. good job. 👌

justfile
holzeis: well.. I don't have a hard opinion of doing it either way. If you like to change that, go for it 🙂

I was thinking some implicit logic like this would be helpful on the pipeline. So that you don't have to provide an additional argument when running the build and release workflow manually.

holzeis:
> thank you for adding this.
> 
> I must say, after initial surprise about the new icon, it's growing on me. good job. 👌

Thanks, the icon was just a quick patch to make it clear that the app is for testing. Happy to change it if somebody with more UI/UX skills makes a proposal 🙂

PR #618, opened 2023-05-10
Add funding button on regtest

klochowicz:
- The button is not visible outside regtest.
- The button by default points to our remote faucet, and `just run` overrides it with local faucet.
- The request is done in Dart, I didn't want to pollute our library with non-production code.

https://github.com/get10101/10101/assets/8319440/bdec0603-ea9d-44a7-a93d-94839ca6f543

### DIFF
--- a/mobile/lib/features/wallet/share_invoice_screen.dart
+++ b/mobile/lib/features/wallet/share_invoice_screen.dart
@@ -1,11 +1,15 @@
+import 'dart:convert';
 import 'dart:developer';
 
+import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:get_10101/features/wallet/create_invoice_screen.dart';
 import 'package:get_10101/features/wallet/domain/wallet_info.dart';
 import 'package:get_10101/features/wallet/wallet_change_notifier.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
 import 'package:get_10101/features/wallet/wallet_screen.dart';
+import 'package:http/http.dart' as http;
 import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 import 'package:qr_flutter/qr_flutter.dart';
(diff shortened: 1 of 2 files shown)


mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -58,6 +63,24 @@ class ShareInvoiceScreen extends StatelessWidget {
                 ),
               ]),
             ),
+            // Faucet button, only available if we are on regtest
+            Visibility(
+              visible: config.network == "regtest",
+              child: OutlinedButton(
+                onPressed: () {
+                  payInvoiceWithFaucet(invoice);
```
luckysori, line 71: :question: Are we not `await`ing here on purpose?

holzeis (approved):

mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -141,4 +164,29 @@ class ShareInvoiceScreen extends StatelessWidget {
...
+    final response = await http.post(
+      Uri.parse('$faucet/lnd/v1/channels/transactions'),
+      headers: <String, String>{'Content-Type': 'application/json'},
+      body: encodedData,
+    );
+
+    if (response.statusCode == 200) {
+      FLog.info(text: response.body);
+    } else {
+      FLog.error(text: "error: ${response.statusCode} body: ${response.body}");
```
luckysori, line 189: :wrench: Since this is something that we know often fails, I think it would be good to give feedback to the user beyond printing an error log.

In this PR that was not merged I [used a snack bar to display the error message](https://github.com/get10101/10101/pull/533/files#diff-e8ba3fa73b3c5a44a61895c9196f7d935a42e33ac87bad9973665d64efdadac7R79-R80) and it was really simple.

It might also be a good idea to not navigate away from the share-invoice screen if this fails.

luckysori (approved):
LGTM, just a QOL suggestion.

mobile/lib/features/wallet/share_invoice_screen.dart
klochowicz: spot on, it's a sync call - doing http requests async in dart requires a lot more wizardry, which is just not practical for a non-production code. 
at least it has a timeout by default.

PR #620, opened 2023-05-10
chore: Add a help text to pay with faucet

holzeis:
Includes
- a helping text, hopefully preventing the user from repeating the pay with faucet flow.
- a subjective improvement on the ui

Screenshot 2023-05-10 at 18 11 06


holzeis:
> 

It also includes an improvement of the button style. Its definitely not the most important change, but lets go for it as its already up.

klochowicz:
I actually prefer just using a `SnackBar` for user feedback and blocking until it finishes, which prevents the problem.

The PR will be up shortly, and I'd prefer that much more than adding this text.

(as for styling, it's subjective, but I prefer the older one too FWIW)

PR #621, opened 2023-05-10
Log transaction broadcast

luckysori:


luckysori:
Bypassing CI since it's just adding a log statement.

PR #626, opened 2023-05-10
Prevent panic by updating `rust-dlc` and `rust-lightning`

luckysori:
See https://github.com/p2pderivatives/rust-lightning/commit/15d48d8243e05fbd198215c6a3a762d2bdc9f38b.

PR #629, opened 2023-05-10
chore: Turn off clippy on bors

klochowicz:
Bors started having trouble with our clippy job.
Turn it off as it's not essential (we run it on github anyway).

This quickfix will buy us more time until we'll move away from bors.

context:
https://github.com/get10101/10101/pull/626#issuecomment-1542742519 (and other today's PRs, where clippy passes locally but fails on `bors`.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -2,6 +2,5 @@ status = [
   "formatting",
   "lint-commits",
   "lint-flutter",
-  "clippy",
   "tests-ln-dlc-node",
 ]


luckysori (approved):
This means that we can end up breaking `main` if we're not careful. But I agree that this is a good idea for now.

luckysori:
Actually, maybe it's not worth disabling that one, since others can also fail instantly.

klochowicz:
> This means that we can end up breaking `main` if we're not careful. But I agree that this is a good idea for now.

not sure whether this is enough if `lint-commits` failed too 😂

klochowicz:
@luckysori I've opened #630 , which might need to do until we go with Github Merge Queue.

PR #630, opened 2023-05-10
chore: Turn off bors and run all tests directly on GitHub

klochowicz:
Bors stopped working reliably and became more of a hindrance than a helpful
companion as it once was.

R.I.P. bors.


klochowicz:
Feel free to block this PR - this is a knee-jerk reaction to the problems we were experiencing today on PR. It doesn't seem that they will go away.

klochowicz:
Since there is no consensus for this radical approach, I've settled with trying to plaster over the most annoying issue with our tests #651 instead.

PR #632, opened 2023-05-10
Allow `just fund` to continue even if LND and coordinator are already connected

luckysori:
It would be more correct to parse the error string to identify which errors can be ignored (such as the already-connected error), but at least this change makes `just fund` and `just fund-regtest` more usable.


bonomat (approved):

luckysori:
bors retry

holzeis (approved):

klochowicz (approved):

Issue #627, opened 2023-05-10
Failed to update channel monitor funding txo on DLC channel update


luckysori:
I've failed to close a position on 7ac6e28. At first glance, it doesn't like a reconnect issue. Here are the relevant logs from the 10101 test app:

```
| 2023-05-10 10:57:35.423 | LogLevel.INFO | Accepting DLC channel collaborative settlement channel_id: 266224002cffce70002ed7a38c43dd9de5bee39eaf8868a58c2f5035b000199d 
| 2023-05-10 10:57:35.425 | LogLevel.ERROR | lightning::chain::chainmonitor: Failed to update channel monitor funding txo: no such monitor registered  
| 2023-05-10 10:57:35.427 | LogLevel.ERROR | Failed to process message: Failed to accept DLC channel close offer for channel 266224002cffce70002ed7a38c43dd9de5bee39eaf8868a58c2f5035b000199d: Invalid parameters were provided: Misuse error: Could not update channel funding transaction. from: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802,kind: SubChannelCloseOffer 
```

---

Could possibly be related to #610, because the error message is similar. But it's slightly different and a completely different error path.

holzeis:
Looks like this error is also happening while opening a position. I just received these logs from one of our users on the test app version.

Build 1.0.17 (1126)

```
| 2023-05-11 17:40:18.606 | LogLevel.INFO | Processing message from: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802,kind: SubChannelOffer 
| 2023-05-11 17:40:18.705 | LogLevel.INFO | Accepting DLC channel offer channel_id: 8905e0f1588dab8aab650d82935f68a5dd8df7d2b766118391f4db5bea71f0a1 
| 2023-05-11 17:40:18.714 | LogLevel.ERROR | lightning::chain::chainmonitor: Failed to update channel monitor funding txo: no such monitor registered  
| 2023-05-11 17:40:18.716 | LogLevel.ERROR | Failed to process message: Failed to accept DLC channel offer for channel 8905e0f1588dab8aab650d82935f68a5dd8df7d2b766118391f4db5bea71f0a1: Invalid parameters were provided: Misuse error: Could not update channel funding transaction. from: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802,kind: SubChannelOffer 
```

luckysori:
[...]

Issue #610, opened 2023-05-09
Failed to update channel monitor: no such monitor registered


holzeis:
Just tested to fund my wallet on the Test VM and run into the following error on the coordinator.

**Version:**
Get10101 Test, 0.0.1 (2)

**Relevant Coordinator Logs**

```
2023-05-09 12:35:05 DEBUG lightning::ln::channel: Received valid commitment_signed from peer in channel 459830fdee6ef87ad647e7f9fa1d0c1f24b1c530ed7e873c6ae0cb3888418449, updating HTLC state and responding with our own commitment_signed and a revoke_and_ack.    
2023-05-09 12:35:05 ERROR lightning::chain::chainmonitor: Failed to update channel monitor: no such monitor registered    
2023-05-09 12:35:05 ERROR lightning::ln::channelmanager: Closing channel 459830fdee6ef87ad647e7f9fa1d0c1f24b1c530ed7e873c6ae0cb3888418449 due to monitor update ChannelMonitorUpdateStatus::PermanentFailure    
2023-05-09 12:35:05 DEBUG lightning::ln::channelmanager: Finishing force-closure of channel with 0 HTLCs to fail    
2023-05-09 12:35:05 ERROR lightning::chain::chainmonitor: Failed to update channel monitor: no such monitor registered    
2023-05-09 12:35:05 ERROR lightning::ln::channelmanager: ChannelMonitor storage failure  
```

luckysori:
Not sure what to do with this one. Did it happen again? I was able to fund a channel a few minutes ago.

Looking at the code in LDK it looks like it should never happen: https://github.com/lightningdevkit/rust-lightning/blob/54a768013f8fab70adacf84ddafb6ed43084794b/lightning/src/chain/chainmonitor.rs#L676-L682.

luckysori:
I've run into the same problem when _closing a position_. These are the logs from my test 10101 app on 7ac6e28:

```
| 2023-05-10 10:57:35.423 | LogLevel.INFO | Accepting DLC channel collaborative settlement channel_id: 266224002cffce70002ed7a38c43dd9de5bee39eaf8868a58c2f5035b000199d 
| 2023-05-10 10:57:35.425 | LogLevel.ERROR | lightning::chain::chainmonitor: Failed to update channel monitor funding txo: no such monitor registered  
[...]

PR #639, opened 2023-05-12
Update `rust-dlc` and `rust-lightning` to add some logs

luckysori:
See https://github.com/p2pderivatives/rust-lightning/commit/5329979edb415c0dfa6c310677f2575ad6bd5f06.

I'm hoping that this will give us more context if we run into #610 or #627.


klochowicz (approved):
fingers crossed 🤞

PR #640, opened 2023-05-12
fix: Add coordinator host to `just run-regtest`

klochowicz:
Now that we open the channel directly to the coordinator, we started relying on
this variable to be set correctly at startup.


luckysori (approved):
Nice, thanks!

PR #643, opened 2023-05-13
fix: fetch coordinator host from config

holzeis:
The host can be correctly fetched from the config. It implies the host either from the `coordinator_p2p_endpoint` or if not set the `coordinator_host`.

This fix is only needed as the `build-ipa` target was incorrect. We could have alternatively also only added the `coordinator_host` argument, but I think its cleaner to not touch the environment variables outside of the environment file and fetch the configs only from the config service.

### DIFF
--- a/justfile
+++ b/justfile
@@ -7,7 +7,6 @@ maker_log_file := "$PWD/data/maker/regtest.log"
 # public regtest constants
 public_regtest_coordinator := "03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@35.189.57.114:9045"
 public_regtest_esplora := "http://35.189.57.114:3000"
-public_coordinator_host:= "35.189.57.114"
 public_coordinator_http_port := "80"
 
 default: gen
@@ -66,7 +65,7 @@ run-regtest args="":
     #!/usr/bin/env bash
     cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
     --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
-    --dart-define="COORDINATOR_HOST={{public_coordinator_host}}" --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
+    --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
 
 fund:
     cargo run --example fund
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -7,7 +7,6 @@ maker_log_file := "$PWD/data/maker/regtest.log"
 # public regtest constants
 public_regtest_coordinator := "03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@35.189.57.114:9045"
 public_regtest_esplora := "http://35.189.57.114:3000"
-public_coordinator_host:= "35.189.57.114"
```
klochowicz, line 10: ...but it doesn't work without it. It might work for you, because you keep your `.env` file. for casual runner (like myself or @luckysori ), we rely on this. 
If you don't mind, let's keep it :-)

klochowicz (changes requested):

mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -85,7 +85,7 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
 
                         final router = GoRouter.of(context);
                         try {
-                          await openCoordinatorChannel();
+                          await openCoordinatorChannel(config.host);
```
klochowicz, line 88: nice one!

justfile
klochowicz: a bit of explanation: `run-regtest` should always use our regtest instance, it doesn't make sense to be changeable via the `.env` file (that's why I added it, and named it as such....).

holzeis: > ...but it doesn't work without it on my machine 😭 (nor for @luckysori )
Did you test it from my branch? 🙂 

> 🤔 Perhaps It might work for you, because you keep your .env file?
It's working because I am fetching the host from the config an not from the environment variable. I confess its a bit hidden in the code, but we already provide the host information through the `public_regtest_coordinator` variable.

The important piece of code is here in the `environment.dart`

```
 String p2pEndpoint = const String.fromEnvironment('COORDINATOR_P2P_ENDPOINT');
    if (p2pEndpoint.contains("@")) {
      final split = p2pEndpoint.split("@");
      coordinatorPublicKey = split[0];
      if (split[1].contains(':')) {
        host = split[1].split(':')[0];
        lightningPort = int.parse(split[1].split(':')[1]);
      }
    }
```
Unless I am missing something obvious, it should work for you too 🙂 .

> If you don't mind too much (minor redundancy after all), perhaps we could keep it as it is?
I don't mind at all, this is not critical at all.

klochowicz: Omg, you're right! Thank you for providing the additional context, I should not have looked at stuff on a phone during the weekend 🤐

klochowicz (approved):

PR #646, opened 2023-05-15
Allow specifying allowed max fee rate when opening a channel

klochowicz:


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -577,6 +577,24 @@ where
...
+                    .settings()
+                    .await
+                    .max_allowed_tx_fee_rate_when_opening_channel
+                {
+                    let current_fee = self
+                        .wallet
+                        .inner()
+                        .get_est_sat_per_1000_weight(CONFIRMATION_TARGET);
+                    if max_allowed_tx_fee < current_fee {
+                        tracing::warn!(%max_allowed_tx_fee, %current_fee, "Not opening a channel because the fee is too high");
```
holzeis, line 593: 🔧 We should fail the payment here.

```
self.channel_manager
          .fail_intercepted_htlc(intercept_id)
          .map_err(|e| anyhow!("{e:?}"))?;
```

holzeis (changes requested):
Great effort, minor things though, I think should be incorporated before merging. 👍

coordinator/src/settings.rs
```diff
@@ -20,6 +20,11 @@ pub struct Settings {
 
     pub fallback_tx_fee_rate_normal: u32,
     pub fallback_tx_fee_rate_high_priority: u32,
+
+    /// At times, we might want to prevent opening new channels if the feel rate is too high.
```
luckysori, line 24: :wrench: Typo: fee rate.

crates/ln-dlc-node/src/ln_dlc_wallet.rs
```diff
@@ -38,8 +38,6 @@ const BDK_CLIENT_CONCURRENCY: u8 = 8;
 
 /// This is a wrapper type introduced to be able to implement traits from `rust-dlc` on the
 /// `ldk_node::LightningWallet`.
-///
-/// We want to eventually get rid of the dependency on `bdk-ldk`, because it's a dead project.
```
luckysori, line 42: Nice.

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
```diff
@@ -158,9 +190,16 @@ pub(crate) async fn send_interceptable_payment(
         payee.connect(coordinator.info).await?;
     }
 
-    payee
-        .wait_for_payment_claimed(invoice.payment_hash())
-        .await?;
+    if let Err(e) = payee.wait_for_payment_claimed(invoice.payment_hash()).await {
```
luckysori, line 193: :thinking: Can we please not add any more branches to these tests? If we want to reuse code and have a different assertion we should just compose things differently.

These enums that activate different test branches are a nightmare for test clarity and they're not strictly necessary.

I know you didn't add the enum, but I would aim to get rid of it rather than continue to extend it.

luckysori (commented):
I like everything except the test-path enum.

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
klochowicz: I agree with your general sentiment, and find TestPaths really annoying.

However, in this case the `TestPath` is used exactly in the same function, so I don't think it makes it any worse.

I will create a ticket to address this technical debt, as refactoring is as such out of scope for this PR.

luckysori: Fair enough! We obviously don't have to do this right now, but hopefully soon :pray:

holzeis (approved):
Nice 👍

klochowicz:
merging as the hanging test is unrelated

PR #647, opened 2023-05-15
Upgrade to `rust-lightning` version `0.0.114`

luckysori:
Two things to take into account:

- The `onboard_from_lnd` test no longer passes, even with the manual node announcement after opening the channel between LND and coordinator.
- I tried to clean up f0513c8 for this PR, but there could still be things that shouldn't be there, as the source is a PR where I was trying to fix force-close of the LN-DLC channel.


luckysori:
Assigning you, @klochowicz. It would be really helpful if you can take this over tomorrow :D

klochowicz:
> Assigning you, @klochowicz. It would be really helpful if you can take this over tomorrow :D

took over and fixed the failing onboarding_from_lnd test 😊 should be ready for merging.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -77,6 +77,11 @@ where
             .payment_hash(sha256::Hash::from_slice(&payment_hash.0)?)
             .payment_secret(payment_secret)
             .timestamp(SystemTime::now())
+            // lnd defaults the min final cltv to 9 (according to BOLT 11 - the recommendation has
+            // changed to 18) 9 is not safe to use for ldk, because ldk mandates that
+            // the `cltv_expiry_delta` has to be greater than `HTLC_FAIL_BACK_BUFFER`
+            // (23).
+            .min_final_cltv_expiry_delta(MIN_CLTV_EXPIRY_DELTA.into())
```
luckysori, line 84: I'm still dumbfounded as to why I removed this :O

Cargo.toml
```diff
@@ -15,3 +15,4 @@ lightning-background-processor = { git = "https://github.com/get10101/rust-light
 lightning-transaction-sync = { git = "https://github.com/get10101/rust-lightning/", rev = "cb1c096e" }
 lightning-net-tokio = { git = "https://github.com/get10101/rust-lightning/", rev = "cb1c096e" }
 lightning-persister = { git = "https://github.com/get10101/rust-lightning/", rev = "cb1c096e" }
+rust-bitcoin-coin-selection = { git = "https://github.com/p2pderivatives/rust-bitcoin-coin-selection" }
```
luckysori, line 18: :+1:

crates/ln-dlc-node/src/tests/onboard_from_lnd.rs
```diff
@@ -53,6 +53,14 @@ async fn onboard_from_lnd() {
         )
         .unwrap();
 
+    assert!(
+        payer
+            .is_channel_active(coordinator.info.pubkey)
+            .await
+            .unwrap(),
+        "Channel is not active"
+    );
```
luckysori, line 62: :question: Was this helpful when debugging? As in, did it ever fail here? If so we can keep it, but I'm keen on keeping the test as simple as possible.

crates/ln-dlc-node/src/tests/mod.rs
```diff
@@ -53,7 +53,7 @@ fn init_tracing() {
     TRACING_TEST_SUBSCRIBER.call_once(|| {
         tracing_subscriber::fmt()
             .with_env_filter(
-                "debug,hyper=warn,reqwest=warn,rustls=warn,bdk=info,lightning=debug,sled=info,lightning::chain::channelmonitor=trace",
+                "debug,hyper=warn,reqwest=warn,rustls=warn,bdk=info,lightning=debug,sled=info,ln_dlc_node=debug,ln_dlc_node::ln=debug,lightning::chain::channelmonitor=trace",
```
luckysori, line 56: :question: Are you sure this makes a difference? If we specify `debug` at the front and no other pattern references `ln_dlc_node` I don't get how this can activate new logs.

coordinator/src/logger.rs
```diff
@@ -25,7 +25,6 @@ pub fn init_tracing(level: LevelFilter, json_format: bool) -> Result<()> {
         .add_directive("rustls=warn".parse()?)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
-        .add_directive("bdk_ldk=debug".parse()?)
```
luckysori: :tada: Thanks.

luckysori (commented):
LGTM. Thanks for fixing my mess, @klochowicz.

crates/ln-dlc-node/src/tests/onboard_from_lnd.rs
klochowicz: the payment was failing, and I wanted make sure it's not the case. I didn't bother with rolling it out in other tests, but I think it's good to check in at least one that the channel gets opened correctly.

luckysori: > I think it's good to check in at least one that the channel gets opened correctly.

This check is already happening when you call `open_channel` on line 24.

luckysori: https://github.com/get10101/10101/blob/51bb3c36fe2f906dc9f035cbcebe4fab5a42e3ea/crates/ln-dlc-node/src/tests/lnd.rs#L80-L94

klochowicz: thank you for pointing me to it - I'll remove the unnneeded patch :)

crates/ln-dlc-node/src/tests/mod.rs
klochowicz: I was under impression that these logs were missing.... but maybe I was wrong? I will remove it from this changeset, let me know if you ever miss these logs ;D

klochowicz:
tests passed locally; these seems to be some issue with running them on CI :(

PR #649, opened 2023-05-17
chore: Add contributing guidelines

klochowicz:
Based on our template from a previous project

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,25 @@
+# Contributing guidelines
+
+Thank you for wanting to contribute to this project!
+
+## Contributing code
+
+There are a couple of things we are going to look out for in PRs and knowing them upfront is going to reduce the number of times we will be going back and forth, making things more efficient.
+
+1. We have CI checks in place that validate formatting and code style.
+   Make sure `just fmt` and `just lint` both finish without any warnings or errors on every commit.
+   - If you don't already have `just` installed, you can obtain it by calling `cargo install just`.
+   - If you don't already have `dprint` installed (utility used for code formatting), you can obtain it [here](https://dprint.dev/install/).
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
+# Contributing guidelines
+
+Thank you for wanting to contribute to this project!
+
+## Contributing code
+
+There are a couple of things we are going to look out for in PRs and knowing them upfront is going to reduce the number of times we will be going back and forth, making things more efficient.
+
+1. We have CI checks in place that validate formatting and code style.
+   Make sure `dprint check` and `cargo clippy` both finish without any warnings or errors on every commit.
```
luckysori, line 10: :question: Should we reference `just` or does that make things harder?

CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
...
+   If you don't already have `dprint` installed, you can obtain it in [various ways](https://dprint.dev/install/).
+2. All text documents (`CHANGELOG.md`, `README.md`, etc) should follow the [semantic linebreaks](https://sembr.org/) specification.
+3. We strive for atomic commits with good commit messages.
+   As an inspiration, read [this](https://chris.beams.io/posts/git-commit/) blogpost.
+   An atomic commit is a cohesive diff with formatting checks, linter and build passing.
+   Ideally, all tests are passing as well but we acknowledge that this is not always possible depending on the change you are making.
+4. If you are making any user-facing changes, include a changelog entry.
```
luckysori, line 17: Haha, does this mean only external contributors have to do this?

CONTRIBUTING.md
klochowicz: thanks, I'll use `just` commands instead!

CONTRIBUTING.md
klochowicz: we want to start doing that as well ;D

PR #651, opened 2023-05-17
chore: Increase ln-dlc-node test timeout for bors

klochowicz:
We often fail in the middle of the test run, causing bors to retry.
Since this timeout is arbitrary, we might as well increase it a bit to reduce
the failure rate.


holzeis:
Are our tests take over 30 minutes now? Last time I checked a successful run took about 5-6 minutes.

holzeis (approved):

klochowicz:
@luckysori pointed me out that this was a hanging test instead, and this would not improve anything. closing.

PR #652, opened 2023-05-17
Configure `demo` flavor for the `regtest` apk build

da-kami:
- Different app id and app name

Note that the logo was not adapted (yet).

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -49,6 +49,16 @@ android {
         versionName flutterVersionName
     }
 
+    flavorDimensions "default"
+
+    productFlavors {
+        demo {
+            dimension "default"
+            resValue "string", "app_name", "10101 (regtest)"
+            applicationIdSuffix ".demo"
+        }
+    }
+
     buildTypes {
         release {
             // TODO: Add your own signing config for the release build.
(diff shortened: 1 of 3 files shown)


mobile/android/app/build.gradle
```diff
@@ -49,6 +49,16 @@ android {
         versionName flutterVersionName
     }
 
+    flavorDimensions "default"
+
+    productFlavors {
+        demo {
```
holzeis, line 55: nit: its a bit confusion that we have test flavor in ios and a demo flavor on Android. 🙂

mobile/android/app/build.gradle
```diff
@@ -49,6 +49,16 @@ android {
...
+            dimension "default"
+            resValue "string", "app_name", "10101 (regtest)"
```
holzeis, line 57: nit: The test app for ios is named 10101 Test

holzeis (approved):

mobile/android/app/build.gradle
da-kami: unfortunately it's a naming issue - in android the flavour cannot start with `test`

mobile/android/app/build.gradle
da-kami: I find it more intuitive to use the bitcoin network, but not sure we want to rename iOS.

PR #655, opened 2023-05-18
fix: Disable close button if position is expired

### LINKED ISSUE #519: Position close button should be disabled on the position is expired
Expired positions are automatically closed by the coordinator. This may take a few minutes before executed. In order to prevent the user from closing the position manually and interfere with the automatism, the close button should be disabled once the position has expired.

holzeis:
resolves #519 

A small fix to ensure that the user cant close a position that has already expired.


klochowicz (approved):
thanks!

klochowicz (approved):
thanks!

PR #661, opened 2023-05-19
feat: Enable anchor outputs

klochowicz:
Anchor outputs are hidden by LDK behind a config flag.

We keep it consistent across all our targets to avoid unnecessary recompilation
for clippy.


klochowicz:
Closing as it can't be finished until LDK 116 lands.

Issue #670, opened 2023-05-20
Never getting a match can result in app being in `Closing` forever


da-kami:
I created an order to close a position, but somehow the order was never matched.

I think the problem is that I lost connection to the orderbook websocket at an unfortunate point in time (unsure why).
The app does reconnect, but it never receives a match, so the position is just stuck.

We should have a timeout for receiving a match and if we don't receive one within x time we should reset the position to open so the user can try again. We will have to be careful to not introduce more weirdness with users triggering multiple "close" orders that could result in a closed and then another open position.

App logs:

```
| 2023-05-19 22:00:16.292 | LogLevel.WARNING | Orderbook WS stream closed with error error: IO error: Software caused connection abort (os error 103) 
| 2023-05-19 22:00:16.297 | LogLevel.DEBUG | Reconnecting to orderbook WS after timeout timeout: 2s 
| 2023-05-19 22:00:16.300 | LogLevel.DEBUG | lightning::ln::channelmanager: Marking channels with 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802 disconnected and generating channel_updates. We believe we can make future connections to this peer.  
| 2023-05-19 22:00:16.302 | LogLevel.DEBUG | Connection lost; reconnecting peer: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@35.189.57.114:9045,reconnect_interval: 1s 
| 2023-05-19 22:00:17.264 | LogLevel.ERROR | Sync failed due to HTTP connection error, retrying: Esplora client error: Reqwest(reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(35.189.57.114)), port: Some(3000), path: "//scripthash/24a8900092ff4c32a27eb7ab03e3b712fffecc2b6573a6511d14af86b1360165/txs", query: None, fragment: None }, source: hyper::Error(Io, Os { code: 103, kind: ConnectionAborted, message: "Software caused connection abort" }) })  
| 2023-05-19 22:00:17.267 | LogLevel.DEBUG | Setting up connection peer: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@35.189.57.114:9045 
| 2023-05-19 22:00:18.259 | LogLevel.DEBUG | Connecting to orderbook API  
[...]

PR #674, opened 2023-05-20
Fail outdated orders

da-kami:
fixes https://github.com/get10101/10101/issues/670 (likely, unless there is some other problem that I don't understand yet...)

If an order remains `Open` for more than `5` minutes we don't expect it to ever get matched. We have a task that runs one per minute to load the orders that are `Open` and determine if the order's creation timestamp is older than `5`minutes; if so we fail the order.


luckysori (approved):

klochowicz (approved):
LGTM!

PR #680, opened 2023-05-20
fix: Change the default concurrency in BDK wallet

klochowicz:
This should substantially reduce the wait time for initial and subsequent sync, improving the app experience.

We had 8 previously, but 4 is the default value in esplora client, so should be
safe to use.

Note: our coordinator can be configured independently now, without neededing to
release a new version.


bonomat (approved):
👍

Issue #682, opened 2023-05-20
Log spam on app


bonomat:
I see 1000s of lines of 
```
| 2023-05-20 14:01:28.338 | LogLevel.DEBUG | lightning::ln::peer_handler: Error handling message from 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156; ignoring: Couldn't find channel for update  
```

bonomat:
Stale and not important at the moment. If it gets too spamy let's re-open.

PR #687, opened 2023-05-22
Do not log LN `peer_handler` ping and pong messages

luckysori:
We update our `rust-lightning` dependencies to be able to keep seeing peer disconnects (which have been promoted to debug).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1433,7 +1433,7 @@ dependencies = [
 [[package]]
 name = "lightning"
 version = "0.0.114"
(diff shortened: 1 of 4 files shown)


klochowicz:
does it also tackle https://github.com/get10101/10101/issues/682 ?

luckysori:
> does it also tackle #682 ?

Nope. I'm not sure we should ignore those errors. This is just about ping and pong.

da-kami (approved):

PR #689, opened 2023-05-22
Move failed-to-broadcast error outside of block-in-place

luckysori:
In case using `tracing` inside `block_in_place` closures does not work.


luckysori:
Just changing logs.

da-kami (approved):

PR #691, opened 2023-05-23
Spawn expensive blocking `rust-dlc` code on blocking threads

luckysori:
This is done to mitigate risks such as starving other async tasks and even deadlocks.


klochowicz (approved):
great job!

PR #696, opened 2023-05-23
Remove unnecessary async

da-kami:
This code was unnecessarily complex and calling `std::thread::spawn` from within an async tokio context is not recommended.


holzeis (approved):
LGTM 👍

PR #707, opened 2023-05-25
Run command with local flavor and configurable IP addr

da-kami:
A simple way to run the app on the phone but connect to a local (any) coordinator via IP.

---

This command makes it easy to quickly change to another machine that runs the setup, assuming the default ports and the checked in coordinator seed.

I am not sure if we want to do anything with this, but it made me think about our configuration in general.

At least for the (local) regtest setup the only thing that usually changes is the host machine's IP, maybe we can simplify the configuration to extract the IP and only set that instead of all the other parameters.

---

Note: I only configured the `local` flavor for Android - if we want to keep this command we could consider adding a flavor iOS as well.

### DIFF
--- a/justfile
+++ b/justfile
@@ -78,6 +78,13 @@ run-regtest args="":
     --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
     --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
 
+run-local-android args="":
+    #!/usr/bin/env bash
+    $LOCAL_IP=(curl -sS ifconfig.co)
+    cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
+    --dart-define="ESPLORA_ENDPOINT=http://${LOCAL_IP}:3000" --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@${LOCAL_IP}:9045" \
+    --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
+
 fund:
     cargo run --example fund
 
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -59,6 +59,14 @@ android {
         }
     }
 
+    productFlavors {
+        local {
+            dimension "default"
+            resValue "string", "app_name", "10101 (local)"
+            applicationIdSuffix ".local"
+        }
+    }
+
     buildTypes {
         release {
             // TODO: Add your own signing config for the release build.


justfile
```diff
@@ -12,6 +12,8 @@ public_regtest_coordinator := "03507b924dae6595cfb78492489978127c5f1e3877848564d
 public_regtest_esplora := "http://35.189.57.114:3000"
 public_coordinator_http_port := "80"
 
+local_ip := "10.60.173.164"
```
klochowicz, line 15: we could probably retrieve this value from the shell instead of hardcoding it? 😅 I don't think it will work for anyone else except yourself on a particular network.

(see run-local suggestion for one way of achieving it)

justfile
```diff
@@ -78,6 +80,13 @@ run-regtest args="":
     --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
     --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
 
+run-local args="":
+    #!/usr/bin/env bash
+    cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
+    --dart-define="ESPLORA_ENDPOINT=http://{{local_ip}}:3000" --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@{{local_ip}}:9045" \
+    --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
```
klochowicz, line 87: we could use cURL to quickly fetch the local ip (it's safe to assume we have it on our dev machines.)

```suggestion
    $LOCAL_IP=(curl -sS ifconfig.co)
    cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
    --dart-define="ESPLORA_ENDPOINT=http://${LOCAL_IP}:3000" --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@{{local_ip}}:9045" \
    --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
```

justfile
```diff
@@ -78,6 +80,13 @@ run-regtest args="":
...
+
+
```
klochowicz, line 89: 🙃 unneeded newline

```suggestion
```

justfile
```diff
@@ -78,6 +80,13 @@ run-regtest args="":
     --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
     --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
 
+run-local args="":
```
klochowicz, line 83: if it's specific to android, let's call it as such
```suggestion
run-local-android args="":
```

klochowicz (approved):
happy for it to go in, provided the feedback is addressed 😊

klochowicz:
@da-kami I'll carry this over the finish line - thank you the PR ❤️

PR #709, opened 2023-05-25
Bunch of small fixes around potentially blocking operations

klochowicz:
See commit messages for details. Hopefully this will reduce our amount of blocks

note: the commits are unrelated, and could come in separate PRs if that's preferred.

### DIFF
--- a/coordinator/src/orderbook/routes.rs
+++ b/coordinator/src/orderbook/routes.rs
@@ -38,10 +38,13 @@ use serde::Serialize;
 use std::fmt;
 use std::str::FromStr;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::sync::broadcast::Sender;
 use tokio::sync::mpsc;
 use tracing::instrument;
 use uuid::Uuid;
+
+const WEBSOCKET_SEND_TIMEOUT: Duration = Duration::from_secs(5);
 
 #[derive(Debug, Deserialize)]
 pub struct AllOrdersParams {
@@ -268,7 +271,12 @@ async fn websocket(stream: WebSocket, state: Arc<AppState>) {
         while let Some(local_msg) = local_receiver.recv().await {
             match serde_json::to_string(&local_msg) {
                 Ok(msg) => {
-                    if let Err(err) = sender.send(Message::Text(msg.clone())).await {
+                    if let Err(err) = tokio::time::timeout(
+                        WEBSOCKET_SEND_TIMEOUT,
+                        sender.send(Message::Text(msg.clone())),
+                    )
+                    .await
+                    {
                         tracing::error!("Could not forward message {msg} : {err:#}");
                         return;
                     }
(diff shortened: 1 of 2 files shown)


coordinator/src/orderbook/routes.rs
```diff
@@ -147,7 +147,7 @@ pub async fn post_order(
     let matched_orders = match_order(order.clone(), all_non_expired_orders)
         .map_err(|e| AppError::InternalServerError(format!("Failed to match order: {e:#}")))?;
 
-    let authenticated_users = state.authenticated_users.lock().await;
+    let authenticated_users = state.authenticated_users.lock().await.clone();
```
da-kami, line 153: ❓ I assume that the `clone` will free the lock immediately, so we are now only locking for the duration of the clone.
Is my understanding correct?

coordinator/src/orderbook/trading.rs
```diff
@@ -152,7 +152,7 @@ fn sort_orders(mut orders: Vec<Order>, is_long: bool) -> Vec<Order> {
 #[autometrics]
 pub async fn notify_traders(
     matched_orders: MatchParams,
-    authenticated_users: HashMap<PublicKey, Sender<OrderbookMsg>>,
+    authenticated_users: &HashMap<PublicKey, Sender<OrderbookMsg>>,
```
da-kami, line 155: ❓ My understanding is, that now this function call just takes a reference, i.e. not consuming the map. 
Is this a pure refactor or does it actually make a difference in terms of blocking / mutex?

coordinator/src/orderbook/routes.rs
```diff
@@ -268,7 +271,12 @@ async fn websocket(stream: WebSocket, state: Arc<AppState>) {
         while let Some(local_msg) = local_receiver.recv().await {
             match serde_json::to_string(&local_msg) {
                 Ok(msg) => {
-                    if let Err(err) = sender.send(Message::Text(msg.clone())).await {
+                    if let Err(err) = tokio::time::timeout(
+                        WEBSOCKET_SEND_TIMEOUT,
+                        sender.send(Message::Text(msg.clone())),
```
da-kami, line 276: ❓how long could it happen that we wait without it? 
i.e. what's the "worst" than can happen when you call `sender.send(...)` and the sender is not available. Can this actually result in long waiting times.

I'm happy to add the timeout regardless :)

da-kami: One note: We have to be careful to avoid scenarios where we stop sending on the server side, but the client actually still thinks they are connected.

I remember we had such kind of scenarios in ItchySats - not entirely sure how we can avoid that, but we are starting to add to connection management on the server side again which might lead to problems on the client side.

da-kami (approved):
Happy to merge this and see if it has any effects. 

I added a few questions and comments - nothing blocking, just making sure I get the intent :)

coordinator/src/orderbook/routes.rs
klochowicz: Yes, exactly. Before we have held the lock until the variable went out of scope, ie. until the end of the function

coordinator/src/orderbook/routes.rs
klochowicz: this PS doesn't add any connection management - this one just makes sure we're not hanging forever with a request.

we might need to be careful when we remove entries from the map, as this should always be communicated to both sides appropriately

PR #711, opened 2023-05-26
Do not spawn new OS thread just for updating fee estimates

luckysori:
It is sufficient to use an asynchronous task, as the work we're doing here is primarily IO-bound: requesting the latest fee rate from using the `esplora` client.

### DIFF
--- a/crates/ln-dlc-node/src/node/mod.rs
+++ b/crates/ln-dlc-node/src/node/mod.rs
@@ -1,5 +1,6 @@
 use crate::disk;
 use crate::dlc_custom_signer::CustomKeysManager;
+use crate::fee_rate_estimator::FeeRateEstimator;
 use crate::ln::app_config;
 use crate::ln::coordinator_config;
 use crate::ln::EventHandler;
@@ -94,6 +95,7 @@ pub struct Node<P> {
     chain_monitor: Arc<ChainMonitor>,
     keys_manager: Arc<CustomKeysManager>,
     pub network_graph: Arc<NetworkGraph>,
+    pub fee_rate_estimator: Arc<FeeRateEstimator>,
     _background_processor: BackgroundProcessor,
     _connection_manager_handle: RemoteHandle<()>,
     _broadcast_node_announcement_handle: RemoteHandle<()>,
@@ -273,13 +275,19 @@ where
         let on_chain_wallet =
             OnChainWallet::new(on_chain_dir.as_path(), network, seed.wallet_seed())?;
 
-        let esplora_client = Arc::new(EsploraSyncClient::new(esplora_server_url, logger.clone()));
+        let esplora_client = Arc::new(EsploraSyncClient::new(
+            esplora_server_url.clone(),
+            logger.clone(),
+        ));
 
         let runtime_handle = tokio::runtime::Handle::current();
+
+        let fee_rate_estimator = Arc::new(FeeRateEstimator::new(esplora_server_url));
         let ln_dlc_wallet = {
             Arc::new(LnDlcWallet::new(
                 esplora_client.clone(),
                 on_chain_wallet.inner,
+                fee_rate_estimator.clone(),
                 storage.clone(),
                 seed.clone(),
                 runtime_handle.clone(),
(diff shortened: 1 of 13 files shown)


crates/ln-dlc-node/src/fee_rate_estimates.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    pub(crate) fn get(&self, target: ConfirmationTarget) -> FeeRate {
+        let (cache, fallbacks) = self.runtime_handle.block_on(async {
+            (
+                self.fee_rate_cache.read().await,
+                self.fallbacks.read().await,
+            )
+        });
```
klochowicz, line 53: 🤔 this seems odd. we can probably get rid of the `block_on` call here? 

(it was the earlier, but I'd like to flag this).

crates/ln-dlc-node/src/fee_rate_estimates.rs
```diff
@@ -0,0 +1,100 @@
...
+            };
+            FeeRate::from_sat_per_kwu(fee_rate as f32)
+        })
+    }
+
+    #[autometrics]
+    pub(crate) async fn update(&self) -> Result<()> {
+        let mut locked_fee_rate_cache = self.fee_rate_cache.write().await;
+
+        let estimates = self.client.get_fee_estimates().await?;
```
klochowicz, line 69: this one should IMHO be in spawn blocking, as it will take some time to finish.

source: https://kerkour.com/rust-async-await-what-is-a-runtime 
```
[...]
What does it mean? Not calling functions that may run for more than [10 to 100 microseconds](https://ryhl.io/blog/async-what-is-blocking/) directly. Instead, spawn_blocking them.
```

crates/ln-dlc-node/src/fee_rate_estimates.rs
```diff
@@ -0,0 +1,100 @@
...
+            tracing::trace!(
+                n_blocks_confirmation = %n_blocks,
+                sats_per_kwu = %fee_rate.fee_wu(1000),
+                "Updated fee rate estimate",
+            );
+        }
+        Ok(())
+    }
+
+    pub async fn update_fallbacks(&self, fallbacks: FeeRateFallbacks) {
```
klochowicz, line 89: I believe you forgot to update it when new settings come from the coordinator - we lost the ability to influence the fallbacks, even at startup :(

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -318,31 +328,25 @@ where
             }
         });
 
-        std::thread::spawn({
+        tokio::spawn({
```
klochowicz, line 331: 👍 I don't think that a whole thread for this was needed. Especially with the new runtime (?!)

holzeis (approved):

crates/ln-dlc-node/src/fee_rate_estimates.rs
klochowicz: the point is that to do this judiciously, if it can block for longer than the time above.
 there's no harm to spawning it on a worker thread instead of the main runtime.

klochowicz (changes requested):
marking as 'changes requested' so we don't push regression in fallback fees update. 

@luckysori I'm happy to give you a hand with that 🙃

crates/ln-dlc-node/src/fee_rate_estimates.rs
luckysori: > this one should IMHO be in spawn blocking, as it will take some time to finish.

This call doesn't do anything other than build and send the HTTP request and it `await`s for all the expensive bits: building, sending and deserialising.

`tokio::spawn_blocking` targets expensive non-`async` code: we won't even be able to call `get_fee_estimates` within the closure!

crates/ln-dlc-node/src/fee_rate_estimates.rs
luckysori: The way forward here is to move away from `tokio::sync::RwLock` in favour of `std::sync::RwLock`. Thanks for leading me there.

crates/ln-dlc-node/src/fee_rate_estimates.rs
luckysori: Yep, I missed that. I've rectified this now.

PR #712, opened 2023-05-26
Avoid using tokio Mutex

klochowicz:
According to tokio documentation, sync Mutex is preferred over
the tokio Mutex.

### DIFF
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -29,8 +29,9 @@ use lightning::chain::WatchedOutput;
 use lightning_transaction_sync::EsploraSyncClient;
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::MutexGuard;
 use std::time::Duration;
-use tokio::sync::Mutex;
 use tokio::sync::RwLock;
 
 pub struct Wallet<D>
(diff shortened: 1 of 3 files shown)


crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -221,20 +226,16 @@ where
...
     pub(crate) fn get_last_unused_address(&self) -> Result<bitcoin::Address, Error> {
         let address_info = tokio::task::block_in_place(move || {
-            self.runtime_handle.block_on(async move {
-                self.inner
-                    .lock()
-                    .await
-                    .get_address(AddressIndex::LastUnused)
-            })
+            self.runtime_handle
+                .block_on(async move { self.bdk_lock().get_address(AddressIndex::LastUnused) })
```
holzeis, line 230: 🤔 there is no await here anymore, is the async still needed?

This also applies to the other methods below.

- get_balance
- on_chain_transaction_list

holzeis (approved):
LGTM

crates/ln-dlc-node/src/ldk_node_wallet.rs
klochowicz: it might be not needed! maybe we can remove the `runtime_handle` as a result!

klochowicz: `runtime_handle` needs to stay for now for the fee estimation work, so I'll not touch it especially that @luckysori has a PR that refactors it completely.

I'll change `on_chain_transaction_list` in a separate follow-up, because there's a lot of unnecessary async calls in the code that will drop away and I don't want to cloud this one.

klochowicz:
I'll investigate the bors failure (whether it passes locally)

klochowicz:
fails locally too 😭 it's unlikely this changes is breaking it then,  I'm going ahead with the merge .

PR #716, opened 2023-05-31
Do not lock `Mutex` to get the network

luckysori:


crates/ln-dlc-node/src/ln_dlc_wallet.rs
```diff
@@ -123,9 +127,7 @@ impl Blockchain for LnDlcWallet {
 
     #[autometrics]
     fn get_network(&self) -> Result<Network, Error> {
-        self.ln_wallet
-            .network()
-            .map_err(|e| Error::BlockchainError(e.to_string()))
+        Ok(self.network)
```
holzeis, line 130: nit: This is now infallible, we could get rid of the result.

holzeis (approved):

crates/ln-dlc-node/src/ln_dlc_wallet.rs
luckysori: You're right! I thought it was a trait method 🤦

luckysori: Ah, nvm. It _is_ a foreign trait method.

klochowicz (approved):
thank you for addressing the TODO ❤️

klochowicz (approved):
thank you for addressing the TODO ❤️

PR #719, opened 2023-05-31
Do not use two tokio runtimes in the mobile app

klochowicz:


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -82,7 +82,7 @@ pub fn get_oracle_pubkey() -> XOnlyPublicKey {
 
 /// Lazily creates a multi threaded runtime with the the number of worker threads corresponding to
 /// the number of available cores.
-fn runtime() -> Result<&'static Runtime> {
+pub fn create_tokio_runtime() -> Result<&'static Runtime> {
```
luckysori, line 85: 🙃 I liked the original better because this one implies that we are building a new runtime every time.

luckysori (approved):

mobile/native/src/ln_dlc/mod.rs
klochowicz: Every time? It does it once, because it stores it in the `Storage` object, even if we called it multiple times... But we don't even call it more that once? Am I missing anything? 🤔

luckysori: Your name change suggests that every time we call it we are creating a new one even though it's not the case as you point out.

klochowicz: thank you for explaining. I indeed made a confusing name, I totally see how it could be understood this way. my apologies! 😅
changed to `get_or_create` to better reflect the name :)

PR #720, opened 2023-05-31
Process incoming DLC messages on a dedicated thread

klochowicz:
Also reduce the wait time to 2s, as we're not going to starve the runtime.

### DIFF
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -97,7 +97,8 @@ async fn main() -> Result<()> {
     let node = Node::new(node, pool.clone());
     node.update_settings(settings.as_node_settings()).await;
 
-    tokio::task::spawn_blocking({
+    // spawn a dedicated thread as this might be CPU-heavy
+    std::thread::spawn({
         let node = node.clone();
         move || loop {
             node.process_incoming_dlc_messages();
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -28,7 +28,7 @@ use time::OffsetDateTime;
 use tracing::metadata::LevelFilter;
 use trade::bitmex_client::BitmexClient;
 
-const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_secs(5);
+const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_secs(2);
 const POSITION_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 const CONNECTION_CHECK_INTERVAL: Duration = Duration::from_secs(30);
 
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -43,7 +43,7 @@ use tokio::runtime::Runtime;
 mod node;
 
 static NODE: Storage<Arc<Node>> = Storage::new();
-const PROCESS_INCOMING_MESSAGES_INTERVAL: Duration = Duration::from_secs(5);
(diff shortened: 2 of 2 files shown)


coordinator/src/bin/coordinator.rs
```diff
@@ -97,7 +97,8 @@ async fn main() -> Result<()> {
     let node = Node::new(node, pool.clone());
     node.update_settings(settings.as_node_settings()).await;
 
-    tokio::task::spawn_blocking({
+    // spawn a dedicated thread as this might be CPU-heavy
+    std::thread::spawn({
```
luckysori, line 101: 🤔 Looking at what we're actually doing in `process_incoming_dlc_messages`, I'm not sure that it makes sense to have a dedicated OS thread.

This OS thread will be mostly idle, since all it has to do is to check for pending messages (a very cheap operation) and process pending messages (probably always expensive) if necessary. But it's important to remember that these messages are only exchanged when opening and closing positions (eventually when updating the position too), which is something that does not happen so frequently.

I would advocate for an async task that spawns a dedicated thread (or perhaps a `rayon` task if we need to parallelise the work with some other heavy computation) when necessary. I understand that it's not cheap to spawn a new OS thread, but it seems wasteful to leave a dedicated thread doing nothing expensive for so long.

klochowicz: I would argue that using the proposed alternative would be even more wasteful.

Rayon would add **4-8** (depending on number of CPUs) threads, all of which would be idle other than these computations, because currently we have no plan to use `rayon` elsewhere, as there's not that much CPU-heavy stuff.

side note: There is of course also a potential of using rayon if we could parallelise this computation, but we'd need to refactor all of the code underneath and the potential for achieving this is low, as the messages from the same node need to be processed sequentially ;/ (maybe from different nodes?)

> I would advocate for an async task that spawns a dedicated thread (or perhaps a rayon task if we need to parallelise the work with some other heavy computation) when necessary. I understand that it's not cheap to spawn a new OS thread, but it seems wasteful to leave a dedicated thread doing nothing expensive for so long.

it is significantly more wasteful to spawn an OS thread multiple times than just have it allocated already.

To put in into perspective: we have `512` threads in the threadpool + `$nproc` threads for the runtime itself. This one more thread is pretty much a rounding error 😅

what we *could* do, if you prefer, is to utilise the same pattern of having a simple periodic `task` spawned that does the `spawn_blocking` for the expensive computation.
The only difference is that then we're using one of the tokio worker threads (`512`) to do so. 
would that be actually faster? I'm not sure, because every second we'd need to instruct a different worker thread to perform this computation; hence the dedicated OS thread solution.

luckysori: > Rayon would add 4-8 (depending on number of CPUs) threads, all of which would be idle other than these computations, because currently we have no plan to use rayon elsewhere, as there's not that much CPU-heavy stuff.

Yeah, using `rayon` here doesn't make sense. I agree.

> it is significantly more wasteful to spawn an OS thread multiple times than just have it allocated already.

But my understanding is that the OS thread is a limited resource. If we're taking one and keeping it effectively idle for 99% of the time, might it not be better to only spawn it when we're about to do some real work? To me this is not so obvious.

In any case, I just don't know haha. We can give this a go as I don't think it's _wrong_.

klochowicz: @luckysori what do you think about this instead? https://github.com/get10101/10101/pull/728 (more consistent with other patterns, and at least it won't show up as long-running task in tokio-console ;) )

luckysori: Tbh I'm not even sure if what you propose here is worse at all, but let's roll with the other for now. Afterwards we can choose spots where a dedicated thread could make more sense.

klochowicz:
#728 got merged, so this one is not needed (can be considered as an alternative if we decide for some reason it's worth it someday).

PR #722, opened 2023-05-31
fix: Stop using bdk fork that introduced 250ms delay

### LINKED ISSUE #717: Failing test: `ln_collab_close`
Probably because of our tendency to skip CI due to #355, we have introduced a regression which causes the `ln_collab_close` test to fail, as reported by @klochowicz in #715.

I was able to reproduce this by running

```
cargo test -p ln-dlc-node -- tests::just_in_time_channel::channel_close::ln_collab_close --exact --ignored
```

which runs into

```
thread 'tests::just_in_time_channel::channel_close::ln_collab_close' panicked at 'assertion failed: `(left == right)`
  left: `1000`,
 right: `0`', crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs:77:5
```

This happens because we are mining an unexpected number of blocks. The test has always only mined:

- 1 block to fund the coordinator's on-chain wallet.
- 1 block to confirm the close transaction.

Somehow I get 4 extra blocks some time after mining the first one.

The logs only show one instance of

```
2023-05-31T06:20:43.353410Z DEBUG ln_dlc_node::tests::bitcoind: Mining n_blocks=1
```

which appears to match the explicit call to `bitcoind::mine` when funding the coordinator.

klochowicz:
Whilst an important quickfix at the time (it allowed us to resume testing on
mainnet), our fork introduced more side-effects than it fixed.

Some of the side-effects observed:
    - bdk wallet sync would take ~15s (instead of ~1-2s)
    - ln_collab_close` test was failing
    - lightning wallet sync took only 3ms to complete (instead ~1000ms normally)

Fixes #717

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,7 +243,8 @@ checksum = "a4a4ddaa51a5bc52a6948f74c06d20aaaddb71924eab79b8c97a8c556e942d6a"
 [[package]]
 name = "bdk"
 version = "0.27.1"
-source = "git+https://github.com/get10101/bdk?rev=83fbc0cb471f8c377442d70bd6d68cea25d2221d#83fbc0cb471f8c377442d70bd6d68cea25d2221d"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51c878ac60a45c41523ff790df555ccb1fbfd634a667220104dcae3e64d7ed9f"
 dependencies = [
  "async-trait",
  "bdk-macros",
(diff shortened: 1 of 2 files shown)


luckysori (approved):
I completely agree that we need to do this, but do we have a strategy to run our own infrastructure already so that we don't hit Blockstream rate limits?

luckysori:
> Fixes #717

I love that it fixes #717, but how come? How can a slow sync lead to the transaction being in a different block than we expect it to? I don't quite get it.

klochowicz:
> I completely agree that we need to do this, but do we have a strategy to run our own infrastructure already so that we don't hit Blockstream rate limits?

I believe the plan is to run our own esplora server - we already have bitcoind synced on mainnet on our machine, so it should be relatively low-hanging fruit.

some ideas for avoid public server rate limits:
I would like to see whether different concurrency / stop gap settings could help us with not hitting the rate limits on public esplora server, as we changed them both at the same time as introducing our fork. 

one more thing that could make a difference would be decoupling bdk and ldk sync to query esplora at different times?

PR #724, opened 2023-06-01
Migrate a few things from async to sync

luckysori:
There might be some overlap with your work @klochowicz as I simply grepped the codebase for `async fn`, but we should be able to put it all together without too much pain.

---

Change to `rust-dlc` here: https://github.com/p2pderivatives/rust-dlc/commit/e4778828c21614dd2bce0b7d687e36345286f382.

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -116,15 +116,11 @@ where
 
                 // Have wallet put the inputs into the transaction such that the output
                 // is satisfied and then sign the funding transaction
-                let funding_tx_result = self
-                    .wallet
-                    .inner()
-                    .create_funding_transaction(
-                        output_script,
-                        channel_value_satoshis,
-                        target_blocks,
-                    )
-                    .await;
+                let funding_tx_result = self.wallet.inner().create_funding_transaction(
+                    output_script,
+                    channel_value_satoshis,
+                    target_blocks,
+                );
 
                 let funding_tx = match funding_tx_result {
                     Ok(funding_tx) => funding_tx,
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -71,7 +71,7 @@ where
     }
 
     #[autometrics]
-    pub async fn propose_dlc_channel_collaborative_settlement(
+    pub fn propose_dlc_channel_collaborative_settlement(
         &self,
         channel_id: [u8; 32],
         accept_settlement_amount: u64,
@@ -84,22 +84,16 @@ where
             "Settling DLC channel collaboratively"
         );
 
-        tokio::task::spawn_blocking({
-            let sub_channel_manager = self.sub_channel_manager.clone();
-            let dlc_message_handler = self.dlc_message_handler.clone();
-            move || {
-                let (sub_channel_close_offer, counterparty_pk) = sub_channel_manager
-                    .offer_subchannel_close(&channel_id, accept_settlement_amount)?;
+        let (sub_channel_close_offer, counterparty_pk) = self
+            .sub_channel_manager
+            .offer_subchannel_close(&channel_id, accept_settlement_amount)?;
 
-                dlc_message_handler.send_message(
-                    counterparty_pk,
-                    Message::SubChannel(SubChannelMessage::CloseOffer(sub_channel_close_offer)),
-                );
+        self.dlc_message_handler.send_message(
+            counterparty_pk,
+            Message::SubChannel(SubChannelMessage::CloseOffer(sub_channel_close_offer)),
+        );
 
-                Ok(())
-            }
-        })
-        .await?
+        Ok(())
     }
 
     #[autometrics]
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -244,12 +244,11 @@ where
 }
 
 #[autometrics]
-pub(crate) async fn process_pending_dlc_actions(
+pub(crate) fn process_pending_dlc_actions(
     sub_channel_manager: Arc<SubChannelManager>,
     dlc_message_handler: &DlcMessageHandler,
-) -> Result<()> {
-    let messages =
-        tokio::task::spawn_blocking(move || sub_channel_manager.process_actions()).await?;
+) {
+    let messages = sub_channel_manager.process_actions();
 
     for (msg, node_id) in messages {
         let msg = Message::SubChannel(msg);
@@ -263,8 +262,6 @@ pub(crate) async fn process_pending_dlc_actions(
 
         dlc_message_handler.send_message(node_id, msg);
     }
-
-    Ok(())
 }
 
 pub fn dlc_message_name(msg: &Message) -> String {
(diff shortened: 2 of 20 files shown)


crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -258,12 +244,11 @@ where
 }
 
 #[autometrics]
-pub(crate) async fn process_pending_dlc_actions(
+pub(crate) fn process_pending_dlc_actions(
```
klochowicz, line 247: 🤔 I'm not sure about making this function `sync`.

I'm tempted to keep it `await` in the hope of using `rayon` parallelism eventually, because we don't need to send the messages one-by-one, especially if they go to different nodes.

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -71,7 +63,7 @@ where
     }
 
     #[autometrics]
-    pub async fn propose_dlc_channel_collaborative_settlement(
+    pub fn propose_dlc_channel_collaborative_settlement(
```
klochowicz, line 66: I can see the motivation of removing `async` if the whole body does a `spawn_blocking`, but I'm not sure whether it will lead us to better code down the line.

Setting aside potential misuse (the caller simply forgetting about `spawn_blocking`, which can happen), we are preventing potential optimisations.

🤔 I'm wondering whether it would be better to have 2 await points, one with task that does the `offer_subchannel_close`, and one that sends the message.

Perhaps we could schedule a meeting and align better on our vision?

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -23,34 +23,26 @@ where
     P: Send + Sync,
 {
     #[autometrics]
-    pub async fn propose_dlc_channel(
+    pub fn propose_dlc_channel(
```
klochowicz, line 26: same as elsewhere, I feel that having it `async` was better,  but we should have *multiple* `spawn_blocking` instead of one (if the whole body would stay like this, then we might as well remove it).

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -90,7 +88,7 @@ where
     }
 
     #[autometrics]
-    async fn match_event(&self, event: Event) -> Result<()> {
+    fn match_event(&self, event: Event) -> Result<()> {
```
klochowicz, line 91: not sure about this either (same concerns as before)

klochowicz (commented):
Any chance of splitting this PR into 2? 
1. All the code that doesn't need to be async at all, additions of `spawn_blocking` in routes etc.
2. Code where you  proactively removed async (particularly by moving `spawn_blocking` into a different layer)

There's a lot of changes that are non-controversial, but I'm not sure whether moving *away* from `async` is a good approach. I have outlined briefly my concerns, and I would like schedule a chat so we can talk this through.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -69,11 +71,6 @@ pub fn get_node_info() -> NodeInfo {
     NODE.get().inner.info
 }
 
-pub async fn update_node_settings(settings: LnDlcNodeSettings) {
-    let node = NODE.get();
-    node.inner.update_settings(settings).await;
-}
```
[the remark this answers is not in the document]
luckysori: It was never used!

crates/ln-dlc-node/src/node/dlc_channel.rs
luckysori: I don't think we need an `async` context to use `rayon` 🤔 

> I'm not sure about making this function sync.

I just don't see how it benefits from being `async`.

crates/ln-dlc-node/src/node/dlc_channel.rs
luckysori: > Setting aside potential misuse (the caller simply forgetting about spawn_blocking, which can happen).

I'm more afraid of the trap that we seem to have fallen into: assuming that it's okay to use `async` by default (particularly when some of our key dependencies (`rust-dlc` and `rust-lightning`) are almost exclusively sync.

> I'm wondering whether it would be better to have 2 await points, one with task that does the offer_subchannel_close, and one that sends the message.

This doesn't make sense to me. Here we have a block that is purely sync. So we should either do the work in a dedicated OS thread in `tokio`'s blocking thread pool. I think we agree on that.

How could splitting the work into two trips to the blocking thread pool help? You will certainly end up with an async task that is almost always idle (which is good), but you didn't actually need it in the first place (there is no async IO happening here) and you could have just done the same work in one trip. I've asked a [related question](https://discord.com/channels/500028886025895936/500336333500448798/1113786926059769916) in the `tokio` Discord. Let's see what they say.

luckysori:
I've rebased on top of `main` after https://github.com/get10101/10101/pull/732 merged.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
klochowicz: we should have used it instead of removing!

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -370,11 +368,10 @@ where
...
                 let min = time_forwardable.as_millis() as u64;
-                tokio::spawn(async move {
-                    let millis_to_sleep = thread_rng().gen_range(min..(min * 5));
-                    tokio::time::sleep(Duration::from_millis(millis_to_sleep)).await;
-                    forwarding_channel_manager.process_pending_htlc_forwards();
-                });
+
+                let millis_to_sleep = thread_rng().gen_range(min..(min * 5));
+                std::thread::sleep(Duration::from_millis(millis_to_sleep));
+                forwarding_channel_manager.process_pending_htlc_forwards();
```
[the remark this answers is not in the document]
luckysori: Not in changeset anymore.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -535,20 +532,27 @@ where
...
-                            .get_peer_node_ids()
-                            .iter()
-                            .any(|(id, _)| *id == target_node_id) {
-                            tracing::info!(%target_node_id, "Found connection to target peer. Continuing HTLCIntercepted event.");
-
-                            return;
-                        }
-                        tracing::debug!(%target_node_id, "Waiting for target node to come online.");
-                        tokio::time::sleep(Duration::from_secs(2)).await;
+                let deadline =
```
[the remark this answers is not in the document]
luckysori: Also not in changeset anymore.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
luckysori: Not with this PR though. Dead code that we remove can always be added later if we actually need it.

luckysori:
We're blocking this on https://github.com/get10101/10101/issues/730, because fixing the wallet takes priority and there will likely be conflicts with this PR.

klochowicz:
> We're blocking this on #730, because fixing the wallet takes priority and there will likely be conflicts with this PR.

it turned out that #730 requires using *sync* interface to bdk, so it should not be blocking after all. IMHO we can merge this.

klochowicz (approved):
thank you for your patience with this @luckysori whilst we evaluated whether we actually can bring back `async` interfaces for bdk first. 

LGTM!

luckysori:
I'm closing this as it's really outdated.

Now that we have a better understanding of how async and non-async code should coexist we can make incremental improvements to the codebase.

PR #727, opened 2023-06-01
Use blocking esplora client

luckysori:
How about this?


klochowicz (approved):
I was meant to do the same! glad I've noticed your PR first 👌

PR #728, opened 2023-06-01
chore: Don't block whole thread for processing DLC messages

klochowicz:
Only spawn_blocking the blocking part.

An alternative solution for the recently submitted PR, where I spawned a
dedicated OS thread.


luckysori (approved):
Let's try this first, sure!

PR #733, opened 2023-06-05
Fix `ln-dlc-node` tests

luckysori:
The fixes are not exhaustive, because I do believe that there is still some flakiness, but I've been able to:

- Turn some flaky tests into reliable tests.
- Fix some tests that were outright failing.

The only reason why we were able to get ourselves into this state is #355, which we must fix asap.


crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
```diff
@@ -72,11 +72,7 @@ async fn ln_collab_close() {
 
     // Give some time for the close transaction to be broadcast before trying to include it in a
     // block
-    tokio::time::sleep(Duration::from_secs(2)).await;
-
-    assert_eq!(payee.get_on_chain_balance().unwrap().confirmed, 0);
-    assert_eq!(payee.get_ldk_balance().available, 0);
-    assert_eq!(payee.get_ldk_balance().pending_close, invoice_amount);
```
klochowicz: can we keep these assertions if they are not troublesome?

I dislike the zealousness approach of only having assertions at the *end* of a test, I don't think that's helpful.

In this way, we clearly expected that at this very moment the balance should be `0`. 
I would very much prefer if we kept it like this.

klochowicz (changes requested):
I'm happy for it to go in without the last commit - I don't think removing assertions there is helpful (and they don't fail anymore anyway; when they failed recently, they have turned out to be useful)

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
klochowicz: I was referring the commit message where it was hinted: `Also, on principle we should aim to assert at the _end_ of a test.`.

IMHO we should rather fix our ldk balance to be more reliable rather than remove an useful assertion. 

also, the on-chain balance assertion removed had nothing to do with that unreliability referred in other part of the commit message.

klochowicz (approved):

klochowicz:
@luckysori it didn't quite fix the tests 🥲. I'm going to merge it anyway, as it's a step forward!

luckysori:
> @luckysori it didn't quite fix the tests 🥲. I'm going to merge it anyway, as it's a step forward!

Yeah, that's not surprising. The fixes were not related to #355.

These changes were for tests that were failing locally, either intermittently or persistently.

PR #734, opened 2023-06-05
Process lightning events asynchronously

luckysori:
I had to backport a few fixes that are only in version 0.0.115 of `rust-lightning`, hence the upgraded `rust-lightning` dependency.


klochowicz (approved):
really nice find with the async background processor 👌

Issue #692, opened 2023-05-23
Coordinator gets stuck eventually after processing a DLC message


holzeis:
When opening / closing a position on regtest as well as on mainnet the coordinator eventually gets stuck. Only some background tasks like node announcement, lightning wallet and onchain wallet sync are keep going.

But the http and p2p apis become unavailable / stuck.

holzeis:
Unfortunately, @bonomat broke the AWS system as well following these steps.

1. open position
2. mine 101 blocks
3. close position
4. closing...

PR #735, opened 2023-06-05
Run ln-dlc-node-tests job on every push

luckysori:
This (at least temporarily) replaces the separate job, so that we run these tests as often as possible. This is important because we run into https://github.com/get10101/10101/issues/692 very frequently when running these tests on CI, whereas it's very hard to reproduce locally.


.github/workflows/tests-ln-dlc-node.yml
```diff
@@ -1,60 +0,0 @@
-name: Slow ln-dlc-node tests
-
-on:
-  push:
-    branches:
-      - "staging"
```
klochowicz, line 6: as an alternative, it might be easier to just run add `-main` branch here too? so they are being run after merging, instead of on all open PRs?

luckysori: Well, my motivation is to see the effect of every push on all PRs. Rather than having to wait for it to merge.

luckysori:
Let's merge it for now. We can always revert it once we're done debugging.

bors r+

PR #741, opened 2023-06-06
Update `rust-dlc` to add even more logs

luckysori:
Very stupid logs that apply to the coordinator side of closing a DLC channel.

See https://github.com/get10101/rust-dlc/commit/37a9265.


luckysori:
We want logs.

PR #743, opened 2023-06-06
chore: Add more logging into rust-dlc

klochowicz:
See: https://github.com/get10101/rust-dlc/commit/a6fc2777d9b8cd1d0a8619c4ddbaad09e0756bee


klochowicz:
going ahead with it, as it just adds more logs.

PR #744, opened 2023-06-06
Add coordinator load test via `ln-dlc-node`

holzeis:
The motivation to add this test is to check that a hosted coordinator is able to trade repeatedly with an app. In this case we simply
simulate the app rather than using the `mobile/native` crate.

Notably, we have to interact with the coordinator's HTTP API. To do so we have to pretend like we got a match from the orderbook. We could involve the orderbook too, but it isn't strictly necessary at this stage.

The test in its current state is much slower than it needs to be. We should work towards replacing the hard-coded sleeps with calls to `wait_until`, wherever possible.

We've added a `load_tests` flag because we probably should refrain from running this type of test as frequently as the rest of the
`ln-dlc-node` tests. Currently it relies on the public regtest `coordinator` being up an running, and sufficiently funded. Eventually
we might want to run this kind of test, for example, before every release.

### DIFF
--- a/crates/ln-dlc-node/src/tests/load.rs
+++ b/crates/ln-dlc-node/src/tests/load.rs
@@ -1,0 +1,208 @@
+use crate::ln::app_config;
+use crate::node::Node;
+use crate::node::NodeInfo;
+use crate::node::PaymentMap;
+use crate::tests::init_tracing;
+use crate::tests::wait_until;
+use anyhow::Result;
+use coordinator::Coordinator;
+use coordinator::Direction;
+use dlc_manager::subchannel::SubChannelState;
+use dlc_manager::Storage;
+use std::borrow::Borrow;
+use std::sync::Arc;
+use std::time::Duration;
+
+mod coordinator;
+
+const ESPLORA_ORIGIN_PUBLIC_REGTEST: &str = "http://35.189.57.114:3000";
+
+#[tokio::test]
+async fn single_app_many_positions_load() {
+    init_tracing();
+
+    let coordinator = Coordinator::new_public_regtest();
+    let app = Arc::new(
+        Node::start_test(
+            "app",
+            app_config(),
+            ESPLORA_ORIGIN_PUBLIC_REGTEST.to_string(),
+        )
+        .unwrap(),
+    );
+
+    tokio::spawn({
+        let app = app.clone();
+        let coordinator_info = coordinator.info();
+        async move { keep_connected(app, coordinator_info).await }
+    });
+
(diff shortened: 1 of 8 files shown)


luckysori:
I've been working on getting this to work, as the hanging bug seems to be really hard to reproduce manually: I traded 20 times in a row or so without trouble.

holzeis (commented):
Thanks for taking over @luckysori

crates/ln-dlc-node/src/tests/load.rs
```diff
@@ -0,0 +1,177 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+
+mod coordinator;
+
+const ESPLORA_ORIGIN_PUBLIC_REGTEST: &str = "http://35.189.57.114:3000";
+
+#[cfg(feature = "load_tests")]
+#[tokio::test]
+async fn single_app_many_positions_load() {
```
klochowicz, line 23: I would consider moving it to an example eventually, because we could use `clap` to pass in arguments, e.g. the amount or even whether to try against local coordinator.

crates/ln-dlc-node/src/tests/load.rs
```diff
@@ -0,0 +1,177 @@
...
+    tokio::time::sleep(Duration::from_secs(5)).await;
+
+    // Operating the bitcoin node remotely is too much of a hassle. Just prepare the environment
+    // before running this test
+    coordinator
+        .open_channel(&app, 200_000, 100_000)
+        .await
+        .unwrap();
+
+    for n in 1..100 {
```
klochowicz, line 51: might be useful to at least put the total amount of positions into a const

klochowicz (approved):
thanks for automating this step! it's test code, so as long as it does the job the sleeps are fine 🤞

crates/ln-dlc-node/src/tests/load.rs
luckysori: I think it's a really good idea to create a dedicated binary for this. It's not really a test that should be treated like the others.

But it's not really an example either. I think it could just be a separate crate, right? Same goes for `fund`.

klochowicz: I don't mind having them as crates; it's also better than as a test as it gives us the same benefits. 
an example for me was a nice way of preventing its compilation during release etc. otherwise they'll need to get *always* compiled along with the workspace - I thought you cared about this part after seeing the feature flag, hence the "example" suggestion.

luckysori:
Can iterate on this later, let's merge it.

PR #746, opened 2023-06-07
Actually assert on DLC channel state in tests

luckysori:
We somehow forgot to assert on the value returned by the `matches!` macro :facepalm:


klochowicz (approved):
LGTM!

PR #748, opened 2023-06-08
Take advantage of parking_lot Mutex

klochowicz:
See commit messages for details;
TL;DR: it's faster, more convenient, widely used and can detect deadlocks.


luckysori (approved):
Nice, let's do this.

klochowicz:
> Nice, let's do this.
cheers.

I'll do the same for that particular peer_state lock in `rust-lightning` fork we're using.

PR #749, opened 2023-06-08
chore: Use rust-lightning fork with parking_lot

klochowicz:
An attempt to try to pin down the reason we're running into deadlock in rust-dlc
rust-lighting fork.


luckysori (approved):
Let's try it!

PR #750, opened 2023-06-08
Add `BdkActor`

### LINKED ISSUE #730: Refactor away Mutex around BDK
It's a huge problem that we can block the whole app because other part is waiting for the Mutex to be unlocked.

Moving the BDK wallet inside an entity that communicates with other threads via channel can avoid this situation. This is essentially the "actor" pattern.
This is what we've done in ItchySats, and we never had a problem with BDK there.

note: There might not be a need to introduce a whole framework to deal with it, it might be enough to hand-craft it: https://ryhl.io/blog/actors-with-tokio/

luckysori:
Fixes #730.

An actor that manages the `bdk::Wallet` resource. It allows us to use the wallet whilst the inevitably expensive on-chain sync is happening in the background. But only for those things that can be cached, such as the balance, wallet history and addresses.

We would like to use the async version of `EsploraBlockchain`, but https://github.com/bitcoindevkit/bdk/issues/165 is still an issue.

The only hacky bit stems from the fact that we still have to implement certain non-async foreign traits and to access any `bdk::Wallet` resource we now have to go via async methods, since the actor is async.

To do so, at some point we have to call an async function "from a sync context". The natural way to do so (for me) would be to use
`runtime.block_on`, which schedules an async task and waits for it to resolve, blocking the thread. *But*, these non-async foreign trait methods are actually called elswhere in _async_ contexts. This leads to `runtime.block_on` panicking because `tokio` is trying to prevent us from blocking a thread in an async context. This is analogous to us misusing async and doing an expensive CPU-bound computation in an async context, but here `tokio` is able to "aid" us since `block_on` is provided by `tokio`.

The solution is to use the following pattern:

```rust
tokio::task::block_in_place(|| {
    tokio::runtime::Handle::current().block_on(async {
        // async code
    })
});
```

From the documentation of `block_in_place`, we are able to run "the provided blocking function on the current thread without blocking the executor". We therefore avoid the panic, as we no longer block the executor when calling `block_on`.

This has one final side-effect, which is that all `ln-dlc-node` async tests now need to go back to using the `multi_thread` flavour.
`block_in_place` works by moving all scheduled tasks to a different worker thread and without `multi_thread` there is only one worker thread, so it just panics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,6 +467,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "89b2fd2a0dcf38d7971e2194b6b6eebab45ae01067456a7fd93d5547a61b70be"
 
 [[package]]
+name = "catty"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dcf0adb3cc1c06945672f8dcc827e42497ac6d0aff49f459ec918132b82a5cbc"
+dependencies = [
+ "spin 0.9.8",
+]
+
+[[package]]
 name = "cc"
 version = "1.0.79"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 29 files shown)


holzeis:
@luckysori @klochowicz Although I think its a reasonable refactor, I wouldn't do that now and rather wait for bdk 1.0.0. I am a bit afraid of regressions.

holzeis:
@luckysori Can we close that PR?

PR #751, opened 2023-06-09
Add background task to report deadlocks

luckysori:
From the `parking_lot` documentation it looks like we need to add something like this to know what deadlocks have been detected.


klochowicz:
thank you for spotting the missing link! ❤️

PR #752, opened 2023-06-09
Build `rust-lightning` with `backtrace` flag

luckysori:
We are temporarily activating this by default to investigate a deadlock.

We have to use a `nightly` toolchain to be able to use this flag.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,6 +4,8 @@
 rustflags = [
   "--cfg",
   "tokio_unstable",
+  "--cfg",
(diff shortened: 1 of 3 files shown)


luckysori:
Building the corresponding Docker image for the coordinator fails: https://github.com/get10101/10101/actions/runs/5218029932/jobs/9418438307#step:6:478.

I am able to build it locally so for now I will:

- Push the image manually to the correct tag to keep testing.
- Revert this patch and just apply it locally when necessary.

Issue #768, opened 2023-06-13
Upgrade to latest `rust-dlc` and `rust-lightning`


luckysori:
With our debugging frenzy around https://github.com/get10101/10101/issues/692 we diverged significantly from upstream[^1] `rust-dlc` and `rust-lightning`.

We introduced our own fix in https://github.com/get10101/10101/pull/756, but since then Tibo has already fixed the same issue on latest `rust-dlc` and `rust-lightning`. By upgrading to latest, we will also be able to use the latest API for force-closing LN-DLC channels, as well as get some fixes surrounding that use-case. Furthermore, we will re-introduce some other fixes related to reconnecting during a DLC channel protocol.

[^1]: By upstream we of course mean the branches and forks maintained by Tibo.

holzeis:
@luckysori can you point me to tibos fix of the deadlock? I am a bit worried of a regression, removing our fix (that proved to be working) for a different fix.

Issue #664, opened 2023-05-19
Follow-up work to #612


luckysori:
Look at the PR description in https://github.com/get10101/10101/pull/612 for context.

This is primarily a reminder for myself that I don't want to forget about the corners we cut when working on #612.

holzeis:
I am removing MVP as the (my) follow up remarks are definitely not critical for MVP :)

luckysori:
[Tibo's recent updates](https://matrix.to/#/!YOUZSBmuJuaGzmxzaA:matrix.org/$Q9MYQy4v7LGbxZfibE2XZy-RYYo5KIlkVZkZjx3cOy4?via=matrix.org) to the forks we are using of `rust-dlc` and `rust-lightning` should be very relevant here.

Issue #663, opened 2023-05-19
Cannot broadcast transaction spending `StaticPaymentOutput`


luckysori:
We get a weird failure when calling `broadcast_transaction`. The very same transaction can be published using `sendrawtransaction` via `bitcoind`.

This only affects the party who did _not_ trigger the force-close of the LN-DLC channel.

holzeis:
related to #591

PR #771, opened 2023-06-14
Stabilise key dependencies

luckysori:
Fixes https://github.com/get10101/10101/issues/766 (they pass locally :D).
Fixes https://github.com/get10101/10101/issues/768.
Fixes https://github.com/get10101/10101/issues/664.
Fixes https://github.com/get10101/10101/issues/663 (we don't run into it and the app gets their money back [^1]).

The main motivation is to stay on top of the latest changes to `rust-lightning` and `rust-dlc`, whilst also definitively removing some of the things that we temporarily added during our latest bug hunt.

Some patches were also removed which might cause some controversy:

- https://github.com/get10101/rust-dlc/commit/ce78c25.
- https://github.com/get10101/rust-dlc/commit/cf3d1eb.
- https://github.com/get10101/rust-dlc/commit/304e85d.

In general reviewers are encouraged to take a look at the differences between these branches:

- For `rust-dlc` we are now using https://github.com/get10101/rust-dlc/commit/8fa3692 [^2] instead of https://github.com/get10101/rust-dlc/commit/2322288.

- For `rust-lightning` we are now using https://github.com/get10101/rust-lightning/commit/c65cb1c5 [^3] instead of https://github.com/get10101/rust-lightning/commit/939ebb43.

The differences are not so obvious because we diverged quite a bit. We still want to keep certain commits that haven't been upstreamed yet.

Other notable changes
---------------------

- Adapting to the new API to force-close the LN-DLC channel provided by `rust-dlc`. As such, we've modified some code around that and updated the `force_close_ln_dlc_channel` test accordingly.

- Introducing `wait_until_dlc_channel_state` to replace a lot of eager sleeps which led to slower and flakier tests.

- Enabling the configuration of the `dlc_manager_periodic_check_interval` and `sub_channel_manager_periodic_check_interval` viat the `LnDlcNodeSettings`. Originally this was added to be able to circumvent another issue, but the issue has now been fixed in
`rust-dlc`.

- Ensuring that we drop the `LnDlcNodeSettings` read lock as soon as possible so that we don't hold onto it during the sleeps of certain periodic tasks. This fixes a bug as in a previous version of this patch we were not able to update the settings in the tests, because there was always at least one read lock being held!

[^1]: The test can still be improved by making sure that the oracle unlocks a CET that pays to both parties. Atm we use an event whose attestation results in the coordinator getting the whole DLC output. This means that we are only verifying that the part of the LN channel that belongs to the app is returned.
[^2]: This branch is temporarily named `feature/ln-dlc-channels-114-tibo`.
[^3]: This branch is temporarily named `split-tx-experiment-114-tibo`.

### DIFF
--- a/crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
+++ b/crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
@@ -1,12 +1,11 @@
-use crate::node::dlc_channel::process_pending_dlc_actions;
+use crate::node::dlc_channel::sub_channel_manager_periodic_check;
 use crate::node::Node;
 use crate::tests::dummy_contract_input;
 use crate::tests::init_tracing;
-use crate::tests::wait_until;
+use crate::tests::wait_until_dlc_channel_state;
+use crate::tests::SubChannelStateName;
 use anyhow::Context;
 use bitcoin::Amount;
-use dlc_manager::subchannel::SubChannelState;
-use dlc_manager::Storage;
 use std::time::Duration;
 
 #[tokio::test]
@@ -47,25 +46,19 @@ async fn reconnecting_during_dlc_channel_setup() {
         .await
         .unwrap();
 
-    tokio::time::sleep(Duration::from_secs(2)).await;
-    coordinator.process_incoming_messages().unwrap();
-
-    app.reconnect(coordinator.info).await.unwrap();
-
-    let sub_channel = wait_until(Duration::from_secs(30), || async {
-        let sub_channels = coordinator
-            .dlc_manager
-            .get_store()
-            .get_offered_sub_channels()?;
-
-        let sub_channel = sub_channels
-            .iter()
-            .find(|sub_channel| sub_channel.counter_party == app.info.pubkey);
-
-        Ok(sub_channel.cloned())
-    })
+    // Process the app's `Offer`
+    let sub_channel = wait_until_dlc_channel_state(
+        Duration::from_secs(30),
+        &coordinator,
+        app.info.pubkey,
+        SubChannelStateName::Offered,
+    )
     .await
     .unwrap();
+
+    // This reconnect does not affect the protocol because the `Offer` has already been delivered to
+    // the coordinator
+    app.reconnect(coordinator.info).await.unwrap();
 
     coordinator
         .accept_dlc_channel_offer(&sub_channel.channel_id)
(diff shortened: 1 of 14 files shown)


holzeis:
> https://github.com/get10101/rust-dlc/commit/cf3d1eb.
> https://github.com/get10101/rust-dlc/commit/304e85d.

I am not against reverting these changes, but what does that mean for our existing production coordinator? Do we need to clear the dlc db for the mainnet coordiantor to work with these changes?

holzeis:
> Fixes https://github.com/get10101/10101/issues/766 (they pass locally :D).

Since the tests are not passing in the CI, I guess your changes are not fixing but only improving #766?

crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
```diff
@@ -76,56 +69,47 @@ async fn reconnecting_during_dlc_channel_setup() {
     app.reconnect(coordinator.info).await.unwrap();
```
holzeis, line 69: 🤔 Can we be sure here, that the coordinator hasn't processed the accept message before the reconnect?

luckysori:
> > In general reviewers are encouraged to take a look at the differences between these branches:
> > 
> > * For rust-dlc we are now using [get10101/rust-dlc@8fa3692](https://github.com/get10101/rust-dlc/commit/8fa3692) [2](https://github.com/get10101/10101/pull/771#user-content-fn-2-bf02b6925d0d4225f01997aa004b35f1) instead of [get10101/rust-dlc@2322288](https://github.com/get10101/rust-dlc/commit/2322288).
> > * For rust-lightning we are now using [get10101/rust-lightning@c65cb1c5](https://github.com/get10101/rust-lightning/commit/c65cb1c5) [3](https://github.com/get10101/10101/pull/771#user-content-fn-3-bf02b6925d0d4225f01997aa004b35f1) instead of [get10101/rust-lightning@939ebb43](https://github.com/get10101/rust-lightning/commit/939ebb43).
> 
> The referenced change seem insignificant, unless the real changes are hiding in the changed dependency to our get10101 fork? Am I missing something?

The linked patches are just heads of branches. The branches themselves are referenced in the footnotes and also if you follow the links:

image

luckysori:
> > Fixes #766 (they pass locally :D).
> 
> Since the tests are not passing in the CI, I guess your changes are not fixing but only improving #766?

Yeah, at the moment we appear to still have at least one other flaky test. I will try to fix it tomorrow.

holzeis (approved):
LGTM 👍 

However, I did not review all changes implicit through the upgraded dependencies.

holzeis:
@luckysori https://github.com/get10101/rust-lightning/commit/6b059a7222997015141fb9aa44788f4a43732957#commitcomment-118021986

crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
```diff
@@ -47,26 +46,20 @@ async fn reconnecting_during_dlc_channel_setup() {
...
+        &coordinator,
+        app.info.pubkey,
+        SubChannelStateName::Offered,
+    )
     .await
     .unwrap();
 
+    // This reconnect does not affect the protocol because the `Offer` has already been delivered to
+    // the coordinator
+    app.reconnect(coordinator.info).await.unwrap();
```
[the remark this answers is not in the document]
luckysori: I honestly would have removed this reconnect from the original test. I didn't quite understand the purpose.

crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
luckysori: We can be sure in the sense that nothing happens unless we call `process_incoming_messages` (bad name because it also sends messages that are queued up) implicitly through `wait_until_dlc_channel_state` or `periodic_check` on the `SubChannelManager` implicitly through `sub_channel_manager_periodic_check`.

But we can certainly check that the message has not arrived yet, by asserting on the state of the app (should be `Offered`?) or maybe checking the coordinator's message queue manually.

crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
```diff
@@ -76,56 +69,47 @@ async fn reconnecting_during_dlc_channel_setup() {
     app.reconnect(coordinator.info).await.unwrap();
 
     // Instruct coordinator to re-send the accept message
-    process_pending_dlc_actions(
+    sub_channel_manager_periodic_check(
```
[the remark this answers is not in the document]
luckysori: > An assertion that a message is pending would be nice here. Otherwise we are just implying that this will resend the accept message

Covered with my response above.

> Also shouldn't it be the confirm message? The accept message is sent by the app.

We never adapted these tests, so we are not fully mimicking what we do in production. That is, the app proposes the channel open and close in most of these tests (if not all).

luckysori:
I've merged #772 and #773 into this PR after getting approvals there.

crates/ln-dlc-node/src/tests/dlc/dlc_setup_with_reconnects.rs
holzeis: so isn't there a background task running automatically trying to process any incoming messages? If so, I think the asserting on the state should be fine. But, I think the `SubChannelState` of the app should be `Confirmed` after the coordinator processes the accept dlc channel offer?

btw. I noticed that in the tests we have the app and the coordinator roles inverted. But I guess that shouldn't have an implications.

luckysori:
> > > Fixes #766 (they pass locally :D).
> > 
> > 
> > Since the tests are not passing in the CI, I guess your changes are not fixing but only improving #766?
> 
> Yeah, at the moment we appear to still have at least one other flaky test. I will try to fix it tomorrow.

@holzeis: I've opened #774 and #775 to capture the two tests that are still flaky. The one related to the JIT payment is way more likely to happen as we could have probably guessed. I think we are safe to close #766 with this PR since it was too general anyway and it has fixed other tests that were intermittently failing because we used static sleeps.

luckysori:
> > > [get10101/rust-dlc@cf3d1eb](https://github.com/get10101/rust-dlc/commit/cf3d1eb).
> > > [get10101/rust-dlc@304e85d](https://github.com/get10101/rust-dlc/commit/304e85d).
> > 
> > 
> > I am not against reverting these changes, but what does that mean for our existing production coordinator? Do we need to clear the dlc db for the mainnet coordiantor to work with these changes?
> 
> I think before upgrading we should force-close all the LN-DLC channels that we have. And then purge the DLC database.

I've introduced https://github.com/p2pderivatives/rust-dlc/commit/3b277dcd7d573eac615fe2cecf0325c45413bcb9 which _should_ allow us to update mainnet without closing the problematic positions.

PR #781, opened 2023-06-15
Use address caches in `LnDlcWallet`

### LINKED ISSUE #776: Always use cached addresses from on-chain wallet
This is to avoid blocking in various places whilst we wait for the on-chain sync to finish. It's particularly important when opening a position.

It's just a band-aid fix until we solve the problem of the wallet blocking all other wallet operations until the on-chain sync is done.

We could opt to simply re-use the same address every time. We leave it up to the implementer.

luckysori:
Fixes #776.

By caching these addresses we are preventing situations where getting an address from the wallet takes a long time because the on-chain sync is ongoing (and therefore has the mutex write lock).

The downside is that we can end up reusing addresses if these caches are accessed more than once during a sync.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -355,9 +355,8 @@ async fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
     Ok(())
 }
 
-pub fn get_new_address() -> Result<String> {
-    let address = NODE.get().inner.get_new_address()?;
-    Ok(address.to_string())
+pub fn get_new_address() -> String {
+    NODE.get().inner.get_new_address().to_string()
 }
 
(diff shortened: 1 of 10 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -356,7 +356,7 @@ async fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
 }
 
 pub fn get_new_address() -> Result<String> {
```
holzeis, line 358: 🤔 Is this returning a result because of a trait, or could we make that infailable?

holzeis (approved):
Nice 👍

mobile/native/src/ln_dlc/mod.rs
luckysori: Good catch!

PR #784, opened 2023-06-16
Prefer unused addresses to new ones (temporarily)

luckysori:
AFAIK generating a new address means that the on-chain wallet must look for that address on the blockchain. Thus, we must only generate a new address if we are sure that we are about to use it.

As such, it is a bad idea to generate new addresses periodically. This would quickly become untenable, as the on-chain sync would grow more and more expensive and slow without good reason.

At the same time, it is best to use new addresses as it ensures that we don't run into address reuse, which is a bad practice as it leads to much worse privacy on-chain. Unfortunately, due to current limitations of `bdk::Wallet` it is not possible to ask the wallet for a new  (or even an unused) address during the on-chain sync. For this reasons we decided to introduce an address cache.

Given all this, it's bad to keep a new-address cache so it has been removed.

Furthermore, given that calls to `get_address(AddressIndex::LastUnused)` already generate new addresses if necessary and that we've already accepted that we are okay with using (probably) unused addresses by introducing the address cache, we have decided to _only_ allow calls to get_address(AddressIndex::LastUnused). This simplifies the code and ensures that we can always rely on the only cache we want to maintain i.e. we never have to wait for the wallet to finish to get an address.

---

We can revisit this change once we upgrade to `bdk:1.0.0-alpha.1` and rewrite our wallet to only take a write lock for a short time to update the wallet database.

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -79,7 +79,7 @@ pub fn router(
             "/api/register_invoice/:target_node",
             post(register_interceptable_invoice),
         )
-        .route("/api/newaddress", get(get_new_address))
+        .route("/api/newaddress", get(get_unused_address))
         .route("/api/node", get(get_node_info))
         .route("/api/invoice", get(get_invoice))
         .route("/api/orderbook/orders", get(get_orders).post(post_order))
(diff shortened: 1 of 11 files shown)


holzeis (approved):
LGTM

PR #785, opened 2023-06-16
Simplify maker binary

luckysori:
The maker currently only creates orders. Letting it do anything else for the time being does not make sense.

### DIFF
--- a/maker/src/bin/maker.rs
+++ b/maker/src/bin/maker.rs
@@ -1,24 +1,9 @@
-use anyhow::Context;
 use anyhow::Result;
-use diesel::r2d2;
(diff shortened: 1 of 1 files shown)


holzeis (approved):
Thank you 👍

bonomat (commented):
Mhm... I always used the maker to open channels when the LND setup failed :/ 

Can we undo this please?

holzeis:
@bonomat how about reverting that change and adding a cli argument to the maker to not start the ldk node? We also removed the ldk node since we did not want to have another service querying the blockstream.info api constantly from the same ip address as the coordinator.

PR #789, opened 2023-06-16
Revert "fix: Fix error-prone periodic checks"

luckysori:
This reverts commit d2202fb469233bc585b5a0c7a8a8bc1eb1a0419b.

The fix was causing certain tests to fail, so it's probably not safe to use.


holzeis (approved):
I'll trust you on that one! :)

PR #794, opened 2023-06-18
Preparation for Q3: Add new issues to new board

bonomat:
Same should be done in other repositories


klochowicz (approved):

klochowicz:
@bonomat - I will do that in other repos too.

PR #800, opened 2023-06-20
Allow concurrent orders in maker

klochowicz:
Maker creates multiple *same* orders based on the latest BitMEX quote.
Their amount is configurable.
When a new quote is received, all orders are updated.

### DIFF
--- a/maker/src/cli.rs
+++ b/maker/src/cli.rs
@@ -40,6 +40,10 @@ pub struct Opts {
     /// If enabled logs will be in json format
     #[clap(short, long)]
     pub json: bool,
+
+    /// Amount of concurrent orders (buy,sell) that maker will create at a time
+    #[clap(long, default_value = "5")]
+    pub concurrent_orders: usize,
 }
 
 #[derive(Debug, Clone, Copy, clap::ValueEnum)]
--- a/maker/src/trading/bitmex_client.rs
+++ b/maker/src/trading/bitmex_client.rs
@@ -12,6 +12,8 @@ use std::time::Duration;
 use time::OffsetDateTime;
 use trade::ContractSymbol;
 
+const BITMEX_RECONNECT_INTERVAL: Duration = Duration::from_secs(10);
+
 pub async fn bitmex(network: Network) -> impl Stream<Item = Result<Quote, Error>> + Unpin {
     let stream = stream! {
         loop {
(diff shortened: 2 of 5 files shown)


maker/src/cli.rs
```diff
@@ -40,6 +40,10 @@ pub struct Opts {
     /// If enabled logs will be in json format
     #[clap(short, long)]
     pub json: bool,
+
+    /// If enabled logs will be in json format
```
holzeis, line 44: 🔧 please adapt the docs

maker/src/trading/mod.rs
```diff
@@ -26,49 +31,54 @@ pub async fn run(orderbook_url: Url, maker_id: PublicKey, network: Network) -> R
...
-            quote.ask(),
-            Direction::Short,
-            maker_id,
-            last_ask,
-            dec!(1000),
-        )
-        .await;
+    };
+
+    while let Some(quote) = price_stream.try_next().await? {
```
holzeis, line 48: 🤔 Beforehand we were firstly adding the new order and then deleting the old ones. Not sure if that was intentional, but this change is reversing the order now.

holzeis (approved):
Let's give it a try!

maker/src/trading/mod.rs
klochowicz: yes, it is intentional - we should first remove the outdated order (we don't want anyone to take it, because the price has changed in the event). 

I will mention this in the commit message to be clear about it.

maker/src/trading/bitmex_client.rs
```diff
@@ -34,9 +36,8 @@ pub async fn bitmex(network: Network) -> impl Stream<Item = Result<Quote, Error>
                 }
             }
 
-            let seconds = 10;
-            tracing::warn!("Disconnected from BitMEX. Reconnecting to BitMEX in {seconds}");
-            tokio::time::sleep(Duration::from_secs(seconds)).await;
+            tracing::warn!("Disconnected from BitMEX. Reconnecting to BitMEX in {seconds} seconds", seconds = BITMEX_RECONNECT_INTERVAL.as_secs());
```
luckysori, line 39: 🔧Shouldn't this be like this?

```rust
tracing::warn!("Disconnected from BitMEX. Reconnecting to BitMEX in {} seconds", BITMEX_RECONNECT_INTERVAL.as_secs());
```

I don't know what the current code does, because it seems to be combining string interpolation with `tracing` attributes.

maker/src/trading/mod.rs
luckysori: Could we split the change into two then? Otherwise it looks like the change is purely a refactor if you look at the commit title.

luckysori (approved):
LGTM.

maker/src/trading/bitmex_client.rs
klochowicz: I don't mind changing.

maker/src/trading/mod.rs
klochowicz: I don't think it's worth it, but I'll try to be more atomic in the future!

luckysori: At least let's rename the commit so that we can more easily spot the change in the future, since we know it's not a refactor.

klochowicz: suggestions welcome!

luckysori: `fix(maker): Delete stale orders before sending new ones`.

PR #802, opened 2023-06-20
Add .gitkeep to data/maker so that run-detached-maker works properly

Restioson:
When setting up the project for the first time in a while, the only issue I came across was that the `run-maker-detached` recipe would try to redirect its logs to `data/maker/regtest.log`, but `data/maker` did not exist. This PR adds a placeholder file .gitkeep to that directory to keep it around, as directories cannot be checked in to git empty.

cc @holzeis since we discussed this on element

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@ mobile/ios/Runner/libnative.a
 
 # ignore data runtime data
 data/**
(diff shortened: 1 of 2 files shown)


holzeis (approved):
👍

klochowicz (approved):

PR #805, opened 2023-06-21
Return early if there are no spendable outputs when handling `SpendableOutputs` event

luckysori:
Otherwise we return an error that is subsequently logged.

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -396,6 +396,10 @@ where
                     })
                     .collect::<Vec<_>>();
 
+                if ldk_outputs.is_empty() {
+                    return Ok(());
+                }
+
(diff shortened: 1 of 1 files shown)


klochowicz (approved):

mobile/native/src/logger.rs
```diff
@@ -28,8 +28,8 @@ pub fn log_base_directives(env: EnvFilter, level: LevelFilter) -> Result<EnvFilt
         // set to debug to show ldk logs (they're also in logs.txt)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
+        .add_directive("lightning=trace".parse()?)
```
klochowicz, line 31: hey, I'm working on this ticket! ;D

luckysori: Good to know! I didn't mean to push that here anyway :D

PR #808, opened 2023-06-21
chore: Log LDK logs on TRACE

klochowicz:
Logging with trace on the app might impact mobile storage; we hope to be able to
revert this change before the next TestFlight release.

### DIFF
--- a/coordinator/src/logger.rs
+++ b/coordinator/src/logger.rs
@@ -25,8 +25,7 @@ pub fn init_tracing(level: LevelFilter, json_format: bool, tokio_console: bool) 
         .add_directive("rustls=warn".parse()?)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
-        .add_directive("lightning::ln::peer_handler=debug".parse()?)
-        .add_directive("lightning::chain=info".parse()?)
+        .add_directive("lightning=trace".parse()?)
         .add_directive("ureq=info".parse()?);
 
     let mut filter = if tokio_console {
--- a/mobile/native/src/logger.rs
+++ b/mobile/native/src/logger.rs
@@ -28,8 +28,7 @@ pub fn log_base_directives(env: EnvFilter, level: LevelFilter) -> Result<EnvFilt
         // set to debug to show ldk logs (they're also in logs.txt)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
-        .add_directive("lightning::ln::peer_handler=debug".parse()?)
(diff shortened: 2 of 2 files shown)


luckysori (approved):
LGTM, thanks!

mobile/native/src/logger.rs
```diff
@@ -28,8 +28,7 @@ pub fn log_base_directives(env: EnvFilter, level: LevelFilter) -> Result<EnvFilt
         // set to debug to show ldk logs (they're also in logs.txt)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
-        .add_directive("lightning::ln::peer_handler=debug".parse()?)
-        .add_directive("lightning::chain=info".parse()?)
+        .add_directive("lightning=trace".parse()?)
```
holzeis, line 31: 🔧 We are only printing debug logs to FLog. This change will increase the logs on the mobile app.

coordinator/src/logger.rs
```diff
@@ -25,8 +25,7 @@ pub fn init_tracing(level: LevelFilter, json_format: bool, tokio_console: bool)
         .add_directive("rustls=warn".parse()?)
         .add_directive("sled=warn".parse()?)
         .add_directive("bdk=warn".parse()?) // bdk is quite spamy on debug
-        .add_directive("lightning::ln::peer_handler=debug".parse()?)
-        .add_directive("lightning::chain=info".parse()?)
+        .add_directive("lightning=trace".parse()?)
```
holzeis, line 28: 🤔  I guess that is only temporary, as such I would have opted to make use of the RUST_LOG env variable instead of changing the code.

Simply changing the config in our cloud_conf might have been sufficient?

luckysori: I'm not sure if it should be _that_ temporary. We're still in a phase during which we are discovering bugs often. If we run into an elusive bug and don't have enough context we will be kicking ourselves.

mobile/native/src/logger.rs
luckysori: Are you saying that this won't work because the app ignores trace logs?

holzeis: yes

klochowicz: I guess we should change that? :) the app is the only consumer of this library, so we might as well have a single place to set the logs. 

of course the trace logs are also seen in the console when you're debugging locally from the app (whether it's native, simulator, or phone connected with usb), so the change makes sense anyway.

luckysori: Fixed here: https://github.com/get10101/10101/pull/826/commits/27f23d230796447b35729ca568b5c7c9f3b6fbd4.

PR #812, opened 2023-06-21
QoL dev improvements

klochowicz:
See commit messages for details

### DIFF
--- a/mobile/lib/features/wallet/share_invoice_screen.dart
+++ b/mobile/lib/features/wallet/share_invoice_screen.dart
@@ -84,8 +84,16 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
                         });
 
                         final router = GoRouter.of(context);
+                        const localCoordinatorPubkey =
+                            "02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9";
                         try {
-                          await openCoordinatorChannel(config.host);
+                          if (config.coordinatorPubkey == localCoordinatorPubkey) {
+                            await payInvoiceWithFaucet(widget.invoice);
+                          } else {
+                            // For remote coordinator, we open the channel
+                            // directly with the coordinator as it's less error-prone
+                            await openCoordinatorChannel(config.host);
+                          }
                           // Pop both create invoice screen and share invoice screen
                           router.pop();
                           router.pop();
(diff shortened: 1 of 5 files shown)


Restioson (approved):
LGTM, I was wondering what the difference between the fund button & instructions here were. Though, both worked.

mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -84,8 +84,16 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
...
 
                         final router = GoRouter.of(context);
+                        const localCoordinatorPubkey = "02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9";
                         try {
-                          await openCoordinatorChannel(config.host);
+                          if ( config.coordinatorPubkey == localCoordinatorPubkey) {
+                            await payInvoiceWithFaucet(widget.invoice);
+                          }
+                          else {
+                            // For remote coordinator, we open the channel
```
holzeis, line 93: 🔧 I wouldn't go for a different setup on our regtest setup than on our localhost setup. IMHO that's only confusing as we would have 3 different ways on how the wallet gets funded. Ideally we should always go the pay invoice path to have it as close as possible to production.

If we feel we can go that path for our local setup, why not also change it for our regtest setup?

klochowicz: This way brings it closer to production in local testing, so it's a major step forward.

I clearly explained that this is needed for 
- specifying custom amount to be sent,
- seeing it in transaction history.

(I need the first one for testing a bug I'm working on, so I'm proceeding).

As for not changing it for regtest - I wanted to be conservative.  Earlier I opted to do the same for the regtest, but it got changed due to us closing the channels with LND randomly on and the team voted for having something that worked more closely. I have no capacity to debug it, nor we're actively pushing 10101-test right now, so this is low on my priority list.
I would be happy if this happened someday though, because I feel it was a hack that we went with bypassing paying the invoice at all.

TL;DR: I'm removing this terrible workaround from the most used path, so we can test more. I hope that we can remove it everywhere.

luckysori:
> Seems like `just fund` takes quite a while; is there a way to make sure it only happens on first run?

I hear you. I would maybe define two recipes:

```just
all-faucet: services gen native fund run
all: services gen native run
```

The names aren't great though D: Alternatively you can just run `just services gen native run` yourself. The explicit recipe is just a convenience

klochowicz:
another alternative would be to move the fund process to the background with `&`. 

overall, I don't think it changes much in the picture compared to the generation of FRB, or actually rebuilding the app (on mac, this takes ages ;/) 

feel free to define your own ones ;D 

as for naming: I would prefer having *all* having it *all*, otherwise the name makes little sense.

Issue #807, opened 2023-06-21
Flaky test `force_close_ln_dlc_channel`


luckysori:
This test sometimes hangs[^1] on CI. I have not been able to reproduce the hang locally.

It first happened on CI here: https://github.com/get10101/10101/actions/runs/5306914796/jobs/9604966058#step:8:1648.

[^1]: This hang is different from #355 in the sense that we are not in the middle of opening or closing the DLC.

luckysori:
Looks like it hasn't happened for a week and [we added some logs](https://github.com/get10101/10101/pull/929) to analyse it in case it resurfaces.

Closing for now; reopen if we see it happen again!

PR #814, opened 2023-06-21
Do not block `tokio` runtime when syncing on-chain

luckysori:
One patch addresses production and the other addresses the tests.

~This could help with #807, but I'm not convinced. [edit]: Indeed it did not help, the same test still hangs on this very PR~.

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -47,6 +47,7 @@ use std::sync::Arc;
 use tokio::sync::broadcast;
 use tokio::sync::mpsc;
 use tokio::sync::RwLock;
+use tokio::task::spawn_blocking;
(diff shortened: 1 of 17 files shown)


holzeis (commented):
LGTM, but given that it does not fix the actual issue, do we still want to merge that?

luckysori:
> LGTM, but given that it does not fix the actual issue, do we still want to merge that?

I would merge both patches as I think what we are currently doing is incorrect.

But specifically c9c9bb85ee97063d771940bc790fda5d394bb0d2 is a fix for production, not for the linked ticket.

holzeis (approved):

klochowicz (approved):

PR #816, opened 2023-06-22
chore: Remove direct channel open workaround for public regtest

klochowicz:
We used this to circumvent the random channel closures / no route errors on
regtest during 10101-test app testing.
Now that the testing is back on mainnet and we focus on stability of the app,
the underlying issues should be tackled properly and this workaround removed.

### DIFF
--- a/mobile/lib/features/wallet/share_invoice_screen.dart
+++ b/mobile/lib/features/wallet/share_invoice_screen.dart
@@ -84,16 +84,8 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
                         });
 
                         final router = GoRouter.of(context);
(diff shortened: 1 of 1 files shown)


klochowicz:
I created this PR after @holzeis feedback in another PR where I removed the workaround for the local dev setup (to bring it closer to production). 

I believe we should investigate why are we closing channels and sometimes failing to find a route on public regtest, as these issues might be the same as we're experiencing on mainnet. 
Merging this will make these issues visible on public regtest again, but at least possible to be addressed.

The fact that we're not onboarding users to the test app at this stage makes it a convenient time to do so.

Thoughts?

luckysori (approved):

holzeis (approved):

PR #817, opened 2023-06-22
chore: Close stale issues after defined period of inactivity

klochowicz:
Issues and PRs without activity will be marked as stale after 30 days, and
closed if the label has not been removed (or any other action happened, such as
comment etc.) after 5 more days.


klochowicz:
If we decide that it does its job well, we can roll it out to other repos.

holzeis (approved):

Issue #714, opened 2023-05-30
Share on twitter encourages the user to close the app while an order is executed


holzeis:
Although our app should be able to withstand a lost in connection during order execution, it's still a bit risky to provoke that behavior actively through the app.

Ideally we ask the user to share on twitter once the order has been executed.

klochowicz:
I would have just hid the button for now if it causes any problems. 
Later makes no sense, because we might need to put more effort into it than it's worth 🤔

bonomat:
We could just add the button to the open position and leave it there at all times.

holzeis:
We should probably improve the UX here so that the users understands that the order is in filling and the app should not be closed. On the successful execution, the user could get a notification that the position has been opened - here the user could share on twitter.

Restioson:
> On the successful execution, the user could get a notification that the position has been opened - here the user could share on twitter.

By notification, do you mean a system notification like a push notification or some other app feedback?

Restioson:
Maybe it would make sense to forbid the user from leaving the order creation screen while it is being filled, if this could cause problems? Some kind of progress bar (one of the ones that is stripy and just moving to indicate no absolute amount of progress) could be used to indicate that the app is busy doing something and shouldn't be closed. This depends on how long it might take the order to be filled, though. I just think that personally, as a user, I don't expect _not_ to be able to close the app after clicking the confirm order button. To me, that seems "done".

Restioson:
Looks like we can use [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications) to send a local notification after an order is filled, if we want, but it is quite verbose and a lot of set up if this is the only use-case for it

holzeis:
I like the idea of informing the user that the order is still being executed and the popup / app should not be closed while its in execution.

I think we did it the other way initially, because we couldn't guarantee that you will get a match immediately. But this is how we currently do it. Either you have a match immediately or the order goes to failed.

So yeah, lets do it how you suggested! 👍

holzeis:
In regards to the notification, I originally thought of an overlay popup similar to the order successfully submitted notification.

bonomat:
> I guess what I mean is that after I see this big green check and "submit order success" with a close button, I think that I can immediately close the app, since it doesn't convey that anything is still 'in progress'. That's what this is communicating to me as a user.
> After the order is filled, the same screen could replace the loading bar/spinner with a share button and the success display.

[...]

PR #820, opened 2023-06-22
Encourage user not to close the app while order is being processed

Restioson:
Share is shown after it is filled, fixing #714.

While order is filling (hard to see but that's a loading spinner on top):
image

Once order has been filled successfully:
image

### DIFF
--- a/mobile/lib/features/trade/order_submission_status_dialog.dart
+++ b/mobile/lib/features/trade/order_submission_status_dialog.dart
@@ -1,0 +1,88 @@
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+enum OrderSubmissionStatusDialogType {
+  pendingSubmit,
+  successfulSubmit,
+  filled,
+  failedFill,
+  failedSubmit
+}
+
+class OrderSubmissionStatusDialog extends StatelessWidget {
+  final String title;
+  final OrderSubmissionStatusDialogType type;
+  final Widget content;
+  final String buttonText;
+  final EdgeInsets insetPadding;
+  final String navigateToRoute;
+
+  const OrderSubmissionStatusDialog(
+      {super.key,
+      required this.title,
+      required this.type,
+      required this.content,
+      this.buttonText = "Close",
+      this.insetPadding = const EdgeInsets.all(50),
+      this.navigateToRoute = ""});
+
+  @override
+  Widget build(BuildContext context) {
+    bool isPending = type == OrderSubmissionStatusDialogType.successfulSubmit ||
+        type == OrderSubmissionStatusDialogType.pendingSubmit;
+
+    Widget closeButton = ElevatedButton(
+        onPressed: () {
+          GoRouter.of(context).pop();
+
+          if (navigateToRoute.isNotEmpty) {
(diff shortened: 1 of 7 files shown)


mobile/lib/features/trade/trade_screen.dart
```diff
@@ -282,7 +298,28 @@ class TradeScreen extends StatelessWidget {
 
   Widget createSubmitWidget(PendingOrder pendingOrder, Amount pnl,
       SubmitOrderChangeNotifier submitOrderChangeNotifier, BuildContext context) {
-    Widget body = Column(
+    String bottomText;
+
+    switch (pendingOrder.state) {
+      case PendingOrderState.submittedSuccessfully:
+      case PendingOrderState.submitting:
+        bottomText = "Please wait while the order is being processed. Do not close the app!";
```
holzeis, line 306: nit: Would be great if the "Do not clsoe the app!" part is more prominent on the popup. e.g. bold and in the next line?

No need to action!

mobile/lib/features/trade/trade_screen.dart
```diff
@@ -282,7 +298,28 @@ class TradeScreen extends StatelessWidget {
...
+        break;
+      case PendingOrderState.orderFailed:
+      case PendingOrderState.submissionFailed:
+        bottomText =
+            "Sorry, something went wrong while processing your order. It has been reverted.";
```
holzeis, line 311:
```suggestion
            "Sorry, we couldn't match your order. Please try again later.";
```

Or something along these lines. Mostly the order fails because the coordinator couldn't find a match.

mobile/lib/features/trade/trade_screen.dart
```diff
@@ -282,7 +298,28 @@ class TradeScreen extends StatelessWidget {
...
+        break;
+      case PendingOrderState.orderFilled:
+        if (pendingOrder.positionAction == PositionAction.close) {
+          bottomText = "Your position has been removed.";
```
holzeis, line 315:
```suggestion
          bottomText = "Your position has been closed.";
```

holzeis (approved):
Nice 👍

mobile/lib/features/trade/order_submission_status_dialog.dart
```diff
@@ -0,0 +1,78 @@
...
+          case OrderSubmissionStatusDialogType.filled:
+            return "Success";
+          case OrderSubmissionStatusDialogType.failedSubmit:
+          case OrderSubmissionStatusDialogType.failedFill:
+            return "Failure";
+        }
+      })()}"),
+      content: content,
+      actions: [
+        ElevatedButton(
```
Restioson, line 65: this should be removed if pending

Restioson:
I've updated the PR to address the review comments as well as to prevent the user from using the back button or clicking outside of the dialogue to close it. This has the unfortunate side effect that clicking outside of it doesn't work even when the order is successful. It can be addressed but then showDialog needs to be replaced with a custom dialogue window basically, from what I am reading online.

holzeis (approved):
I'd say lets give it a try and test it a bit on our test app release. If it feels off, we can always improve it. 👍

PR #823, opened 2023-06-22
Confetti after order confirmed

### LINKED ISSUE #822: Add confetti after order is filled successfully
This is a bit of fun :) Probably does not need to be a priority but I am investigating this now in the hopes that it will be quick.

Restioson:
This is on top of #820 and should wait for that to be merged first. This is intended to add a little hype to the new position and encourage the user to celebrate by sharing on Twitter :)

Right now it is just a little puff of confetti from the tick.

This needs flutter >3.7.2 to work, otherwise it will crash.

Fixes #822.

### DIFF
--- a/mobile/lib/features/trade/order_submission_status_dialog.dart
+++ b/mobile/lib/features/trade/order_submission_status_dialog.dart
@@ -1,3 +1,4 @@
+import 'package:confetti/confetti.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 
@@ -9,7 +10,7 @@ enum OrderSubmissionStatusDialogType {
   failedSubmit
 }
 
-class OrderSubmissionStatusDialog extends StatelessWidget {
+class OrderSubmissionStatusDialog extends StatefulWidget {
   final String title;
   final OrderSubmissionStatusDialogType type;
   final Widget content;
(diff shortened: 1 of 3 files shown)


klochowicz:
It would be good to document the minimum supported flutter version in readme, and someday perhaps use something like: https://fvm.app/ to manage flutter versions

klochowicz:
@Restioson  any chance for a short recording how it looks like? :)

Restioson:
Hi Mariusz, I'm away until the 4th and I can try do this when I get back :)

da-kami:
Just gave this a try locally - looks great! 

Here's a gif that shows it for closing the order (any order filled will trigger it):



Should I fix the lints and we get it merged?

mobile/lib/features/trade/order_submission_status_dialog.dart
```diff
@@ -26,24 +27,47 @@ class OrderSubmissionStatusDialog extends StatelessWidget {
       this.insetPadding = const EdgeInsets.all(50),
       this.navigateToRoute = ""});
 
+  @override
+  State<OrderSubmissionStatusDialog> createState() => _OrderSubmissionStatusDialog();
+}
+
+class _OrderSubmissionStatusDialog extends State<OrderSubmissionStatusDialog> {
+  late final ConfettiController _confettiController;
```
da-kami, line 35: ❓Why are we defining the `ConfettiController` in the `OrderSubmissionStatusDialog` *and* the `TradeScreen`?

My understanding is that the dialog cannot be closed anymore; so I think it should be enough to define the confetti on the `OrderSubmissionStatusDialog`.

I don't think it's wrong to define it on the underlying `Trade`screen as well, but I'm not sure it's necessary.

da-kami (approved):
Just one question comment - not sure important; from my side this PR is ready to merge if we want to.

Note: I had some local changes that are not part of the PR:

```
        modified:   mobile/macos/Podfile.lock
        modified:   mobile/pubspec.lock
```

Note sure this is relevant though; might just be due to me having some outdated versions.

Restioson:
It's probably a hangover from when I was still testing. Unfortunately I did
this last thing before I left on holiday (I am still away with no PC or
laptop) so I didn't have time to clear it up 😅

klochowicz (approved):
looks awesome, let's get this in!

Restioson:
I'd be grateful if someone can remove the unused controller first!

klochowicz:
> I'd be grateful if someone can remove the unused controller first!
> […](#)

awesome, I'll take care of this :)

Issue #799, opened 2023-06-20
Enable `TRACE`-level logs for `rust-lightning`


luckysori:
The motivation would be to be able to provide as much context as possible to LDK devs if something goes wrong.

---

Matt from LDK explicitly asked for louder logs so that the LDK devs can have an easier time helping us debug.

PR #826, opened 2023-06-23
Show TRACE logs on the phone

luckysori:
We need this to actually fix https://github.com/get10101/10101/issues/799.


klochowicz (approved):
makes sense for the time being!

Issue #792, opened 2023-06-17
Failed to close position: Failed to handle SubChannelCloseFinalize


holzeis:
Although my position was shown as closed successfully, opening another position failed as there exist a non-closed position. Checking the logs at the coordinator I found the following error after processing the SubChannelCloseFinalize message from the app.

```
Failed to process message: Failed to handle SubChannelCloseFinalize message from 02aa6fc81e341d69b36174365644a3560bfcc51ce0fb9bb67d93712bc54117cbb8: Invalid state: Misuse error: Close : Got a revoke commitment secret which didn't correspond to their current pubkey
```

It appears my phone lost the connection while the position was closed. That's most likely the reason for the error. The app is now stuck, as I can't open another position or recover the closed on.

```
2023-06-17T17:57:55Z	{"timestamp":"2023-06-17T17:57:55.115809296Z","level":"ERROR","fields":{"message":"Failed to process message: Failed to handle SubChannelCloseFinalize message from 02aa6fc81e341d69b36174365644a3560bfcc51ce0fb9bb67d93712bc54117cbb8: Invalid state: Misuse error: Close : Got a revoke commitment secret which didn't correspond to their current pubkey","from":"02aa6fc81e341d69b36174365644a3560bfcc51ce0fb9bb67d93712bc54117cbb8","kind":"SubChannelCloseFinalize"},"target":"coordinator::node"}
2023-06-17T17:57:55Z	{"timestamp":"2023-06-17T17:57:55.111790453Z","level":"INFO","fields":{"message":"Processing message","from":"02aa6fc81e341d69b36174365644a3560bfcc51ce0fb9bb67d93712bc54117cbb8","kind":"SubChannelCloseFinalize"},"target":"coordinator::node"}
2023-06-17T17:57:55Z	{"timestamp":"2023-06-17T17:57:55.107657568Z","level":"INFO","fields":{"message":"Background sync of Lightning wallet finished in 1ms."},"target":"ln_dlc_node::node"}
[...]

PR #827, opened 2023-06-23
Refactor DLC tests

luckysori:
It's probably not super important since we're moving away from these hopefully soon, but I arrived at this when working on trying to reproduce #792.

The main motivation is to ensure that `create_dlc_channel` and `dlc_collaborative_settlement` can be called without setting up nodes every time. This means that they are more easily composable.

The only downside is that we introduce a bit of repetition when setting up nodes plus the channel between them, but that could be mitigated by introducing another helper function.

### DIFF
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -29,6 +29,8 @@ where
         channel_details: ChannelDetails,
         contract_input: ContractInput,
     ) -> Result<()> {
+        tracing::info!(channel_id = %hex::encode(channel_details.channel_id), "Sending DLC channel offer");
+
         spawn_blocking({
             let oracle = self.oracle.clone();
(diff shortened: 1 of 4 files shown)


crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -29,6 +29,8 @@ where
         channel_details: ChannelDetails,
         contract_input: ContractInput,
     ) -> Result<()> {
+        tracing::info!(channel_id = %hex::encode(channel_details.channel_id), "Sending DLC channel offer");
```
luckysori, line 32: Snuck this in since I was missing this log when debugging.

crates/ln-dlc-node/src/tests/mod.rs
luckysori, line 1: Can revert this now.

klochowicz (approved):
if it makes debugging these tests easier, I don't mind it! 

 hopefully we'll get some higher-level tests in soon though :)

PR #830, opened 2023-06-24
feat: use mempool.space as block explorer

bonomat:
Hard to say if this will have a performance impact or not without deploying and trying it out :)

Related to https://github.com/get10101/meta/issues/164.

We can of course try it out manually first.

### DIFF
--- a/.github/workflows/build-release-binaries.yml
+++ b/.github/workflows/build-release-binaries.yml
@@ -25,7 +25,7 @@ jobs:
     uses: ./.github/workflows/ios.yml
     secrets: inherit
     with:
-      esplora_endpoint: https://blockstream.info/api
+      esplora_endpoint: https://mempool.space/api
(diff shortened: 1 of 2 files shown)


holzeis:
Note: this will only change the esplora endpoint for the mobile apps, the coordinator would still go against blockstream.info. The more critical IMHO is the coordinator though. I would simply do a manual test on production and run it against mempool.space to see if the situtation changes.

luckysori (approved):
I'm happy to try it, but mempool is also rate-limited. I've no idea if the limits are any better. Could we use a mix of both?

klochowicz:
can we get indication what are the rate limits on both? they might be different!

luckysori:
> can we get indication what are the rate limits on both? they might be different!

I think both are undisclosed.

bonomat:
Converting this back to draft: I tried it on the coordinator and it got stuck. I assume we will get a similar result in the app.

holzeis:
@bonomat whats the status of this PR?

bonomat:
We can close this and pick it up again in the future.

PR #835, opened 2023-06-26
Do more to prevent fee-rate-disagreement force-closures

luckysori:
We recently saw another one in production after increasing the background confirmation target from 12 to 24.

### DIFF
--- a/crates/ln-dlc-node/src/fee_rate_estimator.rs
+++ b/crates/ln-dlc-node/src/fee_rate_estimator.rs
@@ -9,6 +9,12 @@ use std::sync::RwLock;
 use std::sync::RwLockReadGuard;
 use std::sync::RwLockWriteGuard;
 
+const CONFIRMATION_TARGETS: [(ConfirmationTarget, usize); 3] = [
+    (ConfirmationTarget::Background, 24),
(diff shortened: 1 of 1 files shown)


crates/ln-dlc-node/src/fee_rate_estimator.rs
```diff
@@ -9,6 +9,14 @@ use std::sync::RwLock;
 use std::sync::RwLockReadGuard;
 use std::sync::RwLockWriteGuard;
 
+const CONFIRMATION_TARGETS: [(ConfirmationTarget, usize); 3] = [
+    // We choose an extremely high background confirmation target to avoid force-closing channels
+    // unnecessarily.
+    (ConfirmationTarget::Background, 1000),
```
holzeis, line 15: @luckysori why not `1008`? 😅

holzeis (approved):
Let's get this in as quick as possible

crates/ln-dlc-node/src/fee_rate_estimator.rs
luckysori: I didn't get that it was meant to be 7 days!

PR #836, opened 2023-06-26
Replace fee rate fallbacks in settings with constant

luckysori:
The fallbacks were only ever used if we could never get an answer from the fee rate server. As such, it didn't make sense for them to be configurable at runtime.

We move them to reduce complexity and because there is minimal benefit in being able to configure values that will very rarely be used.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -29,7 +29,6 @@ use lightning::ln::channelmanager::ChannelDetails;
 use ln_dlc_node::node::dlc_message_name;
 use ln_dlc_node::node::sub_channel_message_name;
 use ln_dlc_node::node::InMemoryStore;
-use ln_dlc_node::FeeRateFallbacks;
 use ln_dlc_node::WalletSettings;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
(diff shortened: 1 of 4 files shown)


holzeis (approved):

klochowicz:
these were added for a specific issue when we actually restarted and could not reach the electrum instance - how come this can not happen again?

luckysori:
> these were added for a specific issue when we actually restarted and could not reach the electrum instance - how come this can not happen again?

From the PR description, my point is that

> it is preferable to not start up the application if we cannot even get up-to-date fee estimates.

Do we disagree?

luckysori:
> How can you ensure that we will not start the app when the up-to-date fee estimates cannot be reached? it happened before due to hitting rate limits, and we did not notice that we were not getting up-to-date fee estimates. 

If we hit rate limits we still do not get up-to-date fee estimates. In such a scenario we are already covered by the cache, unless it happens on startup. If we are still worried that we might hit a rate limit on a restart we can prepopulate the cache with some values.

We could keep reading them from the config file instead of using constants, but they definitely don't need to be configurable during runtime since they can only ever be needed when the `FeeEstimator` is being started.

klochowicz (approved):

Issue #838, opened 2023-06-26
Order matching fees


bonomat:
The coordinator should charge a fee for matching orders and for providing the liquidity. 

This should be a percentage based on the nominal amount (if possible) or a percentage based on the required liquidity. 

Example values to get started with: 
- `0.30% taker`
- `0.15% maker`

Later on we should run experiments with the users to evaluate what are acceptable fees.

da-kami:
Note: I am assuming that we collect the fee separate from the position by paying an invoice. Theoretically we could also adapt the protocol to handle additional fees with the coordinator, but I don't think this is feasible for now. 

A few questions to make it more clear how we expect to implement this:

- *When* are we planning to collect this fee? Right before OR right after execution to open the position or upon closing the position?

- Should the fee be charged with *every execution*? - i.e. do we charge the same fee for opening and for closing? (I assume yes)

---

## Charge before opening position execution

Should we make it an execution condition, that the fee has to be paid before executing?

e.g.: 

1. Taker requests execution with coordinator
2. Maker requests execution with coordinator (timeout for both maker and taker if the other party did not request within X time)
3. Once both have requested: Send invoice to taker and maker
4. Once both have paid: Execute  

This could lead to tricky scenarios, where only one party pays, but the other party fails to pay - what would we do in such a scenario? 

For now we can simplify it to:


## Charge after opening position execution

1. Taker requests execution with coordinator
2. Maker requests execution with coordinator (timeout for both maker and taker if the other party did not request within X time)
3. Execute with both
4. Send invoice to taker
5. Send invoice for maker
6. Record payments, if not paid then eventually blacklist (out of scope for this ticket I guess)

---

Related question on top: 
Are we planning to collect an execution fee for perpetuals in the future as well?
Note: We don't have to decide that now, but if we want to charge on rollovers, it might make sense to just collect one fee upon rollover or closing the position.

PR #839, opened 2023-06-26
chore: github workflows can do this now.

bonomat:
Github workflows can add new issues automatically to the board. This script is not needed anymore which will save us some computing minutes.

See this ticket as an example: https://github.com/get10101/10101/issues/838

Same should be done on other repositories.


holzeis (approved):
Nice 👍

klochowicz (approved):

PR #841, opened 2023-06-26
End to end test proof of concept

klochowicz:
End-to-end tests use the crate 'native' (the Rust-backend) as-is, without the
Flutter dependency.

Changes to `native` crate are minimal, mostly exposing types that were needed
and ensuring that we can link against it for tests (`rlib` target).

The tests live in a separate crate to ensure that we're only calling public APIs.

TestSubscriber ensures that we receive all the events destined for the GUI and
exposes an API to inspect current values for wallet info, position etc.

The end-to-end test needs to be run with `just e2e` command, to ensure
that the necessary components are up and the coordinator is funded.

The vision is that the coordinator and the maker instance will be shared across
all the tests and that each test will spawn its own app that can be run in parallel.

The sample test only checks that the initial walletbalance is 0, but from the
logs one can see that we're connected to the coordinator.

- [x] Fund the app in the test


justfile
```diff
@@ -314,4 +347,11 @@ wipe-prometheus:
     cd services/prometheus
     rm -rf data
 
+
+# end-to-end tests
+e2e: services
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    cargo test -- --test-threads=1
```
luckysori: 🔧 I think you want to specify the package here.

justfile
```diff
@@ -262,6 +262,39 @@ wait-for-electrs-to-be-ready:
...
+    max_attempts=10
+    sleep_duration=1
+
+    check_endpoint() {
+      response=$(curl -s -o /dev/null -w "%{http_code}" "$endpoint")
+      if [ "$response" -eq 200 ]; then
+        echo "Endpoint is ready!"
+        exit 0
+        else
+        echo "Endpoint not ready yet. Retrying..."
```
luckysori, line 280: 🙃 It's helpful if the two `echo`s in this recipe mention the coordinator and not just an "endpoint".

crates/tests-e2e/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tests-e2e"
```
luckysori, line 2: ❓ Why not `e2e-tests`? It sounds more natural.

crates/tests-e2e/src/test_subscriber.rs
```diff
@@ -0,0 +1,140 @@
...
+
+pub struct Senders {
+    wallet_info: watch::Sender<Option<WalletInfo>>,
+    order: watch::Sender<Option<Order>>,
+    order_filled: watch::Sender<Option<Box<TradeParams>>>,
+    position: watch::Sender<Option<Position>>,
+}
+
+// Subscribes to events destined for the frontend (typically Flutter app) and
+// provides a convenient way to access the current state.
```
luckysori, line 20: 🙃 Docstring?

crates/tests-e2e/src/tracing.rs
```diff
@@ -0,0 +1,14 @@
+use std::sync::Once;
+
+pub fn init_tracing() {
+    static TRACING_TEST_SUBSCRIBER: Once = Once::new();
+
+    TRACING_TEST_SUBSCRIBER.call_once(|| {
+        tracing_subscriber::fmt()
+            .with_env_filter(
+                "debug,hyper=warn,reqwest=warn,rustls=warn,bdk=info,lightning=trace,sled=info,lightning::chain::channelmonitor=trace,lightning::ln::peer_handler=debug,ureq=info",
```
luckysori, line 9: 🔧 Let's use multi-line formatting here so that it's easier to read diffs. We do the same in the `ln-dlc-node` tests now.

justfile
```diff
@@ -348,10 +348,11 @@ wipe-prometheus:
...
 
 
+alias e2e := tests-e2e
 # end-to-end tests
-e2e: services
+tests-e2e: services
     #!/usr/bin/env bash
     set -euxo pipefail
-    cargo test -- --test-threads=1
+    cargo test -p tests-e2e -- --ignored --test-threads=1 --nocapture
```
luckysori, line 356: Not sure if we want `--nocapture` by default. Can we pass it in as an argument to the recipe instead?

crates/tests-e2e/src/test_subscriber.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    pub fn position(&self) -> Option<Position> {
+        self.position.borrow().as_ref().cloned()
+    }
+}
+
+impl Subscriber for Senders {
+    fn notify(&self, event: &native::event::EventInternal) {
+        dbg!("ZOSIA");
```
luckysori, line 71: Hahaha.

crates/tests-e2e/src/tracing.rs
```diff
@@ -0,0 +1,14 @@
+use std::sync::Once;
+
+pub fn init_tracing() {
+    static TRACING_TEST_SUBSCRIBER: Once = Once::new();
+
+    TRACING_TEST_SUBSCRIBER.call_once(|| {
```
luckysori, line 6: ❓Do we need this to be a `static`? Would this even be shared between tests? I don't think that's how it works anyway.

luckysori: I realise that we already do this in `ln-dlc-node`, but I also don't understand it there.

luckysori (commented):
Looking good!

crates/tests-e2e/src/tracing.rs
klochowicz: I just happened to copy-paste before you fixed the other occurrence 🙃 
will align.

crates/tests-e2e/Cargo.toml
klochowicz: I feel it will be easier to find on disk when I start with its primary purpose - ie. `tests`.

Was thinking of even leaving out `e2e` from the crate name, and just have it in the justfile - I don't think we'll have different set of tests at all.

justfile
klochowicz: it got fixed in the subsequent commit where I introduced the package.

justfile
klochowicz: added the argument (also aligned ln-dlc-node-tests)

crates/tests-e2e/src/tracing.rs
klochowicz: it needs to if more than one test will be built in one binary, because we should only initialise it once. 
the crucial part is the `Once`, not the static itself - it ensures that we can't accidentally init it more than once.

crates/tests-e2e/src/test_subscriber.rs
klochowicz: yeah, that's just my internal debugging  😅

 shoutout to my friend from Nokia who always used this token to find it in logs all the years ago

klochowicz:
For a while I thought that it would be good to enable them on CI in the same PR, but this became a bit more cumbersome than expected (e.g. requirement to have flutter-rust-bridge). I will spin off a follow-up PR to not muddy the waters; let's concentrate on the test itself here 😊

klochowicz:
Ci addition is handled in #849, which will be rebased after this one lands.

holzeis (approved):
Nice 👍

Issue #846, opened 2023-06-27
Handle persisted spendable outputs


luckysori:
By default we handle these immediately as LDK generates `SpendableOutputs` events, but we persist them in case the corresponding transactions don't get confirmed.

I propose to:
- On node startup, delete from the database all the spendable outputs which have reached sufficient confirmations.
- On node startup, try to broadcast claim transactions for those which are not sufficiently confirmed.

We might want to handle this differently between app and coordinator, but theoretically we should be able to implement it directly in `ln-dlc-node` for both.

---

Related to #798, #844 and #845.

Issue #845, opened 2023-06-27
Persist spendable output descriptors generated by LDK in coordinator


luckysori:
See #798.

We just need to save them to the database.

luckysori:
Merged onto https://github.com/get10101/10101/pull/940 after receiving enough approvals in https://github.com/get10101/10101/pull/938, so effectively done.

Issue #798, opened 2023-06-20
If handling `SpendableOutputs` event fails it is never retried


luckysori:
### Apparent problem

If the node is shut down half-way through processing an event, the event may never be retried again. This might be acceptable/expected for some events, but it's not if it can lead to loss of funds.

### Solution

I think we need to do several things:

1. [ ] Persist the spendable descriptors like the LDK docs indicate. This at least ensures that the claim transaction can be easily rebuilt and broadcast at a later time if it fails for whatever reason (#844 and #845).
2. [ ] Handle the persisted descriptors e.g. by broadcasting corresponding claim transactions on node startup (#846).
3. [ ] Build a dev tool _within_ the app to be able to call `spend_spendable_outputs` on the `KeysManager` with arbitrary data. This must be built on the app because we need the app's keys to build and sign the transaction. The input would be a `SpendableOutputDescriptor` sourced from our app logs.

### Report

A user claims that their on-chain wallet did not get the funds back after force-closing a pure LN channel. As the force-closing party in this case, the app would have to wait for 144 blocks after the publication of the commitment transaction. This test verifies this behaviour[^1]:

https://github.com/get10101/10101/blob/2381782d9e0483935d0feba3988decb787b05034/crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs#L159-L168

The config value is set here:

https://github.com/get10101/10101/blob/2381782d9e0483935d0feba3988decb787b05034/crates/ln-dlc-node/src/ln/config.rs#L17-L19

The logs show that the `DelayedPaymentOutput` was first detected by the `ChannelMonitor`

```
| 2023-06-17 16:00:31.642 | LogLevel.INFO | lightning::chain::channelmonitor: Received spendable output DelayedPaymentOutput 67c198cdcc205c1f531562e5d0c37e9ac22156a839bff7d0a16812792ef38c17:0 marked for spending, spendable at height 794874
```

and later a `SpendableOutputs` event was handled by our Lightning event handler:

```
[...]

PR #847, opened 2023-06-27
Persist spendable output descriptors generated by LDK in app

### LINKED ISSUE #844: Persist spendable output descriptors generated by LDK in app
See https://github.com/get10101/10101/issues/798.

luckysori:
Fixes #844.

This ensures that claiming funds from LN commitment transactions can be retried as much as we need to.

Currently, only the app persists this across restarts. The coordinator still uses an in-memory map, like it does for payments. We still need to fix this.

---

Related to https://github.com/get10101/10101/issues/798.
Next steps captured in #845 and #846.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1970,6 +1970,7 @@ dependencies = [
  "anyhow",
  "base64 0.21.0",
  "bdk",
+ "bitcoin",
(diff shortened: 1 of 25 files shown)


holzeis (approved):
Nice 👍

PR #849, opened 2023-06-27
Enable end-to-end tests on CI

klochowicz:
First run will probably take ages because of lack of cache.
Will closely monitor and optimise onwards.

I'm trying to remove the flutter dependency.

### DIFF
--- a/justfile
+++ b/justfile
@@ -268,7 +268,7 @@ wait-for-coordinator-to-be-ready:
     set +e
 
     endpoint="http://localhost:8000/api/newaddress"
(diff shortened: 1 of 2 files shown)


justfile
luckysori, line 1: 🔧 Can't we split it so that we build separately and only start waiting when we run it?

luckysori (approved):
Nice, it seems to work already!

justfile
klochowicz: I ended up doing it anyway :) but 10s was really arbitrary, so I thought we might as well include it.

PR #851, opened 2023-06-27
fix: Persist position and trade after dlc has been proposed

### LINKED ISSUE #537: Coordinator database has two positions in `Open` state for the same `taker_id`
## Symptom

We just saw this in production for one of Philipp's positions: Whenever we tried to close it it would fail with:

```
| 2023-04-27 10:18:43.304 | LogLevel.ERROR | Failed to execute trade for order Some(ad2540a9-7816-4a80-8eca-576901370e2e): TradeResponse: Could not post trade to coordinator: {"error":"Could not handle trade request: Invalid parameters were provided: Accept balance must be smaller than total collateral in DLC channel."}  
| 2023-04-27 10:18:43.305 | LogLevel.INFO | Updated order state new_state: Failed,order_id: ad2540a9-7816-4a80-8eca-576901370e2e 
```

After investigation we noticed that the values that the coordinator reported when closing the position did not match the position display in the app.

## Cause

There were two `Open` positions for Philipp's `taker_id` in the `coordinator` database.
Since we always just pick the "first" entry in the list when we load the position from the db there was no error, but the *wrong* position was picked and then the values obviously did not add up.

At this point we are not entirely sure how this can happen for this particular instance. 
The inherent problem is that the positions is not tightly enough coupled with the existence of a DLC. 
Once we tie the DLC lifecycle to the position's state this should be resolved - there should never be a position without a DLC.

## Workaround:

I shared in our `production` room how to update the position state in the database if this happens.

holzeis:
Fixes #537

It is more likely that the proposing a dlc will fail. e.g. the oracle is not reachable, the peer lost connection. Hence, it's better to firstly try to propose the dlc and only when this is successful persist the position and trade to the database.

Our current integration test framework does not allow me to easily write a regression test for that. Waiting for the e2e test setup being introduced with https://github.com/get10101/10101/pull/841.

However I verified the fix by running the following manual test.

1. Startup coordinator with an incorrect oracle endpoint
2. Open a position with the app - this order will fail.
3. Check if a position has been created (note, prior to this fix a position would have been created as the propose dlc comes after persisting the position) - No position should have been created.
```
docker exec -it db psql -U postgres -d orderbook -c "SELECT * from positions;"
```
4. Fix the oracle endpoint on the coordinator and restart.
5. Open another position with the app - this order succeeds.
6. Check that a position has been created.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -175,41 +175,6 @@ impl Node {
         let margin_trader = margin_trader(trade_params);
         let margin_coordinator = margin_coordinator(trade_params);
 
-        let liquidation_price = liquidation_price(trade_params);
-
-        let average_entry_price = trade_params
-            .average_execution_price()
-            .to_f32()
-            .expect("to fit into f32");
-        let new_position = NewPosition {
-            contract_symbol: trade_params.contract_symbol,
-            leverage: trade_params.leverage,
-            quantity: trade_params.quantity,
-            direction: trade_params.direction,
-            trader: trade_params.pubkey,
-            average_entry_price,
-            liquidation_price,
-            collateral: margin_coordinator as i64,
-            expiry_timestamp: trade_params.filled_with.expiry_timestamp,
-        };
-        tracing::debug!(?new_position, "Inserting new position into db");
-
-        let connection = &mut self.pool.get()?;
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thank you for reproducing it manually! The fix makes sense to me.

PR #852, opened 2023-06-28
Move our `rust-{dlc, lightning}` branches to `p2pderivatives`

### LINKED ISSUE #837: Migrate to the p2pderivatives repositories
Adopt the new PR workflow on the p2pderivatives repository for `rust-dlc` and `rust-lightning` and archive our own forks.

luckysori:
Fixes #837.

We are still pretty far away from replacing these git dependencies.

Currently we point to our own branches in the `p2pderivatives` repos: `feature/ln-dlc-channels-10101` and `split-tx-experiment-10101`. These are based on `feature/ln-dlc-channels` and `split-tx-experiment` respectively.

They diverge slightly at the moment for convenience and they will inevitably diverge slightly once in a while in the future. This is because we cannot expect a library such as `rust-dlc` to accept changes at the same rate as a product such as 10101 might need.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,7 +891,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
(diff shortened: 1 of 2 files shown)


luckysori:
I will archive our forks as soon as this is merged

klochowicz (approved):

PR #853, opened 2023-06-28
Expand end-to-end tests - add happy path

klochowicz:
Test opening and closing a position.

Make the framework more ergonomic and solve some issues.

### DIFF
--- a/crates/tests-e2e/src/coordinator.rs
+++ b/crates/tests-e2e/src/coordinator.rs
@@ -1,0 +1,43 @@
+use anyhow::Context;
+use anyhow::Result;
+use reqwest::Client;
+
+/// A wrapper over the coordinator HTTP API
+///
+/// It does not aim to be complete, functionality will be added as needed
+pub struct Coordinator {
+    client: Client,
+    host: String,
+}
+
+impl Coordinator {
+    pub fn new(client: Client) -> Self {
+        let host = "http://localhost:8000".to_string();
+        Self { client, host }
+    }
+
+    /// Check whether the coordinator is running
+    pub async fn is_running(&self) -> bool {
+        // We assume that if we can generate new address, the service is running
+        self.get("/api/newaddress").await.is_ok()
+    }
+
+    pub async fn is_node_connected(&self, node_id: &str) -> Result<bool> {
+        let result = self
+            .get(&format!("/api/admin/is_connected/{node_id}"))
+            .await?
+            .status()
+            .is_success();
+        Ok(result)
+    }
+
+    async fn get(&self, path: &str) -> Result<reqwest::Response> {
+        self.client
+            .get(format!("{0}{path}", self.host))
+            .send()
+            .await
+            .context("Could not send request to coordinator")?
+            .error_for_status()
+            .context("Coordinator did not return 200 OK")
+    }
+}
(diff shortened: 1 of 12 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -192,6 +192,8 @@ pub fn run(data_dir: String, seed_dir: String, runtime: &Runtime) -> Result<()>
 
         NODE.set(node);
 
+        event::publish(&EventInternal::Init("Backend initialised".to_string()));
```
holzeis, line 195: nit: How about `10101 is ready!`

holzeis (approved):
Nice 👍

crates/tests-e2e/src/coordinator.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl Coordinator {
+    pub fn new(client: Client) -> Self {
+        let host = "http://localhost:8000".to_string();
+        Self { client, host }
+    }
+
+    /// Check whether the coordinator is running
+    pub async fn is_running(&self) -> Result<()> {
```
da-kami, line 20: nit: I would expect a function that starts with `is_` to return a `bool`.

crates/tests-e2e/tests/collab_close.rs
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    tracing::info!("Closing a position");
+    spawn_blocking(move || api::submit_order(closing_order).unwrap())
+        .await
+        .unwrap();
+
+    wait_until!(app.rx.position().unwrap().position_state == PositionState::Closing);
+
+    // TODO: Assert that the position is closed in the app and the coordinator
```
da-kami, line 63: ❓ Is this TODO meant to be un-actioned?

I guess yes, until we properly introduced a `Closed` state (that should be bound to the DLC state I suppose).

da-kami (approved):
Nice work! 
Great to see the e2e tests coming to life ☺️

crates/tests-e2e/tests/collab_close.rs
klochowicz: I tried to strike enough balance to showcase the vision I had how we could keep writing these tests and still being able to focus on the other tasks.

And you're absolutely right, when we properly introduce a `Closed` state bound to DLC state it will get much easier. 

We'll get there soon. 🦾

mobile/native/src/ln_dlc/mod.rs
klochowicz: changed to `10101 is ready.` - I really think we're overusing the bangs, normal logs don't need to scream 😄

crates/tests-e2e/src/test_flow.rs
```diff
@@ -0,0 +1,33 @@
+/// Waits until the specified condition is met
+#[macro_export]
+macro_rules! wait_until {
+    ($expr:expr) => {
+        // Waiting time for the time on the watch channel before returning error
+        let next_wait_time: std::time::Duration =
+            std::time::Duration::from_secs(if cfg!(debug_assertions) { 60 } else { 30 });
```
luckysori, line 7: 🔧 I guess we can just default to 60.

crates/tests-e2e/tests/happy_path.rs
luckysori, line 1: 🙃 We could name these tests from the perspective of the app:

- `open_position`.
- `close_position`.

Happy path vs collab close doesn't work for me, because collab close is part of the happy path in a way.

luckysori (approved):
Nice!

PR #854, opened 2023-06-28
Fix `SubChannelCloseFinalize` reconnect bug

luckysori:
Fixes https://github.com/get10101/10101/issues/792.

Well, it was already fixed, but let's not talk about that.


holzeis (approved):
😂 🚀

Issue #763, opened 2023-06-13
Do not use remote oracle for testing


luckysori:
We should either mock the oracle or use a test oracle running locally. This is to ensure that our testing (such as CI) is not blocked by the public test oracle being unavailable.

klochowicz:
That would be great. I could not run the test without internet recently :( 

@holzeis, would you mind sharing the docker compose file for the oracle? 
I've got some changeset locally to allow passing in the oracle address and pk, it should be quite trivial to remove this dependency.

PR #855, opened 2023-06-28
chore: add oracle to setup

holzeis:
- Adds the oracle service to docker-compose
- Configurable oracle endpoint and public key on app and coordinator

fixes https://github.com/get10101/cloud-conf/issues/211


.github/workflows/build-release-binaries.yml
```diff
@@ -138,6 +142,8 @@ jobs:
           COORDINATOR_P2P_ENDPOINT: 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156@34.30.20.202:9045
           NETWORK: regtest
           COORDINATOR_PORT_HTTP: 80
+          ORACLE_ENDPOINT: http://api.test.10101.finance:8081
+          ORACLE_PUBKEY: 5d12d79f575b8d99523797c46441c0549eb0defb6195fe8a080000cbe3ab3859
```
da-kami, line 146: ❓Is it on purpose that we specify the public regtest pubkey for android and not the mainnet pubkey (as for ios)?

oracle/default.yml
```diff
@@ -0,0 +1,62 @@
...
+      precision: 0
+      signconfig:
+        base: 2
+        nbDigits: 20
+# configuration for the data feed
+datafeed:
+  cryptoCompare:
+    baseUrl: https://min-api.cryptocompare.com/data
+    # Set your cryptocompare api key here
+    # apiKey: xxxxxxxx
```
da-kami, line 53: ❓ It says we should set an `apiKey`, but we are not actually setting it - is that on purpose?

oracle/key.pem
```diff
@@ -0,0 +1,8 @@
+-----BEGIN EC PRIVATE KEY-----
```
da-kami, line 1: ❓ Just to make sure my understanding is correct - this is the private key for the local regtest default oracle, correct?

crates/ln-dlc-node/src/node/oracle_client.rs
```diff
@@ -4,12 +4,11 @@ use dlc_manager::Oracle;
 use p2pd_oracle_client::P2PDOracleClient;
 use std::str::FromStr;
 
-// TODO: This should come from the configuration.
-const ORACLE_ENDPOINT: &str = "https://oracle.holzeis.me/";
+pub fn build(oracle_endpoint: String) -> P2PDOracleClient {
+    // TODO: fetch public key from oracle at ORACLE_ENDPOINT/oracle/publickey
```
da-kami:
``` suggestion
```

I don't think we should ever fetch the pubkey from the oracle itself but the pubkey should be hardcoded because it should not change; this comment should be removed.

da-kami (approved):
Looking good, few validation questions

.github/workflows/deliverables-push-main.yml
```diff
@@ -47,3 +47,5 @@ jobs:
       coordinator_port_http: 80
       network: regtest
       tag: main
+      oracle_endpoint: http://api.test.10101.finance:8081
```
klochowicz, line 50: 🙃 I know it's subjective, but could we consider some other port number, e.g. `8090` for the oracle? `8081` sounds like a temporary second version during upgrade of whatever we're normally running on port `8080`

klochowicz (approved):
nice, thank you for addressing this!

luckysori (commented):
This seems to fix https://github.com/get10101/10101/issues/763 too.

.github/workflows/build-release-binaries.yml
holzeis: This is building the android release for regtest, hence we use the regtest oracle.

It should be the same for ios. Note you might have mixed it up as we are having an ios regtest and mainnet build.

oracle/default.yml
holzeis: It works without api key.

oracle/key.pem
holzeis: Yes, this is only for the local setup.

crates/ln-dlc-node/src/node/oracle_client.rs
holzeis: It is hardcoded.

Issue #580, opened 2023-05-03
Cannot accept value that would put our exposure to dust HTLCs


holzeis:
Our users currently can't fund their 10101 wallet because of that error.

```
2023-05-03 09:50:28 | {"timestamp":"2023-05-03T09:50:28.484712603Z","level":"INFO","fields":{"message":"Cannot accept value that would put our exposure to dust HTLCs at 10201000 over the limit 5000000 on counterparty commitment tx","log.target":"lightning::ln::channel","log.module_path":"lightning::ln::channel","log.file":"/home/runner/.cargo/git/checkouts/rust-lightning-8ffd0e8d14b5d2cc/6df5761/lightning/src/ln/channel.rs","log.line":2842},"target":"lightning::ln::channel"}
-- | --
2023-05-03 09:50:28 | {"timestamp":"2023-05-03T09:50:28.484820391Z","level":"INFO","fields":{"message":"Cannot accept value that would put our exposure to dust HTLCs at 10201000 over the limit 5000000 on holder commitment tx","log.target":"lightning::ln::channel","log.module_path":"lightning::ln::channel","log.file":"/home/runner/.cargo/git/checkouts/rust-lightning-8ffd0e8d14b5d2cc/6df5761/lightning/src/ln/channel.rs","log.line":2852},"target":"lightning::ln::channel"}
```

holzeis:
It looks like this issue is due to some unreasonable high transaction fees at the moment. I guess whats happening is that the `htlc_timeout_dust_limit` and `htlc_success_dust_limit` are too high for a small payment of 15_000. (that was the biggest payment I tried) and our `max_dust_htlc_exposure_msat` is set to 5_000 sats which is smaller than 15_000.

![Screenshot 2023-05-03 at 12 54 41](https://user-images.githubusercontent.com/382048/235897062-b4150fc5-2d5a-4074-a94c-3315b86574c4.png)

https://github.com/get10101/rust-lightning/blob/6df5761cfc4a27c35ca26a547ff90f656d50d85f/lightning/src/ln/channel.rs#L2831-L2856

```
		let (htlc_timeout_dust_limit, htlc_success_dust_limit) = if self.opt_anchors() {
			(0, 0)
		} else {
			let dust_buffer_feerate = self.get_dust_buffer_feerate(None) as u64;
			(dust_buffer_feerate * htlc_timeout_tx_weight(false) / 1000,
				dust_buffer_feerate * htlc_success_tx_weight(false) / 1000)
		};
[...]

PR #856, opened 2023-06-28
fix: app cannot receive payment with open position

### LINKED ISSUE #498: App cannot receive payment with open position

holzeis:
- **Remove stop-gap on receiving payment with open position**: It seems a change in the past fixed that issue implicitly. Probably upgrading to ldk 114. Anyhow, I tested receiving a payment from the faucet and it worked just fine.
- **Remove constraint on min amount**: Receiving min 50k sats was an arbitrary amount and did at most only mitigate the chances of preventing https://github.com/get10101/10101/issues/580. Given that the current mempool situation does not ask for this constraint anymore, we returning to the original min amount.

fixes #498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Fix panic when processing accept message while peer is disconnected.
+- Removed stop-gap from receiving payments with open position.
+- Reduced min amount of 50k sats on receiving payments.
 
 ## [1.0.20] - 2023-06-16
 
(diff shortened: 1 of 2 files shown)


da-kami (approved):
Happy to remove this for now if we think that mempool fees will remain low in the foreseeable future. 
I think eventually we will have to make these values more dynamic, but that will introduce some complexity.

CHANGELOG.md
luckysori, line 1: 🙃Not that I'm someone who writes changelog entries 😛, but it's nice if you add the changelog entry in the same commit as the change.

luckysori (approved):

klochowicz:
I will quickly add a test that verifies that we can always receive payments when position is open ;D will make the payment of 5k, to showcase that we've lowered the constraint.

klochowicz:
bors retry

PR #858, opened 2023-06-29
Fix constant flickering on app's wallet screen

### LINKED ISSUE #404: Home screen flickers every few seconds
https://user-images.githubusercontent.com/9418575/231685385-b0a96564-5fda-4d09-95c3-16b0b04e28d9.MOV

---

On iPhone SE.

luckysori:
Fix #404.

This was caused by calling `isUserSeedBackupConfirmed` (the future passed to `FutureBuilder`) in `State<WalletScreen>.build`, something that the docs explicitly say we must not do.

As recommended, we call it in `State<WalletScreen>.initState`. This means that the `WalletScreen` is no longer informed about the updated value after confirming that the seed has been backed up.

The simplest way to know is to return a value after `pop`ping (returning from) the `SeedScreen`. This functionality is
only available in a later version of `go_router` so we update the dependency to the latest version. With the returned value we can
update the dependency of `isUserSeedBackupConfirmed`.

Finally, we move the responsibility of keeping the button always visible to the place where we actually enable the button. As such, we don't need to comment out code and the TODO can be moved next to the code that it references.

### DIFF
--- a/mobile/lib/common/app_bar_wrapper.dart
+++ b/mobile/lib/common/app_bar_wrapper.dart
@@ -14,7 +14,7 @@ class AppBarWrapper extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final currentRoute = GoRouter.of(context).location;
+    final currentRoute = GoRouterState.of(context).location;
     const appBarHeight = 32.0;
 
     var actionButtons = <Widget>[];
--- a/mobile/lib/features/wallet/seed_screen.dart
+++ b/mobile/lib/features/wallet/seed_screen.dart
@@ -126,8 +126,8 @@ class _SendScreenState extends State<SeedScreen> {
                       ElevatedButton(
                           onPressed: checked
                               ? () {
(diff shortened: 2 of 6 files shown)


da-kami (approved):
Looks good to me; I trust you tested that it works :)

---

Note: Given that we use `GoRouterState.of(...)` here now; will this need to be adapted in other places of the codebase now that we updated to the latest version of `GoRouter`?

luckysori:
> Looks good to me; I trust you tested that it works :)

Yep, yep. I thought about a gif but it's hard to demonstrate because it's something _not_ happening haha.


> Note: Given that we use `GoRouterState.of(...)` here now; will this need to be adapted in other places of the codebase now that we updated to the latest version of `GoRouter`?

I don't think we need to do anything else in terms of the `go_router` upgrade. This was just a breaking change, but we didn't seem to hit any others as it now builds. The app also looked normal to me after upgrading.

holzeis (approved):
Thanks for fixing this 👍

PR #859, opened 2023-06-29
fix(ui): Remove dummy initial price values from the GUI

klochowicz:
Once introduced for expedience, they started confusing the beta users.

TradeValues allows nullable values for parameters that are taken from the coordinator.

Checks are inserted in place before each `!` defererence (Dart 2 unfortunately
does not detect all not-nullable values properly.).

when coordinator is off at startup - the order cannot be placed because we don't have the counterparty margin:
image

(note: it would be great if we had a higher-level API check from the backend + a notifier in Dart about coordinator status).

### DIFF
--- a/mobile/lib/common/dummy_values.dart
+++ b/mobile/lib/common/dummy_values.dart
@@ -1,9 +1,6 @@
 // Assorted dummy values for expedience with prototyping.
 // Eventually we should replace them with the real values from the backend.
 
-const double dummyBidPrice = 22990.0;
-const double dummyAskPrice = 23010.0;
-
 // TODO replace dummy funding rate with funding rate from backend
 const double fundingRateBuy = 0.003;
 const double fundingRateSell = -0.003;
--- a/mobile/lib/features/trade/application/position_service.dart
+++ b/mobile/lib/features/trade/application/position_service.dart
@@ -11,13 +11,17 @@ class PositionService {
   }
 
   /// Returns the pnl in sat
-  int calculatePnl(Position position, Price price) {
+  int? calculatePnl(Position position, Price price) {
+    if (!price.isValid()) {
+      return null;
+    }
+    final closingPrice = rust.Price(
+      bid: price.bid!,
+      ask: price.ask!,
+    );
     return rust.api.calculatePnl(
         openingPrice: position.averageEntryPrice,
-        closingPrice: rust.Price(
-          bid: price.bid,
-          ask: price.ask,
-        ),
+        closingPrice: closingPrice,
         quantity: position.quantity,
         leverage: position.leverage.leverage,
         direction: position.direction.toApi());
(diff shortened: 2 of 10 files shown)


holzeis:
@klochowicz Can we show something else than **0** in case we don't have the data ready? ideally some kind of spinning wheel, but if thats to cumbersome how about **n/a**?

klochowicz:
> @klochowicz Can we show something else than **0** in case we don't have the data ready? ideally some kind of spinning wheel, but if thats to cumbersome how about **n/a**?

I would prefer doing that in a follow-up that I have mentioned (#860 ). we should do a more comprehensive setup eventually, this code deals with the annoyance of legacy dummy values.
I would even disable `buy`/`sell`, because they're pointless if coordinator is down.

holzeis:
> > @klochowicz Can we show something else than **0** in case we don't have the data ready? ideally some kind of spinning wheel, but if thats to cumbersome how about **n/a**?
> 
> I would prefer doing that in a follow-up that I have mentioned (#860 ). we should do a more comprehensive setup eventually, this code deals with the annoyance of legacy dummy values. I would even disable `buy`/`sell`, because they're pointless if coordinator is down.
> 
> the attached screenshot just shows that we're not crashing the app when the values are not there :)
> 
> I'll proceed to dealing with #860 straight after this, as I feel it's a natural follow-up.

sounds, good. Just a note, that we can also run into not having an order e.g. if the maker is down or haven't sent a new order in some time.

Also now that I come to think of it, did you test how the pnl will be displayed in that case? The screenshot in the bug shows an incorrect pnl because of the dummy values, what will it show now with 0?

klochowicz:
> > > @klochowicz Can we show something else than **0** in case we don't have the data ready? ideally some kind of spinning wheel, but if thats to cumbersome how about **n/a**?
> > 
> > 
> > I would prefer doing that in a follow-up that I have mentioned (#860 ). we should do a more comprehensive setup eventually, this code deals with the annoyance of legacy dummy values. I would even disable `buy`/`sell`, because they're pointless if coordinator is down.
> > the attached screenshot just shows that we're not crashing the app when the values are not there :)
> > I'll proceed to dealing with #860 straight after this, as I feel it's a natural follow-up.
> 
> sounds, good. Just a note, that we can also run into not having an order e.g. if the maker is down or haven't sent a new order in some time.
yes, not having an order should be a state that we support comprehensively.
> 
> Also now that I come to think of it, did you test how the pnl will be displayed in that case? The screenshot in the bug shows an incorrect pnl because of the dummy values, what will it show now with 0?

PnL is for now set to 0. I haven't revamped the widget code anywhere yet except of removing the dummy values, I just made sure that it doesn't crash anywhere and the widgets are ready to use the optional somehow.
Will add these 2 points into the follow-up ticket!

holzeis (approved):
LGTM 👍

PR #861, opened 2023-06-29
chore(mobile): Upgrade flutter packages to support sound null safety

klochowicz:
Note: only the required packages got bumped, not all dependencies.

See:
https://dart.dev/resources/dart-3-migration#null-safety-migration-tools

I have tested both the macOS and iOS versions for opening / closing a position.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,9 +41,9 @@ dependencies = [
 
 [[package]]
 name = "allo-isolate"
-version = "0.1.14"
+version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ed55848be9f41d44c79df6045b680a74a78bc579e0813f7f196cd7928e22fb1"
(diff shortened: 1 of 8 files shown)


mobile/lib/main.dart
```diff
@@ -66,6 +66,12 @@ void main() {
   config.customOpeningDivider = "| ";
 
   FLog.applyConfigurations(config);
+
+  // Based on GoRouter migration guide to preserve old behaviour
+  // https://docs.google.com/document/d/1VCuB85D5kYxPR3qYOjVmw8boAGKb7k62heFyfFHTOvw/edit
+  // TODO: Verify whether we need this setting
+  GoRouter.optionURLReflectsImperativeAPIs = true;
```
luckysori, line 73: 🔧 I tested the app after updating and we didn't need this as far as I could tell.

luckysori (approved):

mobile/lib/main.dart
klochowicz: even better; I erred on the side of caution in the beginning :-) 

thank you for double-checking this!

Issue #688, opened 2023-05-22
Can't get passed receive funds screen


holzeis:
Opening this ticket for a user, who reported that he is stuck on the receive funds screen.

Reported on iPhone 11 Pro (latest OS)

![Image](https://user-images.githubusercontent.com/382048/239961826-6cfe413c-6bc3-4ca5-8780-905d79a8d962.jpg)

Note, I was not able to reproduce that on the iPhone 11 Pro Simulator.

holzeis:
Another user reported issues funding the wallet due to not being able to get past the receive funds screen.



![Image](https://github.com/get10101/10101/assets/382048/d3589955-4ea5-4bc8-9db2-4ae9270cbb91)

PR #862, opened 2023-06-29
Defocus textfield upon clicking anywhere on screen

da-kami:
Fixes https://github.com/get10101/10101/issues/688

The `GestureDetector` records taps within the `SafeArea` of the screen. Setting the `FocusScope` to an unused `FocusNode` removes the focus from the `TextField`. `behavior: HitTestBehavior.opaque,` allows clicking in "empty" parts of the screen; otherwise the focus would only be take away from the `TextField` when e.g. clicking on other text.

This method should work no matter what - in case the screen becomes so full that the text pushes the button out even without the keyboard we would have to make it scrollable. For now I think this is the better solution.

Note that I was not able to reproduce this issue by "just" using larger and bold font through Accessibility settings, so I simulated it by adding dummy text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


holzeis (approved):
Nice 👍 thank you!

holzeis:
@da-kami Can you update the changelog as well please :)

PR #863, opened 2023-06-29
Pay order-matching fee on the app

### LINKED ISSUE #838: Order matching fees

### LINKED ISSUE #884: Only claim the order-matching fee once the trade has been executed
In https://github.com/get10101/10101/pull/863 we don't wait for the trade to be executed before the corresponding order-matching fee is claimed. This is problematic as the app could end up paying for something that didn't happen.

luckysori:
Fixes #838[^1].
Fixes https://github.com/get10101/meta/issues/170.
Fixes https://github.com/get10101/meta/issues/171.
Fixes #884.

We keep it simple for now: the coordinator returns an invoice and we trust the app to pay for it.

Invoice generation rationale
----------------------------
We choose to let the coordinator generate
an invoice rather than send an spontaneous payment from the app so that we can more easily update the coordinator fees without having to release a new version of the app.

Payment timing rationale
------------------------

In patch 7c6a4bbb808267a8e56633ba471487838e74f414 the app was paying the invoice as soon as possible, but this had two downsides:

1. The app could end up paying for a trade that didn't come to fruition.
2. Updating the LN-DLC channel from the `rust-dlc` side and `rust-lightning` side simultaneously is not currently supported, so we could run into channel-breaking issues.

As such, we eventually opted for the simple 74a0acf560a7768c43207cb07d0c688c0771eff8. The solution is not perfect and it assumes that the app will not be rebooted until the payment is sent, but it should be good enough for now.

Other thoughts
--------------

Eventually, the order-matching fee could be included in the payout curve or, perhaps even better, as a separate static output on the buffer transaction (this is currently not supported).

---

These are some TODOs that could be tackled here or in follow-ups:

- [x] Verify that the app can actually pay for both executing the trade and the order-matching fee _before_ executing the trade.
- [x] Only claim the order-matching fee once the trade has been executed. https://github.com/get10101/10101/issues/884
- [x] Tell the user about the estimated fee to be charged before they create the order. https://github.com/get10101/meta/issues/170
- [x] Show the user the charged fee on the success screen after executing the order. https://github.com/get10101/meta/issues/171

[^1]: Given that we can only charge taker fees for now. We should create a separate ticket for the maker once it's possible to do so.

### DIFF
--- a/coordinator/src/node/order_matching_fee.rs
+++ b/coordinator/src/node/order_matching_fee.rs
@@ -2,14 +2,8 @@ use crate::node::Node;
 use anyhow::Result;
 use coordinator_commons::TradeParams;
 use lightning_invoice::Invoice;
+use orderbook_commons::order_matching_fee_taker;
 use orderbook_commons::FEE_INVOICE_DESCRIPTION_PREFIX_TAKER;
-use rust_decimal::prelude::FromPrimitive;
-use rust_decimal::prelude::ToPrimitive;
-use rust_decimal::Decimal;
-use rust_decimal::RoundingStrategy;
-
-/// The order-matching fee per cent for the taker.
-const TAKER_FEE: (i64, u32) = (30, 4);
 
 /// How long the fee invoice will last for.
 const INVOICE_EXPIRY: u32 = 3600;
@@ -19,40 +13,12 @@ impl Node {
         let order_id = trade_params.filled_with.order_id;
         let description = format!("{FEE_INVOICE_DESCRIPTION_PREFIX_TAKER}{order_id}");
 
-        let fee = order_matching_fee(
+        let fee = order_matching_fee_taker(
             trade_params.quantity,
             trade_params.average_execution_price(),
-            Decimal::new(TAKER_FEE.0, TAKER_FEE.1),
         )
         .to_sat();
 
         self.inner.create_invoice(fee, description, INVOICE_EXPIRY)
-    }
-}
-
-fn order_matching_fee(quantity: f32, price: Decimal, fee_per_cent: Decimal) -> bitcoin::Amount {
-    let quantity = Decimal::from_f32(quantity).expect("quantity to fit in Decimal");
-    let price = price;
-
-    let fee = quantity * (Decimal::ONE / price) * fee_per_cent;
-    let fee = fee
-        .round_dp_with_strategy(8, RoundingStrategy::MidpointAwayFromZero)
-        .to_f64()
-        .expect("fee to fit in f64");
-
-    bitcoin::Amount::from_btc(fee).expect("fee to fit in bitcoin::Amount")
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn calculate_order_matching_fee() {
-        let price = Decimal::new(30209, 0);
-
-        let fee = order_matching_fee(50.0, price, Decimal::new(TAKER_FEE.0, TAKER_FEE.1));
-
-        assert_eq!(fee.to_sat(), 497);
     }
 }
--- a/coordinator/src/node/order_matching_fee.rs
+++ b/coordinator/src/node/order_matching_fee.rs
@@ -1,0 +1,58 @@
+use crate::node::Node;
+use anyhow::Result;
+use coordinator_commons::TradeParams;
+use lightning_invoice::Invoice;
+use orderbook_commons::FEE_INVOICE_DESCRIPTION_PREFIX_TAKER;
+use rust_decimal::prelude::FromPrimitive;
+use rust_decimal::prelude::ToPrimitive;
+use rust_decimal::Decimal;
+use rust_decimal::RoundingStrategy;
+
+/// The order-matching fee per cent for the taker.
+const TAKER_FEE: (i64, u32) = (30, 4);
+
+/// How long the fee invoice will last for.
+const INVOICE_EXPIRY: u32 = 3600;
+
+impl Node {
+    pub async fn fee_invoice_taker(&self, trade_params: &TradeParams) -> Result<Invoice> {
+        let order_id = trade_params.filled_with.order_id;
+        let description = format!("{FEE_INVOICE_DESCRIPTION_PREFIX_TAKER}{order_id}");
+
+        let fee = order_matching_fee(
+            trade_params.quantity,
+            trade_params.average_execution_price(),
+            Decimal::new(TAKER_FEE.0, TAKER_FEE.1),
+        )
+        .to_sat();
+
+        self.inner.create_invoice(fee, description, INVOICE_EXPIRY)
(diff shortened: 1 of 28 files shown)


luckysori:
With 4a18708 it looks like this:

image

(Ignore the overflow warnings, those are already happening on native.)

coordinator/src/node/order_matching_fee.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        self.inner.create_invoice(fee, description, INVOICE_EXPIRY)
+    }
+}
+
+fn order_matching_fee(quantity: f32, price: Decimal, fee_per_cent: Decimal) -> bitcoin::Amount {
+    let quantity = Decimal::from_f32(quantity).expect("quantity to fit in Decimal");
+    let price = price;
+
+    let fee = quantity * (Decimal::ONE / price) * fee_per_cent;
```
luckysori, line 37: Is this the fee you expected to charge, @bonomat?

da-kami:
> Verify that the app can actually pay for both executing the trade and the order-matching fee before executing the trade.

Not sure we have to - we can never really guarantee this completely because:

* we have market orders and we cannot guarantee the execution price
* The user might send funds out (we could block that, but that might also be hard.

I think the better solution will be: 

1. The app (user) fails to pay the invoice for the order
2. The coordinator pushes a notification to the app, this will display a warning to the user that the invoice has to be paid otherwise consecutive orders will fail.
3. The coordinator has to keep record of the paymnets in association with the order. We could allow that the user pays the invoice from a different wallet to avoid being blacklisted (...)

da-kami:
>  Tell the user about the estimated fee to be charged before they create the order.

In my opinion we should do this before we release this feature, otherwise it's not transparent enough.

luckysori:
> > Verify that the app can actually pay for both executing the trade and the order-matching fee before executing the trade.
> 
> Not sure we have to - we can never really guarantee this completely because:
> 
>     * we have market orders and we cannot guarantee the execution price
> 
>     * The user might send funds out (we could block that, but that might also be hard.
> 
> 
> I think the better solution will be:
> 
>     1. The app (user) fails to pay the invoice for the order
> 
>     2. The coordinator pushes a notification to the app, this will display a warning to the user that the invoice has to be paid otherwise consecutive orders will fail.
> 
>     3. The coordinator has to keep record of the paymnets in association with the order. We could allow that the user pays the invoice from a different wallet to avoid being blacklisted (...)

Thanks for your thoughts on this.

I agree that we won't be able to fully guarantee that the taker (app) will be able to pay for both executing the trade and the fee invoice.

Since we want this to be atomic anyway and we know that we will eventually do so, I would advocate for keeping it as simple as possible for this non-atomic version. As such I would hesitate to add any complex logic around the current flow.

Given the current state of the project, I still advocate for a very crude check before letting the user place the order because:

1. We can take a reasonable estimate of the order-matching fee client-side, even if the price might fluctuate. We could be conservative and assume that the price might end up being 100 dollars worse, for example.
2. Fees should only be a small number of sats (given the trade sizes we are going for atm), so it's unlikely that we block a valid trade by mistake.

da-kami:
[snip]

> Given the current state of the project, I still advocate for a very crude check before letting the user place the order because:
> 
>     1. We can take a reasonable estimate of the order-matching fee client-side, even if the price might fluctuate. We could be conservative and assume that the price might end up being 100 dollars worse, for example.
> 
>     2. Fees should only be a small number of sats (given the trade sizes we are going for atm), so it's unlikely that we block a valid trade by mistake.

Sure we can - I just wonder how much more work it would be to add the part where the coordinator keeps track. 

I'm happy with the crude check for now; given that we want to display the fee it should be part of the calculation anyway. (sorry, could have pointed that out)
As long as we are aware that we will need some tracking on the coordinator side at some point later, because we cannot guarantee the fee will get paid, that;s fine.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -442,5 +449,36 @@ pub async fn trade(trade_params: TradeParams) -> Result<(), (FailureReason, anyh
...
+        )
+    })?;
+    let order_matching_fee_invoice = order_matching_fee_invoice.parse().map_err(|e| {
+        (
+            FailureReason::TradeResponse,
+            anyhow!("Could not parse order-matching fee invoice: {e:#}"),
+        )
+    })?;
+
+    NODE.get()
```
da-kami, line 467: 🙃 I think it would be better to charge this only once the `Order` was filled!

I think we should easily be able to add / trigger the payment from `trade::order::handler::order:filled` instead of triggering it right after the trade.

The problem I see here is, that the fee payment may succeed, but the execution then may fail because of e.g. insufficient balance. I think it's better to charge once we are sure that the execution succeeded (i.e. upon finalize, that's when we trigger `order_filled`.

da-kami (approved):
Please see my comment, I am OK with merging this version, hence the approval, but I think changing it to pay after `Finalized` (upon updating the order to `Filled`) would be better.

We can also opt to experiment how often the user is charge without the execution succeeding, and if users actually complain.

mobile/native/src/ln_dlc/mod.rs
da-kami: This assumes that we are able to send payments while a position is open - I assume this works.

luckysori: Yep, I agree. I tried to capture this in the TODO in the PR description:

> Only claim the order-matching fee once the trade has been executed.

I will do so before merging this.

holzeis (approved):
Nice 👍

mobile/native/src/ln_dlc/mod.rs
luckysori: > Actually, in the history above the fee is displayed after the position was opened - how is that possible? - am I misunderstanding when the payment is triggered? 

Good catch. I suppose the payment may be sent now, but it may only be claimed a bit later. So in theory you could end up resolving the trade before the fee payment.

luckysori:
I already had approvals before and now I've tackled the extra TODOs. Do take a look at the updated PR, @holzeis and @da-kami if you have time.

I'm merging for now. Any new feedback can be addressed in follow-ups.

bors r+

mobile/native/src/ln_dlc/mod.rs
luckysori: This is all fixed with https://github.com/get10101/10101/pull/863/commits/bb2e157e8abaca2a599764a0c1ea6222b59e01c5, including the visual bug.

PR #864, opened 2023-06-29
feat: charge fee when opening a jit channel

### LINKED ISSUE #577: JIT - Channel Fee

holzeis:
fixes #577 

- The payment of the fee is not yet depicted as a special payment. This will be covered in a follow up PR after aligning with https://github.com/get10101/10101/pull/863
- The funding transaction is only cached in memory

TODOs that will be covered in follow up tickets.
- [ ] https://github.com/get10101/meta/issues/172
- [ ] https://github.com/get10101/10101/issues/873
- [ ] https://github.com/get10101/10101/issues/883

Simulator Screenshot - iPhone 14 Pro - 2023-06-30 at 11 34 28

### DIFF
--- a/mobile/native/src/channel_fee.rs
+++ b/mobile/native/src/channel_fee.rs
@@ -1,0 +1,157 @@
+use crate::commons::reqwest_client;
+use crate::config;
+use crate::event::subscriber::Subscriber;
+use crate::event::EventInternal;
+use crate::event::EventType;
+use crate::ln_dlc;
+use anyhow::anyhow;
+use anyhow::Context;
+use anyhow::Result;
+use bitcoin::Txid;
+use lightning_invoice::Invoice;
+use ln_dlc_node::node::rust_dlc_manager::ChannelId;
+use serde::Deserialize;
+use serde::Serialize;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::sync::Mutex;
+use tokio::runtime::Handle;
+
+#[derive(Clone)]
+pub struct ChannelFeePaymentSubscriber {
+    pub open_channel_tx: Arc<Mutex<Option<EsploraTransaction>>>,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct EsploraTransaction {
+    pub txid: String,
+    pub fee: u32,
+}
+
+impl Subscriber for ChannelFeePaymentSubscriber {
+    fn notify(&self, event: &EventInternal) {
+        let result = match event {
+            EventInternal::ChannelReady(channel_id) => {
+                self.register_funding_transaction(channel_id)
+            }
+            EventInternal::PaymentClaimed(amount_msats) => {
+                self.pay_funding_transaction_fees(*amount_msats)
+            }
+            _ => Ok(()),
+        };
+
+        if let Err(e) = result {
+            tracing::error!("{e:#}");
+        }
+    }
+
+    fn events(&self) -> Vec<EventType> {
+        vec![EventType::ChannelReady, EventType::PaymentClaimed]
+    }
+}
+
+impl ChannelFeePaymentSubscriber {
+    pub fn new() -> Self {
+        Self {
+            open_channel_tx: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    /// Attempts to pay the transaction fees for opening an inbound channel.
+    fn pay_funding_transaction_fees(&self, amount_msats: u64) -> Result<()> {
+        let transaction = match self.get_funding_transaction() {
+            Some(transaction) => transaction,
+            None => {
+                tracing::debug!("No pending funding transaction found!");
+                return Ok(());
+            }
+        };
+
+        tracing::debug!("Trying to pay channel opening fees of {}", transaction.fee);
+        let funding_tx_fees_msats = (transaction.fee * 1000) as u64;
+
+        if funding_tx_fees_msats > amount_msats {
+            tracing::warn!("Trying to pay fees with an amount smaller than the fees!")
+        }
+
+        let invoice_str = tokio::task::block_in_place(|| {
+            Handle::current().block_on(fetch_funding_transaction_fee_invoice(transaction.fee))
+        })?;
+
+        let invoice = Invoice::from_str(&invoice_str).context("Could not parse Invoice string")?;
+        let _payment_hash = invoice.payment_hash();
+
+        match ln_dlc::send_payment(&invoice_str) {
+            Ok(_) => {
+                // unset the funding transaction marking it as being paid.
+                self.unset_funding_transaction();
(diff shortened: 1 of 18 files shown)


holzeis:
It looks like we don't always broadcast the funding transaction when opening a channel. That results into a failed attempt to fetch the funding transaction from esplora, thus the test failing as no channel fee payment can be processed.

see also https://github.com/get10101/10101/issues/873

mobile/native/src/ln_dlc/lightning_subscriber.rs
```diff
@@ -0,0 +1,35 @@
...
+                        continue;
+                    }
+                }
+                Err(e) => {
+                    tracing::error!("Failed to receive event: {e:#}");
+                    continue;
+                }
+            };
+
+            match event {
```
luckysori, line 24: 🙃 If you move this `match` inside of the `if let` above then you don't need to use `continue`s.

mobile/native/src/channel_fee.rs
```diff
@@ -0,0 +1,169 @@
...
+        tracing::debug!("Received new inbound channel with id {channel_id_as_str}");
+
+        match ln_dlc::get_funding_transaction(channel_id) {
+            Ok(txid) => {
+                let transaction: Option<EsploraTransaction> = tokio::task::block_in_place(|| {
+                    Handle::current().block_on(self.fetch_funding_transaction(txid))
+                })
+                .ok();
+
+                if let Some(transaction) = transaction {
```
luckysori, line 101: 🔧 I think we should emit an error if we can't find the transaction on the blockchain.

mobile/native/src/channel_fee.rs
```diff
@@ -0,0 +1,169 @@
...
+use ln_dlc_node::node::rust_dlc_manager::ChannelId;
+use serde::Deserialize;
+use serde::Serialize;
+use std::sync::Arc;
+use std::sync::Mutex;
+use tokio::runtime::Handle;
+
+#[derive(Clone)]
+pub struct ChannelFeePaymentSubscriber {
+    pub open_channel_tx: Arc<Mutex<Option<EsploraTransaction>>>,
```
luckysori, line 19: 🤔 I'm surprised that we need an `Arc<Mutex<_>>` here.

luckysori: 🔧 Perhaps relatedly, we could model this as an enum:

```rust
enum ChannelOpeningFeeState {
    PendingOpen, // before `ChannelReady`
    Owing { fee: u64 }, // after `ChannelReady`, before first `PaymentClaimed`
    Paid, // after first `PaymentClaimed`
}
```

It might help make the code clearer. For instance, every time we get a new `PaymentClaimed` and we have already `Paid` we can be sure that we can safely ignore the event. But if we get it at `PendingOpen` we can emit an error as something has gone very wrong.

mobile/native/src/channel_fee.rs
```diff
@@ -0,0 +1,169 @@
...
+            Handle::current().block_on(self.fetch_funding_transaction_fee_invoice(transaction.fee))
+        })
+        .ok();
+
+        if let Some(invoice) = invoice {
+            match ln_dlc::send_payment(&invoice) {
+                Ok(_) => {
+                    // unset the funding transaction marking it as being paid.
+                    self.unset_funding_transaction();
+                    tracing::info!("Successfully paid funding transaction fees of {funding_tx_fees_msats} msats to {}", config::get_coordinator_info().pubkey);
```
luckysori, line 80: I think this is okay for now, but we need to wait for `Event::PaymentSent` to know that the payment has succeeded.

mobile/native/src/channel_fee.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl ChannelFeePaymentSubscriber {
+    pub fn new() -> Self {
+        Self {
+            open_channel_tx: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    // Attempts to pay the transaction fees for opening an inbound channel.
```
luckysori, line 53: 🙃 Can be a docstring.

luckysori (approved):
Nice!

I've left a bunch of primarily stylistic comments.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -82,6 +85,19 @@ pub fn get_oracle_pubkey() -> XOnlyPublicKey {
...
 
+pub fn get_funding_transaction(channel_id: &ChannelId) -> Result<Txid> {
+    let node = NODE.get();
+    let channel_details = node.inner.channel_manager.get_channel_details(channel_id);
+
+    if let Some(channel_details) = channel_details {
+        if let Some(funding_txo) = channel_details.funding_txo {
+            return Ok(funding_txo.txid);
+        }
+    }
```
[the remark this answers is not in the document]
holzeis: well I guess `original_funding_txo` would be more correct, but at the time where the channel gets opened, the `funding_txo` and the `original_funding_txo` should always be the same.

mobile/native/src/channel_fee.rs
holzeis: Good catch 👍 I am going to change that.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
holzeis: i changed it to the `original_funding_txo`

mobile/native/src/channel_fee.rs
holzeis: After revisiting that issue, we decided to not improve that as this solution is only intermediary.

mobile/native/src/channel_fee.rs
holzeis: I like the idea, but given that this is only an intermediary solution, I am not going to spent more effort into improving the design.

PR #866, opened 2023-06-30
Flutter Analyzer warning fixes

da-kami:
After upgrading to Flutter 3.10.5 these warnings popped up. Fixing them, let's see if CI cares, might be that we don't enforce these, but I would still fix them to clean up the analyzer warnings.


luckysori (approved):
Thanks!

klochowicz:
thanks @da-kami ! I have also fixed it in #867 

I think your still fails - after fixing the lints it complained still about removing const in the children.... - we can either update this one again or just look into #867

klochowicz:
Closing in favour of #867

PR #869, opened 2023-06-30
Remove custom database image

klochowicz:
- Makes sure the local oracle can be started properly.
- cleans up the top-level dir structure a bit

See commits for details.


docker-compose.yml
```diff
@@ -147,11 +147,10 @@ services:
 
   db:
     container_name: db
-    build:
-      context: database
-      dockerfile: Dockerfile
+    image: postgres:15-alpine
     volumes:
       - postgres:/var/lib/postgresql/data
+      - ./services/database:/docker-entrypoint-initdb.d/
```
holzeis, line 153: Mounting the init script like that will not change the fact that the database has to be removed before.

klochowicz: yes, but the database is removed with `just wipe` - the Docker image was not.

holzeis: Doesn't just wipe rub docker-compose down -v?

klochowicz: yes, it runs it, but it does not rebuild the Docker *image*, which was necessary in our old method.

klochowicz: 🤔 did it work for you? or did you end up manually using `psql` to add the table? or force rebuild of the `Dockerfile` ?

holzeis: hmm.. I guess i did run docker-compose up --build -d at some point.

holzeis (approved):
Thanks 👍

klochowicz:
bors ping

klochowicz:
bors was stuck, so I merged the PR without it. tests were passing.

PR #875, opened 2023-06-30
chore: Add original funding txo to list channels response

holzeis:
The funding txo will be different depending on whether a dlc channel is open or closed.

Adding the original funding txo so we can see the actual funding transaction that has been broadcasted on-chain.

### DIFF
--- a/crates/ln-dlc-node/src/ln/channel_details.rs
+++ b/crates/ln-dlc-node/src/ln/channel_details.rs
@@ -13,6 +13,8 @@ pub struct ChannelDetails {
     pub counterparty: PublicKey,
     #[serde(serialize_with = "optional_outpoint_to_str")]
     pub funding_txo: Option<OutPoint>,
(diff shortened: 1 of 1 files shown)


crates/ln-dlc-node/src/ln/channel_details.rs
```diff
@@ -13,6 +13,8 @@ pub struct ChannelDetails {
     pub counterparty: PublicKey,
     #[serde(serialize_with = "optional_outpoint_to_str")]
     pub funding_txo: Option<OutPoint>,
+    #[serde(serialize_with = "optional_outpoint_to_str")]
+    pub original_funding_outpoint: Option<OutPoint>,
```
luckysori, line 17: 🙃To be consistent.

```suggestion
    pub original_funding_txo: Option<OutPoint>,
```

luckysori (approved):
Good, thank you!

PR #877, opened 2023-07-01
feat: add admin api to open channel to other nodes

### LINKED ISSUE #876: Allow coordinator to open channels again
The coordinator needs to be able to open channels to other lightning nodes. This capability was unfortunately removed when the faucet feature was added.

bonomat:
resolves #876

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -6,6 +6,7 @@ use crate::admin::list_channels;
 use crate::admin::list_dlc_channels;
 use crate::admin::list_on_chain_transactions;
 use crate::admin::list_peers;
+use crate::admin::open_channel;
 use crate::admin::send_payment;
 use crate::admin::sign_message;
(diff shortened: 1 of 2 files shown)


coordinator/src/routes.rs
```diff
@@ -92,8 +93,8 @@ pub fn router(
         .route("/api/trade", post(post_trade))
         .route("/api/register", post(post_register))
         .route("/api/admin/balance", get(get_balance))
-        .route("/api/admin/channels", get(list_channels))
-        .route("/api/channels", post(open_channel))
+        .route("/api/admin/channels", get(list_channels).post(open_channel))
+        .route("/api/channels", post(channel_faucet))
```
holzeis, line 97: I think we can remove that endpoint. After merging https://github.com/get10101/10101/pull/816 this shouldn't be used anymore.

holzeis (approved):
👍

PR #889, opened 2023-07-04
Feat/configurable channel capacity

da-kami:
fixes #166

I tried to keep the changeset somewhat contained.

I will see to add some e2e tests in follow up that cover the edge cases.

---

Notes concerning the changes in the app:

1. we could extract the `ChannelInfoService` into a separate `ChangeNotifier`
	- This can enable more testing; at the moment the send and the create-invoice screen can hardly be tested because the `ChannelInfoService` is not injected, so we cannot use a mock there.
2. Se could add events for `ChannelInfo` updates from backend to frontend
	- In the current solution I think it *can* happen that the user triggers a new channel open, then switches to the trade screen but the initialization of the trade screen might not yet receive the `ChannelInfo`. Such a scenario could result in wrongly calculated  trade values. I don't think this scenario is likely to happen because the user usually waits until the balance is updated and then the `ChannelInfo` will be available. 
	- Enables us to use the inbound/outbound liquidity instead of doing calculations with the balance; not sure if this is actually a favourable change because the UX might change slightly (the balance might then not be directly aligned with the capacity)


da-kami:
Converting to draft because when testing with the local setup I ran into some problems with:

- the UI throwing an exception that `ChannelInfo` is not initialized
- I am unable to open jit channel (not sure if that is related to the change, still investigating)

da-kami:
> Converting to draft because when testing with the local setup I ran into some problems with:
> 
>     * the UI throwing an exception that `ChannelInfo` is not initialized

I did not encounter this problem again after cleaning up a `late` initialization that was unnecessary in dart.

> 
>     * I am unable to open jit channel (not sure if that is related to the change, still investigating)

It was unrelated to the changes, the channel was a zombie after funding. Seems this happens on my machine as well sometimes. I overcame it by opening another channel through the faucet interface and could confirm that it worked - @bonomat was my witness.

I have not retested after the last commits were added, but there no changes in the logic, only alignment and refactoring so reopening with the expectation that it still works. Will re-test tomorrow :)

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -629,14 +630,21 @@ where
...
-                // Currently the channel capacity is fixed for the beta program
-                let channel_value = JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT;
+                let channel_value = expected_outbound_amount_msat / 1000 * LIQUIDITY_MULTIPLIER;
+
+                if channel_value > JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX {
+                    tracing::warn!(%intercepted_id, %channel_value, channel_value_maximum=%JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX, "Failed to open channel because maximum channel value exceeded");
+
+                    if let Err(err) = self.channel_manager.fail_intercepted_htlc(intercept_id) {
+                        tracing::error!(%intercepted_id, "Could not fail intercepted htlc {err:?}")
+                    }
```
holzeis, line 646: 🔧  We have to return here, otherwise we will continue to open a channel anyways.

crates/ln-dlc-node/src/ln/mod.rs
```diff
@@ -16,12 +16,19 @@ pub(crate) use logger::TracingLogger;
...
 /// This constant only applies to the coordinator.
-pub(crate) const JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT: u64 = 200_000;
+pub(crate) const JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX: u64 = 200_000;
+
+/// The multiplier to be used by the coordinator to define the just in time channel liquidity
+///
+/// The liquidity provided by the trader will be multiplied with this value to defined the channel
+/// value.
+/// See `JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX` for the maximum channel value.
+pub(crate) const LIQUIDITY_MULTIPLIER: u64 = 2;
```
holzeis: nit: It would be nice if we could configure these values in the `coordinator-settings.toml`

crates/ln-dlc-node/src/ln/mod.rs
```diff
@@ -30,6 +30,11 @@ pub const JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX: u64 = 200_000;
 /// See `JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX` for the maximum channel value.
 pub const LIQUIDITY_MULTIPLIER: u64 = 2;
 
+/// Sats/vbyte rate for position
+///
+/// The coordinator and the app have to align on this to agree on the fees.
+pub const CONTRACT_TX_FEE_RATE: u64 = 4;
```
holzeis, line 36: nit: It would be nice if we could configure that value in the `coordinator-settings.toml`

mobile/ios/Podfile.lock
```diff
@@ -41,8 +41,8 @@ SPEC CHECKSUMS:
   package_info_plus: fd030dabf36271f146f1f3beacd48f564b0f17f7
   path_provider_foundation: eaf5b3e458fc0e5fbb9940fb09980e853fe058b8
   share_plus: 599aa54e4ea31d4b4c0e9c911bcc26c55e791028
-  shared_preferences_foundation: e2dae3258e06f44cc55f49d42024fd8dd03c590c
+  shared_preferences_foundation: 5b919d13b803cadd15ed2dc053125c68730e5126
 
 PODFILE CHECKSUM: cc1f88378b4bfcf93a6ce00d2c587857c6008d3b
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.12.0
```
holzeis, line 48: 🤔 did upgrading your cocoapods not work?

crates/ln-dlc-node/src/ln/mod.rs
da-kami: I think that would be a bigger change because the app also depends on these values. That means that the app initially has to fetch these from the coordinator. That can easily be added on top.

mobile/ios/Podfile.lock
da-kami: It did - I have `1.12.1` installed, no idea what keeps changing this file and why. I'll revert that.

crates/ln-dlc-node/src/ln/mod.rs
da-kami: (see related comment on `LIQUIDITY_MULTIPLIER`)

crates/ln-dlc-node/src/ln/event_handler.rs
da-kami: thanks, good catch!

mobile/lib/common/application/channel_info_service.dart
```diff
@@ -1,36 +1,38 @@
-class ChannelConstraintsService {
-  const ChannelConstraintsService();
+import 'package:get_10101/common/domain/channel.dart';
+import 'package:get_10101/ffi.dart' as rust;
 
-  int getLightningChannelCapacity() {
-    // TODO: Fetch from backend
-    // This value is what we agree on as channel capacity cap for the beta
-    return 200000;
+import '../domain/model.dart';
```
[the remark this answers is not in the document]
klochowicz: can we revise this please?

Official Dart guide clearly recommends the opposite:
https://dart.dev/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory

I don't think there's a good reason to depart from writing code with their best guidelines in mind, esp that we're not pros in Dart.

mobile/native/src/api.rs
```diff
@@ -213,6 +214,26 @@ pub fn force_close_channel() -> Result<()> {
     ln_dlc::close_channel(true)
 }
 
+/// Returns channel info if we have a channel available already
+///
+/// If no channel is established with the coordinator `None` is returned.
+pub fn channel_info() -> Result<Option<ChannelInfo>> {
```
klochowicz, line 220: consider just returning just `Option<ChannelInfo>` - do you really need to distinguish why there's no channel info?

mobile/native/src/commons/api.rs
```diff
@@ -26,3 +27,18 @@ impl From<Price> for trade::Price {
         }
     }
 }
+
+pub struct ChannelInfo {
+    /// The total capacity of the channel as defined by the funding output
+    pub channel_capacity: u64,
+    pub reserve: Option<u64>,
+}
```
klochowicz, line 35: 🤔 could the remaining free functions be part of channel info?

-`coordinator_liquidity_multiplier`
- `max_channel_value`
- `contract_tx_fee_rate?`

I guess then the capacity would need to become an `Option` (and lack of capacity would need that there's no channel that's available & ready).

mobile/pubspec.lock
```diff
@@ -53,10 +53,10 @@ packages:
     dependency: transitive
     description:
       name: build
-      sha256: "3fbda25365741f8251b39f3917fb3c8e286a96fd068a5a242e11c2012d495777"
+      sha256: "80184af8b6cb3e5c1c4ec6d8544d27711700bc3e6d2efad04238c7b5290889f0"
```
klochowicz, line 56: 🧐I don't think bumping all these dependencies belongs to this PR (it's unrelated to advertised changes).

I would encourage creating small PRs that update deps in the future :)

crates/ln-dlc-node/src/ln/mod.rs
klochowicz: in my review, I noted that they could become part of the `ChannelInfo` - this would be one potential way to achieve passing it to the app from the coordinator.

mobile/native/src/api.rs
da-kami: I think we should distinguish between a channel not existing yet `Ok(None)` and errors; those are completely different scenarios and a users might have to be made aware of errors (eventually).

mobile/pubspec.lock
da-kami: Happy to split this off - I thought I would finish this ticket earlier but it had a bit of a rat tail :)

crates/ln-dlc-node/src/ln/mod.rs
da-kami: Fair point - would you be OK if we schedule this in a follow up?

mobile/lib/common/application/channel_info_service.dart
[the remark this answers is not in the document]
da-kami: I had a quick look and I did not see an explicit recommendation to *not* use package imports. Maybe I missed something.
I really don't have a preference - I am basically following what my IDE does. I tried to configure a default to use package imports at some point but could not figure out a way to achieve that in Android studio. Happy if someone can point me to it. I find it pretty annoying that I keep running into this problem ;)

mobile/pubspec.lock
da-kami: done

holzeis (approved):
Thanks for addressing my feedback!

mobile/native/src/commons/api.rs
da-kami: Would address this in a follow up - the current solution works and can be improved.

mobile/pubspec.lock
da-kami: I was hoping to finish this work faster.

Should have split this off properly. The initial commit is actually not complete. There were changes in the Dart code needed after upgrading the packages, but I did not separate these properly. Sorry for that. I'll pay more attention next time.

mobile/lib/common/application/channel_info_service.dart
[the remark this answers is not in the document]
da-kami: Let's continue this discussion outside of this this PR - I'm happy to change our policy and also add a lint, but I don't think the decision should be taken in this PR

Issue #508, opened 2023-04-25
Prometheus metrics


bonomat:
Prometheus came handy in the past to track certain metrics, imho we should add it to our coordinator as well and start tracking: 

(including but not limited to)
- [x] connected peers
- [ ] connected users from our app
- [x] in-bound channel liquidity
- [x] out-bound channel liquidity
- [x] open position (amounts, our_margin, their_margin)
- [ ] if a position gets closed
- [ ] if a position gets opened
- [ ] if a channel gets opened
- [ ] if a channel gets closed
- [x] balance (on/off-chain)
- [ ] amounts traded
- [ ] trades per user
- [ ] deposited money

Note: this is a mix between operational and business metrics. Feel free to seperate them.

bonomat:
Let's stick with the libraries we know for now please. Time is not our friend at the moment and it's better to use tools we know already.

klochowicz:
but this.... saves time? 🤔

luckysori:
We are already using `autometrics` a bit, but we suppose we still want to add some of these bespoke metrics.

bonomat:
Created a follow-up ticket https://github.com/get10101/10101/issues/896 but left out 

- if a position gets closed
- if a position gets opened
- if a channel gets opened
- if a channel gets closed
- amounts traded
- trades per user
- deposited money

which I believe are hard to track using prometheus metrics

PR #890, opened 2023-07-04
Add a few prometheus metrics to coordinator

bonomat:
partly resolves https://github.com/get10101/10101/issues/508

Will open new tickets for the remaining metrics which need more work.


coordinator/src/bin/coordinator.rs
```diff
@@ -128,6 +133,49 @@ async fn main() -> Result<()> {
...
+                spawn_blocking(move || {
+                    let cx = opentelemetry::Context::current();
+                    let channels = node.inner.channel_manager.list_channels();
+                    for channel_detail in channels {
+                        let key_values = [
+                            KeyValue::new("channel_id", hex::encode(channel_detail.channel_id)),
+                            KeyValue::new("is_outbound", channel_detail.is_outbound),
+                            KeyValue::new("is_public", channel_detail.is_public),
+                        ];
+                        CHANNEL_BALANCE_MSATOSHI.observe(
```
klochowicz: nit: do we really want to go with `msat` on our metrics? it makes sense for feature-proofing app in certain places, but for our reporting I can't imagine us caring about these fractions (not that they're used anyway).

coordinator/src/metrics.rs
```diff
@@ -49,6 +51,16 @@ lazy_static! {
         .u64_observable_gauge("node_balance_satoshi")
         .with_description("Node balance in satoshi")
         .init();
+
+    // position metrics
+    pub static ref POSITION_QUANTITY: ObservableGauge<f64> = METER
+        .f64_observable_gauge("position_quantity_contracts")
```
klochowicz, line 57: nit: "position_quantity" or "position_contracts" should be enough. Having both is confusing

coordinator/src/metrics.rs
```diff
@@ -64,12 +76,88 @@ pub fn init_meter() -> PrometheusExporter {
...
+            return;
+        }
+    };
+
+    let mut margin_long = 0;
+    let mut margin_short = 0;
+    let mut quantity_long = 0.0;
+    let mut quantity_short = 0.0;
+
+    // Note: we should filter positions here by BTCUSD once we have multiple contract symbols
```
klochowicz, line 109: the easiest way to not forget is to put the following inside the loop instead of this comment:
 
```rust
debug_assert!(position.contract_symbol == ContractSymbol::BTCUSD' "we should filter positions here by BTCUSD once we have multiple contract symbols`)
```

It will be **not** present in production, but will blow up right after someone enables different contract symbols on their dev machine 🤭

klochowicz (approved):
awesome work!!!

PR #891, opened 2023-07-05
chore(justfile): Improve log browsing experience

klochowicz:
Open by default with `less` in follow mode (so that it refreshes the contents).

It's like vim, one can search using `/` or by highlighting a word with `*`.

In order to stop following and start browsing historical entries
(with pagination!), hit Ctrl+C.


luckysori (approved):
😃

Issue #895, opened 2023-07-05
Position model based on the DLC with associated orders


da-kami:
Recently this topic came up several times, so I decided to create a ticket that sums my vision up:

In ItchySats we always had the problem that the application state was independent of the actual DLC state - and it could diverge leading to a "wrong state" when displaying information.

In 10101 we should overcome this by:

Only have sub-channels and load meta-data for them in our database.
Our database then does not contain any position "state" (such as opened, closing...) but this always comes from the DLC. 
The database however would contain any metadata that the DLC does not store itself, but we load this metadata base on the DLC.
This should include the proper association of the DLC (and position metadata) to it's orders.

I imagine this when we load positions:

1. We load the subchannel data first
2. We load the metadata associated with the subchannel based on the subchannel-id
3. Depending on the state of the subchannel we might do different calculations
4. We build the position objects and and provide them to the application

What does this mean for the creation lifecycle of a position:

1. An order gets matched
2. We request execution
5. Once the order was filled we safe the metadata of the position associated to the subchannel-id. Additionally we associate the order with this metadata in our database.

At the moment we base the position creation / deletion on channel messages being exchanged (i.e. the finalized messages). 
I think this model is OK to *trigger* doing something, but we should act based on the subchannel *state* rather than on the messages. So, we should load the state once a message happend and then act.

da-kami:
CC @luckysori @holzeis 

Please have a look when you have a time - I think we should refine this and start working on it soon - the shortcuts in the current model will become problematic soon :)

holzeis:
see also https://github.com/get10101/10101/issues/530

I think that's a technical debt that we should address. However, I don't see an immediate need for that - the referenced tickets are not critical to justify tackling this immediately. Do you see any immediate issues from our current design that will prevent us achieving our Q3 goal?

da-kami:
Since this is basically a duplicate of #530 I'll close this ticket.

@luckysori if you want to incorporate any of the information I collected here into #530 please do so.

---

@holzeis :

> However, I don't see an immediate need for that - the referenced tickets are not critical to justify tackling this immediately. Do you see any immediate issues from our current design that will prevent us achieving our Q3 goal?

I agree, but I would advise not to push this out for too long. If it won't get priority in Q3 I would recommend to at least do it in Q4.

Note that properly associating the `Order` to the `Position` could be done separately as well.

I think at the moment model changes are still a quite low hanging fruit to fix/change. It might become harder if we built on top of the wrong model for longer.
[...]

Issue #871, opened 2023-06-30
Show realized pnl upon successfully closing a position


holzeis:
Currently we are showing a unrealized pnl on a successfully closed position. That should actually be the realized pnl.

![Image](https://user-images.githubusercontent.com/382048/250012820-28e7ebaf-d504-4865-8a59-bc11ea4cdaf1.png)

da-kami:
I think the easiest solution is to just rename `Unrealized P/L` to `Realize P/L` if we are in a closing scenario - I am not sure the app is aware of which scenario it is though.

---

We can also solve this by calculating the realized P/L based on the execution price of the order. But I am not sure it's worth doing this before we cleanup the model (see below...).

---

To *really* solve this I think we would have to cleanup the model properly - the position should be closely coupled to the DLC (i.e. subchannel).

My vision for this is to only have sub-channels and load meta-data for them from our database.
Our database then does not contain any position "state" (such as opened, closed...) but this always comes from the DLC. The database however would contain any metadata that the DLC does not store itself, but we load this metadata base on the DLC.
This should include the proper association of the DLC (and position metadata) to it's orders.

da-kami:
see: https://github.com/get10101/10101/issues/895

Restioson:
Assigning my self for #897; further work as per @da-kami's comment I'm not sure I can do on my own since I'm still a little unfamiliar with how the Rust side of 10101 works. I can try it with some mentorship though if it's not too time critical

PR #897, opened 2023-07-05
Change "Unrealized P/L" to "P/L" when order closes position and is filled

Restioson:
Addresses #871 but does not calculate the realized P/L from the channel as per #895 (see Daniel's comment on #871).

### DIFF
--- a/mobile/lib/features/trade/trade_screen.dart
+++ b/mobile/lib/features/trade/trade_screen.dart
@@ -300,6 +300,7 @@ class TradeScreen extends StatelessWidget {
   Widget createSubmitWidget(PendingOrder pendingOrder, Amount pnl,
       SubmitOrderChangeNotifier submitOrderChangeNotifier, BuildContext context) {
     String bottomText;
+    String pnlText = "Unrealized P/L";
 
     switch (pendingOrder.state) {
       case PendingOrderState.submittedSuccessfully:
@@ -313,6 +314,9 @@ class TradeScreen extends StatelessWidget {
       case PendingOrderState.orderFilled:
         if (pendingOrder.positionAction == PositionAction.close) {
           bottomText = "Your position has been closed.";
+          // At this point, the position is closed so P/L has been realized
+          // TODO - calculate based on subchannel finalized event
+          pnlText = "P/L";
         } else {
           bottomText = "Congratulations! Your position will be shown in the Positions tab.";
         }
@@ -328,7 +332,7 @@ class TradeScreen extends StatelessWidget {
             runSpacing: 10,
             children: [
               pendingOrder.positionAction == PositionAction.close
-                  ? ValueDataRow(type: ValueType.amount, value: pnl, label: "Unrealized P/L")
+                  ? ValueDataRow(type: ValueType.amount, value: pnl, label: pnlText)
                   : ValueDataRow(
                       type: ValueType.amount,
                       value: submitOrderChangeNotifier.pendingOrderValues?.margin,


da-kami (approved):
I think this might be good enough for now.

It could still be that a wrong realized P/L is shown, but in my understanding only if the price moves very fast (and the execution price becomes outdated).

da-kami:
@Restioson please wait for the review from @holzeis though, I am not sure he had something else in mind :)

holzeis:
hmm I was thinking of actually using the value calculated during closing a position and returning it to the frontend to show the realized pnl.

https://github.com/get10101/10101/blob/1c62f9eeafa231d8c3b807ea101e70d53e8fc3c4/coordinator/src/node.rs#L255-L274

But I guess that might not make a big difference to what we currently have.

da-kami:
> hmm I was thinking of actually using the value calculated during closing a position and returning it to the frontend to show the realized pnl.
> 
> https://github.com/get10101/10101/blob/1c62f9eeafa231d8c3b807ea101e70d53e8fc3c4/coordinator/src/node.rs#L255-L274
> 
> But I guess that might not make a big difference to what we currently have.

The execution is assumed to be based on the execution price anyway - this execution price comes from the orderbook match. The coordinator does not decide the execution price - unless it is a scenario where the position is expired ([see this code for reference](https://github.com/get10101/10101/blob/ba3b50fb8613c956617bb638d4cb5c28851a437b/coordinator/src/bin/coordinator.rs#L150-L201)). But this case is not relevant for this ticket, because the user does not see this success dialog then - and the coordinator triggering a collab close a workaround for the beta after expiry.

If we would associate the orders with our positions correctly already we should just know the execution price and could use it for the calculation.  I don't think we have to return anything from the coordinator to the app, but we have to make the `Position` aware of its closing order (or at least the closing price).

The `pnl` in the `Position` in the UI is currently not aware of orders, it is set here:

https://github.com/get10101/10101/blob/770ed5db32e68ecd4817a6d5637a1b82e1d3b9cf/mobile/lib/features/trade/trade_screen.dart#L53-L58

which falls back to the `PositionChangeNotifier` where we update the pnl based on the latest price.

https://github.com/get10101/10101/blob/caf534d24e1973b6c445ca25f9afb22d8464ddf0/mobile/lib/features/trade/position_change_notifier.dart#L30-L62

In order to use the execution price we would have to make the position aware of its closing order and then lock in the price based on that order instead of using the current price. 

---

@Restioson sorry I did not see this part that clearly in our discussion yesterday.

To achieve using the closing price in a simple way you can:

1. Add some kind of `closing_price` to the `Position`
2. Set this `closing_price` when we update the position in [`update_position_after_order_submitted`](https://github.com/get10101/10101/blob/01f393c2efebfbed45ff61f5d8850a21cb6484f5/mobile/native/src/trade/position/handler.rs#L64-L76); which includes adding the optional `closing_price` to the model so it is communicated in `PositionUpdateNotification`.
3. Add logic to not update the `PnL` if the `closing_price` in the UI `Position1 is not null.

Note: This assumes that the order will be executed with the `closing_price` - if we fail the order execution we would have to not only reset the position to `Open`, but also remove the `closing_price` so we have an unrealized PnL again. We definitely don't always do that (i.e. we don't always re-open a position if the order execution fails), but adding the `closing_price` won't make this any worse, so we can treat that as a separate problem :)

da-kami:
@holzeis after our discussion I think this PR is ready to merge

I captured potential follow-up work here: https://github.com/get10101/10101/issues/922

Feel free to approve and merge :)

holzeis (approved):

PR #901, opened 2023-07-05
Fix just command to run app on android device against locally running services

da-kami:
The `$` was misplaced.
Additionally, fetching the internet address using `ifconfig.co` does not always work - a more resilient solution is to grep ifconfig output to fetch the local network address.

Thanks to stackoverflow https://stackoverflow.com/questions/13322485/how-to-get-the-primary-ip-address-of-the-local-machine-on-linux-and-os-x

### DIFF
--- a/justfile
+++ b/justfile
@@ -80,7 +80,7 @@ run-regtest args="":
 
 run-local-android args="":
     #!/usr/bin/env bash
-    $LOCAL_IP=(curl -sS ifconfig.co)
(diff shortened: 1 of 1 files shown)


da-kami:
reviving this command is pretty great btw. deploy to phone but using the local dev env is nice :)

klochowicz (approved):

Restioson:
Locally this change outputs

```
Target file "--dart-define=ESPLORA_ENDPOINT=http://10.5.0.1
172.20.0.1
172.17.0.1
192.168.100.157
192.168.122.1:3000" not found.
```

PR #907, opened 2023-07-05
Display estimated order-matching fee when trading

luckysori:
This is important as otherwise it's unclear why we are not able to trade with a margin smaller than the usable balance.

On the other hand, I have a hard time making this look nice because my UI design skills are severely lacking. As such, suggestions and even force-pushes to the branch are welcome.

This is how it looks now:

image


holzeis (approved):
LGTM 👍

PR #908, opened 2023-07-05
Ensure that wallet history text does not overflow

luckysori:
At least on Linux, when running `just run` I would consistently get warnings directly on the UI on every wallet history item. Each one would say `Bottom overflowed by 6.0 pixels`. This is annoying when developing, but I believe it's also a sign that we were doing something wrong. Hence this fix.

Before:

image

After:

image

### DIFF
--- a/mobile/lib/features/wallet/wallet_history_item.dart
+++ b/mobile/lib/features/wallet/wallet_history_item.dart
@@ -125,13 +125,13 @@ class WalletHistoryItem extends StatelessWidget {
                         style: TextStyle(
                             color: color,
                             fontFamily: "Courier",
-                            fontSize: 16,
+                            fontSize: 14,
                             fontWeight: FontWeight.bold))
                   ]),
                 ),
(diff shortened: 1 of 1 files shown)


holzeis:
Left without this changes and right with them. to be honest I don't think reducing the font size is an improvement on the phone.

Screenshot 2023-07-05 at 20 43 37

holzeis (commented):
I am against this change as we are targeting mobile users and I don't think this will improve the experience for them.

luckysori:
It does look pretty terrible on the phone, what a shame.

Thanks for testing it out, @holzeis.

klochowicz:
@luckysori perhaps you're missing some fonts and the substitutes are different size? this does not happen on desktop of macOS.

luckysori:
> @luckysori perhaps you're missing some fonts and the substitutes are different size? this does not happen on desktop of macOS.

Could be!

PR #911, opened 2023-07-06
chore(justfile): Improve `just coordinator-detached`

klochowicz:
Wait for coordinator to actually start before stating that it started successfully.

Note: We can't do this with maker for now, as there's no HTTP API there yet.


luckysori (approved):
Thanks!

da-kami (approved):

PR #913, opened 2023-07-06
Lock used utxos

### LINKED ISSUE #873: Channel without funding transaction on chain
I just stumbled upon that issue when testing https://github.com/get10101/10101/pull/864. It appears we are sometimes creating channels which do not succeed in publishing their funding transaction.

The channel itself does not appear to look any different than a channel with the `funding_txo` being published.

Tried to verify if that issue also appears on production, but we do not see the `original_funding_txo` from the `list_channels` api.

see also https://github.com/get10101/10101/pull/875

bonomat:
fixes #873

### DIFF
--- a/crates/ln-dlc-node/src/fee_rate_estimator.rs
+++ b/crates/ln-dlc-node/src/fee_rate_estimator.rs
@@ -32,6 +32,16 @@ pub struct FeeRateEstimator {
     fee_rate_cache: RwLock<HashMap<ConfirmationTarget, FeeRate>>,
 }
 
+pub trait EstimateFeeRate {
+    fn estimate(&self, target: ConfirmationTarget) -> FeeRate;
+}
+
+impl EstimateFeeRate for FeeRateEstimator {
+    fn estimate(&self, target: ConfirmationTarget) -> FeeRate {
+        self.get(target)
+    }
+}
+
 impl FeeRateEstimator {
     /// Constructor for the [`FeeRateEstimator`].
     pub fn new(esplora_url: String) -> Self {
@@ -66,7 +76,7 @@ impl FeeRateEstimator {
         }
     }
 
-    pub(crate) fn get(&self, target: ConfirmationTarget) -> FeeRate {
+    fn get(&self, target: ConfirmationTarget) -> FeeRate {
         self.cache_read_lock()
             .get(&target)
             .copied()
@@ -110,6 +120,6 @@ impl FeeRateEstimator {
 impl FeeEstimator for FeeRateEstimator {
     #[autometrics]
     fn get_est_sat_per_1000_weight(&self, confirmation_target: ConfirmationTarget) -> u32 {
-        (self.get(confirmation_target).fee_wu(1000) as u32).max(FEERATE_FLOOR_SATS_PER_KW)
+        (self.estimate(confirmation_target).fee_wu(1000) as u32).max(FEERATE_FLOOR_SATS_PER_KW)
     }
 }
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -1,11 +1,11 @@
-use crate::fee_rate_estimator::FeeRateEstimator;
+use crate::fee_rate_estimator::EstimateFeeRate;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Error;
 use anyhow::Result;
 use autometrics::autometrics;
 use bdk::blockchain::Blockchain;
-use bdk::blockchain::EsploraBlockchain;
+use bdk::blockchain::GetBlockHash;
 use bdk::blockchain::GetHeight;
 use bdk::database::BatchDatabase;
 use bdk::wallet::AddressIndex;
@@ -25,16 +25,18 @@ use std::sync::Arc;
 use std::time::Instant;
 use tokio::sync::RwLock;
 
-pub struct Wallet<D>
+pub struct Wallet<D, B, F>
 where
     D: BatchDatabase,
+    B: Blockchain,
+    F: EstimateFeeRate,
 {
     // A BDK blockchain used for wallet sync.
-    pub(crate) blockchain: Arc<EsploraBlockchain>,
+    pub(crate) blockchain: Arc<B>,
     // A BDK on-chain wallet.
     inner: Mutex<bdk::Wallet<D>>,
     settings: RwLock<WalletSettings>,
-    fee_rate_estimator: Arc<FeeRateEstimator>,
+    fee_rate_estimator: Arc<F>,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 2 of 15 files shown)


crates/ln-dlc-node/src/fee_rate_estimator.rs
```diff
@@ -32,6 +33,43 @@ pub struct FeeRateEstimator {
     fee_rate_cache: RwLock<HashMap<ConfirmationTarget, FeeRate>>,
 }
 
+#[async_trait]
+pub trait EstimateFeeRate {
+    fn estimate(&self, target: ConfirmationTarget) -> FeeRate;
+    async fn update(&self) -> Result<()>;
+}
```
luckysori, line 40: 🔧 I think this should be defined next to the component that requires this contract i.e. the wallet.

I would keep the `get` method and just call it when implementing the foreign trait method `estimate`.

crates/ln-dlc-node/src/fee_rate_estimator.rs
```diff
@@ -32,6 +33,43 @@ pub struct FeeRateEstimator {
     fee_rate_cache: RwLock<HashMap<ConfirmationTarget, FeeRate>>,
 }
 
+#[async_trait]
+pub trait EstimateFeeRate {
+    fn estimate(&self, target: ConfirmationTarget) -> FeeRate;
+    async fn update(&self) -> Result<()>;
```
luckysori, line 39: 🤔 I don't think `update` needs to be part of this trait. The wallet only cares about getting the estimate, not how it is internally updated.

crates/ln-dlc-node/src/fee_rate_estimator.rs
```diff
@@ -65,35 +103,6 @@ impl FeeRateEstimator {
...
-
-    pub(crate) fn get(&self, target: ConfirmationTarget) -> FeeRate {
-        self.cache_read_lock()
-            .get(&target)
-            .copied()
-            .expect("to have entries for all confirmation targets")
-    }
-
-    #[autometrics]
-    pub(crate) async fn update(&self) -> Result<()> {
```
luckysori, line 77: 🤔 This shouldn't be async anymore. I'll fix that separately.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -25,32 +25,32 @@ use std::sync::Arc;
 use std::time::Instant;
 use tokio::sync::RwLock;
 
-pub struct Wallet<D>
+pub struct Wallet<D, B, F>
```
luckysori, line 29: 🙃 Since we only need the generics for testing, I would isolate the rest of the modules from them by providing a `pub type` which is `Wallet<sled::Tree, EsploraBlockchain, FeeRateEstimator>`.

I can do this afterwards though.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -109,6 +114,11 @@ where
             .fee_rate(fee_rate)
             .enable_rbf();
 
+        let mut locked_out_points = self.locked_outpoints.lock();
+        for out_point in locked_out_points.iter() {
+            tx_builder.add_unspendable(*out_point);
+        }
```
luckysori, line 120: 🙃 Let's use `outpoint` instead of `out_point` consistently.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -132,7 +142,17 @@ where
...
-        Ok(psbt.extract_tx())
+        let transaction = psbt.extract_tx();
+
+        let prev_outpoints = transaction
+            .input
+            .iter()
+            .map(|input| input.previous_output)
+            .collect::<Vec<_>>();
+
+        locked_out_points.extend(prev_outpoints);
```
luckysori, line 153: 🔧 It might be a good idea to take this lock twice rather than holding it for an extended period of time. What do you think, @klochowicz?

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -256,3 +276,166 @@ where
...
+            );
+        }
+
+        let wallet = bdk::Wallet::new(&descriptors.0, None, Network::Regtest, database)?;
+
+        Ok(wallet)
+    }
+
+    #[tokio::test]
+    async fn wallet_with_two_utxo_should_be_able_to_fund_twice_but_not_three_times() {
```
luckysori, line 355: 🙃 Let's move the test to the top of the module, please.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -256,3 +276,166 @@ where
...
+
+    #[async_trait]
+    impl EstimateFeeRate for DummyFeeRateEstimator {
+        fn estimate(&self, _: ConfirmationTarget) -> FeeRate {
+            FeeRate::from_sat_per_vb(1.0)
+        }
+
+        async fn update(&self) -> Result<()> {
+            unimplemented!()
+        }
```
luckysori, line 397: Yeah, I think this confirms that we don't need it in the trait.

luckysori (approved):
LGTM. I've left some comments to clean this up a bit, but I think it's good.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -90,6 +93,8 @@ where
             "Finished on-chain sync",
         );
 
+        self.locked_outpoints.lock().clear();
```
holzeis, line 96: 🤔 is it safe to clear the utxos here? or do we even have to wait for the locked utxo to get a confirmation, before we can safely remove it from the `locked_outputs`?

holzeis (approved):
Nice 👍

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -103,12 +108,17 @@ where
...
         let mut tx_builder = locked_wallet.build_tx();
 
-        let fee_rate = self.fee_rate_estimator.get(confirmation_target);
+        let fee_rate = self.fee_rate_estimator.estimate(confirmation_target);
         tx_builder
             .add_recipient(output_script, value_sats)
             .fee_rate(fee_rate)
             .enable_rbf();
 
+        let mut locked_out_points = self.locked_outpoints.lock();
```
holzeis, line 117: I guess we should implement the same logic at all places where we are building transactions. e.g. the split, glue transactions and commit transaction..

I wonder if it is an issue that we are managing utxos differently for within `ln_dlc_wallet`

https://github.com/get10101/10101/blob/776acb3bf142c30cdfd3a3dfec9fdda123e147a8/crates/ln-dlc-node/src/ln_dlc_wallet.rs#L269-L294

crates/ln-dlc-node/src/ldk_node_wallet.rs
bonomat: I believe it is because during sync with on-chain we should catch-up with all transactions which are in the mempool

crates/ln-dlc-node/src/ldk_node_wallet.rs
bonomat: Thank you, I'l leave it for now then.

PR #921, opened 2023-07-07
Calculate and display open-channel fee estimate

da-kami:
fixes https://github.com/get10101/meta/issues/172

Please note: The screen font may look a bit odd for you because my default test setup is iPhone SE (3rd gen) with **Accessibility Large, Bold Text** activated to ensure that the screen does not break (unless users somehow increase the font even more, not sure if that is possible):

### DIFF
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -9,6 +9,7 @@ use bdk::blockchain::GetBlockHash;
 use bdk::blockchain::GetHeight;
 use bdk::database::BatchDatabase;
 use bdk::wallet::AddressIndex;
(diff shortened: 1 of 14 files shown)


da-kami:
Note: The info message as one text blob is not optimal - we could optimize this and highlight certain parts.

I tried to keep the text short - if you have suggestions on how to shorten it more please add suggestions!

da-kami:
I think it would be great to come to a conclusion for https://github.com/get10101/meta/issues/178 on top of this work. 
For transparency it would be great if all fees would be explained - if there is a channel opening fee, and additionally the coordinator adds fees to the invoice that might be a bit confusing / harder to explain.

klochowicz (approved):
👍 

nit: I got confused in the last commit message due to typo: `s/safe/save`

Issue #915, opened 2023-07-06
Show user if the coordinator is offline


holzeis:
If the coordinator is offline the app is practical only in view mode usable, but the user does not get any information about that.

We should show that the coordinator is offline and disable functionalities like 
- send & receive payment
- open or closing a position

The force close channel button should always be enabled though, informing the user that she has always a way out.

klochowicz:
Closely related to https://github.com/get10101/10101/issues/860

as for force-close channel button - I thought it's not enabled on mainnet?

holzeis:
> as for force-close channel button - I thought it's not enabled on mainnet?

not yet, but it should be considered when implementing that feature. don't panic should always be possible eventually ;)

PR #923, opened 2023-07-07
Handle expired orders in the app

klochowicz:
- prune expired orders in the app when maker or coordinator is offline
- add a test that ensures we detect lack of orders

TODO:
- [x] Disable open/close buttons when there's no 'best price' for a given direction
- [x] disable 'close position' button if there's no price in that direction

image

### DIFF
--- a/mobile/lib/features/trade/position_list_item.dart
+++ b/mobile/lib/features/trade/position_list_item.dart
@@ -3,9 +3,12 @@ import 'dart:async';
 import 'package:expandable/expandable.dart';
 import 'package:flutter/material.dart';
 import 'package:get_10101/common/value_data_row.dart';
+import 'package:get_10101/features/trade/domain/direction.dart';
 import 'package:get_10101/features/trade/domain/position.dart';
+import 'package:get_10101/features/trade/position_change_notifier.dart';
 import 'package:get_10101/features/trade/trade_theme.dart';
 import 'package:intl/intl.dart';
+import 'package:provider/provider.dart';
 
 import 'contract_symbol_icon.dart';
 
@@ -45,9 +48,16 @@ class _PositionListItemState extends State<PositionListItem> {
     formatter.maximumFractionDigits = 2;
 
     TradeTheme tradeTheme = Theme.of(context).extension<TradeTheme>()!;
+    PositionChangeNotifier positionChangeNotifier = context.watch<PositionChangeNotifier>();
 
     // dart cannot promote...
     Position notNullPosition = widget.position!;
+
+    // We're a bit conservative, we only enable action when we have both bid and ask
+    // XXX: Not sure whether ask or bid price here
+    bool priceAvailable = notNullPosition.direction == Direction.long
+        ? positionChangeNotifier.price?.ask != null
+        : positionChangeNotifier.price?.bid != null;
 
     if (!isPositionExpired) {
       Timer(notNullPosition.expiry.difference(DateTime.now().toUtc()), () {
@@ -157,12 +167,13 @@ class _PositionListItemState extends State<PositionListItem> {
             crossAxisAlignment: CrossAxisAlignment.center,
             children: [
               ElevatedButton(
-                onPressed:
-                    notNullPosition.positionState == PositionState.closing || isPositionExpired
-                        ? null
-                        : () async {
-                            await widget.onClose();
-                          },
+                onPressed: notNullPosition.positionState == PositionState.closing ||
+                        isPositionExpired ||
+                        !priceAvailable
+                    ? null
+                    : () async {
+                        await widget.onClose();
+                      },
                 child: notNullPosition.positionState == PositionState.closing || isPositionExpired
                     ? const Row(
                         children: [
--- a/mobile/native/src/orderbook.rs
+++ b/mobile/native/src/orderbook.rs
@@ -8,12 +8,16 @@ use orderbook_commons::best_current_price;
 use orderbook_commons::Order;
 use orderbook_commons::OrderbookMsg;
 use orderbook_commons::Signature;
+use parking_lot::Mutex;
+use std::sync::Arc;
 use std::time::Duration;
+use time::OffsetDateTime;
 use tokio::runtime::Runtime;
 use tokio::task::spawn_blocking;
 use uuid::Uuid;
 
 const WS_RECONNECT_TIMEOUT: Duration = Duration::from_secs(2);
+const EXPIRED_ORDER_PRUNING_INTERVAL: Duration = Duration::from_secs(30);
 
 pub fn subscribe(secret_key: SecretKey, runtime: &Runtime) -> Result<()> {
     runtime.spawn(async move {
@@ -28,14 +32,45 @@ pub fn subscribe(secret_key: SecretKey, runtime: &Runtime) -> Result<()> {
             Signature { pubkey, signature }
         };
 
-        // Consider using a HashMap instead to optimize the lookup for removal/update
-        let mut orders = Vec::new();
+        // Need a Mutex as it's being accessed from websocket stream and pruning task
+        let orders = Arc::new(Mutex::new(Vec::<Order>::new()));
+
+        let _prune_expired_orders_task = {
+            let orders = orders.clone();
+            tokio::spawn(async move {
+                loop {
+                    {
+                        tracing::debug!("Pruning expired orders");
+                        let mut orders = orders.lock();
+                        let orders_before_pruning = orders.len();
+                        *orders = orders.iter().filter(|order| order.expiry >= OffsetDateTime::now_utc()).cloned().collect::<Vec<_>>();
+                        let orders_after_pruning = orders.len();
+
+                        if orders_after_pruning < orders_before_pruning {
+                            let amount_pruned = orders_before_pruning - orders_after_pruning;
+                            tracing::debug!(
+                                orders_before_pruning,
+                                orders_after_pruning,
+                                "Pruned {amount_pruned} expired orders"
+                            );
+
+                            // Current best price might have changed
+                            if let Err(e) = position::handler::price_update(best_current_price(&orders)) {
+                                tracing::error!("Price update from the orderbook failed. Error: {e:#}");
+                            }
+                        }
+                    }
+                    tokio::time::sleep(EXPIRED_ORDER_PRUNING_INTERVAL).await;
+                }
+            })
+        };
 
         loop {
             let url = url.clone();
             let authenticate = authenticate;
             let mut stream =
                 spawn_blocking(move || orderbook_client::subscribe_with_authentication(url, authenticate)).await.expect("joined task not to panic");
+
 
             loop {
                 match stream.try_next().await {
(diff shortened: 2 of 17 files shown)


klochowicz:
converted into draft because CI seems to have trouble with detecting whether maker is online in tests. 
debugging now.

klochowicz:
Two remarks:
- killing a process and spawning it again from the integration test is not great, but does the job - potentially we could have an endpoint on maker and coordinator to "disable" them for test purposes without killing the processes. This would simplify the current workaround.
- @da-kami suggested that it might be good to add an event to the orderbook stream that informs subscribers about the pruning. I was leaning towards this as well, so let me know whether it would be useful to add it now or not.

klochowicz:
@da-kami can you please help me with flutter test, e.g. point me in the right direction? it's quite obscure...

crates/tests-e2e/tests/zzz_stop_orders.rs
```diff
@@ -0,0 +1,49 @@
...
+#[ignore = "need to be run with 'just e2e' command"]
+async fn no_orders_when_counterparty_is_offline() {
+    let setup = TestSetup::new_after_funding().await;
+    wait_until!(has_any_price(&setup.app.rx.prices()));
+
+    tracing::info!("Killing maker process, so that it does not send any orders");
+    kill_process("maker").expect("to be able to kill maker process");
+    assert!(!is_maker_running(), "maker should be stopped by now");
+
+    let wait_time = std::time::Duration::from_secs(70);
```
da-kami, line 34: ❓ Would it be possible to make the expiry time configurable so we can set it to e.g. `5` seconds instead so this test does not have to wait for `70` seconds? 😅

mobile/native/src/orderbook.rs
```diff
@@ -28,15 +32,46 @@ pub fn subscribe(secret_key: SecretKey, runtime: &Runtime) -> Result<()> {
...
+                            let amount_pruned = orders_before_pruning - orders_after_pruning;
+                            tracing::debug!(
+                                orders_before_pruning,
+                                orders_after_pruning,
+                                "Pruned {amount_pruned} expired orders"
+                            );
+
+                            // Current best price might have changed
+                            if let Err(e) = position::handler::price_update(best_current_price(&orders)) {
+                                tracing::error!("Price update from the orderbook failed. Error: {e:#}");
```
da-kami, line 59:
```suggestion
                                tracing::error!("Price update from the orderbook failed after pruning. Error: {e:#}");
```

mobile/test/trade_test.dart
```diff
@@ -128,13 +129,18 @@ void main() {
 
     WalletChangeNotifier walletChangeNotifier = WalletChangeNotifier(walletService);
 
+    PositionChangeNotifier positionChangeNotifier = PositionChangeNotifier(positionService);
+
+    // We have to have current price, otherwise we can't take order
+    positionChangeNotifier.price = Price(bid: 30000.0, ask: 30000.0);
```
da-kami, line 135: nit: I think could potentially solve this cleaner by adding an explicit `Extension` to the `PositionChangeNotifier` in the tests that adds a setter so we can set the price; then we don't have to expose the private `_price`.

mobile/lib/features/trade/position_list_item.dart
```diff
@@ -45,10 +48,18 @@ class _PositionListItemState extends State<PositionListItem> {
...
     TradeTheme tradeTheme = Theme.of(context).extension<TradeTheme>()!;
+    PositionChangeNotifier positionChangeNotifier = context.watch<PositionChangeNotifier>();
+
+    // We're a bit conservative, we only enable action when we have both bid and ask
 
     // dart cannot promote...
     Position notNullPosition = widget.position!;
 
+    // XXX: Not sure whether ask or bid price here
+    bool hasPriceInOrderBook = notNullPosition.direction == Direction.long
```
da-kami, line 59:
``` suggestion
   bool priceAvailable = notNullPosition.direction == Direction.long
```

I find the name `hasPriceInOrderBook` confusing - this has nothing to do with the orderbook, but with the app knowing about the current price.

mobile/lib/features/trade/position_list_item.dart
```diff
@@ -45,10 +48,18 @@ class _PositionListItemState extends State<PositionListItem> {
     formatter.maximumFractionDigits = 2;
 
     TradeTheme tradeTheme = Theme.of(context).extension<TradeTheme>()!;
+    PositionChangeNotifier positionChangeNotifier = context.watch<PositionChangeNotifier>();
+
+    // We're a bit conservative, we only enable action when we have both bid and ask
```
da-kami, line 53: I think this comment is at the wrong line, given that we set `hasPriceInOrderBook` at line `59`

da-kami (approved):
Happy to see this merged, but I think *just* disabling the buttons might be quite confusing. 
The user does not know why the buttons are disabled - we should add some form of indicator that shows the connection status to orderbook / coordinator that makes it obvious that the buttons are disabled because there is not connection.

Alternatively we could also display a toast for now so the user know why when the user enters the screen.

mobile/test/trade_test.dart
klochowicz: I don't think this adds anything, a direct getter and setter `==` a public member....

crates/tests-e2e/tests/zzz_stop_orders.rs
klochowicz: it would be nice indeed, I will create a ticket for this!

klochowicz:
> Happy to see this merged, but I think _just_ disabling the buttons might be quite confusing. The user does not know why the buttons are disabled - we should add some form of indicator that shows the connection status to orderbook / coordinator that makes it obvious that the buttons are disabled because there is not connection.
> 
> Alternatively we could also display a toast for now so the user know why when the user enters the screen.

I agree with you that this would be ideal, there's a ticket scheduled to do this for the upcoming iteration #915. 

I typically try to split the plumbing / backend support for a feature from the ideal UX, to avoid painful rebasing and to be able to concentrate on flutter in the follow-ups. YMMV though.

crates/tests-e2e/tests/zzz_stop_orders.rs
```diff
@@ -0,0 +1,49 @@
...
+            .get(&ContractSymbol::BtcUsd)
+            .expect("BTCUSD price to be present");
+        return btcusd.bid.is_some() || btcusd.ask.is_some();
+    }
+    false
+}
+
+#[tokio::test]
+#[ignore = "need to be run with 'just e2e' command"]
+async fn no_orders_when_counterparty_is_offline() {
```
bonomat, line 26: The test name says that you should not expect an order if the counterparty is offline. This makes sense of course, but I guess what you are testing here is that outdated orders (older than 1m) are not beeing forwarded anymore? 

I don't think there is any use in testing if a terminated binary is not sending orders anymore. 

If you want to test expired orders, you can do a simpler test: 

1. Only start the coordinator and the app
2. Manually post a `NewOrder` to the coordinator using a reqwest client. Here you can set the expiry [to in Xsec](https://github.com/get10101/10101/blob/1eb9efbf3987175a0733fc78f9f21204a9149ca0/crates/orderbook-commons/src/lib.rs#L72).
3. Assert that no new order arrives past this few seconds. 

But even that test might not be that useful...

crates/tests-e2e/tests/zzz_stop_orders.rs
bonomat: I'm against having a test which runs for 70+ seconds :/

crates/tests-e2e/tests/zzz_stop_orders.rs
klochowicz: I wanted to test that we detect lack of orders when the counterparty goes offline. As it turns out, we didn't. So I wrote the code that prunes outdated orders in the app so the feature would be working. 

I like your idea with 'faking' maker; we could have created one with different expiry.
I don't think that the test would be *simpler*, because we would need to kill the maker anyway, and start using a "fake" one, but it certainly would be faster!

I don't mind stripping out the test if you see no value in it.

bonomat: > I don't mind stripping out the test if you see no value in it.

Please treat me as an equal engineer, not as a your superior :)

Thinking about this more, can we write unit tests for this at the places where it matters? e.g. 
- [here](https://github.com/get10101/10101/blob/a1065b46aa6a0107bcdad9c82a6719072ced7ddd/mobile/native/src/orderbook.rs#L63-L80) we write a test that no expired order is being forwarded
- in the frontend we check if the order is still up2date when reading it.

holzeis:
🤔 i am bit late to the party, but is it so bad that we let an order fail if there is no match?

klochowicz:
> 🤔 i am bit late to the party, but is it so bad that we let an order fail if there is no match?

I'm not sure what are you referring to. IMHO you should disable the buttons where there is *no* orders in the orderbook and you cannot do anything.
of course one could hide it in the hope that no-one notices, but that's not good UX.

crates/tests-e2e/tests/zzz_stop_orders.rs
klochowicz: Perhaps let's rephrase it:
> I don't mind stripping out the test if you see no value in it.

What I meant was that I'm OK with stripping out the test if it causes contention for *any* of the reviewers, as it's not necessary for addressing the issue. 

I wrote the test because things were *supposed to* work, and didn't (normal part of engineering process). IMHO it would be great if we keep the test as it's proven to be useful.

I believe integration tests like this are needed; we need to know how we behave when the other party is not available and have that tested.

The criticism for waiting 70s is very much valid, and there's an open ticket that would address it - the maker doesn't have HTTP API to control orders for now, so it was not easy to address in this PR.

🤔Could I have started with an unit test instead? I don't think so - unit tests were there, it's just this case was *missed* during the implementation phase.

If you would like to add additional unit tests, or schedule them for the iteration, I think it's a good idea (the more unit tests the better!).

holzeis:
> > 🤔 i am bit late to the party, but is it so bad that we let an order fail if there is no match?
> 
> 
> 
> I'm not sure what are you referring to. IMHO you should disable the buttons where there is *no* orders in the orderbook and you cannot do anything.
> 
> of course one could hide it in the hope that no-one notices, but that's not good UX.

Well, I guess the issue is that we are using the order price in the first place. That made it look odd as there is no price.

But we could also simply use the actual price (e.g. from bitmex) as reference, hit order and if we can't find a match, we fail the order.

I tend to prefer not disabling the buttons, but no need to action it. We might want to revisit that another time. 🙂

klochowicz:
> > > 🤔 i am bit late to the party, but is it so bad that we let an order fail if there is no match?
> > 
> > 
> > I'm not sure what are you referring to. IMHO you should disable the buttons where there is _no_ orders in the orderbook and you cannot do anything.
> > of course one could hide it in the hope that no-one notices, but that's not good UX.
> 
> Well, I guess the issue is that we are using the order price in the first place. That made it look odd as there is no price.

I don't see it as an issue. If there is no order that can be taken, why would not show this fact? 

> 
> But we could also simply use the actual price (e.g. from bitmex) as reference, hit order and if we can't find a match, we fail the order.
> 
> I tend to prefer not disabling the buttons, but no need to action it. We might want to revisit that another time. 🙂

again, it's poor UX to not disable the buttons when we know for a fact that they cannot succeed. 
I agree with @da-kami that **just** disabling the buttons is not enough, and we should be more clear *why* they are disabled, but this is just a start.

crates/tests-e2e/tests/zzz_stop_orders.rs
klochowicz: @bonomat I have extracted the test into a separate PR, so it's not blocking the functionality:  #937

holzeis:
Again, I do not agree on the bad UX argument. But as mentioned we don't have to action it now.

crates/tests-e2e/tests/zzz_stop_orders.rs
klochowicz: Moved the test into a separate PR #937 

I also realised that I don't need to wait for that long, because maker is likely to have already sent an order.

klochowicz:
> Again, I do not agree on the poor UX argument. But as mentioned we don't have to action it now.

Please state your reasons, as I would love to understand your POV. 

There's very little more frustrating experiences on the web that needing to submit a form to check whether we can submit a form.
if I understood your proposal correctly -  what you are proposing, is needing to submit an order to see whether we can submit an order.

klochowicz:
I'm merging this, as the unsolved issues are captured in the follow-up PR #937 or the follow-up ticket scheduled for this iteration: #915

holzeis:
> > Again, I do not agree on the poor UX argument. But as mentioned we don't have to action it now.
> 
> 
> 
> Please state your reasons, as I would love to understand your POV. 
> 
> 
> 
> There's very little more frustrating experiences on the web that needing to submit a form to check whether we can submit a form.
> 
> if I understood your proposal correctly -  what you are proposing, is needing to submit an order to see whether we can submit an order. 
> 
> 

- Disabling the button presumes that we either have an immediate match or fail.
- Enabled buttons will not guarantee that the order will not fail.
- The order matching is the responsibility of the coordinator. Implying a price from an order is probably wrong in the first place as it is not a match.

Bottom line, by disabling the buttons we already implying that there is no match. This might be the case now as we have a very simplified matching logic, but this will change.

In other words, I think it's perfectly fine to fail an order if there is no match. The App should not implement business logic which is in the responsibility of the coordinator.

holzeis:
Additionally, we will not learn about an trading attempt, indicating more demand.

PR #925, opened 2023-07-07
Default to zero base fee for interceptable invoices

da-kami:
fixes https://github.com/get10101/meta/issues/178

If I understood the conclusion of https://github.com/get10101/meta/issues/178 correctly then this is what we want for now: A zero base fee configuration when creating interceptable invoices.
My understanding is that the coordinator currently uses the `Default` config, so I adapted it there. I did not find a configuration file where we set explicit values - if I overlooked something and we better change it there let me know.

I consider this a follow up of https://github.com/get10101/10101/pull/921 so I gave it a quick crack.
@bonomat if I misunderstood something let me know :)

### DIFF
--- a/coordinator/src/settings.rs
+++ b/coordinator/src/settings.rs
@@ -38,7 +38,8 @@ impl Default for Settings {
         Self {
             jit_channels_enabled: true,
             new_positions_enabled: true,
-            jit_fee_rate_basis_points: 50,
+            // Configured as zero base fee by default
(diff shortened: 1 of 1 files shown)


bonomat (changes requested):
I think you missunderstood something. This is the Fee we charge for opening routing the payment over the newly created JIT channel. It should not be 0. The base fee should be 0.

PR #929, opened 2023-07-10
Add logs to flaky `force_close_ln_dlc_channel` test

luckysori:
This test can hang on CI. From the logs we do see a message corresponding to line 93 (where we mine one block), so it should be sufficient to log everything after that in order to identify where we are hanging.

---

Related to #807.
We don't need to merge this. I just want to keep re-triggering the relevant job to reproduce the hang with extra context.


klochowicz (approved):
there's no harm in merging these - these extra logs will be useful in the event of any crash.

luckysori:
Alright, let's merge this and test it passively.

PR #937, opened 2023-07-12
chore(tests-e2e): Add a test to verify that backends detects lack of orders

klochowicz:


klochowicz:
Follow-up from https://github.com/get10101/10101/pull/923 - see the unresolved issues.

PR #938, opened 2023-07-12
Persist spendable outputs generated by LDK

luckysori:
Fixes https://github.com/get10101/10101/issues/845.

PR #940, opened 2023-07-12
Feat/coordinator storage payments

### LINKED ISSUE #936: Persist payments on coordinator
We are still using `InMemoryStorage` on the coordinator and should save the payments in the database otherwise we might not be able to track metrics properly.

This ticket includes implementing the complete `node::Storage` interface for the coordinator.

da-kami:
Fixes #936. 
Fixes https://github.com/get10101/10101/issues/845 with https://github.com/get10101/10101/pull/940/commits/3df219e5569d2912ca71969fec28c074ab162515.

Store payments by implementing `node::Storage`.

I chucked a few smaller commits in front because I noticed these things.

@luckysori is working on the `todo!`s for spendable outputs - @luckysori can we merge my work without this part or is the coordinator broken then.


da-kami:
I propose to rebase this on @luckysori 's work in #938 so we can merge the combined implementation of the trait

coordinator/src/node/connection.rs
```diff
@@ -28,7 +28,10 @@ pub async fn keep_public_channel_peers_connected(
     }
 }
 
-fn reconnect_to_disconnected_public_channel_peers(node: Arc<Node<InMemoryStore>>) {
+fn reconnect_to_disconnected_public_channel_peers<S>(node: Arc<Node<S>>)
+where
+    S: Storage + Sync + Send + 'static,
```
luckysori, line 33: 🔧 I'm going for a concrete type here. I don't think we need to abstract over this type (in this crate at least).

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -184,6 +184,9 @@ where
                     PaymentPurpose::SpontaneousPayment(preimage) => (Some(preimage), None),
                 };
 
+                // TODO: Check if there is a corrsponding ff payment in the databse and update it to
+                // paid if so TODO: Increase fee counter to track fee revenue
```
luckysori, line 188: 🤔 You won't be able to do this directly in this unfortunately generic event handler.

You can do something like what @holzeis did in the app to subscribe to these events outside of the event handler.

luckysori (approved):
LGTM!

coordinator/src/node/connection.rs
da-kami: Makes sense! I only did this in case we want to re-use these functions in e.g. `test`, but given that this is in the executable that won't be necessary. Please use the concreate type when you rebase :)

crates/ln-dlc-node/src/ln/event_handler.rs
da-kami: Sorry, this TODO should not be part of the changeset but will be done on top! This should have not sneaked in :)

luckysori: Oh yeah, that's okay. I'm just letting you know in case it wasn't already clear.

luckysori (approved):
LGTM!

PR #942, opened 2023-07-12
fix: read scorer

### LINKED ISSUE #797: Panic on reading scorer from disk
My app started to crash after a restart because the scorer file was broken.

xcode://organizer/crashes/downloadPoint?adamId=1671686549&platformId=iOS&installPlatformId=iOS&analyticsPointId=DUp8-dcHhayUiR7pI4Hsm&bundleId=finance.get10101.app

Luckily this can be easily fixed by simply deleting the scorer file.

bonomat:
With this patch we don't read a historically stored scrorer from disk. It's not ideal, but at least users running into panics when loading the scorer will be able to run the app again

Workaround for #797 
fix #797

### DIFF
--- a/crates/ln-dlc-node/src/disk.rs
+++ b/crates/ln-dlc-node/src/disk.rs
@@ -4,8 +4,10 @@ use bitcoin::Network;
 use lightning::routing::scoring::ProbabilisticScorer;
 use lightning::routing::scoring::ProbabilisticScoringParameters;
 use lightning::util::ser::ReadableArgs;
+use std::fs;
 use std::fs::File;
 use std::io::BufReader;
+use std::panic;
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


crates/ln-dlc-node/src/disk.rs
```diff
@@ -15,13 +15,13 @@ pub(crate) fn read_scorer(
...
 ) -> ProbabilisticScorer<Arc<NetworkGraph>, Arc<TracingLogger>> {
     let params = ProbabilisticScoringParameters::default();
-    if let Ok(file) = File::open(path) {
-        let args = (params.clone(), graph.clone(), logger.clone());
-        match ProbabilisticScorer::read(&mut BufReader::new(file), args) {
-            Ok(scorer) => return scorer,
-            Err(e) => tracing::error!("Failed to read scorer from disk: {e}"),
-        }
-    }
+    // if let Ok(file) = File::open(path) {
```
holzeis, line 18: nit: I guess we can remove that commented code and if we need it again we can look at this commit.

holzeis (approved):
@bonomat great, if that fixes the issue 🙂

As we have checked my scorer file was anyways empty, so I guess it doesn't hurt not reading the file from disk?

Are there any big disadvantages of recreating the scorer from scratch every time?

crates/ln-dlc-node/src/disk.rs
bonomat: I think the second commit is more serious. Unfortunately I can't reproduce the panic anymore :/

bonomat:
> Are there any big disadvantages of recreating the scorer from scratch every time?

I guess we could run without reading from disc. This means, we will not learn from past payments and only keep an in-memory picture of the network.

bonomat:
@reviewers 
1. The first commit fixed the problem for me
2. The second commit might be the better fix but I can't test it because it does not panic anymore. 

If we go for the second, let's squash the commits. 
I propose to roll out a new version with this asap, because you really can't use the app if this happens to you.

klochowicz (approved):
I would go for solution 2 - it does not panic anymore, and we're logging potential problems, without sweeping them completely under the rug.

holzeis:
> I propose to roll out a new version with this asap, because you really can't use the app if this happens to you.

@bonomat fyi: https://github.com/get10101/10101/actions/runs/5540939930/jobs/10113681545

Issue #909, opened 2023-07-05
How to charge closing fees without an order match


luckysori:
It is an edge-case, but a trader could simply force-close the channel to avoid paying order-matching fees.

Even if there is no force-close, we do currently enforce frequent collaborative closures that I think circumvent the orderbook. This motivated me to leave the following TODO in https://github.com/get10101/10101/pull/863:

https://github.com/get10101/10101/blob/dadbc4bade0dc3777dafc1e8b9efff5ba7266fb0/mobile/native/src/ln_dlc/node.rs#L248

holzeis:
> It is an edge-case, but a trader could simply force-close the channel to avoid paying order-matching fees.

Can you elaborate on this scenario? How would the user avoid paying the order-matching fees? Also, it looks like this part has nothing to do with the second part.

luckysori:
> Can you elaborate on this scenario? How would the user avoid paying the order-matching fees?

An app user could:

1. Open a position, paying the corresponding order-matching fee invoice.
2. Force-close the LN-DLC channel (currently not supported, but we have to support it).
3. Not have to pay an order-matching fee for closing the position, since force-closing doesn't involve creating an order and getting a match.

With a small position this might not be profitable because the transaction fees might be more expensive than the order-matching fee, but this route might become more attractive with large positions.

> Also, it looks like this part has nothing to do with the second part.

The point of the second part is to highlight that there is another way in which the app user can end up not paying the order-matching fee: by letting it be collaboratively closed on expiry. Such a scenario is related to the first one because in both cases there is no order-matching happening, which is strange in the context of an orderbook. I left it out of https://github.com/get10101/10101/pull/863, because:

- I wasn't sure if we actually wanted to charge the order-matching fee in that case.
- If we do want it, the implementation has to be different because we do not send a trade request from the app to the coordinator (which in the regular flow returns the order-matching fee invoice).

holzeis:
Thanks for the explanation, now the second part also makes sense to me why it is added to this ticket. 👍 

Well in that case, how about only charging a fee for opening a position, but none when closing a position? We could as well charge twice for opening already covering the fees for closing the position upfront - as there is currently no way to not close the position after the expiry. 

We might want to revise that when supporting perpetuals.

bonomat:
Let's try to remove the term: `close position` from our vocabulary and replace it with: 

[...]

PR #946, opened 2023-07-13
Feat/matching fee tracking

da-kami:
Track matching fees in database and report on prometheus as counter.

### DIFF
--- a/coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
+++ b/coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
@@ -1,0 +1,26 @@
+-- Your SQL goes here
+-- insert a payment with amount zero that is a dummy default payment that corresponds to this actual zero-amount invoice:
+-- lnbcrt0m1pj26wczdq50fjhymeqd9h8vmmfvdjsnp4qtwk40kf07d8fzlhdt2s9vqyeczarvk37safua4a0kz7wellkq3vjpp5d7dce9wzhfa3s4a3n7t4xu3ss9slah6sl6mc5ffzqqf45sv82ggqsp54cmyvxklz48ap60guqsln4wuh2ranlap3grg2djt5ykz7005fc6s9qyysgqcqpcxq8z7ps9nqepcz2l2s4qrp6qeks68qry55ylcz542c8a6m6fffmhlmhtn9nnf458ngv83pk0473xfecdk7m7maumqk4jvaymdg7fpgn9tujcehxpqqte97xd
+-- We insert this dummy payment so we are able to sum up fees on already existing trades and make the fee payment hash a mandatory field.
+INSERT INTO
+    payments (
+        payment_hash,
+        htlc_status,
+        amount_msat,
+        flow,
+        payment_timestamp,
+        description
+    )
+VALUES
+    (
+        '6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210',
+        'Succeeded',
+        0,
+        'Inbound',
+        CURRENT_TIMESTAMP,
+        'zero amount payment dummy default value'
+    );
+ALTER TABLE
+    trades
+ADD
+    COLUMN "fee_payment_hash" TEXT NOT NULL DEFAULT '6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210' REFERENCES payments(payment_hash);
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -3,7 +3,7 @@ use anyhow::Result;
 use coordinator::cli::Opts;
 use coordinator::db;
 use coordinator::logger;
-use coordinator::metrics::collect_metrics;
+use coordinator::metrics;
 use coordinator::metrics::init_meter;
 use coordinator::node::connection;
 use coordinator::node::storage::NodeStorage;
@@ -17,6 +17,8 @@ use coordinator::settings::Settings;
 use diesel::r2d2;
 use diesel::r2d2::ConnectionManager;
 use diesel::PgConnection;
+use hex::FromHex;
+use lightning::ln::PaymentHash;
 use ln_dlc_node::seed::Bip39Seed;
 use rand::thread_rng;
 use rand::RngCore;
@@ -135,7 +137,7 @@ async fn main() -> Result<()> {
         async move {
             loop {
                 let node = node.clone();
-                spawn_blocking(move || collect_metrics(node))
+                spawn_blocking(move || metrics::collect(node))
                     .await
                     .expect("To spawn blocking thread");
                 tokio::time::sleep(PROCESS_PROMETHEUS_METRICS).await;
@@ -218,8 +220,23 @@ async fn main() -> Result<()> {
                             }
                         };
 
+                    // Upon collab closing an expired position we cannot charge a fee using an
+                    // invoice. This dummy hash exists in the database to
+                    // represent zero-amount invoices.
+                    let zero_amount_payment_hash_dummy = PaymentHash(
+                        <[u8; 32]>::from_hex(
+                            "6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210",
+                        )
+                        .expect("static payment hash to decode"),
+                    );
+
                     match node
-                        .close_position(position, closing_price, channel_id)
+                        .close_position(
+                            position,
+                            closing_price,
+                            channel_id,
+                            zero_amount_payment_hash_dummy,
+                        )
                         .await
                     {
                         Ok(_) => tracing::info!(
--- a/coordinator/src/db/trades.rs
+++ b/coordinator/src/db/trades.rs
@@ -3,26 +3,80 @@ use crate::orderbook::db::custom_types::Direction;
 use crate::schema::trades;
 use anyhow::Result;
 use autometrics::autometrics;
+use bitcoin::secp256k1::PublicKey;
 use diesel::prelude::*;
+use std::str::FromStr;
+use time::OffsetDateTime;
+
+#[derive(Queryable, Debug, Clone)]
+#[diesel(table_name = trades)]
+struct Trade {
+    id: i32,
+    position_id: i32,
+    contract_symbol: ContractSymbol,
+    trader_pubkey: String,
+    quantity: f32,
+    leverage: f32,
+    collateral: i64,
+    direction: Direction,
+    average_price: f32,
+    timestamp: OffsetDateTime,
+}
 
 #[derive(Insertable, Debug, Clone)]
 #[diesel(table_name = trades)]
-pub struct Trade {
-    pub position_id: i32,
-    pub contract_symbol: ContractSymbol,
-    pub trader_pubkey: String,
-    pub quantity: f32,
-    pub leverage: f32,
-    pub collateral: i64,
-    pub direction: Direction,
-    pub average_price: f32,
+struct NewTrade {
+    position_id: i32,
+    contract_symbol: ContractSymbol,
+    trader_pubkey: String,
+    quantity: f32,
+    leverage: f32,
+    collateral: i64,
+    direction: Direction,
+    average_price: f32,
 }
 
 #[autometrics]
-pub fn insert(conn: &mut PgConnection, trade: Trade) -> Result<()> {
-    diesel::insert_into(trades::table)
-        .values(trade)
-        .execute(conn)?;
+pub fn insert(
+    conn: &mut PgConnection,
+    trade: crate::trade::models::NewTrade,
+) -> Result<crate::trade::models::Trade> {
+    let trade: Trade = diesel::insert_into(trades::table)
+        .values(NewTrade::from(trade))
+        .get_result(conn)?;
 
-    Ok(())
+    Ok(trade.into())
+}
+
+impl From<crate::trade::models::NewTrade> for NewTrade {
+    fn from(value: crate::trade::models::NewTrade) -> Self {
+        NewTrade {
+            position_id: value.position_id,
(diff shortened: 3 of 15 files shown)


holzeis:
fixes: https://github.com/get10101/meta/issues/169

coordinator/src/bin/coordinator.rs
```diff
@@ -256,3 +280,45 @@ async fn main() -> Result<()> {
 
     Ok(())
 }
+
+fn node_event_post_processing(node: Node, mut node_event_receiver: watch::Receiver<Option<Event>>) {
```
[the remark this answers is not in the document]
da-kami: Initially I saw it more as a general post processing where we could do other things than metrics. 
We could clone the receiver and have multiple event loops, but I'm not sure it's good to have really a lot of them. 

However, given that this is only used for metrics at the moment I am happy to move it to the metrics :)

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
```diff
@@ -0,0 +1,26 @@
...
+        'Succeeded',
+        0,
+        'Inbound',
+        CURRENT_TIMESTAMP,
+        'zero amount payment dummy default value'
+    );
+ALTER TABLE
+    trades
+ADD
+    COLUMN "fee_payment_hash" TEXT NOT NULL DEFAULT '6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210' REFERENCES payments(payment_hash);
```
da-kami, line 26: nit: I think we could even give this a `UNIQUE` constraint because each trade fee payment hash should be unique for each trade. What do you think @holzeis ?
That way we have more guarantees that only one record will be returned.

coordinator/src/bin/coordinator.rs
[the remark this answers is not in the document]
da-kami: moved it into `metrics` and renamed it to `collect_metrics_based_on_node_events`

da-kami:
I'm not merging this yet, I think it would be good if @bonomat and @luckysori can have a look :)

coordinator/src/metrics.rs
```diff
@@ -243,3 +257,48 @@ fn node_metrics(cx: &Context, inner_node: Arc<ln_dlc_node::node::Node<NodeStorag
...
+                                // irrelevant, not a fee
+                                continue;
+                            }
+                            Err(e) => {
+                                tracing::error!("Failed to load trades from database in node event post processing: {e:#}");
+                                continue;
+                            }
+                        }
+                    }
+                }
```
luckysori, line 297: 🤔 The `Ok` arm is all blocking code. Particularly `pool.get` waits until we can get a connection from the pool, which _could_ take a while. I would `spawn_blocking` the whole arm.

You won't be able to keep the `continue`s, but you don't need them anyway because the loop ends after the `match` statement.

coordinator/src/metrics.rs
```diff
@@ -243,3 +257,48 @@ fn node_metrics(cx: &Context, inner_node: Arc<ln_dlc_node::node::Node<NodeStorag
         }
     }
 }
+
+pub fn collect_metrics_based_on_node_events(
```
luckysori, line 261: ❓We had a conversation about this yesterday, but I would still like to hear from others on why we think collecting metrics like this is necessary if they are already stored in the database. Cc @bonomat.

I don't see much benefit and this code needs to be maintained.

coordinator/src/db/trades.rs
```diff
@@ -43,6 +50,22 @@ pub fn insert(conn: &mut PgConnection, trade: crate::trade::models::NewTrade) ->
     Ok(trade)
 }
 
+/// Returns the position by trader pub key
+#[autometrics]
+pub fn get_by_fee_payment_hash(
+    conn: &mut PgConnection,
+    payment_hash: PaymentHash,
+) -> QueryResult<Option<crate::trade::models::Trade>> {
```
luckysori, line 58: 🔧 I think we should just return a `bool` and call this `is_associated_with_fee_payment` or something like that. We don't seem to be using the returned data.

coordinator/src/bin/coordinator.rs
```diff
@@ -218,8 +227,23 @@ async fn main() -> Result<()> {
                             }
                         };
 
+                    // Upon collab closing an expired position we cannot charge a fee using an
+                    // invoice. This dummy hash exists in the database to
+                    // represent zero-amount invoices.
```
luckysori, line 232: ❓Is this what we want to do, @bonomat? I'm not totally sure after reading [this comment](https://github.com/get10101/10101/issues/909#issuecomment-1623212216).

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
```diff
@@ -0,0 +1,26 @@
+-- Your SQL goes here
+-- insert a payment with amount zero that is a dummy default payment that corresponds to this actual zero-amount invoice:
+-- lnbcrt0m1pj26wczdq50fjhymeqd9h8vmmfvdjsnp4qtwk40kf07d8fzlhdt2s9vqyeczarvk37safua4a0kz7wellkq3vjpp5d7dce9wzhfa3s4a3n7t4xu3ss9slah6sl6mc5ffzqqf45sv82ggqsp54cmyvxklz48ap60guqsln4wuh2ranlap3grg2djt5ykz7005fc6s9qyysgqcqpcxq8z7ps9nqepcz2l2s4qrp6qeks68qry55ylcz542c8a6m6fffmhlmhtn9nnf458ngv83pk0473xfecdk7m7maumqk4jvaymdg7fpgn9tujcehxpqqte97xd
+-- We insert this dummy payment so we are able to sum up fees on already existing trades and make the fee payment hash a mandatory field.
```
luckysori, line 4: 🙃 It's also for closing expired positions, at least at the moment.

coordinator/src/bin/coordinator.rs
```diff
@@ -218,8 +227,23 @@ async fn main() -> Result<()> {
...
+                    let zero_amount_payment_hash_dummy = PaymentHash(
+                        <[u8; 32]>::from_hex(
+                            "6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210",
```
luckysori, line 235: 🔧 I think it would help if this is private to the `db` module. We could have a `close_expired_position` method without the `fee_payment_hash`. And maybe a `NewTrade::close_expired` encapsulating this stuff.

luckysori (approved):
I think this is well done!

I suggested a couple of things that are not so important and I left a few questions that other teammates might want to address.

coordinator/src/db/trades.rs
da-kami: Fair enough - for me it's more a pattern that an `insert` statement should return the representation of the object that was inserted (i.e. we insert `NewTrade` and return `Trade`) - but I can also remove this for now because it's not used as you say.

coordinator/src/bin/coordinator.rs
da-kami: I actually wanted to *try* charging the user in this scenario as well, but this would require a more elaborate process in terms of messages. The coordinator currently triggers the close on the DLC subchannel level. We could change this so the coordinator proposes the close by sending an event over to the app. The app then triggers the close automatically - that way we would tap into the existing process and would charge the fee. I don't see this as part of this PR though. 

My understanding is, that eventually we want to calculate the fee into the payout (and not charge it with a separate invoice); if we do this at some point this problem will go away.

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
da-kami: Yeah, that was not planned initially - I can adapt the comment or even opt for having *another* dummy value in the db for that. 
I thought about changing the model make the fee optional (removing the `NOT NULL`) constraint instead, but I feel eventually there *should* be a fee for every trade. Or not...? CC @bonomat :)

coordinator/src/metrics.rs
da-kami: Context: @luckysori and me discussed if we *should* actually report this to prometheus at all. If we plan to use it for reporting we will created a report based on the database - what's the motivation to report it to prometheus?

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
da-kami: Ah, I know why I was unable to do that - stupid me. 
The dummy hash (and default...) is obviously always the same, so we can't do that if we want to stick using the dummy 😅

luckysori: Hahah you're right, of course.

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
bonomat: I'm not sure I fully understand the question: are you asking if there should be a fee for every trade? Imho the answer to this question is yes, if it is a trade, i.e. it remove liquidity from the orderbook, then there should be a fee involved.

coordinator/src/bin/coordinator.rs
bonomat: Yea, both ways are better then what we have at the moment 😅
It would be nicer if the app would try to collab close an expired position when it detects it cc @holzeis

coordinator/src/metrics.rs
bonomat: I see these two different types of _reports_: 
- business metrics to track KPIs: these are in the DB and are collected manually once every X (e.g. 1/week)
- liveness metrics: these are reported via prometheus to grafana so that we get a glimpse of what is happening.  

We can combine these: for that I would not send events around but rather have a recurring thread that polls from the db and collects the data and prepares it for prometheus/grafana. 
What do you think?

coordinator/src/db/trades.rs
```diff
@@ -19,10 +35,25 @@ pub struct Trade {
 }
 
 #[autometrics]
-pub fn insert(conn: &mut PgConnection, trade: Trade) -> Result<()> {
-    diesel::insert_into(trades::table)
-        .values(trade)
-        .execute(conn)?;
+pub fn insert(conn: &mut PgConnection, trade: crate::trade::models::NewTrade) -> Result<Trade> {
```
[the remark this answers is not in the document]
da-kami: I think it's good practice to return what was inserted. However, I made a mistake here: should actually return `crate::trade::models::Trade` not the internal `Trade` struct. 
Fixed in `!fixup`

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    )
+VALUES
+    (
+        '6f9b8c95c2ba7b1857b19f975372308161fedf50feb78a252200135a41875210',
+        'Succeeded',
+        0,
+        'Inbound',
+        CURRENT_TIMESTAMP,
+        'zero amount payment dummy default value'
+    );
```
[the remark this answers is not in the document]
da-kami: > Why did you go for CURRENT_TIMESTAMP though? Maybe it doesn't matter at all, but it might be safer to have a static value there to. I think currently it would be relative to when the migration is executed.

I have not thought about it much, but I feel any fixed timestamp would be as arbitrary as the migration timestamp. Do you have any reason why a fixed timestamp would be better?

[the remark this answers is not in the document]
luckysori: I suppose it doesn't matter! I was wondering if the timestamp would confuse us if it was changing after every migration.

coordinator/src/db/trades.rs
luckysori: Yeah, on _inserting_ I do think you make a very good point. But here I'm advocating for replacing this _get_ (whose return value we're only using to see if it's `Some`) with a more dedicated method.

coordinator/src/metrics.rs
da-kami: I completely agree with the DB based reports. That makes sense.

I am not totally convinced about reporting this particular metric to prometheus. Could it be better to query the database from the Grafana board instead of reporting this via prometheus?

If we want to keep reporting this metric to prometheus I think it's better reporting it based on the event rather than a periodic task. If we have a periodic task that loads and reports fees it can happen that we miss reporting one (when we e.g. shut down the coordinator) unless we track what was already reported. This might create scenarios where we get confused because e.g. what is reported on prometheus does not match the report we got from the database. I am not convinced it's worth creating logic to track what we already reported; that's why I opted for reporting to prometheus based on the event.

coordinator/src/bin/coordinator.rs
da-kami: If we agree that the trade fee *should* always be set (but it is not for this particular case at the moment because it's not implemented) then I would keep this in the business logic. The database expects the payment hash to be set and should not "default" it (initial migration excepted, but that's a different case).

coordinator/src/db/trades.rs
da-kami: sorry, this answer was on the wrong comment - I actioned this commend, fixup pushed now :)

coordinator/migrations/2023-07-13-010300_trade_fees/up.sql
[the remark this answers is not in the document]
da-kami: The migration will only be applied once, the respective entry will always have the same timestamp once the migration was applied unless we completely reset the database.

coordinator/src/metrics.rs
da-kami: As discussed offline just now: We won't report `fees` as metrics to Grafana (through Prometheus) because there is nothing to act on here.

PR #949, opened 2023-07-14
Save routing fees in database

da-kami:
fixes https://github.com/get10101/meta/issues/177

Notes:

- I have not added the prometheus counter yet because I am waiting for the result of discussion here: https://github.com/get10101/10101/pull/946#discussion_r1263175979
- Opted for a separate table in the databse instead of adding a table that abstracts over multiple "fees". I think it's better to created views / reports on top

### DIFF
--- a/coordinator/src/db/routing_fees.rs
+++ b/coordinator/src/db/routing_fees.rs
@@ -1,0 +1,69 @@
+use crate::routing_fee;
+use crate::schema::routing_fees;
+use anyhow::Result;
+use bitcoin::hashes::hex::ToHex;
+use diesel;
+use diesel::prelude::*;
+use dlc_manager::ChannelId;
+use hex::FromHex;
+use time::OffsetDateTime;
+
+#[derive(Insertable, Debug, PartialEq)]
+#[diesel(table_name = routing_fees)]
+struct NewRoutingFee {
+    amount_msats: i64,
+    prev_channel_id: Option<String>,
+    next_channel_id: Option<String>,
+}
+
+#[derive(Queryable, Debug, Clone)]
+#[diesel(table_name = routing_fees)]
+struct RoutingFee {
+    id: i32,
+    amount_msats: i64,
+    prev_channel_id: Option<String>,
+    next_channel_id: Option<String>,
+    created_at: OffsetDateTime,
+}
+
+pub fn insert(
+    routing_fee: routing_fee::models::NewRoutingFee,
+    conn: &mut PgConnection,
+) -> Result<routing_fee::models::RoutingFee> {
+    let routing_fee: NewRoutingFee = routing_fee.into();
+    let routing_fee: RoutingFee = diesel::insert_into(routing_fees::table)
+        .values(&routing_fee)
+        .get_result(conn)?;
+
+    Ok(routing_fee.into())
+}
+
+impl From<routing_fee::models::NewRoutingFee> for NewRoutingFee {
+    fn from(value: routing_fee::models::NewRoutingFee) -> Self {
+        Self {
+            amount_msats: value.amount_msats as i64,
+            prev_channel_id: value
+                .prev_channel_id
+                .map(|prev_channel_id| prev_channel_id.to_hex()),
+            next_channel_id: value
+                .next_channel_id
+                .map(|next_channel_id| next_channel_id.to_hex()),
+        }
+    }
+}
+
+impl From<RoutingFee> for routing_fee::models::RoutingFee {
+    fn from(value: RoutingFee) -> Self {
+        Self {
+            id: value.id,
+            amount_msats: value.amount_msats as u64,
+            prev_channel_id: value.prev_channel_id.map(|prev_channel_id| {
+                ChannelId::from_hex(prev_channel_id.as_str()).expect("prev channel id to decode")
+            }),
+            next_channel_id: value.next_channel_id.map(|next_channel_id| {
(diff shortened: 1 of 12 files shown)


coordinator/src/db/routing_fees.rs
```diff
@@ -0,0 +1,69 @@
...
+use bitcoin::hashes::hex::ToHex;
+use diesel;
+use diesel::prelude::*;
+use dlc_manager::ChannelId;
+use hex::FromHex;
+use time::OffsetDateTime;
+
+#[derive(Insertable, Debug, PartialEq)]
+#[diesel(table_name = routing_fees)]
+struct NewRoutingFee {
```
holzeis, line 13: 🤔 is it really necessary to have a separate insertable model for this simple data?

Personally I would prefer one model as I guess it will also reduce complexity.

coordinator/src/node/event_post_processing.rs
```diff
@@ -0,0 +1,47 @@
+use crate::db;
+use crate::node::Node;
+use crate::routing_fee::models::NewRoutingFee;
+use lightning::util::events::Event;
+use tokio::sync::watch;
+
+pub fn spawn(node: Node, mut node_event_receiver: watch::Receiver<Option<Event>>) {
+    tokio::spawn(async move {
```
holzeis, line 8: nit: i would preferred function not to spawn, but leave that to the calling function.

coordinator/src/node/event_post_processing.rs
```diff
@@ -0,0 +1,47 @@
+use crate::db;
```
holzeis, line 1: 🔧 event_post_processing is very generic. Can we make the file name more specific to what it does? E.g. routing_fee.rs

coordinator/src/node/event_post_processing.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::routing_fee::models::NewRoutingFee;
+use lightning::util::events::Event;
+use tokio::sync::watch;
+
+pub fn spawn(node: Node, mut node_event_receiver: watch::Receiver<Option<Event>>) {
+    tokio::spawn(async move {
+        loop {
+            match node_event_receiver.changed().await {
+                Ok(()) => {
+                    // we only handle event where the fee is set, i.e. events that do not concern a
```
holzeis, line 12: 🤔 that comment seems off. Force closing has nothing to do with fees for payment routing.

holzeis (approved):
LGTM 👍

I left minor remarks, but nothing critical to action.

coordinator/src/node/event_post_processing.rs
```diff
@@ -0,0 +1,47 @@
...
+                        let mut conn = match node.pool.get() {
+                            Ok(conn) => conn,
+                            Err(e) => {
+                                tracing::error!("Failed to connect to database during node event post processing event: {e:#}");
+                                continue;
+                            }
+                        };
+
+                        if let Err(e) = db::routing_fees::insert(
+                            NewRoutingFee {
```
bonomat, line 30: Since the struct `NewRoutingFee` is only used here, I think you can remove it and pass in the fields directly into this function.

bonomat (approved):
LGTM

coordinator/src/node/event_post_processing.rs
da-kami: My timebox for addressing this expired :) 
Diesel kept annoying me when I tried to safe the values directly using a tuple. I don't think it's worth investing more time into fighting with Diesel. 

I generally think it's OK to use the same pattern for the tables and accept this bit of boilerplate. If you think it's important I'm happy to address it in a follow up.

coordinator/src/node/event_post_processing.rs
da-kami: I can try to provide more context - my understanding is that `PaymentForwarded` is indeed influenced by channel force-close scenarios, see docs:

```rust
           PaymentForwarded {
		/// The incoming channel between the previous node and us. This is only `None` for events
		/// generated or serialized by versions prior to 0.0.107.
		prev_channel_id: Option<[u8; 32]>,
		/// The outgoing channel between the next node and us. This is only `None` for events
		/// generated or serialized by versions prior to 0.0.107.
		next_channel_id: Option<[u8; 32]>,
		/// The fee, in milli-satoshis, which was earned as a result of the payment.
		///
		/// Note that if we force-closed the channel over which we forwarded an HTLC while the HTLC
		/// was pending, the amount the next hop claimed will have been rounded down to the nearest
		/// whole satoshi. Thus, the fee calculated here may be higher than expected as we still
		/// claimed the full value in millisatoshis from the source. In this case,
		/// `claim_from_onchain_tx` will be set.
		///
		/// If the channel which sent us the payment has been force-closed, we will claim the funds
		/// via an on-chain transaction. In that case we do not yet know the on-chain transaction
		/// fees which we will spend and will instead set this to `None`. It is possible duplicate
		/// `PaymentForwarded` events are generated for the same payment iff `fee_earned_msat` is
		/// `None`.
		fee_earned_msat: Option<u64>,
		/// If this is `true`, the forwarded HTLC was claimed by our counterparty via an on-chain
		/// transaction.
		claim_from_onchain_tx: bool,
	},
```

I adapted the description slightly when refactoring the function; it's not in the doc-comment of `routing_fee::handle`. Feel free to make another suggestion if you have a more understandable description!

holzeis: Thanks for the explanation. 👍

Issue #931, opened 2023-07-10
Track allocated liquidity per user


bonomat:
We want to be able to associate liquidity per user so that we can track cost per user and revenue per locked up liquidity.

Issue #885, opened 2023-07-04
Replace common `EventHandler` with node-specific ones


luckysori:
App and coordinator should handle LDK events differently. For instance:

- The coordinator doesn't want to claim certain payments straight away.
- The coordinator can intercept payments to create JIT channels.
- The app needs to pay a channel-opening fee as soon as a new JIT channel has been created.

These systems will diverge more as we keep evolving 10101. It's a good idea to tackle this as soon as possible so that we don't keep building technical debt unnecessarily. At the moment this lack of separation doesn't even lead to simpler solutions, so it's hard to justify.

---

Related to #884.

bonomat:
Imho we should do this soon because with #926 we will have another actor which might have difference needs for the `EventHandler`.

PR #950, opened 2023-07-14
Feat/track channel fees

holzeis:
That changes turned out to be more complicated than initially expected. The initial scope has been reduced to simply tracking the funding transaction costs. However, after analyzing this issue for quite some time - I decided to introduce a shadow representation of the channel, what should help us moving forward on tracking other on-chain costs related to the channel.

Another pain point when implementing that feature, was that the generic `event_handler` is becoming a hell to maintain as we always have to consider changes in there for both the coordinator and the app. Unfortunately due to certain constraints like the `pending_intercepted_htlcs`, a missing testing framework for the coordinator, etc. it wasn't easy to simply implement handlers for propagated events on the coordinator directly. Instead the shadow channel is now tracked on the app as well as on the coordinator. Although it might not be as much required on the app as on the coordinator.

### todo
- [x] Split up costs into a dedicated on-chain transactions table
- [x] Split up `balance` into `inbound_capacity` and `outbound_capacity`
- [x] Check if an `INDEX` for `channel_id` is needed.
- [x] Make `Channel` database model private
- [x] Only transform to database model on data access layer

resolves https://github.com/get10101/meta/issues/174
resolves https://github.com/get10101/10101/issues/931
resolves https://github.com/get10101/10101/issues/954


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -658,14 +703,43 @@ where
...
+                            Channel::new(user_channel_id, channel_value as i64, target_node_id);
+                        if let Err(err) = self.storage.upsert_channel(new_channel) {
+                            // The channel is already in creation so we need to close the channel.
+                            // The closed channel should then fail the intercepted payment. We could
+                            // already fail the payment here, but it's nicer to have the logic
+                            // consistent at one place, when processing the `Event:ChannelClosed`
+                            tracing::error!("Failed to insert channel to database. Closing channel with temp_channel_id {}! Error: {err:?}", hex::encode(temp_channel_id));
+                            // Note, we can use the temp_channel_id here, as the channel is still in
+                            // creation.
+                            if let Err(err) = self
```
holzeis, line 734: 🔧 its probably better to fail the htlc intercepted instead of closing the channel immediately after creating it. 

In other words, I am going to change that so that the shadow channel will get created before we actually create the channel and fail the intercepted htlc if unsuccessful.

coordinator/src/node/storage.rs
```diff
@@ -112,4 +114,36 @@ impl node::Storage for NodeStorage {
...
+    }
+
+    fn get_channel(&self, user_channel_id: &str) -> Result<Option<Channel>> {
+        let mut conn = self.pool.get()?;
+        crate::db::channels::get(user_channel_id, &mut conn)
+            .map(|c| Some(c.into()))
+            .map_err(|e| anyhow!("{e:#}"))
+    }
+
+    fn all_channels(&self) -> Result<Vec<Channel>> {
```
bonomat, line 130: Maybe I'm missing something, but where is this being used?

mobile/native/src/db/mod.rs
```diff
@@ -347,7 +348,35 @@ pub fn get_spendable_outputs(
...
         .collect::<Result<Vec<_>>>()?;
 
-    tracing::info!(?outputs, "Got all spendable outputs");
+    tracing::debug!(?outputs, "Got all spendable outputs");
 
     Ok(outputs)
 }
+
+pub fn upsert_channel(channel: ln_dlc_node::Channel) -> Result<()> {
+    tracing::debug!(?channel, "Inserting channel");
```
bonomat, line 357: Was this really necessary or could you just have faked this for the App? If it is not being used/needed in the app, you could have simply had an empty funciton here.

coordinator/src/node/storage.rs
holzeis: it's not used at the moment (besides in the tests). are you recommending to remove that function?

mobile/native/src/db/mod.rs
holzeis: well, I guess faking it would have worked, but I think having a shadow of the channel beyond closing a channel is also useful to the app. However, I am not invested into having this already implemented with this change. I am happy to remove this for the app side. I leave it up to the reviewers to decide.

coordinator/src/node/storage.rs
bonomat: I'd remove it if it is not needed as it's dead code which needs to be maintained. I guess `clippy` is not complaining because it is a public trait?

Alternatively you could annotate it with `#[cfg(test)` :)

mobile/native/src/db/mod.rs
bonomat: If we keep it I see us having to maintain it in the future because the db model changed, etc.

> I leave it up to the reviewers to decide.

I'll leave it up to you :P

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -434,14 +441,42 @@ where
...
+                    %user_channel_id,
+                    %channel_id,
                     ?reason,
                     "\nChannel closed",
                 );
+
+                if let Some(mut channel) = self.storage.get_channel(&user_channel_id)? {
+                    if !vec![ChannelState::Pending, ChannelState::Open]
+                        .contains(&channel.channel_state)
+                    {
```
luckysori, line 458: 🔧 I think something like `if matches!(x, Closed | ForceClosedRemote | ForceClosedLocal)` is your friend here.

You could even encapsulate that in a `Channel::is_closed` method, to call `if channel.is_closed()` here.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -434,14 +441,42 @@ where
...
+                        ClosureReason::CounterpartyForceClosed { .. }
+                        | ClosureReason::CommitmentTxConfirmed => {
+                            channel.channel_state = ChannelState::ForceClosedRemote;
+                        }
+                        ClosureReason::HolderForceClosed { .. } => {
+                            channel.channel_state = ChannelState::ForceClosedLocal;
+                        }
+                        _ => {
+                            // TODO: we can't distinguish here whether we or the counter party
+                            // initiated the close of the channel.
```
luckysori, line 472: 🤔 I don't think we need a TODO here. Some of those `ClosureReason`s are not initiated by either party anyway. I would just document what each variant means in `ChannelState`.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -434,14 +441,42 @@ where
...
                 );
+
+                if let Some(mut channel) = self.storage.get_channel(&user_channel_id)? {
+                    if !vec![ChannelState::Pending, ChannelState::Open]
+                        .contains(&channel.channel_state)
+                    {
+                        tracing::warn!(%channel_id, %user_channel_id, "Unexpected state transition. Expected channel state to be either 'Pending' or 'Open', but was '{:?}'", channel.channel_state);
+                    }
+
+                    match reason {
```
luckysori, line 462: 🙃 This could be a `From` implementation.

crates/ln-dlc-node/src/lib.rs
```diff
@@ -105,6 +105,57 @@ pub enum PaymentFlow {
...
+            capacity,
+            trader,
+            created_at: OffsetDateTime::now_utc(),
+            updated_at: OffsetDateTime::now_utc(),
+            costs: 0,
+            channel_id: None,
+            funding_txid: None,
+        }
+    }
+}
```
luckysori, line 157: 🙃 This stuff deserves its own module I think.

crates/ln-dlc-node/src/lib.rs
```diff
@@ -105,6 +105,57 @@ pub enum PaymentFlow {
...
+    Closed,
+    ForceClosedRemote,
+    ForceClosedLocal,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Channel {
+    pub id: Option<i32>,
+    // The `user_channel_id` is set by 10101 at the time the `Event::HTLCIntercepted` when
+    // we are attempting to create a JIT channel.
```
luckysori, line 121: 🔧 All these could be docstrings I think.

crates/ln-dlc-node/src/lib.rs
```diff
@@ -105,6 +105,57 @@ pub enum PaymentFlow {
...
+    // Until the `Event::ChannelReady` we do not have a `channel_id`, which is derived from
+    // the funding transaction. We use the `user_channel_id` as identifier over the entirety
+    // of the channel lifecycle.
+    pub channel_id: Option<String>,
+    pub capacity: i64,
+    // Set at the `Event::ChannelReady`
+    pub funding_txid: Option<String>,
+    pub channel_state: ChannelState,
+    // The counter party of the channel.
+    pub trader: PublicKey,
```
luckysori, line 132: 🤔 Can't be `trader` if this is used by the app too to represent the counterparty.

crates/ln-dlc-node/src/lib.rs
```diff
@@ -105,6 +105,57 @@ pub enum PaymentFlow {
...
+    pub created_at: OffsetDateTime,
+    pub updated_at: OffsetDateTime,
+    // This data will be updated once the fee information is available. Please note, that
+    // this costs is not a source of truth and may not reflect the latest data. It is
+    // eventually implied from the on-chain fees of the channel transactions attached to
+    // this model.
+    pub costs: i64,
```
luckysori, line 139: ❓How can costs be negative?

crates/ln-dlc-node/src/lib.rs
```diff
@@ -105,6 +105,57 @@ pub enum PaymentFlow {
...
+    // The `user_channel_id` is set by 10101 at the time the `Event::HTLCIntercepted` when
+    // we are attempting to create a JIT channel.
+    pub user_channel_id: String,
+    // Until the `Event::ChannelReady` we do not have a `channel_id`, which is derived from
+    // the funding transaction. We use the `user_channel_id` as identifier over the entirety
+    // of the channel lifecycle.
+    pub channel_id: Option<String>,
+    pub capacity: i64,
+    // Set at the `Event::ChannelReady`
+    pub funding_txid: Option<String>,
```
luckysori, line 129: 🤔 I think we should use actual types for the model: `Uuid`, `ChannelId` (it's just a `[u8; 32]` so it's fine I guess), `Txid` and `u64`s for natural numbers.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -697,6 +764,31 @@ where
...
+            .get_channel_details(&channel_id)
+            .ok_or(anyhow!(
+                "Failed to get channel details by channel_id {}",
+                hex::encode(channel_id)
+            ))?;
+        if let Some(mut channel) = self.storage.get_channel(&user_channel_id)? {
+            channel.channel_state = ChannelState::Open;
+            channel.funding_txid = channel_details.funding_txo.map(|txo| txo.txid.to_string());
+            channel.channel_id = Some(hex::encode(channel_id));
+            channel.updated_at = OffsetDateTime::now_utc();
```
luckysori, line 785: 🔧 Might also want to encapsulate this transition in `Channel`.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -434,14 +441,42 @@ where
...
+                let user_channel_id = Uuid::from_u128(user_channel_id).to_string();
                 tracing::info!(
-                    %channel,
+                    %user_channel_id,
+                    %channel_id,
                     ?reason,
                     "\nChannel closed",
                 );
+
+                if let Some(mut channel) = self.storage.get_channel(&user_channel_id)? {
```
luckysori, line 455: 🤔 We should use `spawn_blocking` (preferred when possible) or `block_in_place` anywhere we access storage in an async context.

For all these cases that have been modified (I'm sure it also applies to some other ones too 😛), consider simply wrapping the whole arm in one of those (if there are no `await`s). Otherwise just wrap the potentially blocking code.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -328,6 +328,48 @@ where
                     guard.on_chain_sync_interval
                 });
 
+                let channels = match node_storage.all_channels_without_costs() {
```
luckysori, line 333: 🔧 I would prefer a dedicated task even if this data is related to the on-chain sync. The on-chain sync is critical for many reasons and we've had so many problems in the past with this task. Keeping it simple can only help us.

mobile/native/src/ln_dlc/node.rs
```diff
@@ -77,8 +78,8 @@ impl Node {
         })
     }
 
-    pub async fn get_wallet_histories(&self) -> Result<WalletHistories> {
-        let on_chain = self.inner.get_on_chain_history().await?;
+    pub fn get_wallet_histories(&self) -> Result<WalletHistories> {
```
luckysori, line 81: 🔧 Caller of this no longer needs to be `async` because of this. It would be great if we could update the callers of `keep_wallet_balance_and_history_up_to_date` too to use `spawn_blocking` where needed.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -328,6 +328,48 @@ where
...
+                    Ok(channels) => channels,
+                    Err(e) => {
+                        tracing::error!("Failed to fetch all channels without costs. Error: {e:#}");
+                        vec![]
```
luckysori, line 337: 🔧 I think `continue` is more idiomatic here.

coordinator/src/admin.rs
```diff
@@ -138,7 +138,7 @@ pub async fn list_dlc_channels(
 pub async fn list_on_chain_transactions(
     State(state): State<Arc<AppState>>,
 ) -> Result<Json<Vec<TransactionDetails>>, AppError> {
-    let transactions = state.node.inner.get_on_chain_history().await.map_err(|e| {
+    let transactions = state.node.inner.get_on_chain_history().map_err(|e| {
```
luckysori, line 141: 🔧 We should wrap this in `spawn_blocking`/`block_in_place`.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -249,7 +249,7 @@ pub fn run(data_dir: String, seed_dir: String, runtime: &Runtime) -> Result<()>
     })
 }
 
-async fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
```
luckysori, line 252: 🤔 Haha, caught you! This belongs in the previous patch!

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -257,7 +257,7 @@ where
     }
 
     #[autometrics]
-    pub async fn on_chain_transaction_list(&self) -> Result<Vec<TransactionDetails>> {
+    pub fn on_chain_transaction_list(&self) -> Result<Vec<TransactionDetails>> {
```
luckysori, line 261: 👍 Thank you for cleaning this up!

luckysori (approved):
I've reviewed the first 2 patches and I'm done for the day! I will take a look at the other one tomorrow, sorry!

I left a bunch of comments on small suggestions and some ideas on how to model things a bit differently. Reminder that only the 🤔 faces are more important (in the absence of 📣).

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,21 @@
...
+CREATE TYPE "ChannelState_Type" AS ENUM (
+    'Pending',
+    'Open',
+    'Closed',
+    'ForceClosedRemote',
+    'ForceClosedLocal'
+);
+CREATE TABLE "channels" (
+    user_channel_id TEXT PRIMARY KEY,
+    channel_id TEXT UNIQUE,
```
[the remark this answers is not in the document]
holzeis: The `channel_id` is only created after the funding transaction has been created. In order to track a channel state before that we have to use the `user_channel_id`.

Actually there are several ids.

- `user_channel_id`: Set by the app layer when creating the channel, available on all channel events.
- `temporary_channel_id`: Set when creating the channel, but not available in the event data after the channel is ready. I guess we could look it up using the `channel_id`.
- `channel_id`

In general I have the impression that the use of `temporary_channel_id` and the `channel_id` is a bit unpredictable. Depending on whether the funding transaction succeeded or not you'll have to use one or the other. As we want to manage our own shadow channel, I think its cleaner to use our own id if ldk is even supporting that feature.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,21 @@
+-- Your SQL goes here
+CREATE TYPE "ChannelState_Type" AS ENUM (
+    'Pending',
+    'Open',
+    'Closed',
+    'ForceClosedRemote',
+    'ForceClosedLocal'
+);
+CREATE TABLE "channels" (
+    user_channel_id TEXT PRIMARY KEY,
```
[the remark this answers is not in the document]
holzeis: I guess not, but isn't it implicit when using the `PRIMARY KEY` keyword?

coordinator/src/db/channels.rs
```diff
@@ -0,0 +1,147 @@
...
+
+pub fn upsert(channel: Channel, conn: &mut PgConnection) -> Result<()> {
+    let affected_rows = diesel::insert_into(channels::table)
+        .values(channel.clone())
+        .on_conflict(schema::channels::user_channel_id)
+        .do_update()
+        .set(&channel)
+        .execute(conn)?;
+
+    ensure!(affected_rows > 0, "Could not insert channel");
```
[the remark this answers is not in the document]
holzeis: credit goes to the person who wrote the code that I copied from 😅

coordinator/src/admin.rs
holzeis: Are you suggesting to do both: `block_in_place` and then `spawn_blocking`?

mobile/native/src/ln_dlc/mod.rs
holzeis: damn!

mobile/native/src/ln_dlc/node.rs
holzeis: Addressed here - https://github.com/get10101/10101/pull/950/commits/8fc8096e9a69a7415c9afe4e0b41dd0fabffc7b1

Please have a look if this is what you intended.

coordinator/src/admin.rs
holzeis: Addressed here - https://github.com/get10101/10101/pull/950/commits/8fc8096e9a69a7415c9afe4e0b41dd0fabffc7b1

Please have a look if this is what you intended.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,21 @@
...
+    channel_id TEXT UNIQUE,
+    capacity BIGINT NOT NULL,
```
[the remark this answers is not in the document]
holzeis: Address here - https://github.com/get10101/10101/pull/950/commits/bf5fcc2debfb79e714420aabc705689cad1db109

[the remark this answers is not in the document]
holzeis: Addressed here https://github.com/get10101/10101/pull/950/commits/89e64616bedbc48389553b770285ee4910e0ec24

i am now updating the channel balance on after every wallet sync. I also tried to hock into certain events like `PaymentClaimable`, `PaymentForwarded`, etc. but that turned out to be far to complicated. LDK is simply not consistent in when they are sending what channel information. e.g. most of the Payment events do not include a reference to a channel. so the easiest way to do it was to simply sync the `channel_details` data on a regular basis.

coordinator/src/node/storage.rs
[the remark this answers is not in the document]
holzeis: I think so. This feature was already hell to implement for multiple reasons and the shared event handler was definitely not a lot of help.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -434,14 +441,42 @@ where
             Event::ChannelClosed {
                 channel_id,
                 reason,
-                user_channel_id: _,
+                user_channel_id,
             } => {
-                let channel = hex::encode(channel_id);
+                let channel_id = hex::encode(channel_id);
+                let user_channel_id = Uuid::from_u128(user_channel_id).to_string();
```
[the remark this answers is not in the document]
holzeis: I am parsing the different types of the `user_channel_id` within the Channel impl. Let me know what you think.

crates/ln-dlc-node/src/lib.rs
holzeis: fixed: changed to counterparty

crates/ln-dlc-node/src/lib.rs
holzeis: it can't, its just the database model requires me to use `i64`

holzeis:
Setting this PR to draft to incorporate all feedback.

crates/ln-dlc-node/src/ln/event_handler.rs
holzeis: removed the todo

crates/ln-dlc-node/src/lib.rs
holzeis: changed to docstrings.

crates/ln-dlc-node/src/ln/event_handler.rs
holzeis: encapsulated into the channel impl.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
da-kami: `PRIMARY KEY` is basically a `UNIQUE NOT NULL` constraint that is more explicit. To my knowledge a PK cannot be assigned to multiple columns unless a combined PK is defined as separate constraint :)

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
da-kami: Note: If we use the `channel_id` to query for records (once it is known) we can opt for adding an `INDEX` to optimize querying.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
da-kami: 🙃 I prefer `inbound_capacity` and `outbound_capacity` over `capacity` and `balance`. I feel inbound and outbound is clearer when looking at it - for capacity and balance I always have to think about it.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    balance BIGINT NOT NULL,
+    funding_txid TEXT,
+    channel_state "ChannelState_Type" NOT NULL,
+    counterparty_pubkey TEXT NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    -- this value is stored here for simplicity of creating a sql query. However, it is not the source of truth and can
+    -- be recreated from the various transactions attached to the channel.
+    costs BIGINT NOT NULL DEFAULT 0
```
da-kami, line 21: 🙃 What do we consider "costs" in the context of this table - is this the on-chain transaction fees (only)? Maybe some more words what these "costs" actually represent would be nice.

da-kami (commented):
I had some thoughts when having a look at this PR. You don't have to action this - it's more meant for thinking about a future model. I hope it's not confusing.

---

I am having a hard time understanding what "costs" the `costs` field in the `channels` table aggregates.
I think it would be good if we can stick to the pattern of recording incoming and outgoing sats rather than recording aggregated costs. 
This would however require a model where we record the diff rather than the current state.

When it comes to on-chain transaction costs for the channel it might be better to record all broadcasted transactions and then associate the channel  (referencing the `transaction_id`s) as needed.
That way we should be able to track the channel costs.

---

I had a thought on that ledger table we discussed the other day. I still think it's good to model each cost/profit use case separate but I would try to not represent it as "cost" but more as incoming and outgoing sats. 
We could have a periodic task that loads from all the use-case specific tables and then saves records into a `ledger` table that would be an aggregate that we can easily query. From that ledger table we could query the costs.

Advantages include easier querying, but also some transformation / aggregation based on state. We can for example check if a payment was *actually* paid and only then store it in the ledger - similarly we can check for x confirmations on transactions before we store a record in the ledger.

Disadvantage is that we might introduce a bit more complexity -  we have to make sure that we don't save something in the ledger twice, i.e. we have to keep track of what we stored already. This could be achieved with (lose) relationships to the user case tables (payments, transactions, ...) so we can track if e.g. a payment was already recorded.

Concerning channel state the ledger would have to always record the diff, i.e. incoming and outgoing sats. 
If we have such a table I wonder if we could get rid of the aggregated `cost` field in the `channels` table, because the ledger would record these costs.

crates/ln-dlc-node/src/node/mod.rs
luckysori: Hmm, I assume the sleep is (or should be) outside of the loop. `continue` just goes to the next iteration.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
luckysori: Thanks for the explanation :)

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
luckysori: Makes sense, @holzeis and @da-kami!

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
luckysori: Yeah, my preference is something like `inbound_liquidity` and `outbound_liquidity`. Then balance is just the sum of the two.

luckysori:
I've unresolved some discussions where @da-kami added stuff just for visibility.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
da-kami: Another thought on this.
I started working on https://github.com/get10101/10101/issues/577 and it seems logical to add a `opening_fee_payment_hash` to the `channels` table that would record the `PaymentHash` that associates the JIT channel opening payment with the channel. 
I would not reflect that in the `costs` field though which made me thing if we can find a better name for `costs` - maybe `aggregated_transaction_costs` or similar?

coordinator/src/db/channels.rs
```diff
@@ -0,0 +1,153 @@
...
+    const HAS_STATIC_QUERY_ID: bool = false;
+
+    fn query_id() -> Option<TypeId> {
+        None
+    }
+}
+
+#[derive(Insertable, QueryableByName, Queryable, Debug, Clone, PartialEq, AsChangeset)]
+#[diesel(table_name = channels)]
+pub struct Channel {
```
da-kami, line 45: 🙃 I think it's cleaner to create a `Channel` in the models as well and not leak the database specific representation to the outside. This `Channel` struct should be private within `db::channels`. 
That way the database model can change independent of any business logic changes.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
da-kami: After an (unrelated) chat with @luckysori I came to the conclusion that it would be better to track the transaction costs based on the published transactions rather than this `cost` field.

We recently agreed that we should track all broadcasted transactions - see https://github.com/get10101/10101/issues/954 - if we do this first we can actually associate the transactions with the channel opening.
I think this is the way to go :)

Here's how this could look:

```sql
CREATE TABLE "channels" (
    user_channel_id TEXT PRIMARY KEY,
    channel_id TEXT UNIQUE,
    capacity BIGINT NOT NULL,
    balance BIGINT NOT NULL,
    funding_txid TEXT REFERENCES transactions(txid),
    commit_txid TEXT REFERENCES transactions(txid),
    closing_txid TEXT REFERENCES transactions(txid),
    channel_state "ChannelState_Type" NOT NULL,
    counterparty_pubkey TEXT NOT NULL,
    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
);

-- All the broadcasted transactions
CREATE TABLE transactions (
    txid TEXT PRIMARY KEY
    fee BIGINT NOT NULL DEFAULT 0,
    -- either our or counterparty pubkey
    broadcasted_by: TEXT NOT NULL,
);
```

That way we keep the associations clear:

- On-chain costs must be associated with a transaction
- Off-chain costs (Lightning channel related) must be associated with a payment
- (Trade pnl is virtual and handled outside of payments and transactions)

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
holzeis: The capacity is actually the total capacity of the channel. It does not say who owns how much. I can split it up into a `inbound_liquidity` and `outbound_liquidity`. The `balance` is the `output_liquidity` or `outbound_capacity`.

> Then balance is just the sum of the two.
That's not right. The `balance` would be your `outbound_liquidity`. The sum would be something like the total capacity of the channel.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
holzeis: I fear though we will anyways always have to load all channels to perform those updates.. as long as we can't directly hook to events that trigger an update of a changed channel. I struggled a long time making it work on an event basis, but opted to go back to a stupid regular task that updates the values by brute force.

Hopefully we can come back to that some other time and improve it.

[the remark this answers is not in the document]
holzeis: I will check if I have to query the channel by `channel_id` somewhere and if so I can add the `INDEX`, but I think I am using the `user_channel_id` everywhere.

crates/ln-dlc-node/src/node/mod.rs
holzeis: Not sure if I get you, the sleep wouldn't make much sense outside of the loop, would it?

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
holzeis: I like the idea of storing the transaction separately. That allows us to associate fees directly to the transaction. 👍

holzeis:
> I am having a hard time understanding what "costs" the `costs` field in the `channels` table aggregates. I think it would be good if we can stick to the pattern of recording incoming and outgoing sats rather than recording aggregated costs. This would however require a model where we record the diff rather than the current state.
 
> When it comes to on-chain transaction costs for the channel it might be better to record all broadcasted transactions and then associate the channel (referencing the `transaction_id`s) as needed. That way we should be able to track the channel costs.

I think that makes a lot of sense to me, I will look into incorporating that design.

> I had a thought on that ledger table we discussed the other day. I still think it's good to model each cost/profit use case separate but I would try to not represent it as "cost" but more as incoming and outgoing sats. We could have a periodic task that loads from all the use-case specific tables and then saves records into a `ledger` table that would be an aggregate that we can easily query. From that ledger table we could query the costs.
> 
> Advantages include easier querying, but also some transformation / aggregation based on state. We can for example check if a payment was _actually_ paid and only then store it in the ledger - similarly we can check for x confirmations on transactions before we store a record in the ledger.
> 
> Disadvantage is that we might introduce a bit more complexity - we have to make sure that we don't save something in the ledger twice, i.e. we have to keep track of what we stored already. This could be achieved with (lose) relationships to the user case tables (payments, transactions, ...) so we can track if e.g. a payment was already recorded.
> 
> Channel "costs" would be represented as transaction costs in the ledger (associated to the channel via the transaction id). I am not sure the ledger should depict locked up channel liquidity as well - we could represent that using incoming and outgoing sats that are associated to the channel, but I am not sure it makes sense.

Makes sense to me, I think your main point on the costs, being to generic and less useful when aggregated is perfectly valid. I will introduce a separate table for tracking transactions that we have broadcasted. This way we can associate the channel transactions with broadcasted transactions. I really like analogy to the lightning payments.

da-kami (approved):
Note: I did not do an extremely thorough review reading every line, but focused on the overall solution approach and I think this is good to go in.
I'd say given the size of this we should see to get it in and add improvements on top.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    'Closed',
+    'ForceClosedRemote',
+    'ForceClosedLocal'
+);
+-- All transactions broadcasted by us
+CREATE TABLE "transactions" (
+    txid TEXT PRIMARY KEY,
+    -- the fee is stored here for simplicity of creating a sql query. However, it is not the source of truth and can
+    -- be recreated from looking up the transaction on the blockchain.
+    fee BIGINT NOT NULL DEFAULT 0,
```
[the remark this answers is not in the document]
luckysori: From the comment on line 9, the intention is to store all transactions published by _us_ anyway.

coordinator/src/node/storage.rs
```diff
@@ -105,11 +109,60 @@ impl node::Storage for NodeStorage {
...
 
     fn all_spendable_outputs(&self) -> Result<Vec<SpendableOutputDescriptor>> {
         let mut conn = self.pool.get()?;
-        crate::db::spendable_outputs::get_all(&mut conn)
+        db::spendable_outputs::get_all(&mut conn)
+    }
+
+    // Channel
+
+    fn upsert_channel(&self, channel: Channel) -> Result<()> {
```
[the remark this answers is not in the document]
luckysori: I suppose because of the stupid generic `event_handler`.

I already documented the friction in the relevant issue: https://github.com/get10101/10101/issues/885.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn to_u128(&self) -> u128 {
+        self.0.as_u128()
+    }
+}
+
+impl ToString for UserChannelId {
+    fn to_string(&self) -> String {
+        self.0.to_string()
+    }
+}
```
luckysori, line 48: 🔧 From the [`ToString` docs](https://doc.rust-lang.org/std/string/trait.ToString.html):

> This trait is automatically implemented for any type which implements the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait. As such, ToString shouldn’t be implemented directly: [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) should be implemented instead, and you get the ToString implementation for free.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+impl TryFrom<String> for UserChannelId {
+    type Error = anyhow::Error;
+
+    fn try_from(value: String) -> std::result::Result<Self, Self::Error> {
+        let user_channel_id = Uuid::from_str(value.as_str())?;
+        Ok(UserChannelId(user_channel_id))
+    }
+}
+
+impl Display for Channel {
```
luckysori, line 65: 👍 Good idea.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+            self.user_channel_id.to_string(), self.channel_state, self.counterparty, self.funding_txid, self.created_at, self.updated_at
+        )
+        .fmt(f)
+    }
+}
+
+// We create a shadow copy of the channel as the the ldk channel does not
+// live beyond channel closure. The main purpose of
+// this shadow is to track general meta data of the
+// channel relevant for reporting purposes.
```
luckysori, line 82: 🙃 The formatting here is bothering me haha.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    pub fn update_liquidty(
+        channel: &Channel,
+        channel_details: &Option<ChannelDetails>,
+    ) -> Result<Channel> {
+        let mut channel = channel.clone();
+        match channel_details {
+            Some(channel_details) => {
+                channel.inbound = channel_details.inbound_capacity_msat / 1000;
+                channel.outbound = channel_details.outbound_capacity_msat / 1000;
```
luckysori, line 144: 🤔 I would highly recommend keeping msats around until the moment we want to display something. I think that this has been biting us in the e2e tests (I fixed some of that with #951).

Now it probably doesn't matter so much in terms of reporting, but you never know long term...

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn open_channel(
+        channel: Option<Channel>,
+        channel_details: ChannelDetails,
+    ) -> Result<Channel> {
+        let mut channel = match channel {
+            Some(channel) => channel,
+            None => {
+                let user_channel_id =
+                    UserChannelId::from(channel_details.user_channel_id).to_string();
+                tracing::warn!(%user_channel_id, channel_id = %channel_details.channel_id.to_hex(), public = channel_details.is_public, outbound = channel_details.is_outbound, "Creating a new shadow channel");
```
luckysori, line 162: 🔧 The log message is not giving me warning vibes. Maybe `"Cannot open non-existent shadow channel. Creating a new one"`.

We might also want to check if the state is not `Pending` to log another warning.

I might also consider just printing the channel that you actually created, particularly since you added a nice `Display` implementation.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+                )
+            }
+        };
+
+        tracing::debug!("Updating shadow channel.");
+        channel.channel_state = ChannelState::Open;
+        channel.funding_txid = channel_details.funding_txo.map(|txo| txo.txid);
+        channel.channel_id = Some(channel_details.channel_id);
+        channel.updated_at = OffsetDateTime::now_utc();
+        Ok(channel)
```
luckysori, line 176: 🔧 Probably should log the whole channel after mutating!

```suggestion
        channel.channel_state = ChannelState::Open;
        channel.funding_txid = channel_details.funding_txo.map(|txo| txo.txid);
        channel.channel_id = Some(channel_details.channel_id);
        channel.updated_at = OffsetDateTime::now_utc();

        tracing::debug!(%channel, "Set shadow channel to open");

        Ok(channel)
```

crates/ln-dlc-node/src/lib.rs
```diff
@@ -35,14 +35,17 @@ mod ln_dlc_wallet;
...
 pub mod node;
 pub mod seed;
+pub mod transaction;
 
 pub use ln::CONFIRMATION_TARGET;
 pub use ln::CONTRACT_TX_FEE_RATE;
 pub use ln::JUST_IN_TIME_CHANNEL_OUTBOUND_LIQUIDITY_SAT_MAX;
 pub use ln::LIQUIDITY_MULTIPLIER;
 
+mod shadow;
```
luckysori, line 48: 🔧 Should go with the other regular private modules above.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -658,11 +691,22 @@ where
...
+
+                let new_channel = Channel::new(0, channel_value, target_node_id);
+                tracing::debug!(%new_channel, "Creating shadow channel");
+                if let Err(err) = self.storage.upsert_channel(new_channel.clone()) {
+                    tracing::error!(%intercepted_id, "Failed to insert channel to database. Error: {err:#}");
+                    if let Err(err) = self.channel_manager.fail_intercepted_htlc(intercept_id) {
+                        tracing::error!(%intercepted_id, "Could not fail intercepted htlc {err:?}")
+                    }
+                    return Ok(());
+                }
```
luckysori, line 703: 🙃 We probably need more `block_in_place`s here, but we can clean it up later.

crates/ln-dlc-node/src/ln_dlc_wallet.rs
```diff
@@ -304,7 +309,17 @@ impl dlc_manager::Wallet for LnDlcWallet {
...
+        let transaction = crate::transaction::Transaction {
+            txid: tx.txid(),
+            fee: 0,
+            created_at: OffsetDateTime::now_utc(),
+            updated_at: OffsetDateTime::now_utc(),
+        };
+        match self.node_storage.upsert_transaction(transaction.clone()) {
+            Ok(()) => tracing::info!(%transaction, "Successfully created a shadow transaction."),
+            Err(e) => tracing::error!("Failed to create shadow transaction. Error: {e:#}"),
+        }
```
luckysori, line 322: 🙃 I would just log the error with `if let`.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub channel_id: Option<ChannelId>,
+    pub inbound: u64,
+    pub outbound: u64,
+    /// Set at the `Event::ChannelReady`
+    pub funding_txid: Option<Txid>,
+    pub channel_state: ChannelState,
+    /// The counter party of the channel.
+    pub counterparty: PublicKey,
+    pub created_at: OffsetDateTime,
+    pub updated_at: OffsetDateTime,
```
luckysori, line 100: 🔧 I think this only pertains to the database model.

crates/ln-dlc-node/src/transaction.rs
```diff
@@ -0,0 +1,23 @@
...
+use std::fmt::Formatter;
+use time::OffsetDateTime;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Transaction {
+    pub txid: Txid,
+    pub fee: u64,
+    pub created_at: OffsetDateTime,
+    pub updated_at: OffsetDateTime,
+}
```
luckysori, line 13: 🙃 Here we have another opportunity to encapsulate when the fee is updated with methods etc. The idea would be to be able to remove `pub` from all the fields here (and in `Channel` too).

crates/ln-dlc-node/src/transaction.rs
```diff
@@ -0,0 +1,23 @@
+use bitcoin::Txid;
+use std::fmt;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use time::OffsetDateTime;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Transaction {
+    pub txid: Txid,
+    pub fee: u64,
```
luckysori, line 10: 🔧 I reckon this should be an `Option`.

mobile/native/migrations/2023-07-11-104904_create_channels/up.sql
luckysori, line 1: 🤔 I wonder if we need to save all this (any of it?) for the app.

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
luckysori, line 1: 🤔 As usual I'm not very happy when we start making these tests more complicated. I reckon this should be tested differently, but this is probably the easiest for now.

luckysori (approved):
A few more comments because that's how I roll, but please just merge it and action anything afterwards (or delegate to me for that).

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
da-kami: hm... I wonder if that intention is feasible. 

My thinking: 
If we succeed in publishing a transaction we safe it ✅
But if we fail to publish it it depends *why* - if we fail because it is already published we should still safe it, but in a way that depicts that it was not published by us. 

(Note: In ItchySats we did not record it as failure when we noticed that a transaction that we tried to published was already published and hence the publication failed - I would treat this similar)

crates/ln-dlc-node/src/shadow.rs
```diff
@@ -0,0 +1,71 @@
...
+                channel = Channel::update_liquidty(&channel, &channel_details)?;
+            }
+            channel.funding_txid = Some(funding_txid);
+            channel.updated_at = OffsetDateTime::now_utc();
+            self.storage.upsert_channel(channel.clone())?;
+        }
+        Ok(())
+    }
+
+    pub fn sync_transactions(&self) -> Result<()> {
```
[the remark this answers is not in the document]
holzeis: It is possible, but I think the other way around is better as we will have less db queries.

In the current solution I am only getting transactions that do not have a fee (fee == 0).

coordinator/src/node/storage.rs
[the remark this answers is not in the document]
holzeis: The shared event handler is the main reason.

However this data is also useful to the app side. Especially remembering broadcasted transactions.

@luckysori could that eventually replace remembering the spendable outputs?

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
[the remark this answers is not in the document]
luckysori: It's a bit different in Lightning because the other party doesn't have the same commitment transaction. They have their "local" commitment transaction which is similar, but not exactly the same.

If you end up not being able to publish your local commitment transaction it's because the remote commitment transaction makes it on-chain first. In that case it would be incorrect to save your local commitment transaction.

Eventually we could save the remote commitment transaction as found on-chain, but that's probably outside of the scope of this PR which is already so big.

coordinator/src/node/storage.rs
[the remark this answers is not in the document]
luckysori: > could that eventually replace remembering the spendable outputs?

I would still trust LDK to generate the spendable-output events rather than try to be smarter by getting ahead of it by considering these transactions instead. I think we still need some extra information in the event to actually spend the output straight away.

coordinator/migrations/2023-07-11-104904_create_channels/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+CREATE TABLE "channels" (
+    user_channel_id TEXT PRIMARY KEY,
+    channel_id TEXT UNIQUE,
+    inbound BIGINT NOT NULL,
+    outbound BIGINT NOT NULL,
+    funding_txid TEXT,
```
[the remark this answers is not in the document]
holzeis: Currently we also shadow channels that are pending, meaning the funding transaction does not have to exist.

We could create a separate model for that. But I guess the we should have a data model per channel state, as the same will apply for a commitment tx, etc.

coordinator/src/node/storage.rs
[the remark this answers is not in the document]
holzeis: > I think we still need some extra information in the event to actually spend the output straight away.

The current implementation of the transaction does not save enough data true, but we could extend it.

Aren't we storing the spendable outputs because a broadcasted transaction failed? Or would we run into a risk of losing the transaction after all when the app dies before broadcasting?

mobile/native/migrations/2023-07-11-104904_create_channels/up.sql
holzeis: Are you suggesting @bonomat [proposal](https://github.com/get10101/10101/pull/950#discussion_r1264723049)?

coordinator/src/node/storage.rs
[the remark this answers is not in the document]
luckysori: We store them unconditionally, because you could broadcast a transaction successfully and it might still never get confirmed.

Handling such scenarios is expected to be done with #846.

mobile/native/migrations/2023-07-11-104904_create_channels/up.sql
luckysori: Yeah, exactly. It might be more work now to undo this, so you don't have to.

I suppose my very slight concern is that if we ever split up the event handler we will then remove the `Storage` trait which will let us define interfaces that are actually useful specifically for each role. If we have written all this code already we might believe that we actually need it the way it was originally written, even though it was mostly written to actually implement this trait.

Not very important either way.

holzeis:
@luckysori @da-kami thanks for all your feedback, I've incorporated most of it and only left remarks that haven't been critical or easy to address. As I am having an approval from the both of you, I will go ahead with the merge now. Please feel free to follow up on your remarks if you see them necessary.

bors r+

Issue #927, opened 2023-07-07
Re-model fee rate coordinator


bonomat:
## TL;DR:

Make the routing fee currently hardcoded in the [channel config](https://github.com/get10101/10101/blob/e12d324a241befe8c719afc0f276dfccdc136d92/crates/ln-dlc-node/src/ln/config.rs#L74) configurable from the [settings](https://github.com/get10101/10101/blob/b4e06d514ebb9cd1d2169d8e3c8d5d552d1bd7b1/coordinator/src/settings.rs#L39) and charge the same fee for all channels: 

- when creating new JIT channels
- when opening outbound channels (to non-app users)
- when accepting in-bound channels
- (optional): look into how we can configure routing fee during runtime, document and create a follow-up ticket. 

## Long story

Our node got "exploited" because we were charging 
- 1 sat base fee
- 0% routing fee

Someone used our super cheap routing fee to rebalance their node. 

We should charge a X% routing fee but a 0 sat base fee. This is industry standard as it simplifies path finding. 

We should use the same fee for 
- the routing hint in the interceptable invoice (https://github.com/get10101/10101/blob/a659e992bbbc2054abcfab8c96ea7328a1cd353d/coordinator/src/routes.rs#L144)
- the actual set fee in the channel config (https://github.com/get10101/10101/blob/e12d324a241befe8c719afc0f276dfccdc136d92/crates/ln-dlc-node/src/ln/config.rs#L74)

Right now these two values are not related and we could get weird errors if they diverge. 
The former one needs to be at least as high as the latter one. 

Differently said, since we want this to be configured from the outside, the setting in https://github.com/get10101/10101/blob/b4e06d514ebb9cd1d2169d8e3c8d5d552d1bd7b1/coordinator/src/settings.rs#L39 should be used for all channels (public and private). 

Additionally, it would be good to update routing fees for existing channels. This can be done in a follow-up but might be worth considering.

luckysori:
> Additionally, it would be good to update routing fees for existing channels. This can be done in a follow-up but might be worth considering.

This does not seem to be possible as of today using LDK: https://github.com/lightningdevkit/rust-lightning/blob/df237ba3b455f0ef246604125b8933a7f0074fc5/lightning/src/util/config.rs#L373-L374.

luckysori:
> > Additionally, it would be good to update routing fees for existing channels. This can be done in a follow-up but might be worth considering.
> 
[...]

PR #951, opened 2023-07-17
Configure proportional routing fee consistently

luckysori:
Fixes https://github.com/get10101/10101/issues/927.
Fixes https://github.com/get10101/meta/issues/178.


crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -152,7 +164,7 @@ where
         tracing::info!(peer_id=%target_node, %intercept_scid, "Successfully created intercept scid for payment routing");
         InterceptableScidDetails {
             scid: intercept_scid,
-            jit_routing_fee_millionth: jit_fee_rate_basis_point * 100,
+            jit_routing_fee_millionth: jit_routing_fee_millionth * 100,
```
holzeis, line 167: 🤔 I would remove the `jit` prefix from that variable. As the `just in time` aspect on how the channel got created has little to do on how much we charge on routing fees.

Note, we only need to provide that information here, because 1) the channel might not exist yet and 2) private channels don't get announced, and the sender does not know the required fees for constructing a payment to that channel. The later scenario should be the most common though.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -128,21 +128,33 @@ where
...
+    ///
+    /// - `jit_fee_rate_basis_points`: Fee rate to be charged for opening just in time channels.
+    /// Rate is in basis points, i.e. 100 basis point=1% or 50=0.5%. This argument is optional. If
+    /// not provided, we default to the value configured in [`Node::channel_config`].
     #[autometrics]
     pub fn create_intercept_scid(
         &self,
         target_node: PublicKey,
-        jit_fee_rate_basis_point: u32,
+        jit_fee_rate_basis_point: Option<u32>,
```
holzeis, line 146: ❓ Do we still need that argument? I might have missed something, but isn't it always set to `None`?

holzeis (approved):
LGTM 👍

crates/ln-dlc-node/src/node/invoice.rs
luckysori: Good point, I don't think it's worth keeping it around!

crates/ln-dlc-node/src/node/invoice.rs
luckysori: I think the last patch I am gonna add to this PR will address this.

> Note, we only need to provide that information here, because 1) the channel might not exist yet and 2) private channels don't get announced, and the sender does not know the required fees for constructing a payment to that channel. The later scenario should be the most common though.

I guess this comment reflects the fact we are using this API for multiple purposes. If we route a "regular" payment to the app what you say makes sense. It might mean that we always want/need to add this route hint to the invoice even if we don't want to intercept it!

For the purpose that it was designed for, this information is simply necessary because we _need_ the last `RouteHintHop` to be able to intercept the channel.

luckysori: It was actually merged onto https://github.com/get10101/10101/pull/951/commits/2ddbee01ccb53fe135f3af5497ef239f8693c4b8 in the end, since the test weren't passing anyway.

luckysori:
@holzeis gave me the thumbs up privately.

bors r+

PR #956, opened 2023-07-17
fix: use in-memory scorer for app and persistent for coordinator

bonomat:
Reading the scorer from disk on iOS seems to panic sometimes, hence, we use an in-memory scorer for the app and a persistant scorer for the coordinator.

fix: #797


holzeis (approved):
LGTM 👍 

I guess that should be included when releasing the app pointing to the new coordinator ip.

Issue #843, opened 2023-06-26
JIT - Channel Fee: LSP Flow


bonomat:
Follow-up ticket from #577 

LSP flow: there is work on Bolt X (TBD) which says that nodes may accept payments smaller than amounts in the invoice. Once this is done, we can subtract a channel fee from the payment amount same as Breez does. This is defined in [LSPS2](https://github.com/ZmnSCPxj-jr/lsp/blob/jit-channels/LSPS2/README.md?rgh-link-date=2023-05-02T21%3A21%3A21Z) of the upcoming LSP Spec and will find its way into bolts and all implementations soon. https://github.com/get10101/10101/issues/393

bonomat:
Might work once we upgrade to 0.116 of LDK: https://github.com/lightningdevkit/rust-lightning/issues/1999

PR #958, opened 2023-07-18
Track channel opening transaction fee payments

da-kami:
fixes https://github.com/get10101/meta/issues/175

Upon requesting the invoice the user provides the amount to be paid and the funding transaction id that the fee was extracted from. The coordinator creates the invoice, records the payment in the database and then associates the `payment_hash` with with the `channels` in the database.

This PR is is stacked on top of https://github.com/get10101/10101/pull/950

Notes:

- After a chat with @luckysori we came to the conclusion that the funding transaction fee costs (that the app currently reimburses in a payment) should be the only missing piece for  https://github.com/get10101/meta/issues/175 because any other fees (charged for channel opening) should be covered by the routing fees already. If we want to somehow associate them with the channel as opening fees as well we would have to add additional handling for that. (not sure that's worth it).

### DIFF
--- a/coordinator/src/db/channels.rs
+++ b/coordinator/src/db/channels.rs
@@ -2,6 +2,7 @@ use crate::schema;
 use crate::schema::channels;
 use crate::schema::sql_types::ChannelStateType;
 use anyhow::ensure;
+use anyhow::Context;
 use anyhow::Result;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
@@ -22,6 +23,7 @@ use diesel::QueryableByName;
 use diesel::RunQueryDsl;
 use dlc_manager::ChannelId;
 use hex::FromHex;
+use lightning::ln::PaymentHash;
 use ln_dlc_node::channel::UserChannelId;
 use std::any::TypeId;
 use std::str::FromStr;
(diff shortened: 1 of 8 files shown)


da-kami:
@bonomat is it important that the channel opening routing fee can be distinguished from *other* routing fees we collect?

As stated in the description, if we want that we would have to associate channel open related routing fees with the channels as well. This PR does not implement this :)

coordinator/src/db/channels.rs
```diff
@@ -77,6 +80,26 @@ pub(crate) fn get_all_non_pending_channels(conn: &mut PgConnection) -> QueryResu
...
 
+pub(crate) fn update_payment_hash(
+    payment_hash: PaymentHash,
+    funding_txid: String,
+    conn: &mut PgConnection,
+) -> Result<()> {
+    let channel = channels::table
+        .filter(channels::funding_txid.eq(funding_txid.clone()))
+        .first(conn)
+        .optional();
```
holzeis, line 91: 🙃 you could skip that `optional()` and simplify the match below.

Without optional you would get a `NotFound` Error instead of `None`.

holzeis (approved):
LGTM 👍

bonomat:
> @bonomat is it important that the channel opening routing fee can be distinguished from _other_ routing fees we collect?

I don't think this is important. The routing fee when opening channels will go away in the future, see ticket [here](https://github.com/get10101/10101/issues/843).

bonomat:
> It don't think that fee will go away, as it is still a routing. 

You won't be able to distinguish it anymore because the routing fee is part of the JIT channel opening fee, e.g. he sends 100k and receives only 50k.

PR #959, opened 2023-07-18
Check coordinator status from 10101 wallet (backend)

klochowicz:
- Introduce a mechanism for communicating service status to the frontend.

- Communicate about the orderbook status (websocket stream connection) and
  coordinator status (responsive on HTTP API).

- Verify in e2e test that the status is updated correctly.

Note: The frontend part is well underway, it will come in a separate PR.

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -115,6 +115,7 @@ pub fn router(
         )
         .route("/api/admin/sync", post(post_sync))
         .route("/metrics", get(get_metrics))
+        .route("/health", get(get_health))
         .with_state(app_state)
 }
 
@@ -380,4 +381,10 @@ pub async fn get_metrics(State(state): State<Arc<AppState>>) -> impl IntoRespons
     };
 
     (StatusCode::OK, open_telemetry_metrics + &autometrics)
+}
+
+pub async fn get_health() -> Result<Json<String>, AppError> {
+    // TODO: Implement any health check logic we'd need
+    // So far this just returns if the server is running
+    Ok(Json("Server is healthy".to_string()))
 }
--- a/crates/tests-e2e/src/test_subscriber.rs
+++ b/crates/tests-e2e/src/test_subscriber.rs
@@ -3,9 +3,13 @@ use native::api::ContractSymbol;
 use native::api::WalletInfo;
 use native::event::subscriber::Subscriber;
 use native::event::EventType;
+use native::health::Service;
+use native::health::ServiceStatus;
+use native::health::ServiceUpdate;
 use native::trade::order::Order;
 use native::trade::position::Position;
 use orderbook_commons::Prices;
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::Mutex;
 use tokio::sync::watch;
@@ -19,6 +23,7 @@ pub struct Senders {
     init_msg: watch::Sender<Option<String>>,
     prices: watch::Sender<Option<Prices>>,
     position_close: watch::Sender<Option<ContractSymbol>>,
+    service: watch::Sender<Option<ServiceUpdate>>,
 }
 
 /// Subscribes to events destined for the frontend (typically Flutter app) and
@@ -31,10 +36,12 @@ pub struct TestSubscriber {
     init_msg: watch::Receiver<Option<String>>,
     prices: watch::Receiver<Option<Prices>>,
     position_close: watch::Receiver<Option<ContractSymbol>>,
+    services: Arc<Mutex<HashMap<Service, ServiceStatus>>>,
+    _service_map_updater: tokio::task::JoinHandle<()>,
 }
 
 impl TestSubscriber {
-    pub fn new() -> (Self, ThreadSafeSenders) {
+    pub async fn new() -> (Self, ThreadSafeSenders) {
         let (wallet_info_tx, wallet_info_rx) = watch::channel(None);
         let (order_tx, order_rx) = watch::channel(None);
         let (order_filled_tx, order_filled_rx) = watch::channel(None);
@@ -42,6 +49,7 @@ impl TestSubscriber {
         let (init_msg_tx, init_msg_rx) = watch::channel(None);
         let (prices_tx, prices_rx) = watch::channel(None);
         let (position_close_tx, position_close_rx) = watch::channel(None);
+        let (service_tx, mut service_rx) = watch::channel(None);
 
         let senders = Senders {
             wallet_info: wallet_info_tx,
@@ -51,9 +59,28 @@ impl TestSubscriber {
             init_msg: init_msg_tx,
             prices: prices_tx,
             position_close: position_close_tx,
+            service: service_tx,
         };
 
-        let rx = Self {
+        let services = Arc::new(Mutex::new(HashMap::new()));
+
+        let _service_map_updater = {
+            let services = services.clone();
+            tokio::spawn(async move {
+                while let Ok(()) = service_rx.changed().await {
+                    if let Some((service, status)) = *service_rx.borrow() {
+                        tracing::debug!(?service, ?status, "Updating status in the services map");
+                        services
+                            .lock()
+                            .expect("mutex not poisoned")
+                            .insert(service, status);
+                    }
+                }
+                panic!("service_rx channel closed");
+            })
+        };
+
+        let subscriber = Self {
             wallet_info: wallet_info_rx,
             order_filled: order_filled_rx,
             order: order_rx,
@@ -61,8 +88,10 @@ impl TestSubscriber {
             init_msg: init_msg_rx,
             prices: prices_rx,
             position_close: position_close_rx,
+            services,
+            _service_map_updater,
         };
-        (rx, ThreadSafeSenders(Arc::new(Mutex::new(senders))))
+        (subscriber, ThreadSafeSenders(Arc::new(Mutex::new(senders))))
     }
 
     pub fn wallet_info(&self) -> Option<WalletInfo> {
@@ -91,6 +120,15 @@ impl TestSubscriber {
 
     pub fn position_close(&self) -> Option<ContractSymbol> {
         self.position_close.borrow().as_ref().cloned()
+    }
+
+    pub fn status(&self, service: Service) -> ServiceStatus {
+        self.services
+            .lock()
+            .expect("mutex not poisoned")
+            .get(&service)
+            .copied()
+            .unwrap_or_default()
     }
 }
 
@@ -142,6 +180,12 @@ impl Subscriber for Senders {
                     .send(Some(prices.clone()))
                     .expect("to be able to send update");
             }
+            native::event::EventInternal::ServiceHealthUpdate(update) => {
+                tracing::debug!(?update, "Received service health update event");
+                self.service
+                    .send(Some(*update))
+                    .expect("to be able to send update");
+            }
             native::event::EventInternal::ChannelReady(channel_id) => {
                 tracing::trace!(?channel_id, "Received channel ready event");
             }
@@ -159,6 +203,7 @@ impl Subscriber for Senders {
             EventType::PositionUpdateNotification,
             EventType::PositionClosedNotification,
             EventType::PriceUpdateNotification,
+            EventType::ServiceHealthUpdate,
         ]
     }
 }
--- a/crates/tests-e2e/src/test_subscriber.rs
+++ b/crates/tests-e2e/src/test_subscriber.rs
@@ -134,64 +134,8 @@ impl TestSubscriber {
 
 impl Subscriber for Senders {
     fn notify(&self, event: &native::event::EventInternal) {
-        match event {
-            native::event::EventInternal::Init(init) => {
-                tracing::info!(%init, "Received init message");
-                self.init_msg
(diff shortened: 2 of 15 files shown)


mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
+            ServiceStatus::Online
+        } else {
+            ServiceStatus::Offline
+        };
+
+        // Update the health status on the watch channel
+        tx.send(status).expect("Receiver not to be dropped");
+
+        // Sleep for 1 second before performing the next health check
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
bonomat, line 111: Can we use a constant for this please.

Also, I think 1/second is a bit too often.

klochowicz: of course - is 5 seconds good for you? we can always pass it in from the config too.

mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
+async fn check_health_endpoint(endpoint: String, tx: watch::Sender<ServiceStatus>) {
+    loop {
+        let status = if send_request(&endpoint).await.is_ok() {
+            ServiceStatus::Online
+        } else {
+            ServiceStatus::Offline
+        };
+
+        // Update the health status on the watch channel
+        tx.send(status).expect("Receiver not to be dropped");
```
bonomat, line 108: Where are you checking the `status` right now? I can't seem to find it. 

What do you think of checking the `status` in here and only forward if the service was available or not?

mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
+                }
+            },
+        }
+    }
+}
+
+/// Periodically checks the health of a given service and updates the watch channel
+async fn check_health_endpoint(endpoint: String, tx: watch::Sender<ServiceStatus>) {
+    loop {
+        let status = if send_request(&endpoint).await.is_ok() {
```
bonomat, line 101: I think there is a flaw in this logic: as long as the request returns you any status code (could be a `404`) you treat the service as online.

klochowicz: using the method below (`error_for_status`) is a very convenient way of checking `reqwest::Response` - it returns an `Err` for any response in range 400...599: (and then you can still inspect the `Err` if needed, which I don't need in this case).

source: `https://docs.rs/reqwest/0.11.18/reqwest/struct.Response.html#method.error_for_status

bonomat: Oh, very nice. Sorry that I missed that.

coordinator/src/routes.rs
```diff
@@ -401,3 +402,9 @@ pub async fn get_metrics(State(state): State<Arc<AppState>>) -> impl IntoRespons
 
     (StatusCode::OK, open_telemetry_metrics + &autometrics)
 }
+
+pub async fn get_health() -> Result<Json<String>, AppError> {
+    // TODO: Implement any health check logic we'd need
+    // So far this just returns if the server is running
+    Ok(Json("Server is healthy".to_string()))
```
luckysori, line 409: 🙃 Maybe `{ is_healthy: true }` rather than a string?

mobile/native/src/config/mod.rs
```diff
@@ -13,10 +13,10 @@ use url::Url;
 static CONFIG: Storage<ConfigInternal> = Storage::new();
 
 #[derive(Clone)]
-struct ConfigInternal {
+pub struct ConfigInternal {
     coordinator_pubkey: PublicKey,
     esplora_endpoint: Url,
-    http_endpoint: SocketAddr,
+    pub http_endpoint: SocketAddr,
```
luckysori, line 19: 🔧 Should be a getter like the others. We could even get the `coordinator_health_endpoint` directly by constructing it here.

mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
+        tx.send(status).expect("Receiver not to be dropped");
+
+        // Sleep for 1 second before performing the next health check
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    }
+}
+
+// Returns the status code of the health endpoint, returning an error if the request fails
+async fn send_request(endpoint: &str) -> Result<StatusCode> {
+    tracing::debug!(%endpoint, "Sending request");
```
luckysori, line 117: 🔧 I would demote this to TRACE.

mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
+        let status = if send_request(&endpoint).await.is_ok() {
+            ServiceStatus::Online
+        } else {
+            ServiceStatus::Offline
+        };
+
+        // Update the health status on the watch channel
+        tx.send(status).expect("Receiver not to be dropped");
+
+        // Sleep for 1 second before performing the next health check
```
luckysori, line 110: 🙃 I think this comment can be removed.

mobile/native/src/health.rs
```diff
@@ -22,44 +28,96 @@ pub enum ServiceStatus {
...
                 orderbook: orderbook_tx,
             },
         )
     }
 }
+
+/// Publishes the health status updates for a given service to the event hub
+async fn publish_status_updates(service: Service, mut rx: watch::Receiver<ServiceStatus>) {
+    loop {
+        select! {
```
luckysori, line 80: 🔧 We don't need a `select` if we are only polling one future.

I suppose we can keep it if we know we are gonna use it for something else soon, but I'd rather keep the code simple until we need to make it complicated.

crates/tests-e2e/src/test_subscriber.rs
```diff
@@ -134,77 +134,63 @@ impl TestSubscriber {
...
+            EventType::PositionClosedNotification,
+            EventType::PriceUpdateNotification,
+            EventType::ServiceHealthUpdate,
+        ]
+    }
+}
+
+impl Senders {
+    fn handle_event(&self, event: &native::event::EventInternal) -> anyhow::Result<()> {
+        tracing::trace!(?event, "Received event");
```
luckysori, line 157: 👍 Nice!

luckysori (approved):

coordinator/src/routes.rs
da-kami: Given the current implementation of a simple health endpoint that can be either queried or not (because the server is not available) I would expect an empty `200` as response. 

Eventually we could make a more elaborate endpoint that could return an object that could provide a reason as well (e.g. the oracle is not reachable in the coordinator).

da-kami (approved):
@luckysori already mentioned all the things :D - I only added to his comments

crates/tests-e2e/src/test_subscriber.rs
```diff
@@ -51,18 +59,39 @@ impl TestSubscriber {
             init_msg: init_msg_tx,
             prices: prices_tx,
             position_close: position_close_tx,
+            service: service_tx,
         };
 
-        let rx = Self {
+        let services = Arc::new(Mutex::new(HashMap::new()));
+
+        let _service_map_updater = {
```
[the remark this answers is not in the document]
klochowicz: I prefer to have it for readability - *naming* the task

mobile/native/src/health.rs
klochowicz: I don't think using `select!` is wrong in any way - it's not less performant solution, and it's very clear what the code does....

mobile/native/src/health.rs
```diff
@@ -0,0 +1,65 @@
+use tokio::runtime::Runtime;
+use tokio::select;
+use tokio::sync::watch;
+
+use crate::event::EventInternal;
+use crate::event::{self};
+
+/// Services which status is monitored
```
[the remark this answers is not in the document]
klochowicz: I can, but is it better / preferred?

crates/tests-e2e/src/test_subscriber.rs
```diff
@@ -31,17 +36,20 @@ pub struct TestSubscriber {
     init_msg: watch::Receiver<Option<String>>,
     prices: watch::Receiver<Option<Prices>>,
     position_close: watch::Receiver<Option<ContractSymbol>>,
+    services: Arc<Mutex<HashMap<Service, ServiceStatus>>>,
+    _service_map_updater: tokio::task::JoinHandle<()>,
 }
 
 impl TestSubscriber {
-    pub fn new() -> (Self, ThreadSafeSenders) {
+    pub async fn new() -> (Self, ThreadSafeSenders) {
```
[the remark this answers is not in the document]
klochowicz: I need it, trust me. :-)

mobile/native/src/config/mod.rs
klochowicz: "like the others"? there is no getters for this struct. If I'm not adding anything to the behaviour, just adding getter for getter's sake is pointless. 

Putting the `coordinator_health_endpoint` here makes sense though, and that warrants a method certainly! thank you for your suggestion.

mobile/native/src/health.rs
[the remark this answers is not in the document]
luckysori: According to Grammarly the current version is incorrect:

image

crates/tests-e2e/src/test_subscriber.rs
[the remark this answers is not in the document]
luckysori: That's not how we normally review PRs, sorry.

mobile/native/src/config/mod.rs
luckysori: I suppose getter is not the correct terminology here. What I meant is that we have free functions to access the fields in the static `CONFIG` for the other fields of `ConfigInternal`.

mobile/native/src/health.rs
luckysori: You can just `await` the single future as we are not waiting for multiple concurrent branches, which is what `tokio::select` is for.

mobile/native/src/config/mod.rs
klochowicz: the reason I didn't want to go with that, is that I do not want to expose it to the frontend, and I actually have the config struct, so there's no reason to rely on a global object to retrieve it.

klochowicz: have added a dedicated getter though.

mobile/native/src/health.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Health status of the node
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum ServiceStatus {
+    #[default]
+    Unknown,
+    Online,
+    Offline,
+}
+
+pub type ServiceUpdate = (Service, ServiceStatus);
```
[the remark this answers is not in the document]
klochowicz: It actually had to become a struct, because although Dart can destruct a tuple without a problem, our event dispatch magic does not like this and I had to rework it :) 

will pull that part into this PR.

coordinator/src/routes.rs
klochowicz: The idea is to return an enum with a reason, but it's not in scope of this ticket.

I prefer keeping a string rather than nothing as it's more readable when you manually query it until we have a nicer response there.

mobile/native/src/health.rs
klochowicz: makes sense!

mobile/native/src/orderbook.rs
```diff
@@ -71,6 +78,10 @@ pub fn subscribe(secret_key: SecretKey, runtime: &Runtime) -> Result<()> {
             let mut stream =
                 spawn_blocking(move || orderbook_client::subscribe_with_authentication(url, authenticate)).await.expect("joined task not to panic");
 
+            if let Err(e) =
+                orderbook_status.send(ServiceStatus::Online) {
+                    tracing::warn!("Cannot update orderbook status: {e:#}");
+                };
```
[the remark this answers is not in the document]
klochowicz: 🤔I'm not sure that passing anything but that dedicated watch channel would be a good idea, this code ultimately
I could encapsulate it in a different API to swallow the errors, but we're not doing this anywhere else.

At the end of the day, we could might as well write `expect` as it should not fail, the receiver will not be dropped.

if we want to encapsulate the boilerplate, we could restore our old `log_error` (with a new `log_warn`) macros to avoid this annoying construct! 
thoughts?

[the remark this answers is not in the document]
luckysori: I mean, it already knows about the module because you have to manually set the status to `health::ServiceStatus::Online`.

My proposal was to unburden this module from having to choose types and just give it a method to call. The fact that we are using an enum for the states or a `tokio::sync::watch::Sender` is a bit of an implementation detail.

The error handling is annoying, but we don't have to encapsulate it in the same way. I'm not a huge fan of macros these days, but if you are very keen I think it would make sense to use it here.

mobile/native/src/health.rs
klochowicz: I've defaulted to 10s in production, 1s in tests (all adjustable with a flag passed into flutter).
also, 2s for local running, so we can spot downtimes more easily.

PR #961, opened 2023-07-19
Await `spawn_blocking`, don't just drop the `JoinHandle`

da-kami:
Spotted this when working on another `spawn_blocking` - I think this was just wrong 😬

### DIFF
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -145,7 +145,7 @@ async fn main() -> Result<()> {
                 match node_event_receiver.changed().await {
                     Ok(()) => {
                         let event = node_event_receiver.borrow().clone();
-                        node::routing_fees::handle(node.clone(), event);
+                        node::routing_fees::handle(node.clone(), event).await;
                     }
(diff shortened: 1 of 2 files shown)


luckysori (commented):
You don't need to `await` on the `JoinHandle` returned by `tokio::spawn` or `tokio::spawn_blocking`, unless you want to wait for the task to finish before continuing (or you care about the return type).

da-kami:
> You don't need to `await` on the `JoinHandle` returned by `tokio::spawn` or `tokio::spawn_blocking`, unless you want to wait for the task to finish before continuing (or you care about the return type).

Right - in this case we don't care I suppose - but what about the error handling...
If we don't await it, do we have to make it infallible (like it was implemented)?
Otherwise I don't think anybody would report this error ...? 🤔

da-kami:
Closing because what is in `main` is correct :)
We don't want to await in this case. 
I will amend https://github.com/get10101/10101/pull/958/files#diff-30c66b01d46e4f59ac847e965af472adb7a0a96644820f4645f87a799d2c331c

PR #964, opened 2023-07-20
Fix/cleanup create invoice service

da-kami:
This cleans up the service instantiation and will (eventually) allow us to spin up the UI with purely mocked services :)

@klochowicz were working on this, he will drive this forward and clean up `rust` calls in `main`. :)

### DIFF
--- a/mobile/lib/features/wallet/send_screen.dart
+++ b/mobile/lib/features/wallet/send_screen.dart
@@ -28,26 +28,25 @@ class SendScreen extends StatefulWidget {
 class _SendScreenState extends State<SendScreen> {
   final TextEditingController _textEditingController = TextEditingController();
   final _formKey = GlobalKey<FormState>();
-  final WalletService walletService = const WalletService();
   LightningInvoice? _lightningInvoice;
   bool isDecoding = false;
   bool decodingFailed = false;
-
-  final ChannelInfoService channelInfoService = const ChannelInfoService();
   ChannelInfo? channelInfo;
 
   @override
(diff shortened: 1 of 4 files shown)


mobile/lib/features/wallet/send_screen.dart
```diff
@@ -28,26 +28,25 @@ class SendScreen extends StatefulWidget {
...
+    initChannelValues(channelInfoService);
+    final WalletService walletService = context.read<WalletChangeNotifier>().service;
+    invoiceFromClipboard(walletService);
     super.initState();
   }
 
-  Future<void> initChannelValues() async {
+  Future<void> initChannelValues(ChannelInfoService channelInfoService) async {
     channelInfo = await channelInfoService.getChannelInfo();
   }
```
klochowicz, line 47: it might be easier in fact to not have this method, as line 46 gives more / better information 🤔

klochowicz (approved):
thanks for the cleanup!

mobile/lib/features/wallet/send_screen.dart
da-kami: Adding this method does not change any behavor but it makes me more aware that I am calling something `async`. This is more a stylistic pattern I'm using - it makes me more aware that I am calling `async` code from the `initState` method. As far as I know we cannot `await` in `initState`.
If I just put the call to the underlying service in there I won't see that the call is `async`. If you feel that's not useful feel free to remove it - I'll leave it for now.

PR #965, opened 2023-07-20
chore: Clean up main.dart

klochowicz:


mobile/lib/main.dart
```diff
@@ -424,3 +369,64 @@ Future<void> logAppSettings(bridge.Config config) async {
   String nodeId = rust.api.getNodeId();
   FLog.info(text: "Node ID: $nodeId");
 }
+
+/// Forward the events from change notifiers to the Event service
+void subscribeToNotifiers(BuildContext context) {
+  // TODO: Move this code into an "InitService" or similar; we should not have bridge code in the widget
```
da-kami, line 375: 🙃 I think this is a good start - happy to merge but I think we could also action this TODO as part of this.Shouldn't take long to do

mobile/lib/main.dart
```diff
@@ -424,3 +369,64 @@ Future<void> logAppSettings(bridge.Config config) async {
   String nodeId = rust.api.getNodeId();
   FLog.info(text: "Node ID: $nodeId");
 }
+
+/// Forward the events from change notifiers to the Event service
+void subscribeToNotifiers(BuildContext context) {
```
da-kami, line 374: I like that we pass the `BuildContext` here, that enforces that we can only call this one the build context is available. Makes it more clear when this is needed.

mobile/lib/main.dart
da-kami: Thoughts: I think it might be easiest to pass the service that instantiates this in the constructor of the `TenTenOneApp` widget. That way we can be sure it is initialized when we use it inside. Can also create it as part of the MultiProvider, but for me passing it in the constructor would actually be more intuitive to make the dependency clearer - and we will only use this service within `TenTenOneApp`.

da-kami (approved):

mobile/lib/main.dart
klochowicz: Thank you for the thoughts, I was planning to start with MultiProvider, but might end up going the way you described. I didn't want to spend more time than necessary to get this one closer, the TODO will need to wait for the follow-up.

Issue #932, opened 2023-07-10
Track trading revenue


bonomat:
We should track the revenue per trade/position, aka PnL.

da-kami:
I think the best way to keep track of this state is extending the `Position` state to properly handle a `Closed` state.

This means that we update the position to `Closed` once it is closed and save the `realized` PnL (and potentially the closing price). I would save the PnL based on the DLC state, we could do that in a periodic task (note: the closing price would have to be saved upon `Closing` state and not when updating to `Closed`).

---

Note: I assume that tracking is only concerned about the *realized* PnL for reporting.
The unrealized PnL depends on the current price. We could periodically update the unrealized PnL in the database as well - is this something we would like to have @bonomat ?

da-kami:
I assume that tracking is only concerned about the *realized* PnL for reporting.
The unrealized PnL depends on the current price. We could periodically update the unrealized PnL in the database as well - is this something we would like to have @bonomat ?

bonomat:
Mhm... I would like to know the _unrealized_ PnL as well because this could give us an overview of what's happening before it's too late (aka realized PnL). 

I'm not sure what the best approach is for this.

da-kami:
> Mhm... I would like to know the _unrealized_ PnL as well because this could give us an overview of what's happening before it's too late (aka realized PnL).
> 
> I'm not sure what the best approach is for this.

I already thought about it :)
I propose having two separate nullable fields in the `positions` table, one `realized_pnl` and one `unrealized_pnl`

The`realized_pnl` we update in the databse once we are in `Closed` state (before it is null in the database). The `unrealized_pnl` is set to null then.

The unrealized pnl we can periodically update in the database based on the current price. 
We have to be aware that the unrealized pnl is a snapshot of when the report was created. I think that's the best we can do.

Note: I would not depict the `unrealized_pnl` as part if the `Positions` model so far. The `unrealized_pnl` would only exist in the database (for reporting).
 
Theoretically we could also use the same `pnl` field for realized and unrealized and make it dependent on the state which one it is... - but I think such an "optimiztion" will cause more confusion than it helps :)

da-kami:
Note: I already started adding the `realized_pnl` and the `Closed` state. I will see to add the unrealized part in a separate PR on top once this part is done :)

Issue #941, opened 2023-07-12
Change `direction` in `position` table to reflect the `coordinator`'s position


bonomat:
Right now the field `direction` in the table `position` reflects the position of the trader and not of the coordinator. This should be changed .

da-kami:
I would be careful changing only the direction because it might make it even more confusing.
At the moment the `Position` on the coordinator reflects the trader's position, i.e. all the fields are the trader's (leverage, collateral, ...). If you change only the direction to be the coordinator's then the rest of the fields would still reflect the trader's position.

I would either keep the model as is (potentially making it more clear that the position in the coordinator *is* the trader's position), or change the model to reflect the values of *both* sides (i.e. trader and coordinator).

PR #967, opened 2023-07-20
Periodically update unrealized pnl for position in database

da-kami:
stacked on top of https://github.com/get10101/10101/pull/966

Note: I did not bother setting the `unrealized_pnl` to null when we close the position. Not sure that's actually necessary.

fixes: #932

### DIFF
--- a/coordinator/migrations/2023-07-20-055142_unrealized_pnl_positions/up.sql
+++ b/coordinator/migrations/2023-07-20-055142_unrealized_pnl_positions/up.sql
@@ -1,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE
+    positions
+ADD
+    COLUMN "unrealized_pnl" BIGINT;
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -170,7 +170,9 @@ async fn main() -> Result<()> {
             loop {
                 tokio::time::sleep(UNREALIZED_PNL_SYNC_INTERVAL).await;
                 if let Err(e) = unrealized_pnl::sync(node.clone()).await {
-                    tracing::error!("Failed to sync closed DLCs with positions in database: {e:#}");
+                    tracing::error!(
+                        "Failed to sync unrealized PnL with positions in database: {e:#}"
+                    );
                 }
             }
         }
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -9,6 +9,7 @@ use coordinator::node::closed_positions;
 use coordinator::node::connection;
 use coordinator::node::expired_positions;
 use coordinator::node::storage::NodeStorage;
+use coordinator::node::unrealized_pnl;
 use coordinator::node::Node;
 use coordinator::routes::router;
 use coordinator::run_migration;
@@ -34,6 +35,7 @@ const PROCESS_PROMETHEUS_METRICS: Duration = Duration::from_secs(10);
 const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_secs(5);
 const EXPIRED_POSITION_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 const CLOSED_POSITION_SYNC_INTERVAL: Duration = Duration::from_secs(30);
+const UNREALIZED_PNL_SYNC_INTERVAL: Duration = Duration::from_secs(600);
 const CONNECTION_CHECK_INTERVAL: Duration = Duration::from_secs(30);
 
 #[tokio::main]
@@ -158,6 +160,18 @@ async fn main() -> Result<()> {
                     .await
                     .expect("To spawn blocking thread");
                 tokio::time::sleep(PROCESS_PROMETHEUS_METRICS).await;
+            }
+        }
+    });
+
+    tokio::spawn({
+        let node = node.clone();
+        async move {
+            loop {
+                tokio::time::sleep(UNREALIZED_PNL_SYNC_INTERVAL).await;
+                if let Err(e) = unrealized_pnl::sync(node.clone()).await {
+                    tracing::error!("Failed to sync closed DLCs with positions in database: {e:#}");
+                }
             }
         }
     });
--- a/coordinator/src/node/expired_positions.rs
+++ b/coordinator/src/node/expired_positions.rs
@@ -57,10 +57,7 @@ pub async fn close(node: Node) {
         };
 
         let closing_price = match BitmexClient::get_quote(&position.expiry_timestamp).await {
-            Ok(quote) => match position.direction {
-                trade::Direction::Long => quote.bid_price,
-                trade::Direction::Short => quote.ask_price,
-            },
+            Ok(quote) => quote.get_price_for_direction(position.direction.opposite()),
             Err(e) => {
                 tracing::warn!(
                     "Failed to get quote from bitmex for {} at {}. Error: {e:?}",
--- a/coordinator/src/node/unrealized_pnl.rs
+++ b/coordinator/src/node/unrealized_pnl.rs
@@ -1,0 +1,68 @@
+use crate::db;
+use crate::node::Node;
+use crate::position::models::Position;
+use anyhow::Context;
+use anyhow::Result;
+use diesel::r2d2::ConnectionManager;
+use diesel::r2d2::PooledConnection;
+use diesel::PgConnection;
+use rust_decimal::Decimal;
+use time::OffsetDateTime;
+use trade::bitmex_client::BitmexClient;
+use trade::bitmex_client::Quote;
+use trade::cfd::calculate_pnl;
+use trade::Direction;
+
+pub async fn sync(node: Node) -> Result<()> {
+    let mut conn = node.pool.get()?;
+
+    let positions = db::positions::Position::get_all_open_or_closing_positions(&mut conn)?;
+    let current_quote = BitmexClient::get_quote(&OffsetDateTime::now_utc())
+        .await
+        .context("Failed to fetch quote from BitMEX")?;
+
+    for position in positions.iter() {
+        if let Err(e) = sync_position(&mut conn, position, current_quote.clone()) {
+            tracing::error!(position_id=%position.id, ?current_quote, "Failed to update position's unrealized pnl in database: {e:#}")
+        }
+    }
+
+    Ok(())
+}
+
+fn sync_position(
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    position: &Position,
+    quote: Quote,
+) -> Result<()> {
+    let current_price = match position.direction {
+        trade::Direction::Long => quote.bid_price,
+        trade::Direction::Short => quote.ask_price,
+    };
+
+    let average_entry_price = Decimal::try_from(position.average_entry_price)
+        .context("Failed to convert average entry price to Decimal")?;
+
+    let (long_leverage, short_leverage) = match position.direction {
+        Direction::Long => (position.leverage, 1.0_f32),
+        Direction::Short => (1.0_f32, position.leverage),
+    };
+
+    // the position in the database is the trader's position, our direction is opposite
+    let direction = position.direction.opposite();
+
+    let pnl = calculate_pnl(
+        average_entry_price,
+        current_price,
+        position.quantity,
+        long_leverage,
+        short_leverage,
+        direction,
+    )
+    .context("Failed to calculate pnl for position")?;
+
+    db::positions::Position::update_unrealized_pnl(conn, position.id, pnl)
+        .context("Failed to update unrealized pnl in db")?;
+
+    Ok(())
+}
--- a/coordinator/src/node/unrealized_pnl.rs
+++ b/coordinator/src/node/unrealized_pnl.rs
@@ -35,9 +35,14 @@ fn sync_position(
     position: &Position,
     quote: Quote,
 ) -> Result<()> {
-    let current_price = match position.direction {
-        trade::Direction::Long => quote.bid_price,
-        trade::Direction::Short => quote.ask_price,
+    let closing_price = match position.closing_price {
+        None => match position.direction {
+            trade::Direction::Long => quote.bid_price,
+            trade::Direction::Short => quote.ask_price,
+        },
+        Some(closing_price) => {
+            Decimal::try_from(closing_price).expect("f32 closing price to fit into decimal")
+        }
     };
 
     let average_entry_price = Decimal::try_from(position.average_entry_price)
@@ -53,7 +58,7 @@ fn sync_position(
 
     let pnl = calculate_pnl(
         average_entry_price,
-        current_price,
+        closing_price,
         position.quantity,
         long_leverage,
         short_leverage,
--- a/coordinator/src/node/unrealized_pnl.rs
+++ b/coordinator/src/node/unrealized_pnl.rs
@@ -36,10 +36,7 @@ fn sync_position(
     quote: Quote,
 ) -> Result<()> {
     let closing_price = match position.closing_price {
-        None => match position.direction {
-            trade::Direction::Long => quote.bid_price,
-            trade::Direction::Short => quote.ask_price,
-        },
+        None => quote.get_price_for_direction(position.direction.opposite()),
         Some(closing_price) => {
             Decimal::try_from(closing_price).expect("f32 closing price to fit into decimal")
         }
--- a/coordinator/src/node/unrealized_pnl.rs
+++ b/coordinator/src/node/unrealized_pnl.rs
@@ -6,12 +6,9 @@ use anyhow::Result;
 use diesel::r2d2::ConnectionManager;
 use diesel::r2d2::PooledConnection;
 use diesel::PgConnection;
-use rust_decimal::Decimal;
 use time::OffsetDateTime;
 use trade::bitmex_client::BitmexClient;
 use trade::bitmex_client::Quote;
-use trade::cfd::calculate_pnl;
-use trade::Direction;
 
 pub async fn sync(node: Node) -> Result<()> {
     let mut conn = node.pool.get()?;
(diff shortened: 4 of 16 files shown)


coordinator/src/node/unrealized_pnl.rs
```diff
@@ -0,0 +1,68 @@
...
+use trade::bitmex_client::BitmexClient;
+use trade::bitmex_client::Quote;
+use trade::cfd::calculate_pnl;
+use trade::Direction;
+
+pub async fn sync(node: Node) -> Result<()> {
+    let mut conn = node.pool.get()?;
+
+    let positions = db::positions::Position::get_all_open_or_closing_positions(&mut conn)?;
+    let current_quote = BitmexClient::get_quote(&OffsetDateTime::now_utc())
```
holzeis, line 20: 🔧 We shouldn't fetch new quotes for a position that is state Closing. The execution price should already be set.

coordinator/src/bin/coordinator.rs
```diff
@@ -34,6 +34,7 @@ const PROCESS_PROMETHEUS_METRICS: Duration = Duration::from_secs(10);
 const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_secs(5);
 const EXPIRED_POSITION_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 const CLOSED_POSITION_SYNC_INTERVAL: Duration = Duration::from_secs(30);
+const UNREALIZED_PNL_SYNC_INTERVAL: Duration = Duration::from_secs(60);
```
holzeis, line 37: 🤔 Every minute is quite often isn't it? How about every 10 minutes?

coordinator/src/bin/coordinator.rs
```diff
@@ -162,6 +163,18 @@ async fn main() -> Result<()> {
         }
     });
 
+    tokio::spawn({
+        let node = node.clone();
+        async move {
+            loop {
+                tokio::time::sleep(UNREALIZED_PNL_SYNC_INTERVAL).await;
+                if let Err(e) = closed_positions::sync(node.clone()) {
```
holzeis, line 171: 🤔 Is that a copy paste error? Shouldn't it call `unrealized_pnl::sync` here?

da-kami: 😅 omg, good catch!

coordinator/src/bin/coordinator.rs
da-kami: I guess it depends how accurate we want a report that takes the unrealized pnl into account should be.
I'm happy to put 10 mins :)

coordinator/src/node/unrealized_pnl.rs
da-kami: well... yes and no. The execution price is set, but we only save the *actual* pnl that was collected when we have a `ClosedContract` (based on said contract).
There is no guarantee that a position is `Closing` will actually be closed with that execution price, so I'm not sure it's worth jumping through hoops of setting the execution price and using that. 

I can do it if we really feel it's necessary :)

holzeis: > There is no guarantee that a position in Closing will actually be closed with that execution price.

Not sure I am following. 

Are you referring to a technical reason why we can't guarantee the the execution price?

A match has already been found so there should be no other reason on why we couldn't guarantee the price. Am I missing something?

da-kami: > Are you referring to a technical reason why we can't guarantee the the execution price?

There are multiple reasons why it could fail, I am not sure you can classify all of them as technical. A scenario where e.g. a lot of trades happen at the same time with the same maker could lead liquidity problems with the maker which could lead to execution failure. Same for liquidity on the coordinator.

> A match has already been found so there should be no other reason on why we couldn't guarantee the price. Am I missing something?

Example: We found a match, the app contacts the coordinator but the maker is down / cannot fulfill the trade. The execution just can't get guaranteed. There are other (technical) reasons why it could fail, e.g. the app being offline for some time because the user lost network. 

However, after some thought I think saving the execution price is the right way to go because we will need it for sanity/auditing/validation. We have to make sure that we also remove it in case execution failed. (not part of this PR)
I will see to add saving the execution price with the position once known and will add a separate code path for that reports the `unrealized_pnl` based on the `execution_price` if the `execution_price` is known.

da-kami: Might be good to read https://github.com/get10101/10101/pull/966#issuecomment-1644850594 as well :)

holzeis (approved):

coordinator/src/bin/coordinator.rs
```diff
@@ -162,6 +164,18 @@ async fn main() -> Result<()> {
         }
     });
 
+    tokio::spawn({
+        let node = node.clone();
+        async move {
+            loop {
+                tokio::time::sleep(UNREALIZED_PNL_SYNC_INTERVAL).await;
+                if let Err(e) = unrealized_pnl::sync(node.clone()).await {
+                    tracing::error!("Failed to sync closed DLCs with positions in database: {e:#}");
```
luckysori, line 173: 🔧 This log is the same as the one on line 195.

coordinator/src/node/unrealized_pnl.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    let average_entry_price = Decimal::try_from(position.average_entry_price)
+        .context("Failed to convert average entry price to Decimal")?;
+
+    let (long_leverage, short_leverage) = match position.direction {
+        Direction::Long => (position.leverage, 1.0_f32),
+        Direction::Short => (1.0_f32, position.leverage),
+    };
+
+    // the position in the database is the trader's position, our direction is opposite
```
luckysori, line 56: FML haha.

coordinator/src/node/unrealized_pnl.rs
```diff
@@ -0,0 +1,68 @@
...
+        trade::Direction::Short => quote.ask_price,
+    };
+
+    let average_entry_price = Decimal::try_from(position.average_entry_price)
+        .context("Failed to convert average entry price to Decimal")?;
+
+    let (long_leverage, short_leverage) = match position.direction {
+        Direction::Long => (position.leverage, 1.0_f32),
+        Direction::Short => (1.0_f32, position.leverage),
+    };
```
luckysori, line 54: It's a bit scary that this kind of logic is located in this module. It's probably hard to define where this should be located atm. Maybe we can leave a comment talking about how the position is the trader's position and the coordinator is always leverage 1.

coordinator/src/node/unrealized_pnl.rs
```diff
@@ -0,0 +1,68 @@
...
+
+fn sync_position(
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    position: &Position,
+    quote: Quote,
+) -> Result<()> {
+    let current_price = match position.direction {
+        trade::Direction::Long => quote.bid_price,
+        trade::Direction::Short => quote.ask_price,
+    };
```
luckysori: I can't reason about this. I assume the `direction` corresponds to the trader's direction. Given that this is the coordinator's perspective, does this affect which price it is. I would welcome a comment here.

coordinator/src/node/unrealized_pnl.rs
```diff
@@ -0,0 +1,73 @@
...
+    for position in positions.iter() {
+        if let Err(e) = sync_position(&mut conn, position, current_quote.clone()) {
+            tracing::error!(position_id=%position.id, ?current_quote, "Failed to update position's unrealized pnl in database: {e:#}")
+        }
+    }
+
+    Ok(())
+}
+
+fn sync_position(
```
luckysori, line 33: 🔧 Given my comments below, I think one or two tests would be really helpful, although better if they don't involve the DB at all. Maybe we could move the pure function that calculates the coordinator's PNL based on the (trader's) `Position` to `trade::cfd`?

luckysori (approved):

coordinator/src/node/unrealized_pnl.rs
da-kami: Since it is the price that the trader will get, we have to use that price. We cannot argue from the side of the coordinator here, otherwise we would try to close the position with two different prices.
I'll see to extract a function that we can use throughout the code - it's not the first time this code triggers a lengthy thinking process :)

coordinator/src/node/unrealized_pnl.rs
da-kami: I did a refactoring where I moved the `coordinator_pnl` onto the `Position` struct and added tests.
This particular bit remains within that pnl calculation function. Eventually we might also want to extract this.

Instead of using `1.0` I opted for using the `COORDINATOR_LEVERAGE` constant that I made public in the coordinator's `node` module.

coordinator/src/node/unrealized_pnl.rs
da-kami: see https://github.com/get10101/10101/pull/967#discussion_r1271774521

I think it's better to keep the `cfd` module unaware of the "coordinator". 
I hope the tests are not too confusing - I was confused for a bit until I went back to our famous CFD hedging sheet to check the numbers...

coordinator/src/bin/coordinator.rs
da-kami: was changed to `600` (i.e. 10 mins)

coordinator/src/node/unrealized_pnl.rs
bonomat: This is related: https://github.com/get10101/10101/issues/941

coordinator/migrations/2023-07-20-055142_unrealized_pnl_positions/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE
+    positions
+ADD
+    COLUMN "unrealized_pnl" BIGINT;
```
bonomat, line 5: It might be good to indicate that this is in sats and not percent (if it is).

coordinator/src/position/models.rs
```diff
@@ -49,4 +56,198 @@ pub struct Position {
     /// This field is optional for backwards compatibility because we cannot deterministically
     /// associate already existing contracts with positions.
     pub temporary_contract_id: Option<ContractId>,
+    pub closing_price: Option<f32>,
+}
+
+impl Position {
+    /// Calculates the profit and loss for the coordinator
```
bonomat, line 63: It's my understanding that this function returns the PnL in satoshi. 

```suggestion
    /// Calculates the profit and loss for the coordinator in satoshi
```

coordinator/src/position/models.rs
```diff
@@ -49,4 +56,198 @@ pub struct Position {
...
+            .with_leverage(2.0)
+            .with_quantity(20000.0)
+            .with_average_entry_price(20000.0)
+            .with_direction(Direction::Long);
+
+        let quote = dummy_quote(22000, 0);
+
+        let coordinator_pnl = position.calculate_coordinator_pnl(quote).unwrap();
+
+        assert_eq!(coordinator_pnl, -9_090_909);
```
bonomat, line 147: It's hard for me to follow these amounts :(

bonomat: Ok, it took me a while to understand what is happening and why this value is correct: 

In my words expressed: In this example, the person who went long, bought $20,000 worth of BTC at the price of 20,000, i.e. `1` BTC
At the price of $22,000 he sells $20,000 worth of BTC, i.e. he sells it for `0.909090909` BTC. 
The difference is his profit, i.e. 

`1 BTC` - `0.909090909 BTC` = `0.09090909 BTC` profit

coordinator/src/position/models.rs
```diff
@@ -49,4 +56,198 @@ pub struct Position {
...
+        let quote = dummy_quote(0, 1000);
+
+        let coordinator_pnl = position.calculate_coordinator_pnl(quote).unwrap();
+
+        assert_eq!(coordinator_pnl, 0);
+    }
+
+    /// See also: `given_long_position_when_price_10_pc_up_then_18pc_profit` test in `trade::cfd`
+    #[test]
+    fn given_trader_long_position_when_bid_price_10pc_up_then_coordinator_18pc_loss() {
```
bonomat, line 136: It would be good to also assert the pnl in percent given you mention it in the test function name.

coordinator/src/node/expired_positions.rs
```diff
@@ -57,10 +58,7 @@ pub async fn close(node: Node) {
         };
 
         let closing_price = match BitmexClient::get_quote(&position.expiry_timestamp).await {
```
bonomat, line 60: I know it's not your code, but this price should not be from bitmex but from our orderbook.

coordinator/src/node/expired_positions.rs
```diff
@@ -57,10 +58,7 @@ pub async fn close(node: Node) {
...
-            Ok(quote) => match position.direction {
-                trade::Direction::Long => quote.bid_price,
-                trade::Direction::Short => quote.ask_price,
-            },
+            Ok(quote) => closing_price(position.direction, quote),
```
bonomat, line 61: The `closing_price` is not wrong but for me it adds to the confusion in this whole story and I had to do some mental gymnastics to follow what is happening here 🙈 

What about we do something like this: 

```rust
// 1. get counter direction, i.e. he was long, now he needs to go short, or he was short, now needs to go long
let direction = position.direction.opposite();

// 2. TODO: get latest quote from the orderbook/counter party
let quote = BitmexClient::get_quote(&position.expiry_timestamp)
    .await
    .unwrap();

// 3. get closing price based on order direction
let closing_price = match direction {
    Direction::Long => quote.ask_price,
    Direction::Short => quote.bid_price,
};

```

coordinator/src/position/models.rs
da-kami: I agree with both your statements: It is hard to follow the amounts and it's hard to understand why it's correct.
Any idea how to make it better? 😬

I tried to come up with better numbers, but I feel the numbers are not the problem but more the model and the way we structure and express things. I could not come up with something better so far though. We already had a hard time with such kind of tests in ItchySats as far as I remember.

crates/trade/src/cfd.rs
```diff
@@ -104,9 +105,24 @@ pub fn calculate_pnl(
...
 }
 
+/// The closing price that the trader gets based on bid and ask price
+///
+/// The closing price is always for the trade because the trader requests to close a position.
+/// If the trader closes a long position then the trader gets the best bid price because the trader
+/// is effectively going short to close a long order.
+/// If the trader closes a short position then the trader gets the best ask price because the trader
+/// is effectively going long to close a short order.
+pub fn closing_price(trader_direction: Direction, quote: Quote) -> Decimal {
```
bonomat, line 115: What do you think of implementing this on `Quote`?

coordinator/src/position/models.rs
da-kami: That is a good point, maybe *that* would help understanding it. It's still hard for me to grasp *why* it is 18.x percent though 😅

crates/trade/src/cfd.rs
bonomat: I would then not call it `closing_price` but `price_for_direction` or so.

coordinator/src/position/models.rs
bonomat: It's definitely not intuitive, I agree. I assume you saw already this comment: https://github.com/get10101/10101/pull/967#discussion_r1272132643 which explains the 18%

Here is a simple calculator to play around with number if you haven't found it already: https://btctools.io/calculators/profit-loss

crates/trade/src/cfd.rs
da-kami: It's hard to fit this into my mental model, but I can see how it could make the code easier :)

For me the `Quote` does not care about any underlying financial instrument, it just depicts the best bid and ask price in the market (weighted by volume).
For me "closing" is always related to the position, not to the quote - the quote does not care about closing.
Given that the quote defines the closing price I can see how it could somehow make sense to add it there, but if I would just look at the quote out of context I would be confused.

da-kami: Maybe you can explain your mental model - it might help to make all this code better :)

coordinator/src/position/models.rs
bonomat: I don't know... Maybe add the calculation from my comment and add asserts for that? 

Eitherway... if we have to touch this again, we very likely will need to go through this thought process again 🙈

crates/trade/src/cfd.rs
bonomat: >  For me "closing" is always related to the position, not to the quote - the quote does not care about closing.

Exactly, that's why I want to get rid of the term `closing` :D 

I would add something like: 

```
impl Quote {
    fn get_price_for_direction(&self, direction: Direction) -> Decimal {
        match direction {
            Direction::Long => self.ask_price,
            Direction::Short => self.bid_price,
        }
    }
}
```

and then from the outside, i.e. where you currently call `closing_price` you do: 
```
get_price_for_direction(position.direction.oposite())`
```

Then we know, we want the price for the opposite direction.

coordinator/src/node/expired_positions.rs
da-kami: I see where you are going, but for me using a price from the orderbook only makes sense if there is also an order that is eventually executed. We don't do this at the moment.
The coordinator can still use an orderbook price and just simulate the match - but I find this a bit odd. 
I opened https://github.com/get10101/10101/issues/988 to capture this; feel free to change the description

coordinator/src/position/models.rs
luckysori: I still don't get something. If the coordinator puts up 1 bitcoin (1 leverage) and they lose ~9M sats, isn't that a ~9% loss?

I think the 18 comes from considering the app side, which with leverage 2 doubles their ROE w.r.t leverage 1.

luckysori (approved):
Thank you, I like it!

coordinator/src/position/models.rs
da-kami: That is correct, I was thinking from the taker's point of view. I'll change it to reflect the coordintor's pnl percentage.

coordinator/src/position/models.rs
luckysori: Done [here](https://github.com/get10101/10101/pull/994/commits/581150bd641fa65ea2487d896e7edacf9f0451ab).

PR #970, opened 2023-07-21
Fix outdated schema and correctly handle temporary contract id

da-kami:
Unfortunately https://github.com/get10101/10101/pull/966 merged with an outdated `schema.rs` version in the coordinator. 
This PR fixes this!


luckysori (approved):
I guess we can make this mandatory once all positions have a `temporary_contract_id` set in production.

holzeis (approved):

Issue #883, opened 2023-07-03
Skipping direct route to payee when making payments from app to coordinator


holzeis:
We have seen this problem when testing the app on the public regtest setup.

For instance, when trying to pay the channel-opening fee from app to coordinator this can happen, preventing the app from paying for these costs.

```
| 2023-07-26 18:06:55.383 | LogLevel.DEBUG | Trying to pay channel opening fees of 153  
| 2023-07-26 18:06:55.437 | LogLevel.TRACE | lightning::routing::router: Searching for a route from payer 02a6fec7dfc1c99ec359fb4ee1f413d3de26016fc709ba100ba06daabce2be42cd to payee 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802 with MPP and 1 first hops overriding the network graph  
| 2023-07-26 18:06:55.437 | LogLevel.TRACE | lightning::routing::router: Building path from 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802 (payee) to 02a6fec7dfc1c99ec359fb4ee1f413d3de26016fc709ba100ba06daabce2be42cd (us/payer) for value 153000 msat.  
| 2023-07-26 18:06:55.438 | LogLevel.TRACE | lightning::routing::router: Skipped direct route to payeeSkipped direct route to payee via SCID 1099653251074  
| 2023-07-26 18:06:55.438 | LogLevel.TRACE | lightning::routing::router: Starting main path collection loop with 5 nodes pre-filled from first/last hops.  
| 2023-07-26 18:06:55.439 | LogLevel.TRACE | lightning::routing::router: Skipped direct route to payee via SCID 1099653251074  
| 2023-07-26 18:06:55.440 | LogLevel.TRACE | lightning::routing::router: Starting main path collection loop with 5 nodes pre-filled from first/last hops.  
| 2023-07-26 18:06:55.440 | LogLevel.TRACE | lightning::routing::router: Have now collected 0 msat (seeking 459000 msat) in paths. Last path loop did not find a new path.  
| 2023-07-26 18:06:55.440 | LogLevel.ERROR | Failed to send payment err: RouteNotFound 
[...]

PR #978, opened 2023-07-21
fix: Align funding tx fee invoice parameters with endpoint

holzeis:
The parameters were out of sync resulting in an error which was tried to be parsed as invoice. Which obviously failed.

The `basic.rs` was unfortunately not able to catch that due to a workaround for https://github.com/get10101/10101/issues/883


bonomat (approved):
LGTM. Thanks for fixing

PR #979, opened 2023-07-22
Fail intercepted HTLC backwards when we should

### LINKED ISSUE #943: Fail HTLC backwards in all scenarios when handling `HTLCIntercepted` event
We should consider handling all errors in a single spot so that we can consistently fail the HTLC backwards.

We have recently run into not failing the HTLC backwards in production, which leads to users having to wait a long time for their funds to return to their funding wallets.

### LINKED ISSUE #944: Fail intercepted HTLC backwards if peer disconnects before channel is open
Otherwise we get a pending payment on the funding wallet, which can be really annoying for users and may prevent them from trying to fund the 10101 wallet again.

We have several options when handling the corresponding `ChannelClosed` event. In all cases we need to find the corresponding pending HTLC in our `PendingInterceptedHtlcs` map, indexed by peer ID:

1. Use the `channel_id` to look for the peer ID in the `ChannelDetails` in the `ChannelManager`. This is probably bad because we are no sure if the `ChannelDetails` will consistently be there every time we handle a `ChannelClosed` event.
2. Patch LDK to include the peer ID in the `ChannelClosed` event. There's already an issue for this open in the `rust-lightning` repo.
3. Introduce a shadow representation of the Lightning channel which is persisted before the channel is actually open an remains persisted after it is closed. We want this for other reasons anyway.

luckysori:
Fixes #943.
Fixes #944.

---

This is yet another instance where we have to consider the fact that the same event handler code runs for both app and coordinator even though certain parts need to be specific to each role. I've increased the implicit annoyance counter of https://github.com/get10101/10101/issues/885.

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -31,6 +31,7 @@ use lightning::chain::chaininterface::ConfirmationTarget;
 use lightning::chain::chaininterface::FeeEstimator;
 use lightning::chain::keysinterface::SpendableOutputDescriptor;
 use lightning::ln::channelmanager::InterceptId;
+use lightning::ln::PaymentHash;
 use lightning::routing::gossip::NodeId;
 use lightning::util::config::UserConfig;
 use lightning::util::events::Event;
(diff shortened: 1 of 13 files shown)


crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -458,8 +458,20 @@ where
...
                         self.storage.upsert_channel(channel)?;
+
+                        // Fail intercepted HTLC which was meant to be used to open the JIT channel,
+                        // in case it was still pending
+                        if let Some((intercept_id, _)) =
+                            self.pending_intercepted_htlcs_lock().get(&counterparty)
+                        {
+                            // If this fails it's either because the intercepted HTLC was already
+                            // failed or already claimed
+                            let _ = self.channel_manager.fail_intercepted_htlc(*intercept_id);
```
holzeis: 🤔 Shouldn't we at least log the error if we get one?

holzeis (approved):
Very nice! It would be great if we could add or extend a test for that.

e.g. something similar like that `open_jit_channel_with_disconnected_payee()` with a disconnect after the `Event::FundingGenerationReady`. We could provide an external eventhandler in the tests that calls the disconnect upon that event.

klochowicz (approved):
LGTM! 👍

crates/ln-dlc-node/src/ln/event_handler.rs
luckysori: I would prefer not to. There is nothing we can do with this information and the HTLC will have already been handled properly anyway.

luckysori:
> Very nice! It would be great if we could add or extend a test for that.
> 
> e.g. something similar like that `open_jit_channel_with_disconnected_payee()` with a disconnect after the `Event::FundingGenerationReady`. We could provide an external eventhandler in the tests that calls the disconnect upon that event.

I've added a couple of tests with https://github.com/get10101/10101/pull/979/commits/bb6c7972abb4f387dec3102c957a6a4b8d7c76d6. I would normally try to include these in the original patches, but the interactive rebase is a bit annoying and I don't want to squash everything in between.

holzeis:
> > Very nice! It would be great if we could add or extend a test for that.
> > e.g. something similar like that `open_jit_channel_with_disconnected_payee()` with a disconnect after the `Event::FundingGenerationReady`. We could provide an external eventhandler in the tests that calls the disconnect upon that event.
> 
> I've added a couple of tests with [bb6c797](https://github.com/get10101/10101/commit/bb6c7972abb4f387dec3102c957a6a4b8d7c76d6). I would normally try to include these in the original patches, but the interactive rebase is a bit annoying and I don't want to squash everything in between.

Nice, thank you. Would it be possible to add a test case where we triggered the channel creation already, but the app disconnects in between? We might be able to do that by listening to the `Event:FundingGeneratedEvent` from the tests and disconnect the app from the coordinator then. So the channel should not become ready but closed. (I guess)

luckysori:
> > > Very nice! It would be great if we could add or extend a test for that.
> > > e.g. something similar like that `open_jit_channel_with_disconnected_payee()` with a disconnect after the `Event::FundingGenerationReady`. We could provide an external eventhandler in the tests that calls the disconnect upon that event.
> > 
> > 
> > I've added a couple of tests with [bb6c797](https://github.com/get10101/10101/commit/bb6c7972abb4f387dec3102c957a6a4b8d7c76d6). I would normally try to include these in the original patches, but the interactive rebase is a bit annoying and I don't want to squash everything in between.
> 
> Nice, thank you. Would it be possible to add a test case where we triggered the channel creation already, but the app disconnects in between? We might be able to do that by listening to the `Event:FundingGeneratedEvent` from the tests and disconnect the app from the coordinator then. So the channel should not become ready but closed. (I guess)

Added another test using the trick you described! Merging now.

luckysori:
bors retry

PR #993, opened 2023-07-24
Replace `run-local-android` recipe's LOCAL_IP variable

### LINKED ISSUE #989: Run-local-android doesn't work on linux ubuntu
As mentioned in #901, on my machine, `just run-local-android` outputs

```
Target file "--dart-define=ESPLORA_ENDPOINT=http://10.5.0.1
172.20.0.1
172.17.0.1
192.168.100.157
192.168.122.1:3000" not found.
```

This is because the `grep` commands for `LOCAL_IP` output a whole list of local IPs, some of which are internal, broadcast addresses, or docker addresses.

Restioson:
This uses a combination of answers in https://stackoverflow.com/a/25851186 and https://stackoverflow.com/a/49552792.

Fixes #989 (for me). @da-kami, can you confirm that this still works on OS X?

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,6 +12,14 @@ public_regtest_coordinator := "03507b924dae6595cfb78492489978127c5f1e3877848564d
 public_regtest_esplora := "http://35.189.57.114:3000"
 public_coordinator_http_port := "80"
 
+# command to get the local IP of this machine
+get_local_ip := if os() == "linux" {
+ "ip -o route get to 1 | sed -n 's/.*src \\([0-9.]\\+\\).*/\\1/p'"
+} else {
+ "ifconfig | grep -Eo 'inet (addr:)?([0-9]*\\.){3}[0-9]*' | grep -Eo '([0-9]*\\.){3}[0-9]*' | grep -v '127.0.0.1'"
+}
+
+
 default: gen
 precommit: gen lint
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -80,7 +80,7 @@ run-regtest args="":
 
 run-local-android args="":
     #!/usr/bin/env bash
-    LOCAL_IP=$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1')
+    LOCAL_IP=$(ip -o route get to 1 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')
```
bonomat, line 83: `ip` is unfortunately not installed on a mac :( 
If this works for you, maybe we do a check for the operating system and execute `ip` for `linux` and `ifconfig` for `mac`?

bonomat: E.g. with this it should be relatively easy: https://github.com/casey/just#enabling-and-disabling-recipes

klochowicz: the simplest way to check in a bash script would be checking by calling `uname` from a script. on macs, it returns `Darwin`, so you can match that for mac branch.

Restioson: `just` has a function for this https://github.com/casey/just#system-information

Restioson:
Oh no, I've cursed this branch

da-kami (approved):
works on my machine!

PR #994, opened 2023-07-25
Some cleanup and alignment

da-kami:
A `Quote` is basically a more detailed `Price`; the `Quote` uses the `Price`'s `get_price_for_direction` logic to get the price by direction.

---

@luckysori  comments from https://github.com/get10101/10101/pull/967 should all be worked in

### DIFF
--- a/coordinator/src/position/models.rs
+++ b/coordinator/src/position/models.rs
@@ -31,6 +31,11 @@ pub enum PositionState {
     Closed { pnl: i64 },
 }
 
+/// A trader's position
+///
+/// The position acts as an aggregate of one contract of one user.
+/// The position represents the values of the trader; i.e. the leverage, collateral and direction
+/// are stored from the trader's perspective and not the coordinator's.
 #[derive(Debug)]
 pub struct Position {
     pub id: i32,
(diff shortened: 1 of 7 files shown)


luckysori (approved):
LGTM.

coordinator/src/position/models.rs
```diff
@@ -27,8 +27,16 @@ pub struct NewPosition {
...
 pub enum PositionState {
     Open,
-    Closing,
-    Closed { pnl: i64 },
+    /// The position is in the process of being closed
+    ///
+    /// Once the position is being close the closing price is known.
+    /// For backwards compatibility we make the closing price optional.
+    Closing {
+        closing_price: f32,
```
luckysori, line 35: But it is _not_ optional here.

luckysori (approved):
My first review didn't consider the whole diff for some reason. Still approved though. Thank you!

coordinator/src/position/models.rs
da-kami: Thanks for catching this one! I eventually opted for just defaulting it to `0` when mapping it into the business logic model instead of making it optional there. It is optional in the db model :)

da-kami:
I am merging this because it might be necessary to avoid backwards compatibility issues where the `pnl` is not known in the database when trying to `Close` positions after an upgrade.

PR #995, opened 2023-07-25
bug - cannot fund wallet after open/close position

klochowicz:
After trying to add balance to the wallet after closing a position, it says that
it cannot find a route from the faucet to the app.

This should not be the case.

### DIFF
--- a/crates/tests-e2e/tests/close_position.rs
+++ b/crates/tests-e2e/tests/close_position.rs
@@ -1,5 +1,6 @@
 use native::api::{self};
 use native::trade::position::PositionState;
+use tests_e2e::fund::pay_with_faucet;
 use tests_e2e::setup;
 use tests_e2e::setup::dummy_order;
 use tests_e2e::wait_until;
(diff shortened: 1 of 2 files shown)


holzeis:
@klochowicz Not being able to find a route, does not necessarily have to be a bug. Can you describe how you got to the `no route` error?

crates/tests-e2e/tests/close_position.rs
```diff
@@ -24,4 +25,24 @@ async fn can_collab_close_position() {
...
+
+    // Ensure we sync the wallet info after funding
+    spawn_blocking(move || api::refresh_wallet_info().expect("to succeed"))
+        .await
+        .unwrap();
+
+    let balance_at_closing = test.app.rx.wallet_info().unwrap().balances.on_chain;
+
+    let invoice =
+        spawn_blocking(move || api::create_invoice_with_amount(50_000).expect("to succeed"))
```
bonomat, line 39: I believe the issue is that we initiate the channel with 100k sats, 50k sats go to the app and 50k sats remain at the coordinator. 
This means, you can only receive `50k sats - channel_reserve`.

crates/tests-e2e/tests/close_position.rs
```diff
@@ -24,4 +25,24 @@ async fn can_collab_close_position() {
...
+            .await
+            .unwrap();
+    api::decode_invoice(invoice.clone()).expect("to decode invoice we created");
+
+    let client = reqwest::Client::new();
+    pay_with_faucet(&client, invoice).await.unwrap();
+
+    wait_until!(test.app.rx.wallet_info().unwrap().balances.on_chain > balance_at_closing);
```
bonomat, line 47: Unless I'm missing something, you are not going on-chain here. So this will always fail.

klochowicz:
> @klochowicz Not being able to find a route, does not necessarily have to be a bug. Can you describe how you got to the `no route` error?

I have added test that reproduces this behaviour - what exactly would you like me to describe more?

crates/tests-e2e/tests/close_position.rs
klochowicz: this was it! probably `no_route` was the result of exceeding the channel capacity.

crates/tests-e2e/tests/close_position.rs
klochowicz: it would, indeed, but I never got to that spot due to the `no_route` error beforehand.

klochowicz:
thanks @bonomat for finding the reason why I was not able to pay sats into the wallet anymore (I tried sending too much, e.g. `10_000` would work).

Closing as the confusion has been cleared.

PR #996, opened 2023-07-25
chore: Use network specific quotes everywhere

holzeis:
Use the correct quotes consistently on the coordinator and app depending on the network.

// edit: updated description due to a changed solution approach.

### DIFF
--- a/coordinator/src/node/expired_positions.rs
+++ b/coordinator/src/node/expired_positions.rs
@@ -56,17 +56,18 @@ pub async fn close(node: Node) {
             }
         };
 
-        let closing_price = match BitmexClient::get_quote(&position.expiry_timestamp).await {
-            Ok(quote) => quote.get_price_for_direction(position.direction.opposite()),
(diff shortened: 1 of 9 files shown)


da-kami (approved):
Happy to role with this - as long as you don't actually test hedging against testnet bitMEX this will be fine :)

bonomat:
Not against it but it will mess with our hedging strategy.

holzeis:
Alternatively, we can also use the testnet quotes consistently. e.g. the candlestick chart always shows mainnet prices. Also the testnet quotes do not change that often, thats why I sometime thought the maker was down.

bonomat:
> Alternatively, we can also use the testnet quotes consistently. e.g. the candlestick chart always shows mainnet prices. Also the testnet quotes do not change that often, thats why I sometime thought the maker was down.

If this bothers you, I'd rather go for showing a different candle chart.

bonomat (approved):
Too easy :)

PR #998, opened 2023-07-25
Allow starting frontend without the backend

klochowicz:
@da-kami it's 95% way through; please check why mockito is not quite working....

if you could take it over whilst you're still here, then it would be great :)

### DIFF
--- a/justfile
+++ b/justfile
@@ -174,7 +174,7 @@ maker args="":
     cargo run --bin maker -- {{args}}
 
 flutter-test args="":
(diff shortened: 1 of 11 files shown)


Restioson:
Would need someone new to take over now :sweat_smile:

klochowicz:
> Would need someone new to take over now 😅

yeah, I'll finish this up in the background; right now I want to focus on tasks that are scheduled in this iteration

PR #1003, opened 2023-07-25
Only fund when faucet balance drops below 10 000 000 sats

Restioson:
This speeds up development a little bit. The balance is retrieved by `curl`ing the channel balance.

I left this out of  the `fund-regtest` recipe since it seems to be a manually invoked recipe and not an automated prerequiste, hence it is not really needed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -87,15 +87,15 @@ run-local-android args="":
     --dart-define="ESPLORA_ENDPOINT=http://${LOCAL_IP}:3000" --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@${LOCAL_IP}:9045" \
     --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
 
-fund:
+fund args="":
     #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


bonomat (changes requested):
Not ideal this change: sometimes we will need to fund an already funded node again.

Restioson:
> Not ideal this change: sometimes we will need to fund an already funded node again.

Maybe a new `force-fund` command for that? Or, a new `autofund` command and use that for the prerequisite in `all` etc.

For context, the problem I am trying to solve is the fund process being run *every time* I run `just all` to start the app for debugging

bonomat:
> > Not ideal this change: sometimes we will need to fund an already funded node again.
> 
> Maybe a new `force-fund` command for that? Or, a new `autofund` command and use that for the prerequisite in `all` etc.
> 
> For context, the problem I am trying to solve is the fund process being run _every time_ I run `just all` to start the app for debugging

I never use this command 🙈 I like to keep an eye on what things happen and call each step separately. 

I'm fine with a different function or an extra arg I can pass in to _force_ funding again.

Restioson:
Fair enough! Do you have any preference as to a new command or arg?

bonomat:
> Fair enough! Do you have any preference as to a new command or arg?

Eitherway is fine with me :)

Restioson:
Added arguments -f & --force

bonomat (approved):
lgtm

PR #1005, opened 2023-07-25
Add `await` to fix app initialisation

### LINKED ISSUE #1002: Candlestick chart loads indefinitely
As seen in the screenshot below, the candlestick chart on the trade screen loads indefinitely.

image

Restioson:
If this is not awaited, then something tries to access the database before it is ready.

@klochowicz I think you just left this out by mistake?  Trivial fix in the end!

Fixes #1002.

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -257,9 +257,8 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
 
       subscribeToNotifiers(context);
 
-      runBackend(config).then((_) {
-        FLog.info(text: "Backend started");
-      });
+      await runBackend(config);
+      FLog.info(text: "Backend started");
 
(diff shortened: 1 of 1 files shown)


mobile/lib/main.dart
```diff
@@ -257,7 +257,7 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
 
       subscribeToNotifiers(context);
 
-      runBackend(config).then((_) {
+      await runBackend(config).then((_) {
```
holzeis, line 260: 🙃 if we `await` here we can move the `Flog.info` simply below the `runBackend(config);` line.

bonomat: I did it for you @Restioson

bonomat (approved):
Thank you for the fix. Let's get this in as it fixes the app.

PR #1007, opened 2023-07-26
Manage spendable outputs generated by LDK

### LINKED ISSUE #846: Handle persisted spendable outputs

luckysori:
Fixes #846.

For all `lnd_dlc_nod::Node`s we spin up a task which ensures that LDK spendable outputs that are "owned" by LDK are moved to the `Node`s on-chain wallet.

This functionality could be split between coordinator and app (the two types of `Node` currently conceived), but it does not seem worth it at this stage.


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -679,6 +684,45 @@ where
             remote_handle
         };
 
+        tokio::spawn({
```
holzeis, line 687: 🤔 I guess a task is working for the coordinator, but for the mobile app we may need to trigger it on startup? or allow for manually triggering that sync? I don't think users will have the app open for 30 minutes.

mobile/native/src/db/models.rs
```diff
@@ -1557,6 +1573,17 @@ pub mod test {
...
         assert_eq!(expected, loaded);
+
+        // Verify that we can delete the right spendable output based on its outpoint
+        SpendableOutputQueryable::delete(outpoint, &mut connection).unwrap();
+
+        match SpendableOutputQueryable::get(outpoint, &mut connection) {
+            // Could not find spendable output as expected
+            Err(Error::NotFound) => {}
+            _ => {
+                panic!("Expected to not be able to find deleted spendable output");
```
holzeis, line 1584: ❓ why do we need to panic here?

luckysori: I just followed the pattern of the other test where we delete a row. The point is that we are expecting a specific error when loading a deleted row, so we can fail the test in all other cases.

crates/ln-dlc-node/src/node/mod.rs
luckysori: This is triggered on start-up inherently 😄

I agree that exposing a different mechanism for app users would also be useful. But for now we can just tell them to restart the app if they want to ensure that missing funds are claimed? I didn't want to get into adding stuff to the frontend with this change.

holzeis (approved):

da-kami (approved):

holzeis:
// note, force pushed due to adding an entry to the changelog

Issue #1004, opened 2023-07-25
Some logs are output as logging to the "log" module


Restioson:
Two examples of mistargeted logs:

```
2023-07-25T13:23:58.598547783Z TRACE log: Calling PeerManager's timer_tick_occurred log.target="lightning_background_processor" log.module_path="lightning_background_processor" log.file="/home/restioson/.cargo/git/checkouts/rust-lightning-bee3233422fa4387/a663e1f/lightning-background-processor/src/lib.rs" log.line=472
```

```
2023-07-25T12:39:18.33240107Z  INFO log: Starting transaction sync. log.target="lightning_transaction_sync::esplora" log.module_path="lightning_transaction_sync::esplora" log.file="/home/restioson/.cargo/git/checkouts/rust-lightning-bee3233422fa4387/a663e1f/lightning-transaction-sync/src/esplora.rs" log.line=92
```

This is a problem because they become difficult to filter using env filters.

Restioson:
Possible culprit: https://github.com/get10101/10101/commit/17f0a6e4ba8d3e6a5762e10a4d0e5c6282568980

Restioson:
No response from anyone there, so I'm going to close this as not planned

PR #1016, opened 2023-07-26
Allow setting RUST_LOG to any value in `just all`

Restioson:
While investigating #1004 I discovered that you can't run `RUST_LOG=warn just all` since that makes FRB codegen panic. I just overrode RUST_LOG to either debug or info depending on its actual value for the FRB codegen calls.


klochowicz (approved):
looks good, but maybe we could open a ticket against this totally unreasonable behaviour on FRB as well? 😅 

thanks for sorting this out!

Issue #960, opened 2023-07-18
Save channel creation attempt (scid and userpubkey) into the database


da-kami:
This is important to know if users attempted to open a channel.

PR #1017, opened 2023-07-27
Store channel `fake_scid` with `channels`

da-kami:
fixes https://github.com/get10101/10101/issues/960

Since we want to record the traders *trying* top open a channel as well, we are storing he channel when we prepare the jit channel and hand the user the fake short channel id (scid).
Upon intercepting the channel we update the record if we are able to load a channel for the scid. In case we don't find a channel we print a warning (or error in case we run into an error in the db) but still proceed with opening the channel.

---

I have not implemented storing the fake short channel id in the app, because I don't think it would be very useful. 
If you think this would be good to have please let me know and I'll add it.


coordinator/migrations/2023-07-27-020530_intercept_id_channels/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE
+    channels
+ADD
+    COLUMN "fake_scid" TEXT UNIQUE;
```
holzeis, line 5: 🔧 since we are querying for it, we probably should add an `INDEX` for the fake_scid.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -692,9 +693,25 @@ where
...
 
-        let shadow_channel = Channel::new(0, channel_value, target_node_id);
+        let fake_scid = FakeScid::new(requested_next_hop_scid);
+        let shadow_channel = match self.storage.get_channel_by_fake_scid(fake_scid) {
+            Ok(Some(mut channel)) => {
+                debug_assert!(channel.counterparty == target_node_id);
+                channel.outbound = channel_value;
+                channel
+            }
+
```
holzeis, line 703: nit: useless newline 🙃

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -157,6 +167,18 @@ where
...
 
+    /// Inserts channel placeholder based on scid into the database
+    ///
+    /// This record will be updated once we intercept the HTLC and once the channel will be created.
+    fn insert_channel_from_intercept_id(
+        &self,
+        fake_scid: FakeScid,
+        counterparty: PublicKey,
+    ) -> Result<()> {
+        let channel = Channel::new(0, 0, counterparty, Some(fake_scid));
```
holzeis, line 178: 🤔 That would create a channel in state `Pending`. Shouldn't we introduce a dedicated state for channels that are announced through a created invoice? I feel like we will get a lot of those records 🙈.

e.g. `Announced`

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -148,6 +150,14 @@ where
             htlc_maximum_msat: None,
         };
 
+        // If we fail to record the channel based on intercept id into the database we don't fail
+        // but print an error
+        if let Err(e) =
```
holzeis, line 155: 🔧 We are still creating interceptable invoices for every invoice generated by the app, regardless if a channel exists or not. That may lead to a lot of channel entries that would incorrectly indicate that funding the channel failed.

We should probably add here a check if the user already has a channel. This will still not be 100% correct, but I think will work for the most cases.

da-kami: Do I understand you correctly, that we use this code path for every invoice and not just for opening a channel? 
If so, then I would probably move this code so that we can just do it upon channel creation attempt and not for all invoices.

crates/ln-dlc-node/src/node/invoice.rs
da-kami: That's reasonable I think.

crates/ln-dlc-node/src/node/invoice.rs
da-kami: Can confirm that all invoices are created by calling the `prepare_jit_channel` endpoint. 

I thought about changing this to two separate endpoints, one for initial channel creation and one for consecutive invoices. But I came to the conclusion that it's better to role with your proposal of just adding a check in the coordinator that decides if this is an initial channel open or not. That gives the coordinator more control which is good.

Might still be a good idea to rename `prepare_jit_channel` to a more intuitive name for what it is used for.

mobile/native/src/ln_dlc/node.rs
```diff
@@ -411,6 +412,11 @@ impl node::Storage for NodeStorage {
         db::get_channel(user_channel_id)
     }
 
+    fn get_channel_by_fake_scid(&self, _fake_scid: FakeScid) -> Result<Option<Channel>> {
+        // The app does not support loading channels by fake scid
+        unimplemented!("The app does not support loading channels by fake scid")
+    }
```
luckysori, line 418: 👍I think this is okay, although it reminds me of how much I hate the `Storage` trait I added.

coordinator/src/db/channels.rs
```diff
@@ -128,6 +142,7 @@ impl From<ln_dlc_node::channel::ChannelState> for ChannelState {
     fn from(value: ln_dlc_node::channel::ChannelState) -> Self {
         match value {
             ln_dlc_node::channel::ChannelState::Pending => ChannelState::Pending,
+            ln_dlc_node::channel::ChannelState::Announced => ChannelState::Announced,
```
luckysori, line 145: 🙃 Super nit.
```suggestion
            ln_dlc_node::channel::ChannelState::Announced => ChannelState::Announced,
            ln_dlc_node::channel::ChannelState::Pending => ChannelState::Pending,
```

crates/ln-dlc-node/src/channel.rs
```diff
@@ -88,6 +118,7 @@ pub struct Channel {
     /// the funding transaction. We use the `user_channel_id` as identifier over the entirety
     /// of the channel lifecycle.
     pub channel_id: Option<ChannelId>,
+    pub fake_scid: Option<FakeScid>,
```
luckysori, line 121: 🔧 We could comment that this is only `Some` if we are dealing with a JIT channel.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -148,6 +150,21 @@ where
...
+            .iter()
+            .any(|channel| channel.counterparty.node_id == target_node);
+        if is_initial_channel_open {
+            let channel =
+                Channel::new_from_initial_channel_open(target_node, FakeScid::new(intercept_scid));
+            // If we fail to record the channel based on intercept id into the database we don't
+            // fail but print an error
+            if let Err(e) = self.storage.upsert_channel(channel) {
+                tracing::error!("Failed to insert channel based on scid: {e:#}");
+            }
```
luckysori, line 165: 🔧 I would just fail here. Obviously storing the fake SCID is not super crucial, but I don't see why it should fail anyway.

mobile/native/src/db/models.rs
```diff
@@ -1139,6 +1141,7 @@ impl From<Channel> for ln_dlc_node::channel::Channel {
             channel_id: value
                 .channel_id
                 .map(|cid| ChannelId::from_hex(&cid).expect("valid channel id")),
+            fake_scid: None,
```
luckysori, line 1144: 🤔 I think we do have this value in `value`.

da-kami: This is on the app, and on the app we don't have it; the model was not extended to reflect this value.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -692,9 +693,25 @@ where
...
+            Ok(None) => {
+                tracing::warn!(%fake_scid, "Unable to find intercepted channel by fake scid in database, inserting without fake scid instead");
+                Channel::new(0, channel_value, target_node_id, None)
+            }
+            Err(e) => {
+                tracing::error!(%fake_scid, "Failed to load channel by fake scid: {e:#}");
+                Channel::new(0, channel_value, target_node_id, None)
+            }
+        };
```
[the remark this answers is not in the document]
da-kami: Early in the process, like the scenario below, I think it's OK to fail. But here it would be a shame if we have to fail creation the channel because we don't find the `scid` or there is an error. I would keep this; I don't think it does any harm.

crates/ln-dlc-node/src/node/invoice.rs
da-kami: I wonder if the above check is exhaustive enough to say we fail here. I wonder if we can have cases where we don't have a usable channel *just yet*; then the invoice creation would fail (because we cannot insert the same channel twice) event though the user just wants to make a payment. I felt it might be better to not make the invoices creation depend on this insert...

da-kami: I think eventually it would be good to have two different pathes for initial channel open and payments over an existing channel. But I did not want to refactor this as part of this PR...

mobile/native/src/db/models.rs
luckysori: Ah, good point. Not ideal to have a shared model, but we can clean that up at some point in the future.

crates/ln-dlc-node/src/node/invoice.rs
luckysori: I see where you are coming from.

I think ideally we would only call this function when we know that the invoice will be used to open a JIT channel. Then we wouldn't need to reason about the possibility that this might be for a regular payment.

It's probably okay to leave this as is for now then.

luckysori (approved):
LGTM.

My only disagreement is on just logging errors in the event handler. I think it's a bit better to error because the feature we're adding should just work (even if it's not super crucial). In theory we could miss problems with the feature if the tests pass with these errors being logged.

luckysori:
Assigning myself because we still have a merge freeze and @da-kami is leaving us 😢

luckysori:
Let's wait for the release of version 1.2 from `main` before merging this.

Issue #764, opened 2023-06-13
Can't fund wallet through just in time channel


holzeis:
On regtest we get regularly no route issues due to `no hop data found` errors. This seems to happen not deterministically, meaning that sometimes it works right away and sometimes it doesn't.

The assumption is that upon opening a channel a node announcement of some sorts needs to happen for the payment to get successfully routed.

This is very likely also the issue that our `onboard_from_lnd` is failing every now and then.

bonomat:
Related to https://github.com/get10101/10101/issues/774

holzeis:
Note, in production this issue seems to be solely related to our liquidity or connectivity.

klochowicz:
All end-to-end tests use the funding step now on regtest, let's monitor whether they fail at all due to this issue - I will update it if I spot any failures due to this reason.

klochowicz:
I managed to replicate the issue for `no_route` when paying an invoice *after* the funding (see: #880 )

bonomat:
> From #999:
> 
> This consistently fails on an android emulator for me. Invoice is still payable manually using the faucet. ![image](https://user-images.githubusercontent.com/6688948/255915992-0abd64b7-68d3-46ab-abed-be904b03cd46.png)

Does it work for you natively?

Restioson:
> Does it work for you natively?

Yes, it does

bonomat:
Very interesting. Thanks for sharing.

holzeis:
@Restioson can you check the nodeinfo on the lnd node? Is synced_to_graph true?

Restioson:
In my case, it actually _was_ an android setup issue, since in android the REGTEST_FAUCET environment variable is not set up correctly

PR #1021, opened 2023-07-27
fix: no_route found in local dev setup

bonomat:
We need to wait until the node announcement was processed correctly by lnd. This can be checked if lnd is aware of the coordinator's `node_alias`.

In the below patches we change the node announcement to 1/hour and allow triggering it manually using a http enpoint. With this I was so to always get into a working setup :)

fixes https://github.com/get10101/10101/issues/764

### DIFF
--- a/crates/ln-dlc-node/src/node/mod.rs
+++ b/crates/ln-dlc-node/src/node/mod.rs
@@ -7,6 +7,7 @@ use crate::ln::EventHandler;
 use crate::ln::TracingLogger;
 use crate::ln_dlc_wallet::LnDlcWallet;
 use crate::node::dlc_channel::sub_channel_manager_periodic_check;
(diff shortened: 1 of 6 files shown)


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -7,6 +7,7 @@ use crate::ln::EventHandler;
 use crate::ln::TracingLogger;
```
holzeis, line 7: 🙃 the commit message actually says something different than what the change entails 😅 

"move helper function out of mod" would be correct 🙃

holzeis (approved):
Nice 👍

crates/ln-dlc-node/src/node/mod.rs
bonomat: in-out, same same but different

PR #1023, opened 2023-07-27
Check if client outdated

### LINKED ISSUE #916: Inform user in the app if she is running on an outdated version
It happened that a user was running on an outdated version. It might help if the app tells the user directly that a new version is available.

We could event go so far to disable some functionalities if the app is not on the latest version.

Restioson:
This PR is based on the advice from @klochowicz in #916. It introduces a new `/api/version` endpoint to the coordinator, which returns its version as recorded in `coordinator/Cargo.toml`. This, in turn, is kept in sync by the `draft_new_release` action (I am not quite sure how to test it locally). Then, on startup, the mobile client will perform a GET request to `/api/version`, compare it to its own version, and if it is out of date it will create a snackbar notification.

I'm not sure if we want to add an action to direct the user to an update page, or to add something else to the snackbar.

There might also be considerations for local/regtest development where we might want to handle this differently.

In line with what was discussed, this PR does not disable anything, merely informing the user that they are out of date. A future PR could introduce a confirm prompt for taking actions (or on startup) if the user is deemed "too" out of date such that it would be dangerous.

Fixes #916.

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1,6 +1,9 @@
+import 'dart:convert';
+
 import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
 import 'dart:io';
+import 'package:http/http.dart' as http;
 import 'package:flutter_native_splash/flutter_native_splash.dart';
 import 'package:flutter_rust_bridge/flutter_rust_bridge.dart';
 import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
@@ -50,6 +53,7 @@ import 'package:get_10101/features/trade/domain/order.dart';
 import 'package:get_10101/features/trade/domain/price.dart';
 import 'package:get_10101/features/wallet/domain/wallet_info.dart';
 import 'package:get_10101/ffi.dart' as rust;
+import 'package:version/version.dart';
 
 final GlobalKey<NavigatorState> _rootNavigatorKey = GlobalKey<NavigatorState>(debugLabel: 'root');
 final GlobalKey<NavigatorState> _shellNavigatorKey = GlobalKey<NavigatorState>(debugLabel: 'shell');
@@ -104,6 +108,9 @@ class TenTenOneApp extends StatefulWidget {
 }
 
 class _TenTenOneAppState extends State<TenTenOneApp> {
+  final GlobalKey<ScaffoldMessengerState> scaffoldMessengerKey =
+      GlobalKey<ScaffoldMessengerState>();
+
   final GoRouter _router = GoRouter(
       navigatorKey: _rootNavigatorKey,
       initialLocation: WalletScreen.route,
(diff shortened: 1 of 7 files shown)


mobile/lib/main.dart
```diff
@@ -221,7 +228,34 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
...
+      try {
+        final response = await http.get(
+          Uri.parse('http://${config.host}:${config.httpPort}/api/version'),
+        );
+
+        final clientVersion = Version.parse(packageInfo.version);
+        final coordinatorVersion = Version.parse(jsonDecode(response.body));
+        FLog.info(text: "Coordinator version: ${coordinatorVersion.toString()}");
+
+        if (coordinatorVersion > clientVersion) {
```
holzeis, line 249: 🤔 will that work for something like "1.0.21" and "1.1.0" ?

mobile/lib/main.dart
```diff
@@ -221,7 +228,34 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
...
+          messenger.showSnackBar(SnackBar(
```
holzeis, line 250: 🔧 I think a snack bar is not the most prominent way of informing the user that she is running an outdated version. How about a popup, that the user has to confirm?

holzeis (commented):
Can you add a screenshot? I am not sure if a snackbar is sufficient to inform the user about an outdated version.

mobile/lib/main.dart
Restioson: This is using a SemVer parsing library, and I believe 1.0.21 should compare as less than 1.1.0 in SemVer

mobile/lib/main.dart
holzeis: I would argue for it, but we don't have to do it in this change.

Restioson: I'm happy to circle back to this tomorrow and get it in!

Restioson: It now shows an alert dialogue

Restioson:
image

holzeis:
Looks like there is unnecessary space in front of the second paragraph.

mobile/lib/main.dart
```diff
@@ -221,7 +228,48 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
...
+
+        if (coordinatorVersion > clientVersion) {
+          FLog.warning(text: "Client out of date. Current version: ${clientVersion.toString()}");
+          showDialog(
+              context: _shellNavigatorKey.currentContext!,
+              builder: (context) => AlertDialog(
+                      title: const Text("Update available"),
+                      content: Text("A new version of 10101 is available: "
+                          "${coordinatorVersion.toString()}.\n\n"
+                          " Please note that if you do not update 10101, the app"
```
holzeis, line 257:
```suggestion
                          "Please note that if you do not update 10101, the app"
```

holzeis (approved):

Restioson:
I'll merge this shortly, accidentally excluded pubspec.yml

PR #1025, opened 2023-07-27
Enable "Send Payment" button only when the invoice is validated

### LINKED ISSUE #1024: "Send payment" can be clicked before invoice is parsed/ready, leading to no-op
If you click "send payment" fast enough, you return to the wallet screen and nothing happens. This is slightly confusing, since the button should always either succeed or fail.

Restioson:
This is still a work-in-progress, because if a valid but already paid invoice is pasted, then the "Send Payment" button still exhibits the weird behaviour described in #1024. Fixes #1024.

### DIFF
--- a/mobile/lib/features/wallet/send_payment_change_notifier.dart
+++ b/mobile/lib/features/wallet/send_payment_change_notifier.dart
@@ -11,6 +11,7 @@ enum PendingPaymentState {
 
 class PendingPayment {
   PendingPaymentState state;
+  bool displayed = false;
   final String rawInvoice;
   final LightningInvoice decodedInvoice;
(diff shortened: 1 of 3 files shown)


holzeis (approved):
LGTM 👍 Thanks!

Restioson:
Still one more thing I want to get to here so I'd appreciate it if you can have a look after :)

Restioson:
Fixed the other issue now in [6852939](https://github.com/get10101/10101/pull/1025/commits/6852939404abc53b54709e3fd390e4c4dde0c4b4) :)

da-kami (approved):

PR #1039, opened 2023-07-30
feat: release regtest to google play store using fastlane

bonomat:
Let's try this sometimes on a different machine. There is no rush though :)


holzeis:
I'll give it a go today and see if it works on my machine.

bonomat:
Replaced by https://github.com/get10101/10101/pull/1219

PR #1040, opened 2023-07-31
fix: ci scripts were pointing to wrong ip

bonomat:


luckysori (approved):

.github/workflows/build-release-binaries.yml
```diff
@@ -139,7 +139,7 @@ jobs:
           SIGNING_STORE_PASSWORD: ${{ secrets.ANDROID_UPLOAD_SIGNING_STORE_PASSWORD }}
           ENCODED_KEYSTORE: ${{ secrets.ANDROID_UPLOAD_KEYSTORE }}
           ESPLORA_ENDPOINT: https://blockstream.info/api
-          COORDINATOR_P2P_ENDPOINT: 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156@34.30.20.202:9045
+          COORDINATOR_P2P_ENDPOINT: 022ae8dbec1caa4dac93f07f2ebf5ad7a5dd08d375b79f11095e81b065c2155156@46.17.98.29:9045
```
holzeis, line 142: 🔧I guess the pubkey was already wrong. Regtest should not point to the prod environment.

holzeis (approved):

.github/workflows/build-release-binaries.yml
bonomat: This is the production build. The thing we had to fix in this PR was to update the IP to our new production machine.

holzeis: But the line below points to regtest. Either the ip is wrong or the network.

bonomat: Lol... or even better... this part is not even being used 🙈 It's for the andoroid build

PR #1046, opened 2023-08-01
chore: Update dprint to latest version

klochowicz:
Old Rust plugin got archived and awaits a rewrite in wasm sandbox.
In the meantime, the recommended way is to directly call `rustfmt.

See: https://github.com/dprint/dprint-plugin-rustfmt

Add `rustfmt.toml` to preserve our custom formatting.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,6 @@
+wrap_comments = true
+comment_width = 100
+max_width = 100
+imports_granularity = "Item"
+group_imports = "One"
+edition = "2021"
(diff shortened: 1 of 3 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+wrap_comments = true
+# comment_width = 100
+# max_width = 100
+# imports_granularity = "Item"
+# group_imports =  "One"
+# edition = "2021"
```
luckysori, line 6: ❓Can we enable these?

luckysori (approved):

rustfmt.toml
klochowicz: oops! I temporarily disabled them as I was debugging problems with my local `rustfmt` which was hanging. 
fixed :)

PR #1047, opened 2023-08-01
Add ScrollableSafeArea to scroll widgets if keyboard would squish them

### LINKED ISSUE #1026: Widgets overflow when keyboard is expanded
When using the emulator for the Pixel (original), the text entry box overflows by 47 pixels when the keyboard is expanded. I assume this is a screen size thing and probably something about how flutter handles keyboard expansion.

 image

```
══╡ EXCEPTION CAUGHT BY RENDERING LIBRARY
╞═════════════════════════════════════════════════════════
The following assertion was thrown during layout:
A RenderFlex overflowed by 47 pixels on the bottom.

The relevant error-causing widget was:
  Column
  Column:file:///home/restioson/IdeaProjects/10101/mobile/lib/features/welcome/welcome_s
  creen.dart:40:20

To inspect this widget in Flutter DevTools, visit:
http://127.0.0.1:9100/#/inspector?uri=http%3A%2F%2F127.0.0.1%3A38525%2FYDLK3WMrB-c%3D%2F
&inspectorRef=inspector-0

The overflowing RenderFlex has an orientation of Axis.vertical.
The edge of the RenderFlex that is overflowing has been marked in the rendering with a
yellow and
black striped pattern. This is usually caused by the contents being too big for the
RenderFlex.
Consider applying a flex factor (e.g. using an Expanded widget) to force the children of
the
RenderFlex to fit within the available space instead of being sized to their natural
size.
This is considered an error condition because it indicates that there is content that
cannot be
seen. If the content is legitimately bigger than the available space, consider clipping
it with a
ClipRect widget before putting it in the flex, or using a scrollable container rather
than a Flex,
like a ListView.
The specific RenderFlex in question is: RenderFlex#89770 relayoutBoundary=up4
OVERFLOWING:
  needs compositing
  creator: Column ← Padding ← ColoredBox ← Container ← _FormScope ← WillPopScope ←
    Form-[LabeledGlobalKey<FormState>#4b040] ← MediaQuery ← Padding ← SafeArea ←
    KeyedSubtree-[GlobalKey#b8770] ← _BodyBuilder ← ⋯
  parentData: offset=Offset(20.0, 20.0) (can use size)
  constraints: BoxConstraints(0.0<=w<=371.4, 0.0<=h<=268.2)
  size: Size(371.4, 268.2)
  direction: vertical
  mainAxisAlignment: start
  mainAxisSize: max
  crossAxisAlignment: end
  textDirection: ltr
  verticalDirection: down
```

Same occurs for this widget in the RX screen:
image

and the trade screen
image

Restioson:
Fixes #1026. This just makes the content scroll vertically if it would be too small to fit on the screen (due to its size or due to the on-screen keyboard being opened).

### DIFF
--- a/mobile/lib/features/trade/trade_bottom_sheet.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet.dart
@@ -30,11 +30,13 @@ tradeBottomSheet({required BuildContext context, required Direction direction}) 
                 currentFocus.unfocus();
               }
             },
-            child: SizedBox(
-                // TODO: Find a way to make height dynamic depending on the children size
-                // This is needed because otherwise the keyboard does not push the sheet up correctly
-                height: 450,
-                child: TradeBottomSheet(direction: direction)),
+            child: SingleChildScrollView(
+              child: SizedBox(
+                  // TODO: Find a way to make height dynamic depending on the children size
+                  // This is needed because otherwise the keyboard does not push the sheet up correctly
+                  height: 450,
+                  child: TradeBottomSheet(direction: direction)),
+            ),
           ),
         ),
       );
--- a/mobile/lib/features/wallet/create_invoice_screen.dart
+++ b/mobile/lib/features/wallet/create_invoice_screen.dart
@@ -13,6 +13,7 @@ import 'package:go_router/go_router.dart';
 import 'package:provider/provider.dart';
 import 'package:get_10101/common/modal_bottom_sheet_info.dart';
 import 'package:get_10101/common/domain/channel.dart';
+import 'package:get_10101/common/scrollable_safe_area.dart';
 
 class CreateInvoiceScreen extends StatefulWidget {
   static const route = "${WalletScreen.route}/$subRouteName";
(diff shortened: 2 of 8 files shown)


Restioson:
I think this has some issues and I need to test it again

Restioson:
Fixed an issue with the bottom sheets (they now use just a normal `SingleScrollView` since the height is fixed), as well as the backup seed page. I just undid the changes on the backup seed page since `CrossAxisAlignment.baseline` doesn't work with the `IntrinsicHeight` widget used in `ScrollableSafeArea`. This should be fine since this page should fit on most devices, even small ones, since the keyboard can't be opened.

luckysori (commented):
Can you post a couple of screenshots/gifs so that we can see how it looks after the changes?

Restioson:
> Can you post a couple of screenshots/gifs so that we can see how it looks after the changes?


[device-2023-08-02-134317.webm](https://github.com/get10101/10101/assets/6688948/ba0f48de-3936-4c7e-b16d-8534fe640525)

Here I show some screens which don't fit on the device and some that do

Restioson:
One thing - the QR code size might change with this. I should test that quickly

mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -65,13 +66,14 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
...
                 ),
-                Expanded(
-                  child: Center(
-                    child: QrImageView(
-                      data: widget.invoice.rawInvoice,
-                      version: QrVersions.auto,
-                      padding: const EdgeInsets.all(qrPadding),
-                    ),
+                Container(
+                  height: qrWidth,
```
Restioson, line 70: This line implies that `qrWidth` is the width of the QR code: https://github.com/get10101/10101/blob/b91fa44a02625b4d9efd5c61323f4f269cd53b3d/mobile/lib/features/wallet/share_invoice_screen.dart#L82

But that actually isn't true :sweat_smile:  So forcing it into these dimensions keeps it quite small

Restioson: On main:
image

On this branch: 

image

The "Amount: ..." text has got nothing to do with the QR code size on `main`, but it works as per the comment on this branch. This makes me think the original code is buggy and this PR just "fixes" that bug.

The reason this is needed in the first place is because the QrImageView causes errors in flutter if it isn't assigned a height and it is a child of an `IntrinsicHeight` widget (as used here to prevent specifying the whole viewport height manually)

luckysori: I wonder how that looks on the phone. I guess the new version wastes too much screen real state at the moment.

Could we just skip changing this page for now? I don't think one can open the keyboard on this screen anyway.

mobile/lib/features/wallet/create_invoice_screen.dart
```diff
@@ -103,118 +104,115 @@ class _CreateInvoiceScreenState extends State<CreateInvoiceScreen> {
...
-                      Expanded(
-                        child: AmountInputField(
-                          value: amount != null ? amount! : Amount(0),
-                          hint: "e.g. ${formatSats(Amount(5000))}",
-                          label: "Amount",
-                          controller: _amountController,
-                          onChanged: (value) {
-                            if (value.isEmpty) {
-                              return;
+          child: ScrollableSafeArea(
```
luckysori, line 107: 😆 This part of the diff is brutal when all that has likely changed is this bit at the top.

luckysori (approved):
LGTM and thanks for the video!

mobile/lib/features/wallet/share_invoice_screen.dart
Restioson: > I wonder how that looks on the phone. I guess the new version wastes too much screen real state at the moment.

Well, the phone usually won't be as big as on desktop anyway 😅 so I doubt it makes much of a difference

> Could we just skip changing this page for now? I don't think one can open the keyboard on this screen anyway.

We can, yea

mobile/lib/features/trade/trade_bottom_sheet.dart
```diff
@@ -30,11 +30,13 @@ tradeBottomSheet({required BuildContext context, required Direction direction})
...
-                // This is needed because otherwise the keyboard does not push the sheet up correctly
-                height: 450,
-                child: TradeBottomSheet(direction: direction)),
+            child: SingleChildScrollView(
+              child: SizedBox(
+                  // TODO: Find a way to make height dynamic depending on the children size
+                  // This is needed because otherwise the keyboard does not push the sheet up correctly
+                  height: 450,
+                  child: TradeBottomSheet(direction: direction)),
+            ),
```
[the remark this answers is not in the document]
Restioson: > ❓Why do we not use a ScrollableSafeArea here?

I did try to use it here but it causes the TradeBottomSheet to expand to fill the entire screen (not what we want), and we don't need it here anyway. First of all, this wasn't using a safe area anyway. Second of all, ScrollableSafeArea is mostly SafeArea + Scroll + some hacks to make things work inside scroll. Since this is already sized, we don't need those hacks, so we just need scroll. That's why SingleChildScrollView fits perfectly

mobile/lib/features/wallet/create_invoice_screen.dart
Restioson: I also removed the 

```
            child: Container(
              constraints: const BoxConstraints.expand(),
```

because it doesn't seem to do much really (neither before nor after this change...)

PR #1048, opened 2023-08-01
Compile android for x86 too

Restioson:
Some older API versions use x86 by default in the emulator. Also deletes the old libs in `jniLibs` in `clean` task.


holzeis (approved):
LGTM 👍

PR #1049, opened 2023-08-02
Run `rustfmt` with nightly

luckysori:
We must do this if we want `dprint` to respect some of the unstable options we have configured in our `rustfmt.toml`. This is what we get if we run `cargo fmt` (as opposed to `cargo +nightly fmt`):

```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 100`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Item`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = One`, unstable features are only available in nightly channel.
```

After running `dprint fmt` with the new configuration we can see that the autoformatter consistently applies `group_imports = One` to our codebase.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@nightly
+        with:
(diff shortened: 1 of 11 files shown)


bonomat (approved):
Fine with me

PR #1050, opened 2023-08-02
Update `ChannelConfig` of open channels

### LINKED ISSUE #952: Update proportional routing fee for existing channels
In #951 we allow the coordinator to update the proportional routing fee, but this does not affect channels that are already open. There seems to be a way to do this via [this method](https://github.com/lightningdevkit/rust-lightning/blob/8e2b70d91a0f7322042d1b0935527c90c59afe45/lightning/src/ln/channel.rs#L1159).

_Originally posted by @luckysori in https://github.com/get10101/10101/issues/927#issuecomment-1637319219_

luckysori:
Fixes #952.


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -770,6 +771,29 @@ where
             settings,
         })
     }
+
+    pub fn update_ldk_settings(&self, ldk_config: UserConfig) {
```
holzeis, line 775: 🤔 shouldn't we differentiate between public and private channels here? Otherwise we will always have on or the other if that gets executed.

holzeis (commented):
Nice, what parts of the ldk config can we change with that specifically? I guess that will have no impact on the handshake?

crates/ln-dlc-node/src/node/mod.rs
luckysori: Good question! What this PR adds shouldn't have an effect on whether channels are public or private because that is controlled by `ChannelHandshakeConfig::announced_channel` which is not controlled by `ChannelConfig`. In general, I'm not sure there is a mechanism to turn a private channel into a public one.

luckysori: In relation to updating the `UserConfig` in general, we already have a default value for `announced_channel` for both coordinator and app. In the case of the coordinator we default to public channels and only override this when opening JIT channels in the event handler.

In theory we could change this default value with this API, but that would be weird. It's actually currently impossible to change it via the coordinator settings, which only support updating the `forwarding_fee_proportional_millionths` value:

https://github.com/get10101/10101/blob/2e574bf83ed40c48b6e6b72b7bdb1155c2c606c7/coordinator/src/settings.rs#L99-L110

luckysori:
> Nice, what parts of the ldk config can we change with that specifically?

With this PR the only thing that we are changing is the ability to override the `ChannelConfig` for open channels. Since this feature is only used by the coordinator and the coordinator currently only cares about updating the proportional routing fees, that's all that we can do.

Eventually we might want to use this feature to change other bits.

> I guess that will have no impact on the handshake?

Yep.

holzeis (approved):

PR #1053, opened 2023-08-02
Do not attempt to publish transaction spending no LDK spendable outputs

luckysori:
Otherwise we see errors in the logs, because transactions spending no UTXOs are obviously invalid!


holzeis (approved):
Thanks. 👍

bonomat (approved):

PR #1055, opened 2023-08-03
chore: Extract functions in Node initialisation

klochowicz:
The amount of implicit dependencies we have is unwieldy.

Extract closures into methods.


luckysori (approved):
Hard to argue with these changes! Thanks ♥️

PR #1062, opened 2023-08-04
Allow using different event handlers

klochowicz:
The main motivation was as such:
Separate the boring things (e.g. initialisation of bajilion pieces of LDK) from the customisation points, which might be different.

As such, `ln_dlc_node::Node` is only a collection of entities needed to be used to run a LDK node with DLC capable.
For now, there's a single 'start' method into which different EventHandlers can be passed in.

It's easy to imagine that if we ever need/want to just have separate free functions in various node implementations for starting the node. They can be composed quite easily with the aid of the functions extracted from Node in a PR made recently.
As such, this is not the scope of the ticket, and I had to stop the refactor somewhere 😀

### DIFF
--- a/crates/ln-dlc-node/src/ln/event_handler.rs
+++ b/crates/ln-dlc-node/src/ln/event_handler.rs
@@ -130,52 +130,14 @@ where
                 output_script,
                 user_channel_id,
             } => {
-                let user_channel_id = Uuid::from_u128(user_channel_id).to_string();
-                tracing::info!(
-                    %user_channel_id,
-                    %counterparty_node_id,
-                    "Funding generation ready for channel with counterparty {}",
-                    counterparty_node_id
-                );
-
-                let target_blocks = CONFIRMATION_TARGET;
-
-                // Have wallet put the inputs into the transaction such that the output
-                // is satisfied and then sign the funding transaction
-                let funding_tx_result = self
-                    .wallet
-                    .inner()
-                    .create_funding_transaction(
-                        output_script,
-                        channel_value_satoshis,
-                        target_blocks,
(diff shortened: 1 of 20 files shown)


bonomat (approved):
Looks good. I trust that you verified that this also works in the simulator.

klochowicz:
> Looks good. I trust that you verified that this also works in the simulator.

yup, works with the simulator, checked it. I have quite high degree of confidence in our test suite ever since e2e tests got added, as they test the same code.

klochowicz:
I will wait for the release though (and for a final look from @luckysori )

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -888,3 +890,18 @@ impl<S> EventHandler<S> {
             .expect("Mutex to not be poisoned")
     }
 }
+
+#[async_trait]
+pub trait EventHandlerTrait: Send + Sync {
```
luckysori, line 895: I am bit saddened that we have to introduce another abstraction, but I guess there's no better way to do this if we don't tear down the `ln-dlc-node` crate.

luckysori (approved):

crates/ln-dlc-node/src/ln/event_handler.rs
klochowicz: I thought about this quite a bit and came to the conclusion that this indeed is the lesser evil. 
Not a big trade-off really for reusing a lot of the code.

if we ever have time to look at this again, I'd be keen to see where you push it - at least it should be easier to refactor from this point :)

PR #1063, opened 2023-08-04
Improve some log messages

bonomat:


mobile/native/src/db/mod.rs
```diff
@@ -310,7 +313,15 @@ pub fn get_payments() -> Result<Vec<(lightning::ln::PaymentHash, ln_dlc_node::Pa
 
     let payment_hashes = payments.iter().map(|(a, _)| a).collect::<Vec<_>>();
 
-    tracing::debug!(?payment_hashes, "Got all payments");
+    let formatted_payment_hashes = payment_hashes
+        .clone()
+        .iter()
+        .map(|hash| hex::encode(hash.0))
+        .collect::<Vec<_>>();
```
klochowicz, line 320: 🙃

given that on line 314 we only mapped the vector for the purpose of displaying, we don't need to do it twice.

We can remove the line above and use a combination of both:

```rust
let formatted_payment_hashes = payments.iter().map(|(hash, _)| hash).map(|hash| hex::encode(hash.0)).collect::<Vec<_>>();
```

mobile/native/src/event/event_hub.rs
```diff
@@ -44,7 +44,7 @@ impl EventHub {
 
     /// Publishes the given event to all subscribers. Note, that this will be executed in a loop.
     pub fn publish(&self, event: &EventInternal) {
-        tracing::debug!("Publishing event {:?}", event);
+        tracing::debug!(event = %event, "Publishing an internal event");
```
klochowicz, line 47: 🤔 wouldn't `event = %event` be equivalent to just `%event`? I thought that the `a = %b` notation just renames b to a in tracing?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -306,7 +306,7 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
     });
 
     let off_chain = off_chain.iter().filter_map(|details| {
-        tracing::info!(?details, "Off-chain payment details");
+        tracing::debug!(details = %details, "Off-chain payment details");
```
klochowicz, line 309:
```suggestion
        tracing::debug!(%details, "Off-chain payment details");
```

klochowicz (approved):
looks good, just some nits

mobile/native/src/event/event_hub.rs
bonomat: Yes it's the same. Personally I prefer the explicit way of writing, i.e. `event = %event` over `%event`. I find it easy to grep over the code if I want to find log messages with `event` in it.

mobile/native/src/db/mod.rs
bonomat: Good catch :)

mobile/native/src/event/event_hub.rs
klochowicz: I have no strong feelings, so if it works for you better, it's fine :) just wasn't sure whether you were aware of this!

PR #1064, opened 2023-08-04
test: Reproduce force-closure with reversed order

bonomat:
Resolves https://github.com/get10101/10101/issues/874

This test runs now on my machine.

I kept https://github.com/get10101/10101/pull/1064/commits/d3b53e3d762fd87f1de9dedae6974ba0bfde7359 separate for @holzeis to see.

Issue #1031, opened 2023-07-30
Give app user an indication that their channel is being closed on-chain


holzeis:
Instead of trying to get back into a recoverable mode of the app we want go for an emergency scenario: 

If the app detects that the DLC channel has been force closed (or is currently being force closed), we show an emergency screen. The emergency screen shows something like this: 

> Your DLC channel has been force closed. Don't you worry, you will be able to get your funds. Please keep the app open or come back to check when the transactions have been included. You can then recover your funds in a new app using your seed phrase. 

Recovering from the seed phrase is covered in https://github.com/get10101/10101/issues/1079

bonomat:
To be precise: we should only remove the position once the corresponding CET was successfully spent (or is spendable), i.e. once the oracle attests to the price. Before that we do not know the PnL.

bonomat:
Deemed as not critical for now and de-prioritized. We want to move forward with features that increase trading and enable perpetuals. 
Removed from the milestone.

PR #1067, opened 2023-08-04
Release version 1.2.0

1010Tom:
Hi @bonomat!
This PR was created in response to a manual trigger of the release workflow here: https://github.com/get10101/10101/actions/runs/5762281370.
I've bumped the versions in the manifest files in this commit: 0533ac91c7c17614dd447fbfb1c33f63792988e6.
Merging this PR will create a GitHub release!

closes https://github.com/get10101/10101/milestone/3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.2.0] - 2023-08-04
+
 - Automatically retry spendable outputs if not successfully published
 - Permit the closure of the LN-DLC channel in any intermediate state.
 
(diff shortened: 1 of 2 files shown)


klochowicz (approved):

bonomat:
Needs a rebase after https://github.com/get10101/10101/pull/1064 was merged.
I'd say https://github.com/get10101/10101/issues/1031 can go into the a patch release so that we can trigger a release over the weekend and have it reviewed till monday.

holzeis:
maybe blocked by https://github.com/get10101/10101/issues/1074 ?

bonomat:
> maybe blocked by #1074 ?

Do you believe it is a regression which came in between 1.1.0 and now? If not, I'd say this is a "normal" bug which we need to fix but independently of 1.2.0

holzeis:
> > maybe blocked by #1074 ?
> 
> Do you believe it is a regression which came in between 1.1.0 and now? If not, I'd say this is a "normal" bug which we need to fix but independently of 1.2.0

well, i don't know. It looks like it. Downgrading to 1.1.0 (1726) is also crashing my app. Obviously it's not happening in all scenarios.. Maybe we release and check if it happens as well on the main version. If so, we don't have to publish it for external users and work on a patch.

bonomat:
Rebased!

PR #1073, opened 2023-08-05
Use different event handlers for app and coordinator

klochowicz:
Individual event handlers are shareable to minimise the possibility of fixing a bug in one place and not others. Each EventHandler is fully customisable.

I didn't go *all* way with extracting them into their separate crates - if needed this can be done as a follow-up.


klochowicz:
I have added this PR against the other one (when the #1062 gets merged, this will get automatically rebased).

klochowicz:
converting to draft, seems like there's a regression (still useful to see the full picture of the changes).

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -0,0 +1,216 @@
...
+                )?;
+            }
+            Event::OpenChannelRequest {
+                temporary_channel_id,
+                counterparty_node_id,
+                funding_satoshis,
+                push_msat,
+                ..
+            } => {
+                handlers::handle_open_channel_request_trusted_0_conf(
```
bonomat, line 92: The coordinator should not accept 0-conf channels. Only the app should accept 0-conf channels from the coordinator. 

I guess we had this logic before?

klochowicz: we didn't have any more logic, we only had a comment:
`TODO: only accept 0-conf from the coordinator.`

klochowicz: #1075 fixes, but I can include it into this PR. it is the first behavioural change I've done, all the other stuff was "just" a refactor, ie. I tried to have zero behavioural changes.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -196,94 +193,16 @@ impl<S> Node<S>
...
-        oracle: OracleInfo,
-    ) -> Result<Self> {
-        let mut ldk_config = coordinator_config();
-
-        // TODO: The config `force_announced_channel_preference` has been temporarily disabled
-        // for testing purposes, as otherwise the app is not able to open a channel to the
-        // coordinator. Remove this config, once not needed anymore.
-        ldk_config
-            .channel_handshake_limits
-            .force_announced_channel_preference = false;
```
luckysori, line 261: I guess the tests will reveal if we don't need this anymore.

crates/ln-dlc-node/src/config.rs
```diff
@@ -1,9 +1,48 @@
+use std::time::Duration;
+
```
luckysori: I think `rustfmt` should be able to fix this after our recent config changes!

crates/ln-dlc-node/src/config.rs
```diff
@@ -1,9 +1,48 @@
...
+pub const CONFIRMATION_TARGET: ConfirmationTarget = ConfirmationTarget::HighPriority;
+
+/// When handling the [`Event::HTLCIntercepted`], the user might not be online right away. This
+/// could be because she is funding the wallet through another wallet. In order to give the user
+/// some time to open 10101 again we wait for a bit to see if we can establish a connection.
+///
+/// This constant specifies the amount of time we are willing to delay a payment.
+pub(crate) const HTLC_INTERCEPTED_CONNECTION_TIMEOUT: Duration = Duration::from_secs(30);
+
 pub fn app_config() -> UserConfig {
```
luckysori, line 45: 🔧 Might be worth renaming this to `app_ldk_config` since we've lost the context of the module name.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -279,626 +274,673 @@ where
...
+impl<S> EventHandlerTrait for EventHandler<S>
+where
+    S: Storage + Send + Sync + 'static,
+{
+    async fn handle_event(&self, event: Event) {
+        self.handle_event(event).await
+    }
+}
+
+// Get the handlers
```
luckysori: 🔧 It seems like this doesn't belong here.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -12,44 +9,14 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::watch;
 
-type PendingInterceptedHtlcs = Arc<Mutex<HashMap<PublicKey, (InterceptId, u64)>>>;
+pub type PendingInterceptedHtlcs = Arc<Mutex<HashMap<PublicKey, (InterceptId, u64)>>>;
+pub type EventSender = watch::Sender<Option<Event>>;
 
 #[async_trait]
 pub trait EventHandlerTrait: Send + Sync {
```
luckysori, line 16: 🙃 That's a rough name haha.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -251,7 +46,6 @@ where
 /// don't require custom behaviour
 pub mod handlers {
     use super::*;
```
luckysori, line 48: 🔧 I think it's not recommended to use this kind of import outside of test modules.

crates/ln-dlc-node/src/ln/event_handler.rs
```diff
@@ -997,3 +792,17 @@ pub mod handlers {
...
 }
+#[async_trait]
+impl<T: EventHandlerTrait + ?Sized> EventHandlerTrait for Arc<T> {
+    async fn match_event(&self, event: Event) -> Result<()> {
+        (**self).match_event(event).await
+    }
+
+    async fn handle_event(&self, event: Event) {
+        (**self).handle_event(event).await
+    }
```
luckysori: ❓Why do we need two methods?

crates/ln-dlc-node/src/tests/load.rs
```diff
@@ -28,7 +30,13 @@ async fn single_app_many_positions_load() {
     init_tracing();
 
     let coordinator = Coordinator::new_public_regtest();
+
+    let app_event_handler = |node, event_sender| {
+        Arc::new(AppEventHandler::new(node, event_sender)) as Arc<dyn EventHandlerTrait>
+    };
```
luckysori, line 36: 🔧 Can we compose this in a  `Node::start_test_app` method? It's pretty noisy atm.

luckysori (approved):
I think this works!

I like that we can share code between app and coordinator when needed. I don't like that we had to introduce certain abstractions to do so. I still hope that we can get rid of the `ln_dlc_node::Node` one day 🙏

crates/ln-dlc-node/src/node/mod.rs
klochowicz: yeah, if the tests pass, it's probably not needed :)

crates/ln-dlc-node/src/config.rs
klochowicz: the config's clean, I must've only called dprint later. I will adjust my IDE to use the same config to prevent this in the future.

crates/ln-dlc-node/src/ln/event_handler.rs
klochowicz: it became evident in the following commit, where I made sure that every event handler has the option to send the events (to avoid reimplementing that.
`handle_event()` has a default impl that's shared, and the implementers only need to implement:

- `event_sender() -> &EventSender` - how to retrieve `EventSender` 
- `match_event()` - how to implement actual event logic.

crates/ln-dlc-node/src/tests/load.rs
klochowicz: this is the only place where it's outside, because of custom requirements.

`Node::start_test_app` has different needs (which has this inside it), and cannot be used here :/

The other way I contemplated was to have `AppEventHandler::new_arc()` method that just does this for you, but then I realised it's only needed in tests; we don't rely on dynamic dispatch at all in the production code.
side note - perhaps we could refactor it away, but I can't justify the effort in doing that right now 🙃

crates/ln-dlc-node/src/ln/event_handler.rs
klochowicz: why would this not be recommended? it's perfectly valid syntax, and used for, well... organising the code? Happy to learn more about the reasons if you don't mind sharing some links.

Moved it into `common_handlers.rs`.

luckysori: @klochowicz: There's a Clippy lint that explains this: https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard_imports.

(Unresolved conversation for visibility.)

klochowicz: thank you! 

....shall we enable it then? :)

klochowicz: aaaaaa I think we were talking past each other - I thought your comment was about having a module defined inside module (with `mod handlers {}`, and you were talking about wildcard imports!

luckysori: Oh lol. I actually did not mind the internal submodule.

PR #1080, opened 2023-08-07
e2e test for force-closing dlc channel

bonomat:
resolves [#1077](https://github.com/get10101/10101/issues/1077)


mobile/native/src/event/mod.rs
```diff
@@ -27,6 +27,7 @@ pub fn publish(event: &EventInternal) {
 pub enum EventInternal {
     Init(String),
     Log(String),
+    NodePubkey(String),
```
klochowicz, line 30: 🔧

I don't think we need to define a whole `Event` for this - calling `api::get_node_id().0 in the test` should suffice! (this is how we can acess the `SyncReturn` in rust tests. 
This would also be the same way that we are retrieving the node id in the app, so we will slightly increase test coverage 😉

this should decrease the size of this PR substantially.

crates/tests-e2e/tests/force_close_position.rs
```diff
@@ -0,0 +1,51 @@
...
+#[tokio::test]
+#[ignore = "need to be run with 'just e2e' command"]
+async fn can_force_close_position() {
+    let test = setup::TestSetup::new_with_open_position().await;
+    let coordinator = &test.coordinator;
+    let app = &test.app;
+    let bitcoin = &test.bitcoind;
+
+    wait_until!(app.rx.node_pubkey().is_some());
+    let app_pubkey = app.rx.node_pubkey().unwrap();
```
klochowicz, line 16: (see other comment for explanation).

```suggestion
    let app_pubkey = api::get_node_id().0;
```

klochowicz (commented):
looks good

crates/tests-e2e/tests/force_close_position.rs
bonomat: Nice, I didn't know this. Thanks.

klochowicz (approved):
LGTM!

crates/tests-e2e/tests/force_close_position.rs
```diff
@@ -0,0 +1,50 @@
...
+    bitcoin.mine(100).await.expect("To be able to mine blocks");
+    // Let the coordinator catch-up with the blocks
+    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+    coordinator
+        .get_dlc_channels()
+        .await
+        .expect("to be able to retrieve dlc channels from coordinator")
+        .iter()
+        .any(|chan| {
+            chan.counter_party == app_pubkey && chan.state == SubChannelState::OnChainClosed
```
luckysori, line 48: Interestingly this just checks that the subchannel is closed, but the DLC is still alive (pending the oracle attestation and some more confirmations).

luckysori (approved):
Thanks! I think this helps even if it's not checking everything.

crates/tests-e2e/tests/force_close_position.rs
bonomat: I spent a few minutes looking into how to mock the oracle and dropped that idea again... but yea, ideally we would go all the way to also spend the CETs

PR #1083, opened 2023-08-08
Make order expiry configurable

klochowicz:


mobile/lib/features/trade/domain/order.dart
```diff
@@ -76,6 +76,8 @@ class Order {
         direction: bridge.Direction.Long,
         orderType: bridge.OrderType.market(),
         state: bridge.OrderState.Open,
-        creationTimestamp: 0);
+        creationTimestamp: 0,
+        orderExpiryTimestamp: 0
```
bonomat, line 80: Is this field needed for something?

bonomat (approved):
LGTM

luckysori (approved):

mobile/lib/features/trade/domain/order.dart
klochowicz: it can be used to detect expired orders from the orderbook. Not used now, but it wouldn't compile if I didn't change it ;/

PR #1088, opened 2023-08-09
feat: release regtest ios app on push to main using fastlane

bonomat:
Note: it is a bit slow unfortunately because we have to build the IPA file twice. 
See inline comment. 

Works on my fork: https://github.com/bonomat/10101/actions/runs/5805589465/job/15736938345

### DIFF
--- a/.github/workflows/ios-fastlane.yml
+++ b/.github/workflows/ios-fastlane.yml
@@ -27,10 +27,6 @@ on:
         required: true
         description: "The public key of the oracle. Must match with the public key returned at http://${oracle_endpoint}/oracle/publickey"
         type: string
-      app_scheme:
-        required: true
-        description: "For regtest enter 'test' or mainnet enter 'Runner'"
-        type: string
       fastlane_provisioning_profile_specifier:
         required: true
         description: "For regtest enter 'match AppStore finance.get10101.app.test' or mainnet enter 'match AppStore finance.get10101.app 1691412333'"
@@ -161,6 +157,13 @@ jobs:
           # inputs
           FASTLANE_DEVELOPER_APP_IDENTIFIER: ${{ inputs.fastlane_developer_app_identifier }}
           FASTLANE_PROVISIONING_PROFILE_SPECIFIER: ${{ inputs.fastlane_provisioning_profile_specifier }}
-          FASTLANE_APP_SCHEME: ${{ inputs.app_scheme }}
+          NETWORK: ${{ inputs.network }}
         run: |
+          if [ "$NETWORK" = "mainnet" ]; then
+            FASTLANE_APP_SCHEME="Runner"
+          else
+            FASTLANE_APP_SCHEME="test"
+          fi
+          
+          echo "APP_SCHEME is set to $FASTLANE_APP_SCHEME"
           just publish-testflight-fastlane
--- a/.github/workflows/build-release-binaries.yml
+++ b/.github/workflows/build-release-binaries.yml
@@ -34,7 +34,6 @@ jobs:
       oracle_pubkey: 16f88cf7d21e6c0f46bcbc983a4e3b19726c6c98858cc31c83551a88fde171c0
       fastlane_developer_app_identifier: finance.get10101.app
       fastlane_provisioning_profile_specifier: match AppStore finance.get10101.app 1691412333"
-      app_scheme: Runner
 
   build_android_apk:
     runs-on: macos-latest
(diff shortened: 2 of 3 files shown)


RELEASE.md
```diff
@@ -49,3 +49,22 @@ just release-testflight
...
 
 Once uploaded, log into `appstoreconnect.apple.com/apps/` and approve testing the new version.
+
+## Using fastlane
+
+Make sure that all environment variables are set in the `.env` file.
+
+You will also need to install [`fastlane`](https://fastlane.tools/).
+
+1. We need to build the IPA file first without code signing because fastlane does not support `--dart-define`.
```
bonomat, line 59: This is quite unfortunate. An alternative solution would be to manually edit the files which are being edited if `--dart-define` is being used. See answer here: https://stackoverflow.com/questions/64396002/flutter-fastlane-ios-how-to-pass-dart-define

klochowicz: I guess we can try living with it for now....if the overhead becomes too big, we can always try the other way.

am I correct in thinking that these IPA files can also be used to upload to a phone and test that way?

mobile/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -796,9 +805,11 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = "AppIcon-test";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = V3D64P9D6W;
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
+				CODE_SIGN_STYLE = Manual;
+				CURRENT_PROJECT_VERSION = 1805;
```
klochowicz, line 810: 🤔 can't we use the $(FLUTTER_BUILD_NUMBER) anymore? it seems to be a bit of nuisance to update this one manually in multiple places - it will be forgotten sooner or later 🙁

in the worst case we could do this manually with a `just set-ipa-version $VERSION` command that just runs `sed` under the hood

klochowicz (approved):

.github/workflows/ios-fastlane.yml
```diff
@@ -0,0 +1,166 @@
...
+        required: true
+        type: string
+      oracle_pubkey:
+        required: true
+        description: "The public key of the oracle. Must match with the public key returned at http://${oracle_endpoint}/oracle/publickey"
+        type: string
+      app_scheme:
+        required: true
+        description: "For regtest enter 'test' or mainnet enter 'Runner'"
+        type: string
```
luckysori, line 33: ❓This is a bit weird :O Why do we need to pass it in if we already know the values based on the network?

mobile/ios/Runner.xcodeproj/project.pbxproj
bonomat: I believe it is autogenerated now and not controlled via Flutter anymore, i.e. it will be overwritten everytime you build it. Hence, the change in this file. But maybe I just change it back since we don't check in the build changes after the CI run.

.github/workflows/ios-fastlane.yml
bonomat: Fair, we could just write a bash script for that.

mobile/ios/Runner.xcodeproj/project.pbxproj
klochowicz: does this mean that our build numbering will change?

will the version seen inside the app match the new one?

bonomat: We are still using dynamic build number which are generated the same way as before just in a different place: 

https://github.com/get10101/10101/blob/09e2b63a99104ab76bd311940d67494a2e121855/mobile/ios/fastlane/Fastfile#L49-L53

mobile/ios/fastlane/Gemfile.lock
```diff
@@ -0,0 +1,283 @@
...
+      minitest (~> 5.1)
+      tzinfo (~> 1.1)
+    addressable (2.8.5)
+      public_suffix (>= 2.0.2, < 6.0)
+    algoliasearch (1.27.5)
+      httpclient (~> 2.8, >= 2.8.3)
+      json (>= 1.5.1)
+    artifactory (3.0.15)
+    atomos (0.1.3)
+    aws-eventstream (1.2.0)
```
holzeis, line 18: 🤔 that are an awful lot of dependencies. Do we really need all of them?

holzeis (approved):
Nice 👍

mobile/ios/fastlane/Gemfile.lock
bonomat: That's autogenerated and we've got no control over it.

PR #1089, opened 2023-08-10
fix: Fix recent db migration in the app

klochowicz:
If the orders table was not empty,
the migration would never succeed as sqlite does not allow setting
non-const default values unfortunately.

Given that all the historical orders in the DB would have been
outdated in 60 seconds anyway, set the default expiry to 0 instead.


luckysori (approved):
LGTM.

PR #1090, opened 2023-08-10
chore(justfile): Don't invalidate the cache between iOS and native builds

klochowicz:
It was a major slowdown to wait for the project to recompile as cache was being invalidated.

By specifying a different target directory we ensure build artefacts will be
preserved when switching between `just ios` and `just native`.

Note: Only the debug build has been adjusted, as this should not matter for the releases.

### DIFF
--- a/justfile
+++ b/justfile
@@ -75,8 +75,8 @@ android-release:
 
 # Build Rust library for iOS (debug mode)
 ios:
-    cd mobile/native && cargo lipo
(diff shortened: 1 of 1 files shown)


klochowicz:
I have tested it (it does not invalidate the builds anymore). `cargo clean` still works, as it's a subdirectory in the `target` dir.

luckysori (approved):

Issue #983, opened 2023-07-23
Receive screen validation is changing inconsistently 


holzeis:
Tested on the latest test version. 

1. Fund wallet with 56300 sats
2. Opened and closed a few positions
3. Opened and closed the app a couple of times in between.
4. Tried to receive funds, which shows me initially that I can receive up to 66,113
5. Enter some amount and the value what I can receive up to changes to 0.

![Screenshot 2023-07-23 at 09 03 57](https://github.com/get10101/10101/assets/382048/7dfe4f42-608d-4e78-be50-ad034ba22b3a)

da-kami:
I think it would be best to action this ticket https://github.com/get10101/10101/issues/1008, reproduce the bug in a test and *then* add a fix.

Restioson:
Trying to reproduce this locally right now and it seems to work for me. Here is my history:

![image](https://github.com/get10101/10101/assets/6688948/2e3d1560-8e4d-4b06-be42-2983e85fd9e2)
![image](https://github.com/get10101/10101/assets/6688948/877a119b-016a-4b85-b867-5870ea5ba803)

The latest payment is after receiving money.

If you fund with 56 300 sats and then later transfer out 22 413 sats to get to 33 887 sats, it is my understanding that you should be able to receive 22 413 sats, correct?

Restioson:
New test case.

Wallet history:
![image](https://github.com/get10101/10101/assets/6688948/d4afe2ef-9887-40cc-9ab4-304ea42dc00b)

Initial allowance:
![image](https://github.com/get10101/10101/assets/6688948/c2a357ce-ae65-41e2-bcfa-5233bc35bbd1)

Updated allowance:
![image](https://github.com/get10101/10101/assets/6688948/b8d9c109-b35d-4906-a5f7-b8d9e967bcff)

Restioson:
Is anyone able to reproduce this? I can't from a clean setup

Restioson:
> It is definitely reproducible that the receivable amount is changing once an amount is entered.

Yes, I think we can action that swiftly but the other half of the ticket (which is somewhat more important I'd argue) I can't reproduce yet

holzeis:
Got ya! Well I think this might have only been a special case of the bug.

With can't receiving I meant the amount I can receive is 0. purely uk validation.

klochowicz:
@holzeis I've updated the ticket name, please doublecheck if it still conveys the problem

holzeis:
Sounds good 👍

[...]

PR #1094, opened 2023-08-14
Wumbo trade: aka configurable channel size

bonomat:
With this change we can configure the max channel size aka, we can enable wumbo trades.

I've also made a small frontend change: the info text is gone and replaced with an info icon. This solved the problem that the `maxChannelSize` is not immediately available .

https://github.com/get10101/10101/assets/224613/293eaa2b-dd3c-44a5-bdaa-3cd476d599a0

This will help us with running the increased-channel-size experiment.

resolves https://github.com/get10101/10101/issues/983

### DIFF
--- a/mobile/lib/common/amount_text_input_form_field.dart
+++ b/mobile/lib/common/amount_text_input_form_field.dart
@@ -2,8 +2,9 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
 import 'domain/model.dart';
+import 'modal_bottom_sheet_info.dart';
 
-class AmountInputField extends StatelessWidget {
+class AmountInputField extends StatefulWidget {
   const AmountInputField(
       {super.key,
       required this.controller,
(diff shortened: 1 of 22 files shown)


mobile/lib/common/amount_text_input_form_field.dart
```diff
@@ -2,56 +2,72 @@ import 'package:flutter/material.dart';
...
       required this.value,
+      required this.isLoading,
+      this.infoText,
       this.validator});
 
   final Amount value;
   final TextEditingController controller;
   final String label;
   final String hint;
+  final String? infoText;
```
Restioson, line 23: Could we use the nullness of infoText as a proxy for whether it has finished loading? I.e, it is null until it has loaded, so we can get rid of the separate tracker bool

Restioson (commented):
LGTM except that the loading spinner is comically large :sweat_smile:

mobile/lib/common/amount_text_input_form_field.dart
bonomat: I thought about it first, but then decided against it because `loading` could be generally useful.

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -46,6 +46,9 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
 
   ChannelInfo? channelInfo;
 
+  /// The max channel capacity as received by the LSP or if there is an existing channel
```
klochowicz, line 49: 🤔 not sure if I follow this comment.

Does it mean that this is etheir:
- (if no channel open): potential channel capacity advertised by the LSP?
- (if channel open): current channel capacity (which could be different from the currently advertised by the LSP, right?)

klochowicz (approved):
nice one!

thanks for cleaning up the settings with `serde_as`, that's a neat trick. 🫠

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
bonomat: Yes, that's what I meant :)

PR #1097, opened 2023-08-15
chore: Use automatic codesign

klochowicz:
The previous setting never worked for me.
Setting it to automatic, as the error message suggested, works when I'm
deploying locally.

If this works for others & for our fastlane / release process, it would be
    really convenient.

### DIFF
--- a/mobile/ios/Runner.xcodeproj/project.pbxproj
+++ b/mobile/ios/Runner.xcodeproj/project.pbxproj
@@ -445,11 +445,9 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CODE_SIGN_STYLE = Manual;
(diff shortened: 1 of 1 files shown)


klochowicz:
@bonomat @holzeis can you please check whether this works on your machines & when releasing the test app?

klochowicz:
Apparently using fastlane might be the way to go instead.
@bonomat , can we still build for ios for flutter with hot reload capabilities ? (like with `flutter run`?). Because this is what I tried to fix in this PR.

holzeis:
> @bonomat @holzeis can you please check whether this works on your machines & when releasing the test app?

We need the manual signing for the github actions pipeline - automatic signing only ever worked on the local machine. see also 
- https://medium.com/team-rockstars-it/the-easiest-way-to-build-a-flutter-ios-app-using-github-actions-plus-a-key-takeaway-for-developers-48cf2ad7c72a
- https://damienaicheh.github.io/flutter/github/actions/2021/04/22/build-sign-flutter-ios-github-actions-en.html

It's very critical to touch this signing configuration - Did you test to build the app from the pipeline on a fork?

bonomat (changes requested):
Please only make these changes if the CI release build worked in a fork. 
As Richie pointed it out - we need manual code signing for the CI. If you want to manually release it locally, you can temporarly enable automatic signing.

PR #1099, opened 2023-08-15
Display channel status in app

### LINKED ISSUE #1031: Give app user an indication that their channel is being closed on-chain

luckysori:
Fixes #1031.

There are many reasons why the app might end up with a channel (vanilla or split) closed on-chain, particularly at this stage of the project.

As such we want to give some visibility to the app user, so that they know what to expect if they run into this situation. The main objective is to prevent loss of funds, something that could happen if the user deletes the app data before certain transactions have been confirmed.

Interestingly, for the DLC channel branch of the split channel, the user could choose to trust 10101, as the coordinator will eventually publish the correct CET, partially paying to the user's on-chain wallet.

The LN branch is more dangerous, as the user must wait for the funds to be claimed from the commitment transaction by their own LN node. Since we currently do not back up this data, deleting the app data at the wrong time could lead to loss of funds.

The main limitation of this patch, captured in a TODO, is that we currently do not model the LN _closing_ channel state in our shadow channel representation. This means that we will end up telling the user that the LN channel is closed before it is fully closed, which is dangerous.

The user can also use the channel status information to know if they are able to open a new channel _with the same app_. If the channel status is `NotOpen` they can create a new JIT channel. This information is useful because they will know that they can continue using 10101 whilst their previous channel might still be being closed in the background.

Also:

- Register new incoming JIT channels with a `UserChannelId`. This is important for this patch because otherwise we cannot identify the channel, which prevented us from marking it as closed in its shadow representation.

---

Notes:

- I struggled a lot with settling on a way to model this just combining the shadow channel state and the subchannel state. It will be even harder once we use the DLC channel state and the DLC contract state. For now I don't think it is a requirement because the user can for now trust the coordinator to publish the right CET, so modelling that stuff perfectly is not as crucial.

- I think we need to model the `Closing` state of the LN channel for this feature to be complete. I decided to leave it out because that is not an obvious change. This PR moves us closer to user safety, but it's not sufficient. I opened a ticket to track this: https://github.com/get10101/10101/issues/1105.

---

If the LN-DLC channel is closing the wallet status thermostat icon turns red indicating that there's a problem:

red-thermostat

On the status page I've added a section for the channel status, which renders an explanation if the channel is closing:

channel-status-screen

Some decisions I made:

- Not to remove the position from the trade screen because it's still useful information. Also the user won't be able to do anything with it because new orders will fail instantly.
- Not to show a more invasive screen because it would require a lot more design work. I thought it would be a good idea to reuse a page we already integrated.


crates/ln-dlc-node/src/ln/common_handlers.rs
```diff
@@ -254,8 +254,16 @@ where
...
-            .notify_ln_channel_closed(channel_id)?;
+        match node
+            .sub_channel_manager
+            .notify_ln_channel_closed(channel_id)
+        {
+            Ok(()) => {}
+            Err(dlc_manager::error::Error::InvalidParameters(msg)) => {
+                tracing::debug!("Irrelevant LDK closure notification: {msg}");
+            }
+            e @ Err(_) => e.context("Failed to notify subchannel manager about LDK closure")?,
```
klochowicz, line 265: `e @ Err(_)` - what sort of wizardry is this? never seen it before! 😳

mobile/lib/common/app_bar_wrapper.dart
```diff
@@ -17,9 +19,13 @@ class AppBarWrapper extends StatelessWidget {
...
     const appBarHeight = 35.0;
 
+    ChannelStatusNotifier channelStatusNotifier = context.watch<ChannelStatusNotifier>();
+
     var actionButtons = [
       IconButton(
-        icon: const Icon(Icons.thermostat),
+        icon: channelStatusNotifier.isClosing()
+            ? const Icon(Icons.thermostat, color: Colors.red)
+            : const Icon(Icons.thermostat),
```
klochowicz, line 28: This is really cool ❤️! I'll make use of colours for other states when after this PR gets it.

mobile/lib/common/channel_status_notifier.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:f_logs/f_logs.dart';
+import 'package:flutter/material.dart';
+import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
+import 'package:get_10101/common/application/event_service.dart';
+import 'package:get_10101/ffi.dart';
+
+class ChannelStatusNotifier extends ChangeNotifier implements Subscriber {
```
klochowicz, line 7: 🙃  would you mind inserting a doc comment here about the purpose of this class

mobile/lib/common/status_screen.dart
```diff
@@ -20,39 +21,78 @@ class _StatusScreenState extends State<StatusScreen> {
...
+                text: "Your channel with 10101 is being closed on-chain!\n\n",
+                style: TextStyle(fontWeight: FontWeight.bold)),
+            TextSpan(
+                text:
+                    "Your Lightning funds will return back to your on-chain wallet after some time. You will have to reopen the app at some point in the future so that your node can claim them back.\n\n"),
+            TextSpan(
+                text:
+                    "If you had a position open your payout will arrive in your on-chain wallet soon after the expiry time. \n")
+          ]))));
+    }
```
klochowicz, line 87: 🙃 I like extraction of the widgets, but could we please make it into a function instead of variable? then it's easier to find it in the IDE

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -0,0 +1,197 @@
...
+use ln_dlc_node::node::rust_dlc_manager::subchannel::SubChannel;
+use std::borrow::Borrow;
+use std::time::Duration;
+
+const UPDATE_CHANNEL_STATUS_INTERVAL: Duration = Duration::from_secs(5);
+
+/// The status of the app channels.
+#[derive(Debug, Clone, Copy)]
+pub enum ChannelStatus {
+    /// No channel is open.
```
klochowicz, line 16: have you considered having an `Unknown` channel status as the enum variant?

That's how I went with `ServiceStatus`. I'm not sure if any of the 2 options is better, but it would be good to align on one approach.

Personally, I prefer not sending `null` over (although apparently we have null safety in Dart3), but this might be my bias after years of pain in C/C++ world. It also ensures we can stringify enums nicer etc, as we can't forget about that wariant in these places  - with `null`, things can get a bit ad-hoc and inconsistent as they might be done in multiple places.

Happy to be convinced otherwise though!

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -0,0 +1,197 @@
...
+    // TODO: Add support for `LnForceClosing` status and ensure that channel status remains in
+    // `LnDlcForceClosing` whilst the LN channel is still closing.
+    // LnForceClosing,
+    /// The LN-DLC channel is in the process of being force-closed.
+    LnDlcForceClosing,
+    /// The LN-DLC channel is in an inconsistent state.
+    Inconsistent,
+}
+
+pub async fn track_channel_status(node: impl Borrow<Node>) {
```
klochowicz, line 44: 🤔 why not simply: 
```suggestion
pub async fn track_channel_status(node: &Node) {
```
is there any additional benefit for using your construct?

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -0,0 +1,197 @@
...
+    let subchannels = node.list_dlc_channels()?;
+
+    let status = derive_ln_dlc_channel_status(ln_channel, subchannels);
+
+    Ok(status)
+}
+
+// TODO: We currently only look at the state of the subchannel. We should also take into account the
+// state of the DLC channel and the contract. Otherwise we won't be able to convey that the CET is
+// not yet confirmed.
```
klochowicz, line 91: have you considered sending them all, 
and simplifying the representation in the UI?

seems like it's mostly an UX issue to convey enough information to the user to not overwhelm them, but still for example show enough for the power users and testers to be able to help them.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -0,0 +1,197 @@
...
+                    SubChannelState::ForceClosing => ChannelStatus::LnDlcForceClosing,
+                },
+            }
+        }
+        // If the LN channel does not have an ID associated with it, we assume that it's still not
+        // open, for it should not be usable yet.
+        Channel {
+            channel_id: None, ..
+        } => ChannelStatus::NotOpen,
+        // TODO: Do not share shadow channel model between app and coordinator.
```
klochowicz, line 161:
```suggestion
```

I don't think such a TODO is needed here; we have already encoded in the type system that the problematic state can't be achieved, so it's firmly in "good enough" in my book  :)

mobile/lib/common/channel_status_notifier.dart
```diff
@@ -0,0 +1,31 @@
...
+  bridge.ChannelStatus? latest;
+
+  ChannelStatusNotifier();
+
+  bridge.ChannelStatus? getChannelStatus() {
```
klochowicz, line 12: 🤔 it might be nicer if we didn't need to return a nullable type here (see other discussion topic )

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let status = derive_ln_dlc_channel_status(ln_channel, subchannels);
+
+    Ok(status)
+}
+
+// TODO: We currently only look at the state of the subchannel. We should also take into account the
+// state of the DLC channel and the contract. Otherwise we won't be able to convey that the CET is
+// not yet confirmed.
+fn derive_ln_dlc_channel_status(
```
klochowicz, line 92: 🔧 this is a great function to cover with unit test to showcase how it behaves.

klochowicz (approved):
thanks for tackling this! as it's by no means a small PR, I'm happy to deal with the comments in a follow-up as nothing is critical.

mobile/native/src/ln_dlc/channel_status.rs
luckysori: AFAIK you can't pass an `Arc` in then.

luckysori:
> > Some decisions I made:
> > 
> > * Not to remove the position from the trade screen because it's still useful information. Also the user won't be able to do anything with it because new orders will fail instantly.
> 
> What information do you think is still useful on the position?

I prefer to keep the position because at least it shows that it's still open, even if the LN-DLC channel is not functional. It would be a bit surprising if the position disappeared, and you would no longer be able to know what you did.

In any case, it's not exactly ideal so we can iterate on this. Perhaps displaying a `Closing on-chain` tag would suffice?

crates/ln-dlc-node/src/ln/common_handlers.rs
luckysori: And it's cooler than I thought! Look at the second example here: https://doc.rust-lang.org/rust-by-example/flow_control/match/binding.html.

mobile/lib/common/app_bar_wrapper.dart
luckysori: I'm glad you like it. I wasn't sure if it was gonna be received well haha.

mobile/native/src/ln_dlc/channel_status.rs
luckysori: I definitely considered it! Since you are suggesting it too I'm happy to try to model it differently.

mobile/native/src/ln_dlc/channel_status.rs
luckysori: I guess I already chose to build a complete model in the backend rather than forwarding everything to the UI and letting it decide what to do with it. My main motivation was to keep the difficult things in Rust.

I think both approaches do require some relatively complex modelling already and it would only get worse if we introduce the two other dimensions mentioned here.

With what you propose I do like being able to show more detailed info to expert users though. In general I am open to solving this issue differently, but we can probably roll with this PR at this stage?

mobile/native/src/ln_dlc/channel_status.rs
luckysori: I can move the TODO as this is just the site of a symptom, but I do think it's still relevant.

The fact that we have to handle this enum variant in the app means that we have not encoded in the type system that we can't have this variant here. Unless I'm missing something, by adding the `unreachable` we are just saying "trust us programmers, we haven't made a mistake".

mobile/native/src/ln_dlc/channel_status.rs
luckysori: I agree. I extracted it so that it could be tested, but I was too lazy at the time.

mobile/lib/common/status_screen.dart
luckysori: Do you mind refactoring this yourself? I'm not sure I understand what function you want to define here. I only extracted it so that we could add the last widget depending on the channel status.

mobile/native/src/ln_dlc/channel_status.rs
luckysori: I'm gonna be pragmatic here and say that we can add this unit test later because it's a pain to build the `SubChannel` coming from `rust-dlc`.

I think to make it more testable we need to map the foreign types to other module-private ones which only have the data we care about for `derive_ln_dlc_channel_status`.

crates/ln-dlc-node/src/ln/common_handlers.rs
klochowicz: really nice. I will add it to my repertoire ;)

mobile/lib/common/status_screen.dart
klochowicz: no probs! will do that as a follow-up

mobile/native/src/ln_dlc/channel_status.rs
klochowicz: fair enough!  I guess one can also pass in an  `Arc` by doing `track_channel_status(&*node)`, but I see the appeal now.

mobile/native/src/ln_dlc/channel_status.rs
klochowicz: good point. It might just not be worth it the added complexity, especially when there's only consumer of this library.

I don't think there's anything wrong with this approach, so we might just roll with it 👍

for our future selves: 
The potential future exploration would be more like  a free function in the Rust API that takes the complex type and simplifies it, so that we can split the responsibilities of:
1) monitoring the states
2) simplifying the representation for the user (the latter could have its own unit test.

mobile/native/src/ln_dlc/channel_status.rs
klochowicz: fair enough, if it's not trivial we can move it.
Maybe if we split this function into 2? keep the complex representation still internal to the module, but at least test the function that "simplifies" the state? 
(doesn't need to happen now).

mobile/lib/features/trade/order_change_notifier.dart
```diff
@@ -31,11 +31,11 @@ class OrderChangeNotifier extends ChangeNotifier implements Subscriber {
       orders[order.id] = order;
 
       sortOrderByTimestampDesc();
+
+      notifyListeners();
```
klochowicz, line 35: nice catch!!!!

klochowicz (approved):
LGTM!

PR #1100, opened 2023-08-15
fix(justfile): Remove skipping funding

klochowicz:
At times, the script would pass with a non-functioning faucet.
The few seconds saved on skipping funding is not worth the extra debugging
    effort, as `just fund` catches these occurrences.

### DIFF
--- a/justfile
+++ b/justfile
@@ -106,15 +106,7 @@ run-local-android args="":
     --dart-define="REGTEST_FAUCET=http://${LOCAL_IP}:8080" --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
 
 fund args="":
-    #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


luckysori (commented):
Damn, I was excited to discover today that we had already added this...

Maybe we can pass an argument to `all` to selectively call `fund`? When developing it is overwhelmingly more common to not need to fund the faucet again. Those seconds do add up over time.

Restioson:
> Maybe we can pass an argument to all to selectively call fund

What do you mean? Like the -f option for just fund?

klochowicz:
> Damn, I was excited to discover today that we had already added this...
> 
> Maybe we can pass an argument to `all` to selectively call `fund`? When developing it is overwhelmingly more common to not need to fund the faucet again. Those seconds do add up over time.

If you'd like to keep it, please fix it - there is a problem currently when you lose way more time if things are down and the current version of the script does not detect problems accurately. I am quite confident that I'm losing more seconds by just doublechecking whether things should work instead of knowing that things work.

IMHO it would be a much better saving to fix mockito mock generation which takes ages instead.

luckysori (approved):

PR #1102, opened 2023-08-15
fix: it was printing the type but not the balance as intented

bonomat:


mobile/lib/features/wallet/share_invoice_screen.dart
```diff
@@ -37,7 +37,7 @@ class _ShareInvoiceScreenState extends State<ShareInvoiceScreen> {
     WalletInfo info = context.watch<WalletChangeNotifier>().walletInfo;
     final bridge.Config config = context.read<bridge.Config>();
 
-    FLog.debug(text: "Refresh receive screen: ${info.balances.onChain}");
+    FLog.debug(text: "Refresh receive screen: ${info.balances.onChain.sats}");
```
Restioson, line 40: formatSats could be used here maybe

1010Tom (approved):
LGTM

Issue #872, opened 2023-06-30
Add transaction details to transactions in wallet history


holzeis:
Upon clicking a transaction one should see some details to the transaction. There are different types of transactions for which a details view is required.

- [ ] On chain transaction (potentially add a link to mempool)
- [ ] Off-chain transaction (payment or received)
  - [ ] Invoice description (moved from #886)
- [ ] Position transaction (open and close)
- [ ] JIT Channel opening fee (potentially add a link to the funding transaction on mempool)
- [ ] Order matching fee

Figma mock: 

https://www.figma.com/file/Sr3SfZ46sZstKjSwftmTtc/10101-v1-(incl-bitcoin.design)?type=design&node-id=7375%3A45885&mode=design&t=tFOk1Qr2JvlNIaBY-1

E.g. like this

https://github.com/get10101/10101/assets/224613/1e989c69-6ba3-4c32-9117-f7f2ba850024

Restioson:
With reference to the 'fee' row in the lightning payment example - is it not the case that lightning fees are per-channel and not per-transaction?

holzeis:
> With reference to the 'fee' row in the lightning payment example - is it not the case that lightning fees are per-channel and not per-transaction?

You are paying fees for outgoing transactions. The amount of fees is defined through the route (channels) you are using.

Restioson:
> > With reference to the 'fee' row in the lightning payment example - is it not the case that lightning fees are per-channel and not per-transaction?
> 
> You are paying fees for outgoing transactions. The amount of fees is defined through the route (channels) you are using.

Ah, I see, thanks!

Restioson:
Do we want to model other timestamps than just the time we 'first learned' about this payment? E.g invoice creation timestamp

Restioson:
With #1213, we will model all information required to fulfill this ticket. Now, onto prettifying it as per the Figma :)

Restioson:
Nit on figma: "you sent -1000" sats would imply that you _received_ 1000 sats, wouldn't it? Maybe we can convey this in a different way?

Restioson:
> Nit on figma: "you sent -1000" sats would imply that you _received_ 1000 sats, wouldn't it? Maybe we can convey this in a different way?

Thought: we could have outgoing transactions have red amount text and incoming have green?

PR #1103, opened 2023-08-15
Show JIT channel fee in wallet history

### LINKED ISSUE #982: Show JIT Channel fee as fee in the wallet history
Currently the JIT Channel fee is only shown as payment with the payment hash as description. We should align that with the order matching fees. e.g.

- [ ] Dedicated icon (or the same as for order matching)
- [ ] Description that the fee was for opening an inbound channel.

Screenshot 2023-07-23 at 06 34 06

Restioson:
image

Fixes #982

### DIFF
--- a/coordinator/src/node/channel_opening_fee.rs
+++ b/coordinator/src/node/channel_opening_fee.rs
@@ -6,6 +6,7 @@ use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::ThirtyTwoByteHash;
 use lightning::ln::PaymentHash;
 use lightning_invoice::Invoice;
+use ln_dlc_node::channel::JIT_FEE_INVOICE_DESCRIPTION_PREFIX;
 use ln_dlc_node::PaymentInfo;
 
 impl Node {
@@ -13,14 +14,12 @@ impl Node {
         &self,
         amount: u64,
         funding_txid: String,
-        description: Option<String>,
         expiry: Option<u32>,
     ) -> Result<Invoice> {
-        let invoice = self.inner.create_invoice(
-            amount,
-            description.unwrap_or_default(),
-            expiry.unwrap_or(180),
-        )?;
+        let description = format!("{JIT_FEE_INVOICE_DESCRIPTION_PREFIX}{funding_txid}");
+        let invoice = self
+            .inner
+            .create_invoice(amount, description, expiry.unwrap_or(180))?;
         let payment_hash = invoice.payment_hash().into_32();
         let payment_hash_hex = payment_hash.to_hex();
 
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -208,7 +208,6 @@ pub struct InvoiceParams {
 pub struct OpenChannelFeeInvoiceParams {
     pub amount: u64,
     pub channel_funding_txid: String,
-    pub description: Option<String>,
     pub expiry: Option<u32>,
 }
 
@@ -237,12 +236,7 @@ pub async fn get_open_channel_fee_invoice(
 ) -> Result<String, AppError> {
     let invoice = state
         .node
-        .channel_opening_fee_invoice(
-            params.amount,
-            params.channel_funding_txid,
-            params.description,
-            params.expiry,
-        )
+        .channel_opening_fee_invoice(params.amount, params.channel_funding_txid, params.expiry)
         .await
         .map_err(|e| AppError::InternalServerError(format!("Failed to create invoice: {e:#}")))?;
 
--- a/crates/ln-dlc-node/src/channel.rs
+++ b/crates/ln-dlc-node/src/channel.rs
@@ -14,6 +14,10 @@ use std::str::FromStr;
 use time::OffsetDateTime;
 use uuid::Uuid;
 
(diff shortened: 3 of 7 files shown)


holzeis:
@Restioson That was not the intention of that ticket 🙈. I suggested a similar display as for the matching fees 😅 

Screenshot 2023-08-16 at 13 40 49

This change goes more into the direction of showing transaction details. https://github.com/get10101/10101/issues/872 Did you maybe linked the wrong ticket?

Restioson:
I didn't link the wrong ticket - check main since #1054 has been merged. This PR extends #1054 by separating JIT opening fee from regular payments

crates/orderbook-commons/src/lib.rs
```diff
@@ -21,7 +21,11 @@ pub use crate::price::Price;
 pub use crate::price::Prices;
 
 /// The prefix used in the description field of an order-matching fee invoice to be paid by a taker.
-pub const FEE_INVOICE_DESCRIPTION_PREFIX_TAKER: &str = "taker-fee-";
+pub const FEE_INVOICE_DESCRIPTION_PREFIX_TAKER: &str = "taker-fee";
```
luckysori, line 24: 📣 I appreciate that it's probably nicer to remove this dash, but I think this is a breaking change. Probably worth avoiding that headache.

As a side note, I am pretty sure the prefix idea that I introduced is not very good and there were better alternatives 🙈

crates/orderbook-commons/src/lib.rs
```diff
@@ -21,7 +21,11 @@ pub use crate::price::Price;
...
+
+/// The prefix used in the description field of an JIT channel opening invoice to be paid by the
+/// client.
+pub const JIT_FEE_INVOICE_DESCRIPTION_PREFIX: &str = "jit-channel-fee";
```
luckysori, line 28: 🤔 Hmmm, this is weird because the JIT channel is orthogonal to the orderbook. Maybe put this in `ln-dlc-node` for now?

luckysori (commented):
LGTM! I left a couple of remarks that are worth considering.

crates/orderbook-commons/src/lib.rs
Restioson: > 📣 I appreciate that it's probably nicer to remove this dash, but I think this is a breaking change. Probably worth avoiding that headache.

Yea, most definitely :sweat_smile: 

Maybe for now I should add a legacy match.

The reason I added this was because there isn't really a way to match on multiple prefixes cleanly, whereas if you define the syntax for a structured payment invoice description as 

```
{tag} {additional_data}
```

then you can just `split_once` by space and match on the tag to get what kind of invoice it is.

However, it's worth noting that both this implementation _and_ the previous implementation could lead to some strange results. 

For instance, what would happen if an invoice were paid by a mischievous (or even malicious) user with the description `taker-fee-NotAnOrderId?!/.sidfj`? At best this would display wrong, worse it could crash the app, and at worst it could introduce a security vulnerability.

It's probably worth workshopping how this should work :sweat_smile:

crates/orderbook-commons/src/lib.rs
Restioson: True, can do!

Restioson:
I've pushed a commit to revert, and happy to squash up once we've resolved the discussion. Please don't merge until then

holzeis:
> I didn't link the wrong ticket - check main since #1054 has been merged :slightly_smiling_face: This PR extends #1054 by separating JIT opening fee from regular payments

Hmm.. on main it's now simply shown as an outgoing payment instead of showing directly in the title that it is a JIT channel open fee payment. It seems inconsistent that we show a matching fee in the title but not for the JIT channel opening fee.

Restioson:
I might not be understanding you correctly here so bear with me as I repeat what I understand from what you said :sweat_smile: 

> Hmm.. on main it's now simply shown as an outgoing payment instead of showing directly in the title that it is a JIT channel open fee payment

Agree, that's what this PR is changing.

> It seems inconsistent that we show a matching fee in the title but not for the JIT channel opening fee

On this branch, it does show as channel open fee. Let me show another screenshot in case it wasn't clear:

image

holzeis:
> I might not be understanding you correctly here so bear with me as I repeat what I understand from what you said :sweat_smile: 
> 
> 
> 
> > Hmm.. on main it's now simply shown as an outgoing payment instead of showing directly in the title that it is a JIT channel open fee payment
> 
> 
> 
> Agree, that's what this PR is changing.
> 
> 
> 
> > It seems inconsistent that we show a matching fee in the title but not for the JIT channel opening fee
> 
> 
> 
> On this branch, it does show as channel open fee. Let me show another screenshot in case it wasn't clear:
> 
> 
> 
> image
> 
> 

Oh that's great! I just saw the other screenshot and thought this was the only change 😅

Well done 👍 looks nice. Sorry for the confusion 😅

Restioson:
No problem!

So I guess it's just the query about prefixes vs splitting to resolve? Maybe that should go in a follow-up and just stick to prefixes for now, though. In that case I can rebase this and squash it up before merge

holzeis (approved):
LGTM 👍

PR #1106, opened 2023-08-16
Ensure that broadcast TXs are always stored

### LINKED ISSUE #1101: Ensure that transactions are registered in DB table
Not all transactions are recorded because we have more than trait implementation of `BroadcasterInterface`.

luckysori:
Fixes #1101.

We needed to align the implementations of the two traits where we broadcast transactions. There were several places where we were not storing broadcast transactions because of this.

### DIFF
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -1,4 +1,5 @@
 use crate::fee_rate_estimator::EstimateFeeRate;
+use crate::node::Storage;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Error;
@@ -40,6 +41,7 @@ where
     settings: RwLock<WalletSettings>,
     fee_rate_estimator: Arc<F>,
     locked_outpoints: Mutex<Vec<OutPoint>>,
+    node_storage: Arc<dyn Storage + Send + Sync + 'static>,
 }
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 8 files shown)


holzeis (approved):
LGTM 👍

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -40,6 +41,7 @@ where
     settings: RwLock<WalletSettings>,
     fee_rate_estimator: Arc<F>,
     locked_outpoints: Mutex<Vec<OutPoint>>,
+    node_storage: Arc<dyn Storage + Send + Sync + 'static>,
```
klochowicz, line 44: can this be done with static dispatch? We probably lose a bit performance here, but I guess we're not using that too often for us to matter.

crates/ln-dlc-node/src/transaction.rs
```diff
@@ -4,12 +4,60 @@ use std::fmt::Display;
 use std::fmt::Formatter;
 use time::OffsetDateTime;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq)]
```
klochowicz, line 7: Why would we want to `Copy` this type?

klochowicz (approved):

crates/ln-dlc-node/src/ldk_node_wallet.rs
luckysori: I just followed the same pattern we were using in `ln_dlc_wallet`. But that doesn't mean it's necessarily good!

You are right that we could use static dispatch, but then we have to propagate the type parameter and I will be unhappy. Ideally we get rid of the ill-conceived `Storage` trait and then we don't have to worry about it anymore :D

crates/ln-dlc-node/src/transaction.rs
luckysori: It means we don't have to `clone` here:

https://github.com/get10101/10101/blob/ee2d5e8c67a0c109145a234fc83b97c5e5928080/crates/ln-dlc-node/src/shadow.rs#L58

PR #1111, opened 2023-08-16
Model more transaction history details 

### LINKED ISSUE #1140: Show failed transactions in transaction history
It looks like we are not showing failed off-chain transactions in the wallet. It would be good to see them.

Restioson:
Addresses #872 by modelling more history details for different kinds of events (transactions, trades, fees, etc).

I don't think this PR should address the Figma mock beyond adding all the info that it has to the display and that should rather be done in a followup, but happy to change this if desired.

- [x] Model on-chain payment data
  - [x] Confirmations
  - [x] Fee
      - [x] Check if the fee reported is correct   
- [x] Model lightning payment data
    - ~~[ ] From node~~ I do not believe this to be possible without an explicit proof-of-payment protocol: https://bitcoin.stackexchange.com/a/85555
    - [x] Invoice description
    - [x] Payment preimage
    - [x] Invoice itself
    - [x] Show expiry time 
    - [x] Show expired payments
        - [x]  Test
    - [x] Handle failed payments (fixes #1140)
        - [x] Test
- [x] JIT fee - blocked on #1103
- [x] Order matching fee
    - [x] Order ID

Questions:
- [x] What should we do about unpaid invoices in the database?

Related: #1103

### DIFF
--- a/crates/ln-dlc-node/src/lib.rs
+++ b/crates/ln-dlc-node/src/lib.rs
@@ -99,6 +99,7 @@ pub struct PaymentInfo {
     pub flow: PaymentFlow,
     pub timestamp: OffsetDateTime,
     pub description: String,
+    pub invoice: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy)]
@@ -142,6 +143,7 @@ impl From<Invoice> for PaymentInfo {
                 InvoiceDescription::Direct(direct) => direct.to_string(),
                 InvoiceDescription::Hash(hash) => hash.0.to_hex(),
             },
+            invoice: Some(value.to_string()),
         }
     }
 }
--- a/crates/ln-dlc-node/src/node/invoice.rs
+++ b/crates/ln-dlc-node/src/node/invoice.rs
@@ -64,18 +64,21 @@ where
     pub fn create_interceptable_invoice(
         &self,
         amount_in_sats: Option<u64>,
-        invoice_expiry: u32,
+        invoice_expiry: Option<Duration>,
         description: String,
         final_route_hint_hop: RouteHintHop,
     ) -> Result<Invoice> {
+        let invoice_expiry = invoice_expiry
+            .unwrap_or_else(|| Duration::from_secs(lightning_invoice::DEFAULT_EXPIRY_TIME));
         let amount_msat = amount_in_sats.map(|x| x * 1000);
         let (payment_hash, payment_secret) = self
             .channel_manager
-            .create_inbound_payment(amount_msat, invoice_expiry, None)
+            .create_inbound_payment(amount_msat, invoice_expiry.as_secs() as u32, None)
             .map_err(|_| anyhow!("Failed to create inbound payment"))?;
         let invoice_builder = InvoiceBuilder::new(self.get_currency())
             .payee_pub_key(self.info.pubkey)
             .description(description)
+            .expiry_time(invoice_expiry)
             .payment_hash(sha256::Hash::from_slice(&payment_hash.0)?)
             .payment_secret(payment_secret)
             .timestamp(SystemTime::now())
--- a/crates/ln-dlc-node/src/node/invoice.rs
+++ b/crates/ln-dlc-node/src/node/invoice.rs
@@ -101,6 +101,12 @@ where
             })
             .map_err(|_| anyhow!("Failed to sign invoice"))?;
         let invoice = Invoice::from_signed(signed_invoice)?;
+
+        self.storage.insert_payment(
+            payment_hash,
+            invoice.clone().into(),
+        )?;
+
         Ok(invoice)
     }
 
--- a/crates/ln-dlc-node/src/node/invoice.rs
+++ b/crates/ln-dlc-node/src/node/invoice.rs
@@ -102,10 +102,8 @@ where
             .map_err(|_| anyhow!("Failed to sign invoice"))?;
         let invoice = Invoice::from_signed(signed_invoice)?;
 
-        self.storage.insert_payment(
-            payment_hash,
-            invoice.clone().into(),
-        )?;
+        self.storage
+            .insert_payment(payment_hash, invoice.clone().into())?;
 
         Ok(invoice)
     }
@@ -227,6 +225,7 @@ where
                 flow: PaymentFlow::Outbound,
                 timestamp: OffsetDateTime::now_utc(),
                 description,
+                invoice: Some(format!("{invoice}")),
             },
         )?;
 
--- a/mobile/native/src/db/models.rs
+++ b/mobile/native/src/db/models.rs
@@ -672,6 +672,8 @@ pub(crate) struct PaymentInsertable {
     pub updated_at: i64,
     #[diesel(sql_type = Text)]
     pub description: String,
+    #[diesel(sql_type = Nullable<Text>)]
+    pub invoice: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, FromSqlRow, AsExpression)]
@@ -782,6 +784,7 @@ pub(crate) struct PaymentQueryable {
     pub created_at: i64,
     pub updated_at: i64,
     pub description: String,
+    pub invoice: Option<String>,
 }
 
 impl PaymentQueryable {
@@ -812,6 +815,7 @@ impl From<(lightning::ln::PaymentHash, ln_dlc_node::PaymentInfo)> for PaymentIns
             created_at: timestamp,
             updated_at: timestamp,
             description: info.description,
+            invoice: info.invoice,
         }
     }
 }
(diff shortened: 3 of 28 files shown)


Restioson:
With https://github.com/get10101/10101/pull/1111/commits/ec9b7de468b8ab96ad039d001768a165a6e72604, I insert the invoice into the `payments` table upon creation. This is required to know the description once the payment has been claimed. However, it also leads to the payment just sticking around in the DB if it's unpaid. I'm not sure what we want to happen here from a UX perspective if the user creates an invoice and it isn't paid.

Restioson:
This is more-or-less ready, aside from the above comment

holzeis:
> With https://github.com/get10101/10101/pull/1111/commits/ec9b7de468b8ab96ad039d001768a165a6e72604, I insert the invoice into the `payments` table upon creation. This is required to know the description once the payment has been claimed. However, it also leads to the payment just sticking around in the DB if it's unpaid. I'm not sure what we want to happen here from a UX perspective if the user creates an invoice and it isn't paid (expiry date & periodic sweep to remove? notify user?).

I would remove it once it has expired. I think our invoices are already created with an expiry.

crates/ln-dlc-node/src/lib.rs
```diff
@@ -142,6 +143,7 @@ impl From<Invoice> for PaymentInfo {
                 InvoiceDescription::Direct(direct) => direct.to_string(),
                 InvoiceDescription::Hash(hash) => hash.0.to_hex(),
             },
+            invoice: Some(format!("{value}")),
```
luckysori, line 146: 🔧 A bit simpler?
```suggestion
            invoice: Some(value.to_string()),
```

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -103,6 +103,19 @@ where
...
             .map_err(|_| anyhow!("Failed to sign invoice"))?;
         let invoice = Invoice::from_signed(signed_invoice)?;
+
+        self.storage.insert_payment(
+            payment_hash,
+            PaymentInfo {
+                preimage: Some(
+                    self.channel_manager
+                        .get_payment_preimage(payment_hash, payment_secret)
+                        .map_err(|_| anyhow!("Failed to get payment preimage for new invoice"))?,
```
luckysori, line 113: 🤔 I'm not sure it's such a good idea to store the preimage from the perspective of the invoice generator. If this is leaked before the final HTLC is claimed then people can lose money. I would leave this as `None` for now and consider removing it from the model altogether.

mobile/lib/main.dart
```diff
@@ -333,6 +333,7 @@ class _TenTenOneAppState extends State<TenTenOneApp> {
       FlutterNativeSplash.remove();
     }
     await walletChangeNotifier.refreshWalletInfo();
+    FLog.info(text: "Bitcoin address: ${walletChangeNotifier.service.getUnusedAddress()}");
```
luckysori, line 336: ❓Do we want to keep this?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -278,29 +278,42 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
...
+    let blockchain_height = node.get_blockchain_height()?;
     let on_chain = on_chain.iter().map(|details| {
         let net_sats = details.received as i64 - details.sent as i64;
 
         let (flow, amount_sats) = if net_sats >= 0 {
-            (api::PaymentFlow::Outbound, net_sats as u64)
+            (api::PaymentFlow::Inbound, net_sats as u64)
         } else {
-            (api::PaymentFlow::Inbound, net_sats.unsigned_abs())
+            (api::PaymentFlow::Outbound, net_sats.unsigned_abs())
```
luckysori, line 288: 👍 Oh, god. Thanks!

luckysori (approved):
LGTM, except for the preimage. I would leave that out completely :D 

Can we get some screenshots too? 👀

mobile/lib/main.dart
Restioson: Oops, this was just for testing

crates/ln-dlc-node/src/node/invoice.rs
Restioson: Fair enough - I just added it because it was on the Figma mock - @bonomat thoughts?

Restioson:
> Can we get some screenshots too? 👀

I can make some, although it isn't pretty as this is just the modeling side of things and not the display

luckysori:
> > Can we get some screenshots too? 👀
> 
> I can make some, although it isn't pretty as this is just the modeling side of things and not the display

Fair enough, we can look at the screenshots once that's nicer.

crates/ln-dlc-node/src/lib.rs
Restioson: Right, I forgot about that trait impl :)

crates/ln-dlc-node/src/node/invoice.rs
Restioson: I'll change it to use the preimage from PaymentClaimed event - is this sufficient? If I understand correctly, this is after funds have been transferred, so it should be fine (?)

Restioson:
> I would remove it once it has expired. I think our invoices are already created with an expiry.

See these lines in `native::ln_dlc::create_invoice`:
https://github.com/get10101/10101/blob/e30aa48796ee940b8a6f27a44d388cd15f79b0ec/mobile/native/src/ln_dlc/mod.rs#L578-L584

How should we interpret this? Apparently `lightning` (the crate) adds two hours (gleaned by reading the code). 

I'm not sure if we can rely on the +2 hours, but maybe we just should?

Restioson:
image
image
image
image

@luckysori some screenshots! Missing on chain because it's a pain :laughing: but I can take one if we need. It requires restarting the app to show up.

Restioson:
> I would remove it once it has expired. I think our invoices are already created with an expiry.

Actually, I'm not so sure of this. I think an expired invoice is semantically a failed payment (or maybe we can add a new HTLCStatus for this called expired), or something similar.

User story for this (proposed) feature: 
- As a wallet user,
- I want to check on the history of invoices I've sent
- so I know if the transfer was paid, expired, or failed

If the user creates an invoice, sends it off, and it isn't paid, they should be able to go back and see that it wasn't paid instead of silently disappearing, imo

Restioson:
I've pushed https://github.com/get10101/10101/pull/1111/commits/720fc3675dd8a47cf18809be82ebcf476ab4efc3 but it's untested as I still need to wait the hour for it to expire (I could make one expire on purpose but since it's Friday, I may as well wait till Monday :wink:)

mobile/native/src/db/models.rs
```diff
@@ -672,6 +672,8 @@ pub(crate) struct PaymentInsertable {
     pub updated_at: i64,
     #[diesel(sql_type = Text)]
     pub description: String,
+    #[diesel(sql_type = Nullable<Text>)]
+    pub invoice: Option<String>,
```
holzeis, line 676: 🤔 Do we really need the full invoice here? The payment already contains most of its properties anyways.

Restioson: We are displaying the raw invoice, so I would say yes. Maybe this means we can remove the description column though for instance.

holzeis: but is the raw invoice really useful? why not simply display the contents of the invoice which is already mostly present on the payment record.

Restioson:
Example of a failed payment

image

Restioson:
I think this is ready, pending review. That is to say, I am happy with it now.

mobile/native/src/db/models.rs
Restioson: > but is the raw invoice really useful? why not simply display the contents of the invoice which is already mostly present on the payment record.

I simply implemented it because it's there in the Figma - maybe the user wants to inspect for their own details or for debugging purposes. In any case, IMO its helpful for our own debugging to have the real invoice around. This is the authoritative truth of what the payment is.

Ultimately, maybe @bonomat can clarify the reasoning here as he made the Figma mock?

bonomat: > Ultimately, maybe @bonomat can clarify the reasoning here as he made the Figma mock?

Mhmm.. better show and and don't need it than not show it and need it? 🤷‍♂️

I've modeled it after screens I've seen on other lightning apps.

Restioson:
Are we happy with this? If so, I can resolve conflicts for merge

luckysori (approved):
LGTM, thank you!

bonomat (approved):
I guess we can make it look nice in a follow-up PR :)

Restioson:
Ah, maybe I should have a crack at displaying the fees for lightning transactions

klochowicz:
@Restioson can we try to merge this soon to avoid the need for rebasing? we can always keep improving in a follow-up, and these changes can be useful for others :)

Restioson:
> @Restioson could we try to merge this soon to avoid the need for rebasing? we can always keep improving in a follow-up, and these changes can be useful for others :)


Sure, I'll do that today!

PR #1114, opened 2023-08-17
chore: merge settings screen and fix transition direction

bonomat:
This bugged me for a while: 
We had two settings routes which lead to the same page with dividers which did not make sense. Also the transition to settings came in from the wrong side.

New flow: 

Simulator Screen Recording - iPhone 14 Pro Max - 2023-08-17 at 17 32 09


mobile/lib/common/app_bar_wrapper.dart
```diff
@@ -71,3 +70,21 @@ class AppBarWrapper extends StatelessWidget {
...
+      const begin = Offset(-1.0, 0.0);
+      const end = Offset.zero;
+      const curve = Curves.ease;
+
+      var tween = Tween(begin: begin, end: end).chain(CurveTween(curve: curve));
+
+      return SlideTransition(
+        position: animation.drive(tween),
+        child: child,
+      );
```
klochowicz, line 87: nice animation! ❤️

klochowicz (approved):
I believe that original idea was to have different settings for 'trade` and for 'wallet`, and depending on which tab was active, go to particular settings.
I was never fan of this, so I'm happy to see it go 😁

I'll happily do the same with the status screen (I modelled this one after settings to remain consistent).

Issue #1079, opened 2023-08-07
Recover app on-chain wallet from seed


bonomat:
Implement a way to create a new 10101 app based on an existing 10101 seed.

For example, after a fresh install we confront the user with the choice to either create a new wallet or to import an existing seed.

Caveat: this only affects the on-chain wallet for now. To recover the channels we need to [back up more data](https://github.com/get10101/10101/issues/1098) and allow the user to import this data on a fresh install.

luckysori:
We should also seriously consider backing up the `rust-dlc` and `rust-lightning` data needed to reconstruct the node remembering all the keys needed to recover funds that are not yet owned by the on-chain wallet. This would make it much safer for people to have to reinstall their app.

luckysori:
I put it back into `New` so that we can pay attention to this next time we plan an iteration.

@holzeis and I had a chat about this yesterday and we agreed that **the fact that our seed phrase is non-standard** is something worth addressing soon.

We currently use BIP39 to generate a mnemonic phrase, but instead of using it directly we hash it in two different ways to derive one seed for the on-chain wallet and another for LDK's `KeysManager`. The advantage of this is that we only have to back up one seed for two purposes. In practice this is not actually useful because (1) we have not yet implemented a way to recover a 10101 wallet using this seed and, more importantly, (2) you would not be able to recover the whole 10101 wallet just with this seed[^1].

The main disadvantage of the current approach is that a user cannot use their mnemonic phrase to recover their on-chain funds using a different wallet. If they do want to recover their on-chain funds they would need to use something like https://github.com/get10101/bip39-to-bip32/ to learn the BIP32 key of their on-chain wallet. Then they could import it into a tool such as Sparrow. This whole process is unnecessarily complicated.

I would propose to simply generate one BIP39 seed for the on-chain wallet and a separate seed for LDK and to **only expose the on-chain seed to the user in the `Back up your wallet` screen**. The LDK seed is not useful on its own AFAIK and we should probably back it up with the rest of the LDK data[^1].

[^1]: https://github.com/get10101/10101/issues/1098.

luckysori:
@luckysori to split up this ticket.

PR #1115, opened 2023-08-18
Allow switching between mainnet and regtest networks.

klochowicz:
It needs a restart, which is clearly communicated in the app.
Eventually we might have "expert" mode that gets unlocked with a toggle or something, so it can't be accidentally clicked.

Removes the need of deploying two versions and mostly removes the need for
dart_define from the release process.

(I almost finished vergen support that would remove other dart_define, will add it in the morning!). This would unblock us to proper continuous delivery for iOS.

note: I've aligned the distance between the buttons to be consistent (10px) after running the screenshot tool.

Screenshot 2023-08-17 at 10 31 50 pm

Screenshot 2023-08-17 at 10 31 42 pm

### DIFF
--- a/justfile
+++ b/justfile
@@ -87,14 +87,13 @@ ios-release:
 run args="":
     #!/usr/bin/env bash
     cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
-    --dart-define="REGTEST_FAUCET=http://localhost:8080" --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2" \
+    --dart-define="REGTEST_FAUCET=http://localhost:8080" \
+    --dart-define="DEV_MODE=true"
 
 # Run against our public regtest server
 run-regtest args="":
     #!/usr/bin/env bash
     cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
-    --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
-    --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}"
 
 [unix]
 run-local-android args="":
@@ -102,6 +101,7 @@ run-local-android args="":
     LOCAL_IP=$({{get_local_ip}})
     echo "Android app will connect to $LOCAL_IP for 10101 services"
     cd mobile && flutter run {{args}} --dart-define="COMMIT=$(git rev-parse HEAD)" --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
+    --dart-define="DEV_MODE=true" \
     --dart-define="ESPLORA_ENDPOINT=http://${LOCAL_IP}:3000" --dart-define="COORDINATOR_P2P_ENDPOINT=02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9@${LOCAL_IP}:9045" \
     --dart-define="REGTEST_FAUCET=http://${LOCAL_IP}:8080" --dart-define="COORDINATOR_PORT_HTTP=8000" --flavor local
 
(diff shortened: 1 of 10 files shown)


bonomat (changes requested):
I'm not sure we should merge this as is and I wish we would have had the discussion before putting effort into this. 

I'm sorry if these efforts were triggered because of my post [on elements](https://matrix.to/#/!LeqkIiXcIvhcREDcDP:matrix.org/$yYpMVYgUdjj-Ez95xi_vsjAG1b5am-K611UGTS_3tXs?via=matrix.org&via=breadpirates.chat&via=lightninghackers.com).

A few points I can think of from the top of my heads which we need to think through: 
 
- backup of data: right now it is not safe to remove/uninstall the app because it will delete your data directory. Because of this I would not want to run the regtest build (which needs cleansing once in a while) next to my production environment.
- incompatible APIs: a continues delivery build might not be always compatible with the latest release. Because of this we potentially need to run a "released"-regtest version next to our continuously deployed version? In this case the feature would make sense but we still would need to release a separate bundle to test against our CD version? 
- incompatible DB migrations: I sometimes install old regtest versions of the app just to see if a bug was present in an older version. For that a full-wipe might be necessary (see point 1). If not, a DB downgrade/migration undo, might not be safe in all cases.

Unfortunately I don't think the time is ready for this feature and we have things which are more important than that.

klochowicz:
> I'm not sure we should merge this as is and I wish we would have had the discussion before putting effort into this.
> 

That's very reasonable, I should have marked it as a draft, as this was a proof of concept rather than ready for merging idea. For quite some time I believed (now I'm not sure whether that belief was shared) that it would be very hard to switch between the mainnet/testnet versions of the app (earlier called "demo mode" in ItchySats). 
When I realised I could apply the change *only after restart* I realised it might be a solvable problem, so I tried it out. 
My intention was to spark a discussion, but perhaps it was not a very good idea without opening at least a GitHub discussion around it.
I've created one now: https://github.com/get10101/10101/discussions/1127

> I'm sorry if these efforts were triggered because of my post [on elements](https://matrix.to/#/!LeqkIiXcIvhcREDcDP:matrix.org/$yYpMVYgUdjj-Ez95xi_vsjAG1b5am-K611UGTS_3tXs?via=matrix.org&via=breadpirates.chat&via=lightninghackers.com).

It catalysed my thinking into this direction, but ultimately, I'm to blame for going outside of the scope for this iteration, for which I take responsibility. See discussion points for the reasons I tried it out. 
> 
> A few points I can think of from the top of my heads which we need to think through:
> 
> * backup of data: right now it is not safe to remove/uninstall the app because it will delete your data directory. Because of this I would not want to run the regtest build (which needs cleansing once in a while) next to my production environment.

Off the top of my head, I would say that I would try to restructure our flow to *not* remove the app / uninstall in these situations - the most obvious way would be to always provide a fresh regtest after switching to mainnet and back (with new seed and db).
Of course, it goes without saying that our backup workflow is an issue and should be dealt with and at the very least we should tackle https://github.com/get10101/10101/issues/1079 as well.

> * incompatible APIs: a continues delivery build might not be always compatible with the latest release. Because of this we potentially need to run a "released"-regtest version next to our continuously deployed version? In this case the feature would make sense but we still would need to release a separate bundle to test against our CD version?

I thought about this one for a while, and realised that we could fully embrace CD and wouldn't need to do releases anymore whilst in the closed beta. (other than for marketing purposes, and done retrospectively, when we feel particular build is good).
I shared these (somewhat controversial) ideas around that in the discussion thread.

> * incompatible DB migrations: I sometimes install old regtest versions of the app just to see if a bug was present in an older version. For that a full-wipe might be necessary (see point 1). If not, a DB downgrade/migration undo, might not be safe in all cases.

for this one, we could consider:  
 - testing on different phone than production builds
 - using Simulator
 - using native builds (desktop) 

🤔 I think that it can *already* happen, one of our testers (or us) might try to go to earlier versions themselves and inadvertently break their app.

Ideally, we could at least show a message upon failed migration that states something like
"DB migration failed. Please switch back to build XXXX which worked on this database" to make the recovery / export easier.


> 
> Unfortunately I don't think the time is ready for this feature and we have things which are more important than that.

Ultimately it's your call, however please have a read over the reasons this could speed up delivery / experiment workflow / bug fixing and many other aspects of the app. The investment might actually pay off.

klochowicz:
Closing this PR because it doesn't seem like it would get sufficient buy-in.
It can be revived at a later stage if we decide to go into this direction.

PR #1117, opened 2023-08-18
Add Lightning-node capabilities to the maker

### LINKED ISSUE #926: Add channel capabilities to the maker
The PR https://github.com/get10101/10101/pull/785 removed the channel capabilities from the maker. 

This was useful because we could test another service opening channels with the coordinator besides LND. 

We will also need this in the long run.

### LINKED ISSUE #1087: Numbers don't always add up after funding in e2e tests
See e.g.: https://github.com/get10101/10101/actions/runs/5801924993/job/15727244113) (had different tests exhibiting this behaviour).

before it fails at the assertion, we print out what's the actual value - in this case: 

`2023-08-08T21:18:13.559000Z  INFO tests_e2e::setup: Successfully funded app with faucet funded_amount=49847 ln_balance=49779`

grepping for this string shows that other tests (which pass) get slightly more into their lightning wallet;

`2023-08-08T21:17:39.712423Z  INFO tests_e2e::setup: Successfully funded app with faucet funded_amount=49847 ln_balance=49847`

To unblock myself i'll plaster over this by temporarily relaxing the assertion.

luckysori:
Fixes #926.
Fixes #1087.


maker/src/routes.rs
```diff
@@ -34,12 +35,13 @@ pub fn router(
 
     Router::new()
         .route("/", get(index))
-        .route("/api/newaddress", get(get_unused_address))
+        .route("/api/new-address", get(get_unused_address))
```
bonomat, line 38: Why the rename? In the coordinator it's `newaddress`

bonomat (approved):
Very nice. Thanks

maker/src/routes.rs
luckysori: I was just in the mood to make it more readable (and less typeable 😛), but I definitely don't have to change it.

luckysori: I have walked back the change. What I really wanted to rename was the `dlc_channels` endpoint to use kebab-case, but that's completely unrelated haha.

klochowicz (approved):
nice work!

PR #1118, opened 2023-08-18
Check for new DLC-related messages 25 times more often

### LINKED ISSUE #1107: Speed up subchannel setup and settlement
At the moment we are checking for new `rust-dlc`-related messages every 5 seconds. This can really slow down subchannel setup and settlement.

To speed this up we should either:
  1. Reduce the timeout (and change the internals of `rust-dlc` to replace a `Mutex` with an `RwLock` so that checking more frequently is safe).
  2. Implement our own `DlcMessageHandler` so that we can expose a stream of messages that consumers can `await` on, rather than having to poll the internal queue periodically.

We can start with 1 and eventually transition to 2.

luckysori:
Fixes #1107.

I think this could be very, very helpful in preventing disconnection-related issues in production. In theory this could improve the speed of the worst-case scenario for the current version of the subchannel open/close protocol by ~25 seconds!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Add support for push notifications
-- Added new setting to coordinator to configure max channel size to traders
+- Add support for push notifications.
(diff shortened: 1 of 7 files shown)


luckysori:
Hmm, maybe our e2e tests are catching the fact that we are starving the producers by checking so often! I'll take a look on Monday.

PR #1119, opened 2023-08-18
chore: bump DLC fee rate

bonomat:
A fix fee rate of 4 is not safe anymore as the mempool has been congested since a long time purging transactions with a low fee. We clearly need anchor outputs, but this is an attempt to at least be able to force close positions

### DIFF
--- a/crates/ln-dlc-node/src/config.rs
+++ b/crates/ln-dlc-node/src/config.rs
@@ -16,7 +16,7 @@ pub const LIQUIDITY_MULTIPLIER: u64 = 2;
 /// Sats/vbyte rate for position
 ///
 /// The coordinator and the app have to align on this to agree on the fees.
-pub const CONTRACT_TX_FEE_RATE: u64 = 4;
+pub const CONTRACT_TX_FEE_RATE: u64 = 7;
 
 /// The speed at which we want a transaction to confirm used for feerate estimation.
 ///


bonomat:
@luckysori / @holzeis : any thoughts on this?

luckysori (approved):
LGTM.

We could alternatively let the coordinator calculate it depending on the state of mempool, although I don't know how useful that is given that the contract expiry is set to a day later. With rollover coming soon we will also be able to update it! Which is pretty much the same as Lightning renegotiating fees.

One correction: I think the coordinator sets the fee unilaterally. This constant is also used in the app's frontend, but they don't _need_ to agree.

crates/ln-dlc-node/src/config.rs
```diff
@@ -16,7 +16,7 @@ pub const LIQUIDITY_MULTIPLIER: u64 = 2;
 /// Sats/vbyte rate for position
 ///
 /// The coordinator and the app have to align on this to agree on the fees.
-pub const CONTRACT_TX_FEE_RATE: u64 = 4;
+pub const CONTRACT_TX_FEE_RATE: u64 = 7;
```
holzeis, line 19: 🤔 that's very conservative.. I would go for something like 20, to be on the safe side - since this value can't be changed for a stuck dlc channel afterwards.

holzeis:
> LGTM.
> 
> We could alternatively let the coordinator calculate it depending on the state of mempool, although I don't know how useful that is given that the contract expiry is set to a day later. With rollover coming soon we will also be able to update it! Which is pretty much the same as Lightning renegotiating fees.
> 
> One correction: I think the coordinator sets the fee unilaterally. This constant is also used in the app's frontend, but they don't _need_ to agree.

We have to keep in mind that the fees are substracted from the margin the user puts in. So changing that during a rollover will have an impact on the contracts or the execution price.

crates/ln-dlc-node/src/config.rs
bonomat: 7 seems to look ok-ish for now, but it's not ideal as we do not know how the mempool situation develops. 
20 seems to be an overkill for me as we would always overpay.

bonomat:
> We have to keep in mind that the fees are substracted from the margin the user puts in. So changing that during a rollover will have an impact on the contracts or the execution price.

Good point....

bonomat:
> LGTM.
> 
> We could alternatively let the coordinator calculate it depending on the state of mempool, although I don't know how useful that is given that the contract expiry is set to a day later. With rollover coming soon we will also be able to update it! Which is pretty much the same as Lightning renegotiating fees.
> 
> One correction: I think the coordinator sets the fee unilaterally. This constant is also used in the app's frontend, but they don't _need_ to agree.

That;s a good idea, but for that to be really useful we should have relatively short running DLCs, as it's hard to predict what fee rate we will have in a week or more.

Issue #1068, opened 2023-08-04
Rollover DLC


bonomat:
In order to have perpetual futures (CFDs) we need to extend our DLCs. 

In a nutshell, a rollover is closing the current DLC and reopening it again with a new expiry and the same values.
Note: _same_ is not 100% correct but for this ticket the funding rate is out of scope.  

Funding rate ticket: https://github.com/get10101/10101/issues/1069

Note: we only rollover a non-expired position.

holzeis:
`rust-dlc` already exposes an API to update a dlc channel. see also https://github.com/p2pderivatives/rust-dlc/blob/master/dlc-manager/src/channel_updater.rs#L954.

Using that API we don't have to collaboratively close and reopen the position, which simplifies the protocol significantly. The api also expects a `payout_amount` parameter which can come in handy when implementing the funding rate.

holzeis:
```mermaid
sequenceDiagram

actor u as User
participant a as App
participant c as Coordinator

Note right of a: Position is about to expire. e.g. in 2 hours

c->>u: Push notification: Come online to rollover your position.
u->>a: Starts 10101 app
a-->>u: Show dialog for extending position

alt Extend position 
a->>c: extend position
c->>a: Renew offer
a-->>c: Renew accept
c->>a: Renew confirm
a-->>c: Renew finalize
c->>a: Renew revoke
a->>a: Update position
a-->>u: Show finished extending position
else Close position
a->c: collaborative close position
a->>a: Remove position
a-->>u: Show finished close position
end






```

Issue #1108, opened 2023-08-16
`OrderSubmissionStatusDialog` can hang forever


luckysori:
If the subchannel open/close protocol does not finish (either successfully or unsuccessfully), the `OrderSubmissionStatusDialog` will remain open indefinitely.

This is a problem because the same dialog says that the user should not close the app, but that's the only way to move forward.

luckysori:
> I feel like this is technically not a dialog/UI issue, but rather a backend issue. We need some kind of timeout to prevent this, and I think it should come in the form of an event or failure type from the backend to the frontend.

Good point. @holzeis and I did discuss this yesterday and I think we didn't come to a consensus on whether this should be solved in the UI or the backend.

I tend to agree with you that it should be driven by the backend though.

Restioson:
The issue I have with it being driven from the frontend is that it's difficult to tell whether it's just hanging or really just taking a long time. I feel like the logic for introspecting the backend (and deciding whether hang or taking long) should live there and not in the frontend. 

For _now_ it might just be "is it taking longer than 1 minute", but in the future there may be indicators of progress.

bonomat:
@holzeis : I've unnasigned you and move it into `todo`. Imho it is clear what todo here.

klochowicz:
> @holzeis : I've unnasigned you and move it into `todo`. Imho it is clear what todo here.

Is it clear? with the recent changes the whole existance of the Dialog is questionable - perhaps it would be enough to have just a  dismissible dialog with the result of the operation? then the problem goes away entirely.

bonomat:
> Is it clear? 

If it is not clear to you, feel free to ask and we can add more context

> with the recent changes the whole existance of the Dialog is questionable - perhaps it would be enough to have just a dismissible dialog with the result of the operation? then the problem goes away entirely.

As long as the protocol execution is not instant I'm in favor of having this dialog while it is being executed.

luckysori:
Did you fix this recently, @holzeis?

holzeis:
Not for the order submission, but it should be fairly easy to add.

However that would be on the frontend side, but I guess that's the best we can do.

Issue #1029, opened 2023-07-30
Make position expiry configurable


holzeis:
At the moment any position expires at the next days midnight. Meaning that a position is at least 24 hours and at most 48 hours open.

This logic is hard-coded. It would be great if we could configure that, not only for testing purposes.

klochowicz:
Duplicate of #935

holzeis:
> Duplicate of #935 

This is not a duplicate. The order expiry is unrelated to the position expiry.

klochowicz:
> > Duplicate of #935
> 
> This is not a duplicate. The order expiry is unrelated to the position expiry.

it's different, but of course it's related; they should be changed via the same API, and most likely tackled at the same time.

bonomat:
I don't think so. 
The order is the thing which is in the orderbook, the position is the thing the trader has locally. 
Both expiries are different, e.g. the order may expire in 1 minute, while the position expires in ~24h.

klochowicz:
Aren't they both essentially terms, on which maker wants to offer anything? 
Of course they're used in different times, but I would fully expect for both being configurable from the maker side. 
After all, in the ideal world they'd set how long their orders are active and how long to set positions up for (minus platform constraints).

bonomat:
Hi @klochowicz : you've been assigned to this since 2 weeks. Are you working on it? If not, please unassign yourself

klochowicz:
> Hi @klochowicz : you've been assigned to this since 2 weeks. Are you working on it? If not, please unassign yourself

I was 90% in, but then noticed that it was removed from iteration *after* last objective alignment, so I focused on more pressing things. I will post it today, because it's close to completion.

bonomat:
Sounds good. Thanks for clarification.

bonomat:
Deprioritized for now.

klochowicz:
I've linked the work that I've done on this ticket, to whoever wants to pick it up.

holzeis:
Can we close that ticket, now that we are using a logic to calculate the expiry time?

bonomat:
I think so, changing this calculation logic with a config might be a bit too complex.

bonomat:
I think so, changing this calculation logic with a config might be a bit too complex.

Issue #977, opened 2023-07-21
Perpetual Futures


holzeis:
At the moment we have an expiry timestamp on every position. After it has been expired we will trigger a collaborative closure once the user is back online. With perpetual positions we want to add the possibility to "rollover" a position and extend it. 
```[tasklist]
### Tasks
- [ ] https://github.com/get10101/10101/issues/1068
- [ ] https://github.com/get10101/10101/issues/1069
- [ ] https://github.com/get10101/10101/issues/1044
- [ ] https://github.com/get10101/10101/issues/1185
- [ ] https://github.com/get10101/10101/issues/1070
- [ ] https://github.com/get10101/10101/issues/1071
- [ ] https://github.com/get10101/10101/issues/133
- [ ] https://github.com/get10101/10101/issues/1123
- [ ] https://github.com/get10101/10101/issues/1184
```

holzeis:
Below a proposal of how perpetual futures could look like for the users. I'

```mermaid
sequenceDiagram

actor u as User
participant a as App
participant c as Coordinator

Note right of a: Position is about to expire. e.g. in 2 hours

c->>u: Push notification: Come online to rollover your position.
u->>a: Starts 10101 app
a-->>u: Show dialog for extending position

alt Extend position 
a-->>u: Show finished extending position
else Close position
a->c: collaborative close position
a->>a: Remove position
a-->>u: Show finished close position
end






```

klochowicz:
what's the advantage of asking the user a question instead of extending it automatically until the user wants to close the trade?

bonomat:
Mostly done :)

PR #1122, opened 2023-08-18
feat: Rollover DLC

holzeis:
This introduces the rollover of a signed dlc. It uses the renew offer api of the rust dlc. 

- The payout amount is calculated from the original average entry price - as we are rolling over a position nothing should change. However, this is most likely the place where we would like to add paying the funding rates for the rollover.

- The state `Rollover` is introduced to the position on the coordinator and the app side. It helps to show a rollover in progress as with the dlc creation a lot of messages are exchanged in between. We also need that state to prevent the coordinator from accidentally setting the position to `Closed` as the underlying contract is Closed after the new one is set. Therefore the `temporary_contract_id` is also updated.

I left a some todos regarding the dlc messages. It would be nice if we could process them similar to the lightning messages through the or a event handler. That would allow for the design to be nicer decoupled.

resolves https://github.com/get10101/10101/issues/1068


coordinator/src/rollover.rs
```diff
@@ -0,0 +1,361 @@
...
+        // Sets the position state to rollover indicating that a rollover is in progress.
+        let mut connection = self.pool.get()?;
+        db::positions::Position::rollover_position(
+            &mut connection,
+            rollover.counterparty_pubkey.to_string(),
+            &rollover.maturity_time(),
+        )
+    }
+
+    /// Finalizes the rollover protocol with the app setting the position to open.
```
luckysori, line 114: 🔧 This talks about the app but this is coordinator code.

coordinator/src/routes.rs
```diff
@@ -260,6 +264,32 @@ pub async fn post_trade(
     Ok(invoice.to_string())
 }
 
+#[instrument(skip_all, err(Debug))]
+#[autometrics]
+pub async fn rollover(
```
luckysori, line 269: ❓Are we certain that we want to trigger rollover by calling an endpoint on the coordinator? I'm not sure, but I'm not sure it is wrong either.

crates/tests-e2e/tests/rollover_position.rs
```diff
@@ -0,0 +1,39 @@
...
+        .rx
+        .position()
+        .map(|p| PositionState::Rollover == p.position_state)
+        .unwrap_or(false));
+    wait_until!(test
+        .app
+        .rx
+        .position()
+        .map(|p| PositionState::Open == p.position_state)
+        .unwrap_or(false));
```
luckysori, line 38: 🔧 I think we need a few more assertions to ensure that rollover has actually taken place. In theory we could got to the rollover state and back to the open state because rollover failed?

mobile/native/src/ln_dlc/node.rs
```diff
@@ -189,6 +192,37 @@ impl Node {
             }
         };
 
+        // todo(holzeis): It would be nice if dlc messages are also propagated via events, so the
+        // receiver can decide what events to process and we can skip this component specific logic
+        // here.
+        if let Message::Channel(channel_message) = msg.clone() {
```
luckysori, line 198: 🔧 I think we can avoid this potentially very expensive `clone` by only cloning the bits that are needed.

coordinator/src/rollover.rs
holzeis: well the coordinator talks with the app to finalize the rollover protocol. what would you want to change here?

coordinator/src/routes.rs
holzeis: I proposed the flow like this https://github.com/get10101/10101/issues/977#issuecomment-1682080659. Based on the learnings from closing expired positions, I think its best to let the app trigger this action.

coordinator/src/rollover.rs
luckysori: I thought you were meaning to say that the _coordinator_ would set the position to open.

coordinator/src/routes.rs
luckysori: I think your reasoning makes sense, but it's weird to have to grant permission to keep your position alive when we are offering perpetual swaps.

It's unfortunate because I think this is mostly motivated by implementation details (brittle protocol and frequent disconnects). We can roll with this for now, but we might want to revisit this down the line.

crates/tests-e2e/tests/rollover_position.rs
holzeis: > 🔧 I think we need a few more assertions to ensure that rollover has actually taken place. 
I've added an assertion on the extended expiry, what else would you suggest to assert on?

> In theory we could got to the rollover state and back to the open state because rollover failed?
Only if we implement it, but currently there is no flow from `Rollover` to `Open` if the the renew protocol fails. That reminds me that we need to consider the renew protocol as well for our timeout handling. (see also https://github.com/get10101/10101/issues/1108)

coordinator/src/rollover.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub fn event_id(&self) -> String {
+        let maturity_time = self.maturity_time().unix_timestamp();
+        format!("{}{maturity_time}", self.contract_symbol)
+    }
+
+    /// Calculates the maturity time based on the current expiry timestamp.
+    ///
+    /// todo(holzeis): this should come from a configuration https://github.com/get10101/10101/issues/1029
+    pub fn maturity_time(&self) -> OffsetDateTime {
+        let tomorrow = self.expiry_timestamp.date() + Duration::days(2);
```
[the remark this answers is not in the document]
holzeis: Given that there is already a ticket to rework that, I am opting to skip that remark and address it with #1029.

coordinator/src/rollover.rs
```diff
@@ -0,0 +1,361 @@
...
+use time::OffsetDateTime;
+use trade::ContractSymbol;
+
+#[derive(Debug, Clone)]
+struct Rollover {
+    counterparty_pubkey: PublicKey,
+    contract_descriptor: ContractDescriptor,
+    expiry_timestamp: OffsetDateTime,
+    margin_coordinator: u64,
+    margin_trader: u64,
```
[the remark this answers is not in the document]
holzeis: Everything except the expiry timestamp will remain the same though. Would you still add the `original_` prefix?

coordinator/src/rollover.rs
holzeis: Yes it does set it from Rollover to Open 😅

coordinator/src/rollover.rs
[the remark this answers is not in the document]
bonomat: In the future we might want to change the margin, i.e. deduct the funding rate from it. Alternatively we could let the user pay from _the outside_ using lightning invoices. 
The remaining values will remain the same.

PR #1124, opened 2023-08-19
Change thermostat colour when services are down

klochowicz:
Thermostat changes colour to yellow if any of them is unknown, and to red if at least one of the services is down.


luckysori (approved):
LGTM (caveat: I'm still not very confident with Flutter).

bonomat (approved):
As usual, for UI changes screenshots/videos would be useful. 

Otherwise, looks good to me :)

mobile/lib/common/status_icon_button.dart
```diff
@@ -0,0 +1,48 @@
...
+    final serviceStatusNotifier = context.watch<ServiceStatusNotifier>();
+
+    final overallStatus = serviceStatusNotifier.overall();
+
+    return IconButton(
+      icon: channelStatusNotifier.isClosing() || overallStatus == rust.ServiceStatus.Offline
+          ? const Icon(Icons.thermostat, color: Colors.red)
+          : overallStatus == rust.ServiceStatus.Unknown
+              ? const Icon(Icons.thermostat, color: Colors.yellow)
+              : const Icon(Icons.thermostat),
```
Restioson, line 23: This might be cleaner as an if/else or switch - this is a bit nested which makes it harder to read

Restioson, line 23: What colour is this by default?

mobile/lib/common/status_screen.dart
```diff
@@ -67,25 +65,26 @@ class _StatusScreenState extends State<StatusScreen> {
               ),
             ],
           )),
+      Visibility(
```
Restioson, line 68: This is cool, I didn't know about this widget :)

Restioson (approved):

mobile/lib/common/status_screen.dart
klochowicz: yeah, saves painful array constructs - we've used it in a couple of places in the past

mobile/lib/common/status_icon_button.dart
klochowicz: the theme colour (blends in, does not stand out.)

klochowicz: because of the indentation I find it quite clear, but of course it might be subjective. I'll keep it in mind for the future.

mobile/lib/common/status_screen.dart
Restioson: Makes sense!

PR #1112, opened 2023-08-16
fixes: get a releasable build

bonomat:
... more to come...

PR #1135, opened 2023-08-21
fix: wrong provisioning profile

bonomat:
Release [1.2.1 failed](https://github.com/get10101/10101/actions/runs/5928785922/job/16075130621)

Did not test it on the fork but locally. The problem was that when I've provided the [last fix ](https://github.com/get10101/10101/pull/1112/commits/005fd65f29d8375e61722069fceb301f147272b7) for fastlane I forgot to change the provisioning profile in the xcode config.

Uploading version 1.2.1 from local setup as this is being created


holzeis (approved):
LGTM, I guess 😅

PR #1136, opened 2023-08-21
Fix Flutter hot restart

klochowicz:
It still shows some warnings in Firebase (they will be fixed by #1104 ), but at least it unblocks rapid Flutter development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,13 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # this might remove tools that are actually needed,
+          # when set to "true" but frees about 6 GB
+          tool-cache: false
+          large-packages: false
       - uses: actions/checkout@v3
       - uses: extractions/setup-just@v1
       - name: Setup rust toolchain
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,9 @@ jobs:
         with:
           flutter-version: "3.10.5"
           channel: "stable"
+          cache: true
+          cache-path: '${{ runner.tool_cache }}/flutter/:channel:-:version:-:arch:' # optional, change this to specify the cache path
+          architecture: x64 
       - name: Install FFI bindings
         run: just deps-gen
       - name: Generate FFI bindings
--- a/justfile
+++ b/justfile
@@ -260,7 +260,10 @@ services: docker run-coordinator-detached run-maker-detached fund
 
 # Run everything at once (docker, coordinator, native build)
 # Note: if you have mobile simulator running, it will start that one instead of native, but will *not* rebuild the mobile rust library.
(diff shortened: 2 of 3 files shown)


luckysori (approved):

Restioson:
I can't seem to open or close positions after a hot restart - the buttons do nothing and are greyed out.

bonomat (commented):
It looks like hot restart is working without this fix.

klochowicz:
> I can't seem to open or close positions after a hot restart - the buttons do nothing and are greyed out.

It takes a while for things to work properly, in my experience sometimes ~30s.

klochowicz:
> It looks like hot restart is working without this fix.
> Every now and then we get an error that some files are corrupt though.



> It looks like hot restart is working without this fix.
> Every now and then we get an error that some files are corrupt though.

just double-checked it and it does not (at least on macOS desktop build):

here is the relevant log:
```

flutter: | 2023-08-22 11:43:55.861 | LogLevel.INFO | Attempt 3 failed: FfiException(RESULT_ERROR, Failed to start the backend

Caused by:
    IO error: could not acquire lock on "/Users/mariusz/Library/Containers/finance.get10101.app/Data/Library/Application Support/finance.get10101.app/regtest/db": Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }, null)

flutter: | 2023-08-22 11:43:55.863 | LogLevel.ERROR | Max retries reached, backend could not start.

Lost connection to device.
```

recent master, with freshly wiped directories.
perhaps you got confused because we try 3 times to start the backend with some timeout in between before we stop the app completely now? 🤔

klochowicz:
> I can't seem to open or close positions after a hot restart - the buttons do nothing and are greyed out.

what's the status screen saying? "unknown" state of orderbook?

Restioson:
> what's the status screen saying? "unknown" state of orderbook?

It's night here so I'll check tomorrow - but if I recall, it was a theme-coloured indicator. I don't think I clicked into the screen, or if I did I don't remember what it said.

bonomat:
> just double-checked it and it does not (at least on macOS desktop build):

Best is to always test on the simulator because at the end we build for the phone and not for desktop.

.github/workflows/ci.yml
```diff
@@ -128,6 +128,13 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
+      - name: Free Disk Space (Ubuntu)
```
bonomat, line 131: It would have been nice to not have this in this PR but in a separate one.

Restioson:
> Best is to always test on the simulator because at the end we build for the phone and not for desktop.

While that is true, it's usually a bit easier to develop on the desktop than on the simulator (which is where hot restart is most used). Maybe this is just for me, since my emulator is pretty slow

klochowicz:
> > just double-checked it and it does not (at least on macOS desktop build):
> 
> Best is to always test on the simulator because at the end we build for the phone and not for desktop.

I test on both before submitting the PR and primarily on Simulator when I'm doing layout. 

when I need a rapid iteration loop whilst developing, I start with desktop - why would I slow myself down?

PR #1137, opened 2023-08-22
fix: Adjust navigator usage within Flutter

klochowicz:
Using `useRootNavigator` flag was fragile, because it assumed that
GoRouter will *always* remain be the top navigator.
This has changed with introducing `BetterFeedback` (similar issue would have
    been observed with Firebase Analytics, if we merged that already).

By selecting a navigator in dialog windows that was above GoRouter we could not access it as we only had access to the parent context.

Allow specifying global navigators directly if needed (e.g. when they need to be
    on top of the widget stack) by referring to their keys.


luckysori (approved):

luckysori:
Merging because we need this for testing!

Issue #1131, opened 2023-08-21
App stuck on fill order success dialog


holzeis:
After opening a position the app just stuck on the success dialog. see screenshot.

Branch: main
Commit: 03e0b844fdf60ada6c8c41cd851e6ac8e1271927
Build Number: 1880
Build Version: 1.2.0

![IMG_C77F3F3148DD-1](https://github.com/get10101/10101/assets/382048/60b14fd6-f005-471e-87f3-f744795cba4c)

Restioson:
Related error message: `No GoRouter found in context`

Restioson:
Bisected to 13d1a48b1c03d9c699464526088e8b2f692b84d1

klochowicz:
> Bisected to [13d1a48](https://github.com/get10101/10101/commit/13d1a48b1c03d9c699464526088e8b2f692b84d1) (cc @klochowicz).
> 
> Side note: bisecting this took _forever_ because of all the recompiles & wipes I needed to do. Not likely to do it again 😅 I think just searching for a cause would have been faster. However, the error message was incredibly vague so I didn't really know where to start.

 Thanks, I'll look into it.
It might be easiest to try to reproduce these things on different versions of 10101-test app (unfortunately only in iOS's TestFlight).

holzeis:
Sorry for not sharing the logs earlier.

https://nextcloud.holzeis.me/s/EfJ5T7nxPps5dTW

Issue #1129, opened 2023-08-21
Can't open account overview


bonomat:
On regtest (build 1868) I can't open the account overview anymore. If I click on the drop down, nothing happens.

bonomat:
Can't reproduce it on `master` using simulator nor deploying right to the phone. :/ Same commit hash as used on the phone also works on the simulator. :/

klochowicz:
what's the 'account overview'? the settings screen?

bonomat:
The wallets/accounts in the wallet overview, i.e. bitcoin and lightning

![image](https://github.com/get10101/10101/assets/224613/3eb4cb1c-00f3-4918-a48a-fa3b6fdc8dd5)

Restioson:
Possibly related: #1131

klochowicz:
> Possibly related: #1131. I reckon this is the same routing issue

if it is indeed the same, then we should see whether #1137 will fix this one too.

Restioson:
hmm, this is actually an expansion panel so it shouldn't interact with routing :thinking:

bonomat:
It did not fix it.

bonomat:
I'm currently testing removing the feedback plugin again and see if this fixes the issue.

bonomat:
Unfortunately the PR #1144 did not fix it :/ 
I found this ticket https://github.com/flutter/flutter/issues/132759 which seems to be related and verified that locally I'm running flutter version `3.10.6` and on CI we used `3.13.`

Restioson:
Hm, I am experiencing this locally - wonder if it's because I updated my flutter? :grimacing:

klochowicz:
> Hm, I am experiencing this locally - wonder if it's because I updated my flutter? 😬

yes, that is most likely the case.

Restioson:
It's somewhat worrying that this doesn't work on later flutter though - presumably we'll need to upgrade at some point

klochowicz:
> It's somewhat worrying that this doesn't work on later flutter though - presumably we'll need to upgrade at some point

it might be a temporary regression - might be worth reporting somewhere. 
I don't think it's a great cause to worry as we have done the biggest upgrade (Dart 3) and the other part can stay solid for a while.
it would be great to use `fvm` or some other way to pin the flutter version for all the contributors, otherwise we can get more bug reports like that. ;/

PR #1144, opened 2023-08-22
fix: remove feedback plugin

### LINKED ISSUE #1152: Dialog can now be clicked away
Changing the `useRootNavigator` to `false` broke the feature of not being able to click away the loading dialog while a position is being opened. 

Differently said, you can now click on the bottom bar to navigate away from the loading dialog and warning will be closed. 

https://github.com/get10101/10101/assets/224613/b82dacff-e1b6-437e-a8c1-f6454bc551c2

bonomat:
I can confirm, the feedback plugin breaks the account drop down. Since it also broke other places in the code, e.g. the confirm dialog after trade which required an extra fix I went ahead and removed it again.

By removing the feedback plugin we don't need those fixes neither. All of this was undone in this commit.

reverts 13d1a48b1c03d9c699464526088e8b2f692b84d1
reverts 8fb02bafb16bbe58364e4ab2b03c443f60881e9c

Resolves https://github.com/get10101/10101/issues/1129
Resolves https://github.com/get10101/10101/issues/1131
Resolves #1152

Tested in the manually deployed version 
- build: 1
- version 1.2.2
- commit 3620a3ebe47e8365b5c19d42f7b269b62c611712


klochowicz (changes requested):
please separate the fixes for feedback plugin from removing the functionality.

if you've read the commit descriptions, then you might have read that we have been using wrong way to bring things to the front and it would have break itself then too (ie. we'll need for #1104 ).

klochowicz:
it's sad to see a piece of functionality aimed to get feedback from our users go, but it's ultimately your call of course.

Just to confirm: the account overview was the only outstanding problem, yes? You do realise that the other one was due to our lack of proper usage of Flutter as we cobbled things together very quickly, and perhaps this one was too?

bonomat:
> Just to confirm: the account overview was the only outstanding problem, yes? You do realise that the other one was due to our lack of proper usage of Flutter as we cobbled things together very quickly, and perhaps this one was too?

Unfortunately I found another bug which was caused by this: https://github.com/get10101/10101/issues/1152. Not critical but it's unfortunately not clear what else is broken because of the changes related to the plugin.

klochowicz:
just to confirm @bonomat - you don't want the #1104 to lang anytime soon either? because this will likely trigger the same fixes needed (I'm happy to take care of them all in a comprehensive way, which I believe being the best way forward, ie. understanding Flutter usage more.

holzeis (approved):
Let's remove this for now so we unblock other things and come back to it another time by also addressing the bugs that this feedback plugin has introduced.

Restioson:
> Let's remove this for now so we unblock other things and come back to it another time by also addressing the bugs that this feedback plugin has introduced.

I think it would be a good idea to write up a list of bugs somewhere along with steps to reproduce so we can test it on the next  relevant PR

PR #1149, opened 2023-08-23
fix: actually run native tests with `just native-test`

Restioson:
Seems like an accidental omission

### DIFF
--- a/justfile
+++ b/justfile
@@ -199,8 +199,9 @@ maker args="":
 flutter-test:
     cd mobile && flutter pub run build_runner build && flutter test
 
+# Tests for the `native` crate
 native-test:
-    cd mobile/native
+    cd mobile/native && cargo test
 
 test: flutter-test native-test
 


klochowicz (approved):
certainly an omission - whilst you're at there, maybe it be good to add a comment what these tests are? 
because they're certainly not "native" - this poor naming came from the FRB template we used for the project.

it's actually "tests for the backend part of the 10101 app"

holzeis (approved):
😅

Restioson:
>  because they're certainly not "native" - this poor naming came from the FRB template we used for the project.

Well, they are 'native' in that they test the native crate :) Can add a note

klochowicz:
> > because they're certainly not "native" - this poor naming came from the FRB template we used for the project.
> 
> Well, they are 'native' in that they test the `native` crate :) Can add a note

yup, the "native" crate is misnamed indeed. :)

Restioson:
> Well, they are 'native' in that they test the native crate :) Can add a note

@klochowicz: done

PR #1155, opened 2023-08-24
chore: change contract duration to 7 days

### LINKED ISSUE #1154: Change contract duration to 7 days
Right now our contract duration is ~24h (till tomorrow midnight). 
For stable sats this is too short, we should increase this to 7 days. 

Hardcoding is fine for now, but we then should go for https://github.com/get10101/10101/issues/1029

holzeis:
This is also a good reference for implementing #1029. The expiry is currently scattered through the code and would be ideally only configured at one place.

resolves #1154


bonomat (approved):
LGTM and thanks for the changelog update :)

PR #1156, opened 2023-08-24
fix: ensure fix flutter version in ci

bonomat:
it looks like the wrong version is being used. Also adding cache which should speed up ci speed slightly.

Verified in own fork: https://github.com/bonomat/10101/actions/runs/5962763768/job/16174600731

Resolves https://github.com/get10101/10101/issues/1129

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
+  FLUTTER_VERSION: "3.10.5"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
+  FLUTTER_VERSION: "3.10.5"
```
klochowicz, line 14: nice, env variable makes a lot of  sense here!

klochowicz (approved):

Issue #1141, opened 2023-08-22
Failed to handle `SubChannelFinalize` message


luckysori:
Some apps have run into a broken state because of the coordinator being unable to process their `SubChannelFinalize` message. It's strange because the reason given is that the coordinator is in the `Offered` state, but one would expect the coordinator to be in `Confirmed` by then.

On top of that, the coordinator then receives a `SubChannelAccept` message (in theory because of a reconnect which rolls back the state). This seems to be a good idea, as the coordinator appears to have already rolled back to `Offered`, but handling this message fails with `Invalid commitment tx signature from peer`.

Logs here: https://itchysats.grafana.net/goto/0ckTe0gIR?orgId=1.

---

Related to https://github.com/get10101/10101/issues/1074.

luckysori:
This is the reason why our coordinator panics
---------------------------------------------
-   See https://github.com/get10101/10101/issues/1074.
-   Every time these two peers reconnect the coordinator is restarted because of the panic.
-   [Logs](https://itchysats.grafana.net/goto/Kf9j4agSR?orgId=1) as proof.

[Logs](https://itchysats.grafana.net/goto/qUtEVagSR?orgId=1) from one user[^1]
-------------------------------------------------------------------------------------

I've trimmed down the logs to only keep the most important bits.

    2023-08-17T20:24:13Z	INFO Received peer Init message from 03365df35cd56a0f3b3b085039e22b5ec6e3d5f71c2a92266ca582bb0768d42b29
    2023-08-17T20:27:26Z	INFO Opening position peer_id=03365df35cd56a0f3b3b085039e22b5ec6e3d5f71c2a92266ca582bb0768d42b29
    2023-08-17T20:27:33Z	INFO Processing message kind=SubChannelAccept from=03365df35cd56a0f3b3b085039e22b5ec6e3d5f71c2a92266ca582bb0768d42b29
    2023-08-17T20:27:34Z	INFO Sending message kind=SubChannelConfirm from=03365df35cd56a0f3b3b085039e22b5ec6e3d5f71c2a92266ca582bb0768d42b29
    
    ;; Reconnect
    2023-08-17T20:27:55Z	INFO Received peer Init message from 03365df35cd56a0f3b3b085039e22b5ec6e3d5f71c2a92266ca582bb0768d42b29
    
[...]

Issue #1074, opened 2023-08-05
Aborting after panic: Commitment txids are unique outside of fuzzing, where hashes can collide


holzeis:
After upgrading my test app to the latest build, my app is crashing at startup when updating the channel monitor.  

```
2023-08-05T07:40:46.992981Z ERROR native::api: Aborting after panic in task info=panicked at 'Commitment txids are unique outside of fuzzing, where hashes can collide', /Users/holzeis/.cargo/git/checkouts/rust-lightning-bee3233422fa4387/26db954/lightning/src/chain/channelmonitor.rs:2115:17
```

See full logs below

```
flutter: | 2023-08-05 09:40:45.711 | LogLevel.INFO | Built on commit: 261cf2aca2293cc80ca24b804b9406f7cc462094
flutter: | 2023-08-05 09:40:45.711 | LogLevel.INFO | Built on branch: main
flutter: | 2023-08-05 09:40:45.711 | LogLevel.INFO | Build number: 1.1.0
flutter: | 2023-08-05 09:40:45.711 | LogLevel.INFO | Build version: 1.1.0
flutter: | 2023-08-05 09:40:45.711 | LogLevel.INFO | Network: regtest
flutter: | 2023-08-05 09:40:45.713 | LogLevel.INFO | Esplora endpoint: http://35.189.57.114:3000
flutter: | 2023-08-05 09:40:45.713 | LogLevel.INFO | Coordinator: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@35.189.57.114:9045
flutter: | 2023-08-05 09:40:45.714 | LogLevel.INFO | Oracle endpoint: http://127.0.0.1:8081
flutter: | 2023-08-05 09:40:45.714 | LogLevel.INFO | Oracle PK: 16f88cf7d21e6c0f46bcbc983a4e3b19726c6c98858cc31c83551a88fde171c0
flutter: | 2023-08-05 09:40:45.716 | LogLevel.INFO | Node ID: 034de0c34cc7941dc8fb901b2923d8f4480e4358ec08f1c08726f12facc1f4fd79
flutter: | 2023-08-05 09:40:45.719 | LogLevel.DEBUG | Last login was at 2023-08-05 05:37:50 +00:00:00
[...]

PR #1157, opened 2023-08-24
Drop pending `SubChannelFinalize` message after reconnect

holzeis:
For now we have decided not to merge the `rust-dlc` changes into the `feature/ln-dlc-channels` branch, because we have not been able to prove that they fix the bugs listed in https://github.com/get10101/10101/issues/1141 and https://github.com/get10101/10101/issues/1074. We instead now depend on an updated version of `feature/ln-dlc-channels-10101`.

We propose to merge this hoping that it does fix those problems and knowing that it should not have a negative effect regardless. At the very least we will not process stale messages unnecessarily and we will emit fewer error logs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -953,7 +953,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
-source = "git+https://github.com/p2pderivatives/rust-dlc?rev=22b7052#22b70527f5d499153710c3aaac1a7a7315be32f3"
+source = "git+https://github.com/p2pderivatives/rust-dlc?rev=b76802c#b76802c99e060c7e55705dd5ec5ee3775ed29936"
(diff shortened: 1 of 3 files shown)


holzeis:
With 21ccaabfcd9ea50e5f2f057b927dc30826df0b92 I was able to reproduce #1141 demonstrating the that the fix in `rust-dlc` does not fix the issue.

holzeis:
With 3caa33c2f3e8527e712878628739f35e39ea7b5a I was able to reproduce #1074 

cc: @luckysori

holzeis:
It appears this issue is happening if a pending accept message is processed after the channel has been re-established. This also applies to the `CloseAccept` message.

holzeis:
Closing this PR in favor of https://github.com/get10101/10101/pull/1169

PR #1158, opened 2023-08-24
Add Synthetic USD balance into account overview

klochowicz:
https://github.com/get10101/10101/assets/8319440/9fb836cd-cf47-4048-b210-615826f91e28

### DIFF
--- a/mobile/lib/features/trade/position_change_notifier.dart
+++ b/mobile/lib/features/trade/position_change_notifier.dart
@@ -5,6 +5,8 @@ import 'package:get_10101/common/application/event_service.dart';
 import 'package:get_10101/common/domain/model.dart';
 import 'package:get_10101/features/trade/application/position_service.dart';
 import 'package:get_10101/features/trade/domain/contract_symbol.dart';
+import 'package:get_10101/features/trade/domain/direction.dart';
+import 'package:get_10101/features/trade/domain/leverage.dart';
 
 import 'domain/position.dart';
 import 'domain/price.dart';
@@ -15,6 +17,32 @@ class PositionChangeNotifier extends ChangeNotifier implements Subscriber {
   Map<ContractSymbol, Position> positions = {};
 
   Price? price;
+
+  /// Amount of stabilised bitcoin in terms of USD (fiat)
+  double getStableUSDAmountInFiat() {
+    if (hasStableUSD()) {
+      final positionUsd = positions[ContractSymbol.btcusd];
+      return positionUsd!.quantity;
+    } else {
+      return 0.0;
+    }
+  }
+
+  Amount getStableUSDAmountInSats() {
+    if (hasStableUSD()) {
+      final positionUsd = positions[ContractSymbol.btcusd];
+      return positionUsd!.getAmountWithUnrealizedPnl();
+    } else {
+      return Amount(0);
+    }
+  }
+
+  bool hasStableUSD() {
+    final positionUsd = positions[ContractSymbol.btcusd];
+    return positionUsd != null &&
+        positionUsd.direction == Direction.short &&
+        positionUsd.leverage == Leverage(1);
+  }
 
   Future<void> initialize() async {
     List<Position> positions = await _positionService.fetchPositions();
--- a/mobile/lib/features/wallet/balance_row.dart
+++ b/mobile/lib/features/wallet/balance_row.dart
@@ -2,6 +2,9 @@ import 'package:flutter/material.dart';
 import 'package:flutter_svg/svg.dart';
 import 'package:get_10101/common/amount_text.dart';
 import 'package:get_10101/common/domain/model.dart';
+import 'package:get_10101/common/fiat_text.dart';
+import 'package:get_10101/features/stable/stable_screen.dart';
+import 'package:get_10101/features/trade/position_change_notifier.dart';
 import 'package:get_10101/features/wallet/create_invoice_screen.dart';
 import 'package:get_10101/features/wallet/domain/wallet_history.dart';
 import 'package:get_10101/features/wallet/send_screen.dart';
@@ -40,8 +43,10 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
   Widget build(BuildContext context) {
     WalletTheme theme = Theme.of(context).extension<WalletTheme>()!;
     WalletChangeNotifier walletChangeNotifier = context.watch<WalletChangeNotifier>();
-    TextStyle normal = const TextStyle(fontSize: 16.0);
-    TextStyle bold = const TextStyle(fontWeight: FontWeight.bold, fontSize: 16.0);
+    const normal = TextStyle(fontSize: 16.0);
+    const bold = TextStyle(fontWeight: FontWeight.bold, fontSize: 16.0);
+
+    PositionChangeNotifier positionChangeNotifier = context.watch<PositionChangeNotifier>();
 
     Amount amount;
     String name;
@@ -53,6 +58,12 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
       rowBgColor = theme.lightning;
       icon = SvgPicture.asset("assets/Lightning_logo.svg");
       amount = walletChangeNotifier.lightning();
+    } else if (widget.walletType == WalletHistoryItemDataType.stable) {
+      name = "Synthetic-USD";
+      rowBgColor = theme.lightning;
+      icon = SvgPicture.asset("assets/USD_logo.svg");
+      // this in unused for now, other than to initialise value (and satisfy the compiler)
+      amount = positionChangeNotifier.getStableUSDAmountInSats();
     } else {
       name = "On-chain";
       rowBgColor = theme.onChain;
@@ -65,12 +76,14 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
     double buttonSpacing = 10;
 
     BalanceRowButton send = BalanceRowButton(
+      type: widget.walletType,
       flow: PaymentFlow.outbound,
       enabled: _expanded,
       buttonSize: buttonSize,
     );
 
     BalanceRowButton receive = BalanceRowButton(
+      type: widget.walletType,
       flow: PaymentFlow.inbound,
       enabled: _expanded,
       buttonSize: buttonSize,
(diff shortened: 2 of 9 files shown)


mobile/native/src/event/api.rs
```diff
@@ -84,7 +84,6 @@ pub struct FlutterSubscriber {
 /// Subscribes to event relevant for flutter and forwards them to the stream sink.
 impl Subscriber for FlutterSubscriber {
     fn notify(&self, event: &EventInternal) {
-        tracing::debug!(event = %event, "Forwarding event to flutter");
```
bonomat, line 87: Thanks :)

mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -138,10 +142,30 @@ class _WalletScreenState extends State<WalletScreen> {
...
+                                            amount: walletChangeNotifier.total(),
+                                            textStyle: const TextStyle(
+                                                fontSize: 20.0, fontWeight: FontWeight.bold)),
+                                        Visibility(
+                                            visible:
+                                                positionChangeNotifier.getStableSatsAmount() != 0.0,
+                                            child: Row(
+                                              children: [
+                                                const SizedBox(width: 5),
+                                                const Text("+"),
```
holzeis, line 157: 🔧 Combining the balances of sats and $ is very irritating for me. IMHO this balance should only ever show one value. In other wallets (e.g. blink or zeus) they let you select what currency you prefer and show you the balance correspondingly. 

Equally, I would like to see our balance to do the same. If we want to show the synthetic usd balance we should simply add it on the balance in the currency selected. btc, usd, eur, etc.

holzeis (commented):
It's still weird to me that we show the synthetic usd balance in the wallet screen, because it does not match the transaction history showing an outgoing transaction.

Overall it does not feel consistent, maybe it would be better to first go to figma and have a look there? Also considering that this is not necessarily required for the experiment.

mobile/lib/features/trade/position_change_notifier.dart
```diff
@@ -16,6 +17,12 @@ class PositionChangeNotifier extends ChangeNotifier implements Subscriber {
 
   Price? price;
 
+  /// Amount of stabilised bitcoin in terms of USD (fiat)
+  double getStableSatsAmount() {
```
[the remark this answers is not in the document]
Restioson: Has this been used by another company already or?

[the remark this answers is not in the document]
holzeis: Yes, that's why I wouldn't call it like that.

mobile/lib/features/trade/position_change_notifier.dart
```diff
@@ -16,6 +17,12 @@ class PositionChangeNotifier extends ChangeNotifier implements Subscriber {
...
+    final position = positions[ContractSymbol.btcusd];
+    return position != null && position.direction == Direction.short ? position.quantity : 0.0;
```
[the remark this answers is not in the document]
klochowicz: good point!

mobile/lib/features/wallet/wallet_screen.dart
klochowicz: can we try and see whether people prefer it? in the spirit of experiment? 

it's not a simple UX matter, because other wallets you mentioned do not have the synthetic stablecoin version. 
Please note that Galloy in their stablesats put exactly this approach (although in a nicer UI, to be honest!)


the idea was: 
- make it always visible that user

> Equally, I would like to see our balance to do the same. If we want to show the synthetic usd balance we should simply add it on the balance in the currency selected. btc, usd, eur, etc.

this hides the whole stablesats feature and entirely misses the point. 


The discussion whether to display currencies is close to me, and I think we should do that separately anyway, if we want to be a *good* lightning wallet.

mobile/lib/features/trade/position_change_notifier.dart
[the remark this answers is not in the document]
klochowicz: sure, happy to change it if it works better for you.  :)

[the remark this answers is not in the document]
klochowicz: removed all the references in code, text and commit messages.

mobile/lib/features/wallet/balance_row.dart
```diff
@@ -53,6 +60,17 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
       rowBgColor = theme.lightning;
       icon = SvgPicture.asset("assets/Lightning_logo.svg");
       amount = walletChangeNotifier.lightning();
+    } else if (widget.walletType == WalletHistoryItemDataType.stable) {
+      final positionBtc = positionChangeNotifier.positions[ContractSymbol.btcusd];
+      final hasStable = positionBtc != null && positionBtc.direction == Direction.short;
```
Restioson, line 65: I haven't looked much into how stabilized sats are implemented here - would this mean that any short against bitcoin (even one that isn't semantically 'stabilizing sats' but just a regular bet) would show up as stablesats? Or is this not the case

mobile/lib/features/wallet/domain/wallet_history.dart
```diff
@@ -2,7 +2,14 @@ import 'package:get_10101/common/domain/model.dart';
...
 import 'package:get_10101/bridge_generated/bridge_definitions.dart' as rust;
 
-enum WalletHistoryItemDataType { lightning, onChain, trade, orderMatchingFee, jitChannelFee }
+enum WalletHistoryItemDataType {
+  lightning,
+  onChain,
+  trade,
+  orderMatchingFee,
+  jitChannelFee,
+  stable
```
Restioson, line 11: I can't remember if it's merged yet but at some point I made a new enum WalletType for this... By name, this enum should only really be used for transaction history types I think

mobile/lib/features/wallet/wallet_screen.dart
Restioson: Personally I don't really mind either way. The benefit of displaying sats + dollars is ease of view - stabilised sats seem to feel more "first class" this way.

Restioson (approved):

mobile/lib/features/wallet/balance_row.dart
holzeis: Yes at the moment we simply display a short position leverage 1 as synthetic usd.

klochowicz: good catch, this one should be abstracted away into `positionChangeNotifier` too. 
Having business logic in widget code was not a good idea 😔

mobile/lib/features/wallet/wallet_screen.dart
holzeis: > it's not a simple UX matter, because other wallets you mentioned do not have the synthetic stablecoin version.

Have you had a look at blink? It's the galoy stablesats wallet.

mobile/lib/features/wallet/domain/wallet_history.dart
klochowicz: I don't mind if you refactor this further, it might be cleaner! I added it here as this is the one that's used currently in the `balance_row` . 
ideally, we should distinguish creating a stable position as a different history type!

mobile/lib/features/wallet/wallet_screen.dart
klochowicz: > > it's not a simple UX matter, because other wallets you mentioned do not have the synthetic stablecoin version.
> 
> Have you had a look at blink? It's the galoy stablesats wallet.

yup, they have both always visible by default *and* the feature of showing everything in USD (which we don't). I don't have an Android, so I can't really test it out.

mobile/lib/features/wallet/balance_row.dart
klochowicz: fixed.

mobile/lib/features/wallet/domain/wallet_history.dart
Restioson: Ah yes, the it's in one of my PRs still :)

mobile/lib/features/wallet/wallet_screen.dart
holzeis: I am not against showing the Synthetic usd balance, it's just the way it is shown with your change seems inconsistent. 

I see the following issues

- We don't show a balance for the long positions (As you have pointed out yourself)
- We sometimes show a balance with multiple currencies. Contrary to your comment, blink does not do the same, it shows multiple accounts which we do as well with the difference that we collapse them, but it does not show two different balances in on row.
- The sats for the synthetic usd is still shown as an outgoing transaction.

I really would prefer to go back to Figma with that and think it through before releasing this change as it's not critical for the experiment and tackles a component that needs a redesign anyways.

mobile/lib/features/wallet/balance_row.dart
Restioson: > Yes at the moment we simply display a short position leverage 1 as synthetic usd.

Where should we discuss this UX further? @holzeis

mobile/lib/features/wallet/wallet_screen.dart
klochowicz: If you want to bring it in quickly and can't live with that part of the PR, just change the Visibility widget to 'false'. 
I think we'd have exactly the same conversation in Figma, I can't imagine that changing anything- happy to split it , or perhaps *iterate* on it seeing feedback for customers.

holzeis: > If you want to bring it in quickly and can't live with that part of the PR, just change the Visibility widget to 'false'.

what do you mean by that?

> I think we'd have exactly the same conversation in Figma, I can't imagine that changing anything- happy to split it , or perhaps iterate on it seeing feedback for customers.

I think we will have a different discussion there, because we can think about the bigger picture, including how this should look like in the history, if we should stick with a collapsible balance, how we should show long positions (e.g. keep them in the balance), show a balance in a single currency, etc.

klochowicz: > what do you mean by that?

Sorry, I was on my phone, so tried to reply briefly before. I meant changing the `Visibility` widget's `visible` property to `false`, because this is the sticking point for this issue.

Are you actively against this PR? I don't mind iterating on this further at all, because I don't think we've nailed it, but I certainly think this is step in the right direction. 

I would like to get some clarification from you before I go ahead and merge it. Right now it's in an awkward limbo state.

PS the lack of Figma / big picture discussion also applies to the whole stablesats feature, not just this part of it. It's a quick experiment / hack to gather reactions and early feedback.

holzeis: > Are you actively against this PR? I don't mind iterating on this further at all, because I don't think we've nailed it, but I certainly think this is step in the right direction.

I am only against the part where we show the balance with "45100 sats + 10$". I am fine with the rest though. But that's just my opinion. I am not going to block that PR, because of it.

klochowicz: noted, thank you for the feedback. 
let's give it a spin towards until the end of the week, and I'm happy to remove it if it is weird after that period of internal testing, ok? 
let's see how that feels on the device.

PR #1160, opened 2023-08-25
feat: make contract fee rate configurable

### LINKED ISSUE #628: iOS status bar is not visible
We obscure the status bar, both on light and dark system theme.

IMG_F4420D64634C-1

Probably the best way is changing the background to purple (or a little lighter colour).

bonomat:
While we can't update existing contracts, we can at least change the settings when needed for new contracts.

Resolves https://github.com/get10101/10101/issues/1159
Resolves #628 

image

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -1,6 +1,4 @@
 PODS:
-  - device_info_plus (0.0.1):
-    - Flutter
   - Firebase/CoreOnly (10.12.0):
     - FirebaseCore (= 10.12.0)
   - Firebase/Messaging (10.12.0):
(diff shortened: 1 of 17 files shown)


mobile/ios/Podfile.lock
```diff
@@ -1,6 +1,4 @@
 PODS:
```
holzeis, line 1: 🙃 You could have put this into a separate commit

holzeis (approved):
LGTM! Thanks 👍

mobile/ios/Podfile.lock
bonomat: 🙈

PR #1167, opened 2023-08-26
feat: Allow delivering messages in the background handler

klochowicz:
Configuring this should allow us to deliver messages and push notifications even if the app is terminated.


holzeis (approved):
Nice, let's give it a try a test if it works 👍

Issue #1139, opened 2023-08-22
Ensure that broadcasted transactions are re-broadcasted


bonomat:
If a transaction was not included yet in the mempool, the app/coordinator should try to rebroadcast the transaction, e.g. on restart.

Alternatively, we could store all broadcasted transactions in the DB and do it manually when needed.

bonomat:
Critical because if we do not do this yet we might lose the posibility to broadcast old transactions. Our logs get pruned after 30 days.

holzeis:
@bonomat is this really fixed? your change only stores the raw transaction but does not try to re-broadcast them, or does it?

bonomat:
Correct. I descoped this part and only focused on storing it so that we can manually re-broadcast when needed.

PR #1168, opened 2023-08-28
feat: store raw transaction in db

bonomat:
for transactions we are broadcasting we should store the raw transaction so that we can re-broadcast it in the future if needed.

Manually extracting it is a [a lot of work](https://www.notion.so/get10101/Extracted-raw-transactions-b913770b883b4f10a7f1071a09c3ca1f) and only doable for the past 30 days. 

Resolves https://github.com/get10101/10101/issues/1139 by storing the raw tx in the database


holzeis (approved):
Nice 👍

PR #1180, opened 2023-08-29
feat(maker): Monitor and expose status of services connected

klochowicz:
Monitor the following services:
    - coordinator
    - orderbook (we don't subscribe to orderbook feed yet)

in order to derive health status of the maker.

Health status is exposed via the HTTP API as well as reported as
dedicated prometheus metrics.

- [ ] Subscribe to orderbook websocket stream to be able to detect if it is down
- [ ] Subscribe to bitmex client stream to be able to spot outdated information

the TODO listed above will be wired only when bitmex and orderbook streams will be integrated into maker, and as such are *not* a dependency of this PR - we can wire them in the corresponding PRs or in a follow-up.

### DIFF
--- a/crates/tests-e2e/src/maker.rs
+++ b/crates/tests-e2e/src/maker.rs
@@ -29,8 +29,12 @@ impl Maker {
         Self::new(client, "http://localhost:18000")
     }
 
+    /// Check whether maker is running and that it is connected to all services
+    /// it depends on
     pub async fn is_running(&self) -> bool {
-        self.get("/").await.is_ok()
+        self.get("/health")
+            .await
+            .is_ok_and(|r| r.status().is_success())
     }
 
     pub async fn sync_on_chain(&self) -> Result<()> {
--- a/maker/src/health.rs
+++ b/maker/src/health.rs
@@ -1,0 +1,138 @@
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+use reqwest::Client;
+use reqwest::StatusCode;
+use reqwest::Url;
+use serde::Serialize;
+use std::time::Duration;
+use tokio::sync::watch;
+
+/// Health status of a service
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize)]
+pub enum ServiceStatus {
+    #[default]
+    Unknown,
+    Online,
+    Offline,
+}
+
+/// Health monitoring for the node
+///
+/// Simple endpoint querying is handled by provided configuration, for more complex health checks
+/// the transmitters are exposed to be plugged in the services that need to publish their health
+/// status.
+
+#[derive(Clone)]
+pub struct Health {
+    /// Coordinator HTTP API status
+    coordinator_rx: watch::Receiver<ServiceStatus>,
+    /// Orderbook websocket stream status
+    orderbook_rx: watch::Receiver<ServiceStatus>,
+    /// Bitmex pricefeed stream status
+    bitmex_pricefeed_rx: watch::Receiver<ServiceStatus>,
+}
+
+/// Transmitters that need to be plugged in the services that need to publish their health status.
+pub struct Tx {
+    pub orderbook: watch::Sender<ServiceStatus>,
+    pub coordinator: watch::Sender<ServiceStatus>,
+    pub bitmex_pricefeed: watch::Sender<ServiceStatus>,
+}
+
+/// Struct returned by maker's health endpoint.
+#[derive(Debug, Serialize)]
+pub struct OverallMakerHealth {
+    coordinator: ServiceStatus,
+    orderbook: ServiceStatus,
+    bitmex_pricefeed: ServiceStatus,
+}
+
+impl OverallMakerHealth {
+    pub fn is_healthy(&self) -> bool {
+        self.coordinator == ServiceStatus::Online
+            && self.bitmex_pricefeed == ServiceStatus::Online
+            && self.orderbook == ServiceStatus::Online
+    }
+}
+
+impl Health {
+    pub fn new() -> (Self, Tx) {
+        let (orderbook_tx, orderbook_rx) = watch::channel(ServiceStatus::Unknown);
+        let (coordinator_tx, coordinator_rx) = watch::channel(ServiceStatus::Unknown);
+        let (bitmex_pricefeed_tx, bitmex_pricefeed_rx) = watch::channel(ServiceStatus::Unknown);
+
+        (
+            Self {
+                coordinator_rx,
+                orderbook_rx,
+                bitmex_pricefeed_rx,
+            },
+            Tx {
+                orderbook: orderbook_tx,
+                coordinator: coordinator_tx,
+                bitmex_pricefeed: bitmex_pricefeed_tx,
+            },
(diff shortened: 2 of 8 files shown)


klochowicz:
I'm not sure whether we actually need the orderbook status (we don't seem to rely on the websocket stream at all so far? 🤔), and bitmex status can be added when we actually start hedging (it was not mentioned in the ticket, but I think it's an essential service for the maker to be monitored).

crates/tests-e2e/src/maker.rs
```diff
@@ -29,8 +29,10 @@ impl Maker {
         Self::new(client, "http://localhost:18000")
     }
 
+    /// Check whether maker is running and that it is connected to all services
+    /// it depends on
     pub async fn is_running(&self) -> bool {
-        self.get("/").await.is_ok()
+        self.get("/health").await.is_ok()
```
Restioson, line 35: Do we want to interpret the healthcheck result here? Or is it always okay if it returns 200?

maker/src/health.rs
```diff
@@ -0,0 +1,120 @@
...
+            },
+            Tx {
+                orderbook: orderbook_tx,
+                coordinator: coordinator_tx,
+            },
+        )
+    }
+
+    // TODO: Any ideas of the most useful way of sending this information?
+    // - perhaps a serialized struct (json) with the status of each service?
```
Restioson, line 60: Even just a struct here could make sense - it could then later be serialized if needed, or logged just like that with `{:?}`

Restioson (approved):

maker/src/health.rs
bonomat: Same thought: why are you not returning a struct serialized to json?

maker/src/routes.rs
```diff
@@ -52,6 +56,7 @@ pub fn router(
         .route("/api/pay-invoice/:invoice", post(pay_invoice))
         .route("/api/sync-on-chain", post(sync_on_chain))
         .route("/metrics", get(get_metrics))
+        .route("/health", get(get_health))
```
bonomat, line 59: For whom is this new endpoint?

maker/src/health.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    pub fn get_coordinator_status(&self) -> ServiceStatus {
+        *self.coordinator_rx.borrow()
+    }
+
+    pub fn get_orderbook_status(&self) -> ServiceStatus {
+        *self.orderbook_rx.borrow()
+    }
+    // TODO: Add bitmex status? would anything else be useful here?
```
bonomat, line 86: Should this rather be captured in a ticket?

maker/src/health.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut health_info = String::new();
+        health_info.push_str(&format!(
+            "Coordinator: {:?}\n",
+            self.get_coordinator_status()
+        ));
+        health_info.push_str(&format!("Orderbook: {:?}\n", self.get_orderbook_status()));
+
+        if self.get_coordinator_status() == ServiceStatus::Online
+        // TODO: Uncomment this line when we retrieve the orderbook status
+        // && (self.get_orderbook_status() == ServiceStatus::Online)
```
bonomat, line 71: Does this work already?

maker/src/metrics.rs
```diff
@@ -58,12 +69,42 @@ pub fn init_meter() -> PrometheusExporter {
...
+    match service_status {
+        ServiceStatus::Unknown => {
+            gauge.observe(cx, 0, &[KeyValue::new("status", "unknown")]);
+        }
+        ServiceStatus::Online => {
+            gauge.observe(cx, 1, &[KeyValue::new("status", "online")]);
+        }
+        ServiceStatus::Offline => {
+            gauge.observe(cx, 0, &[KeyValue::new("status", "offline")]);
+        }
```
bonomat, line 97: 🚨 I don't think this works like you want: 

Our grafana agent will poll this metric ever X seconds but it will not reset an old value it has seen before, which means, with this setup, we might have a service which is online and offline at the same time, i.e. 

at 10:00:00 service was offline: we create the metric: 
`coordinator_status{status=offline}=0`

at 11:00:00 the service is online, we create the metric
`coordinator_status{status=online}=1`

This means, at 11:00:00, in our grafana dashboard we will have both values set: 
```
coordinator_status{status=offline}=0
coordinator_status{status=online}=1
```

Because `coordinator_status{status=offline}` and `coordinator_status{status=online}` are two different metrics.

Instead, you need to use the same metric to hold different states, e.g. 

- `coordinator_status=0` = offline
- `coordinator_status=1` = online
- `coordinator_status=2` = unknown --> with this we could then indicate different error codes.

bonomat (changes requested):
There are a few comments/todos in the code and I believe the usage of the metrics is not correct.
Let me know if my comment is not clear.

crates/tests-e2e/src/maker.rs
klochowicz: added interpreting of the result.

maker/src/health.rs
klochowicz: changed to return serialised json.

maker/src/routes.rs
klochowicz: it's useful in development, testing, and manual checking if we want to quickly see the state.

maker/src/health.rs
klochowicz: this was not in the original scope of the ticket, so I have left it as a comment first - wasn't sure whether we wanted it! 
it seems it can be addressed rather quickly, so I'll add support for it in this PR.

maker/src/health.rs
klochowicz: if by "orderbook" we mean websocket stream that other people look at orders, then not, we are still not subscribing to it from the maker. I was hoping that other work done for this milestone would encompass this, but nothing got still merged.

@luckysori do I remember correctly that you were adding the orderbook websocket stream?

maker/src/metrics.rs
klochowicz: indeed, I had a wrong mental model how attributes for metrics work. Thank you for explanation, I've fixed it now.


cool idea with mapping `2 == unknown` .

maker/src/health.rs
klochowicz: edit: I have removed the TODO from this place and hard-coded `Online` status for the orderbook where it should be plugged in (there's another TODO there anyway).

klochowicz:
> There are a few comments/todos in the code and I believe the usage of the metrics is not correct.
> Let me know if my comment is not clear.

I believe I fixed the issue.
the only `todo` left in code is about us not using the orderbook websocket at all yet, so there's not much I can do about it. I've added bitmex stream.

bonomat (approved):
LGTM

bonomat:
> > There are a few comments/todos in the code and I believe the usage of the metrics is not correct.
> > Let me know if my comment is not clear.
> 
> I believe I fixed the issue. the only `todo` left in code is about us not using the orderbook websocket at all yet, so there's not much I can do about it. I've added bitmex stream monitoring.

Thanks for addressing.

PR #1186, opened 2023-08-29
chore: delete some autometrics

bonomat:
I don't think we need to annotate all these functions. I tried to keep them on
- important functions
- functions which cold potentially run longer
- functions where we had problems in the past


klochowicz (approved):
seems like a reasonable trade-off.

holzeis (approved):

PR #1192, opened 2023-08-30
temporary debugging aid

klochowicz:
This is purely for debugging. we might add it with some dart-define flag afterwards, but I didn't bother cleaning this whilst we're still bug hunting

### DIFF
--- a/mobile/ios/Runner.xcodeproj/project.pbxproj
+++ b/mobile/ios/Runner.xcodeproj/project.pbxproj
@@ -444,11 +444,9 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = "";
-				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = V3D64P9D6W;
+				DEVELOPMENT_TEAM = V3D64P9D6W;
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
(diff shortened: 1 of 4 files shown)


mobile/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -444,11 +444,9 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_STYLE = Automatic;
```
bonomat, line 447: Please don't commit this.

mobile/native/src/api.rs
```diff
@@ -248,6 +248,19 @@ pub fn run(
...
     let runtime = ln_dlc::get_or_create_tokio_runtime()?;
+
+    // TODO: Hide this behind some debug flag, it shouldn't run in production
+    // (or maybe log once a minute in release build?)
+    let handle = runtime.handle();
+    let runtime_monitor = tokio_metrics::RuntimeMonitor::new(handle);
+    let frequency = Duration::from_secs(5);
+    runtime.spawn(async move {
+        for metrics in runtime_monitor.intervals() {
+            tracing::debug!(?metrics, "tokio metrics");
```
bonomat, line 259: What's the idea of how to use these metrics?

klochowicz: it tells you all the information about the current runtime - descriptions of the metrics are here:

https://docs.rs/crate/tokio-metrics/latest

one of the most interesting for us are for example:
- [workers_count](https://docs.rs/tokio-metrics/0.2.*/tokio_metrics/struct.RuntimeMetrics.html#structfield.workers_count) - how many worker threads we actually have on our runtime (e.g. I was surprised seeing my iphone having 6!)
- https://docs.rs/tokio-metrics/0.2.*/tokio_metrics/struct.RuntimeMetrics.html#structfield.max_busy_duration - maximum amount the worker thread was busy. 

also, if the queue of tasks scheduled but not executed would be growing we could see some contention with things falling behind.

also:task metrics might be interesting too.

Restioson: > maximum amount the worker thread was busy

Interesting, does this include blocking threads?

Regardless, it will be good to make sure we aren't accidentally doing blocking work on the runtime!

mobile/ios/Runner.xcodeproj/project.pbxproj
klochowicz: have you read the description of this PR? it's just a draft for aiding debugging of the problem we're having 🙃 none of it gets in.

klochowicz: I have changed the name of the PR to make it less ambiguous, I realised that it would have helped to filter out the PR before you jumped into it.

mobile/native/src/api.rs
klochowicz: this is not intended to be merged; I have now made it clear by renaming the PR!

this is more of a showcase of another tool that one can have on their side branch.

Issue #1193, opened 2023-08-30
Synthetic stable coin amount not shown


bonomat:
I believe this used to work 

![image](https://github.com/get10101/10101/assets/224613/60b9e0fd-feab-4d9a-9830-4d2c2ac086b2)

![image](https://github.com/get10101/10101/assets/224613/13acf33a-ff4f-409d-bf06-8d3a363fdb79)

holzeis:
Its also not shown in the wallet balance.

![Screenshot 2023-08-30 at 07 30 23](https://github.com/get10101/10101/assets/382048/fd790519-31e9-41ed-a0df-87b61d0db019)

PR #1197, opened 2023-08-30
fix: show balance of synth usd 

bonomat:
in `hasStableUSD()` we are comparing the leverage of the position to be equal the Leverage(0), i.e. `positionUsd.leverage == Leverage(1);`. For this to work we need to implement equal on `Leverage`.

Resolves https://github.com/get10101/10101/issues/1193

image

image


holzeis (approved):
Nice 👍

PR #1199, opened 2023-08-30
fix: Add buffer delay after outbound capacity is ready

holzeis:
It could happen that the channel already shows outbound capacity due to the received payment, but the channel updates are not completely finished yet. We suspect that this could lead to issues when trying to pay the JIT channel opening fees at the same time. Resulting in force closures or worst case and unresponsive app (deadlock?).

By waiting additional 500 milliseconds after the outbound capacity is set, we hopefully prevent paying an invoice while the channel is still updated.

### DIFF
--- a/mobile/native/src/channel_fee.rs
+++ b/mobile/native/src/channel_fee.rs
@@ -79,8 +79,15 @@ impl ChannelFeePaymentSubscriber {
                 funding_tx_fees_msats,
                 "Waiting for outbound capacity on channel to pay jit channel opening fee.",
             );
-            Handle::current()
-                .block_on(self.wait_for_outbound_capacity(channel_id, funding_tx_fees_msats))
+            Handle::current().block_on(async {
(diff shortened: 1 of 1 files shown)


bonomat (approved):

mobile/native/src/channel_fee.rs
```diff
@@ -79,8 +79,15 @@ impl ChannelFeePaymentSubscriber {
                 funding_tx_fees_msats,
                 "Waiting for outbound capacity on channel to pay jit channel opening fee.",
             );
-            Handle::current()
-                .block_on(self.wait_for_outbound_capacity(channel_id, funding_tx_fees_msats))
+            Handle::current().block_on(async {
+                self.wait_for_outbound_capacity(channel_id, funding_tx_fees_msats)
+                    .await?;
+                // We add another sleep to ensure that the channel has actually been updated after
```
klochowicz, line 85: it would be nice if we could tag these as `FIXME` or something like that, to find these spots easier, especially when we're bug-hunting in the future!

PR #1201, opened 2023-08-31
chore(deps): Bump ureq package to solve security advisory

klochowicz:
https://github.com/advisories/GHSA-8qv2-5vq6-g2g7

Turned out to be really simple to fix it, by bumping one of our dependencies (`webpki` was only a transitive one of ours)


luckysori (approved):
Thanks!

PR #1206, opened 2023-09-01
Fix panic at startup when asking for node id from the backend

klochowicz:
Sometimes the app  would panic at startup (for local deployments for me), and I
    figured it was trivial to make it more resilient in this case.

We now log an error if for whatever reason we couldn't display the node id at
    startup (e.g. node wasn't ready) - this should not be a showstopper by any means, and it gets
    logged at later stage anyway.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -100,8 +100,12 @@ pub fn get_node_key() -> SecretKey {
     NODE.get().inner.node_key()
 }
 
-pub fn get_node_info() -> NodeInfo {
-    NODE.get().inner.info
+pub fn get_node_info() -> Result<NodeInfo> {
+    Ok(NODE
+        .try_get()
+        .context("NODE is not initialised yet, can't retrieve node info")?
+        .inner
+        .info)
 }
 
 pub async fn update_node_settings(settings: LnDlcNodeSettings) {
--- a/crates/tests-e2e/tests/force_close_position.rs
+++ b/crates/tests-e2e/tests/force_close_position.rs
@@ -12,7 +12,7 @@ async fn can_force_close_position() {
     let coordinator = &test.coordinator;
     let bitcoin = &test.bitcoind;
 
-    let app_pubkey = api::get_node_id().0;
+    let app_pubkey = api::get_node_id().unwrap().0;
 
     let dlc_channels = coordinator.get_dlc_channels().await.unwrap();
 
--- a/crates/tests-e2e/tests/rollover_position.rs
+++ b/crates/tests-e2e/tests/rollover_position.rs
@@ -13,7 +13,7 @@ async fn can_rollover_position() {
     let test = setup::TestSetup::new_with_open_position().await;
     let coordinator = &test.coordinator;
     let dlc_channels = coordinator.get_dlc_channels().await.unwrap();
-    let app_pubkey = api::get_node_id().0;
+    let app_pubkey = api::get_node_id().unwrap().0;
 
     tracing::info!("{:?}", dlc_channels);
 
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -442,8 +442,12 @@ Future<void> logAppSettings(bridge.Config config) async {
   FLog.info(text: "Oracle endpoint: ${config.oracleEndpoint}");
   FLog.info(text: "Oracle PK: ${config.oraclePubkey}");
 
-  String nodeId = rust.api.getNodeId();
-  FLog.info(text: "Node ID: $nodeId");
+  try {
+    String nodeId = rust.api.getNodeId();
+    FLog.info(text: "Node ID: $nodeId");
+  } catch (e) {
+    FLog.error(text: "Failed to get node ID: $e");
+  }
 }
 
 /// Forward the events from change notifiers to the Event service
@@ -500,7 +504,7 @@ Future<void> runBackend(bridge.Config config) async {
   final seedDir = (await getApplicationSupportDirectory()).path;
 
   // We use the app documents dir on iOS to easily access logs and DB from
-  // the device. On other plaftorms we use the seed dir.
+  // the device. On other platforms we use the seed dir.
   String appDir = Platform.isIOS
       ? (await getApplicationDocumentsDirectory()).path
       : (await getApplicationSupportDirectory()).path;
--- a/mobile/native/src/api.rs
+++ b/mobile/native/src/api.rs
@@ -360,8 +360,8 @@ pub fn decode_invoice(invoice: String) -> Result<LightningInvoice> {
     })
 }
 
-pub fn get_node_id() -> SyncReturn<String> {
-    SyncReturn(ln_dlc::get_node_info().pubkey.to_string())
+pub fn get_node_id() -> Result<SyncReturn<String>> {
+    Ok(SyncReturn(ln_dlc::get_node_info()?.pubkey.to_string()))
 }
 
 pub fn get_channel_open_fee_estimate_sat() -> Result<u64> {
--- a/mobile/native/src/trade/order/mod.rs
+++ b/mobile/native/src/trade/order/mod.rs
@@ -148,7 +148,7 @@ impl Order {
 impl From<Order> for orderbook_commons::NewOrder {
     fn from(order: Order) -> Self {
         let quantity = Decimal::try_from(order.quantity).expect("to parse into decimal");
-        let trader_id = ln_dlc::get_node_info().pubkey;
+        let trader_id = ln_dlc::get_node_info().expect("to have info").pubkey;
         orderbook_commons::NewOrder {
             id: order.id,
             // todo: this is left out intentionally as market orders do not set a price. this field
@@ -175,7 +175,7 @@ impl From<OrderType> for orderbook_commons::OrderType {
 /// Enroll the user in the beta program
 pub async fn register_beta(email: String) -> Result<()> {
     let register = RegisterParams {
-        pubkey: ln_dlc::get_node_info().pubkey,
+        pubkey: ln_dlc::get_node_info()?.pubkey,
         email: Some(email),
         nostr: None,
     };
--- a/mobile/native/src/trade/position/handler.rs
+++ b/mobile/native/src/trade/position/handler.rs
@@ -35,7 +35,7 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
     tracing::debug!(?order, ?filled, "Filling order with id: {}", order.id);
 
     let trade_params = TradeParams {
-        pubkey: ln_dlc::get_node_info().pubkey,
+        pubkey: ln_dlc::get_node_info()?.pubkey,
         contract_symbol: order.contract_symbol,
         leverage: order.leverage,
         quantity: order.quantity,


holzeis (commented):
When `logAppSettings` is called the node should definitely be initialised. Are we swallowing an error during `runBackend`?

Restioson:
Seems a bit weird that we could get to that point without having the node initialised - shouldn't we wait for it or something?

klochowicz:
@holzeis, to be honest I haven't dig into this very deep; the node got initialised soon after and this has unblocked me. I don't think there's any reason to be so strict with our invariant for something that is essentially just a log statement, as we make no use with it anyway.

Restioson:
Might be worth a followup in case this could cause other issues down the line?

klochowicz:
> Might be worth to get this in now and add followup ticket in case the node not being initialised by then could cause other issues down the line?

this PR can't cause any issues (it just fixes the logging thing) and everything worked well during extensive testing I've done on top of this. 
I suspect it might be something to running it with a debugger, which is not our typical production use-case.

I see no point in us *crashing the application* if we cannot log node id early at startup.

Restioson:
> this PR can't cause any issues (it just fixes the logging thing) and everything worked well during extensive testing I've done on top of this.

Sorry if I was confusing, but I didn't mean this PR, but rather what this PR _addresses_ could cause issues. This PR addresses the symptom of the node not being initialised. However, the node not being initialised may in and of itself cause other issues down the line. So, I agree that this should go right in, but we should also investigate the underlying cause (with lower priority).

lgleasain:
Just ran into this as well 

```
I/flutter (21854): | 2023-09-07 18:06:15.036 | LogLevel.INFO | Oracle endpoint: http://127.0.0.1:8081 
I/flutter (21854): | 2023-09-07 18:06:15.040 | LogLevel.INFO | Oracle PK: 16f88cf7d21e6c0f46bcbc983a4e3b19726c6c98858cc31c83551a88fde171c0 
F/libc    (21854): Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 21906 (1.ui), pid 21854 (ce.get10101.app)
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'alps/full_rlk6580_we_c_m/rlk6580_we_c_m:11/685/685:user/release-keys'
Revision: '0'
ABI: 'arm64'
Timestamp: 2023-09-07 18:06:15-0400
pid: 21854, tid: 21906, name: 1.ui  >>> finance.get10101.app <<<
uid: 10165
signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
    x0  0000000000000000  x1  0000000000005592  x2  0000000000000006  x3  0000007af761e930
    x4  0000000000000000  x5  0000000000000000  x6  0000000000000000  x7  7f7f7f7f7f7f7f7f
    x8  00000000000000f0  x9  0000007df97fb800  x10 ffffff80fffffbdf  x11 0000000000000001
    x12 0000000000000000  x13 0000000000000000  x14 0000007acfb9b2d4  x15 0000000000000005
    x16 0000007df9894c80  x17 0000007df98763a0  x18 0000007aee58a000  x19 00000000000000ac
    x20 000000000000555e  x21 00000000000000b2  x22 0000000000005592  x23 00000000ffffffff
    x24 0000007acfb97108  x25 0000000000000001  x26 0000007ad263d898  x27 000000790899a8e0
    x28 0000000800000079  x29 0000007af761e9b0
    lr  0000007df9829544  sp  0000007af761e910  pc  0000007df9829574  pst 0000000000000000
backtrace:
      #00 pc 000000000004e574  /apex/com.android.runtime/lib64/bionic/libc.so (abort+180) (BuildId: 3dd99fe7a181e7428f58a10f32f9fd0e)
      #01 pc 0000000002f27ac4  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #02 pc 0000000002f210f4  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #03 pc 0000000001342d64  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #04 pc 0000000002f24920  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #05 pc 0000000002f246ac  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #06 pc 0000000002f23278  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #07 pc 0000000002f24430  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #08 pc 0000000002f4327c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #09 pc 0000000002f43048  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #10 pc 0000000001039324  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #11 pc 000000000134a500  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #12 pc 0000000000ea3d04  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #13 pc 000000000133b3e4  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #14 pc 00000000010ffd9c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #15 pc 00000000011c8b50  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #16 pc 00000000011c818c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #17 pc 0000000000ffa44c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #18 pc 0000000001010e30  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #19 pc 0000000000fec9a8  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #20 pc 00000000013461b4  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #21 pc 00000000011c7e7c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #22 pc 0000000000ffe9ac  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #23 pc 0000000001010e30  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #24 pc 0000000000ff6cf8  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #25 pc 000000000134738c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #26 pc 00000000011c7670  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #27 pc 0000000000f24d8c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so
      #28 pc 0000000000e32c0c  /data/app/~~JqNuMu6QuOM2VkRlS6Uh9Q==/finance.get10101.app-FUZBdN8RqrEjQGH9KYDx7Q==/lib/arm64/libnative.so (wire_get_node_id+4)
      #29 pc 00000000000076a4  [anon:dart-code]
Lost connection to device.
```

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -100,8 +100,12 @@ pub fn get_node_key() -> SecretKey {
     NODE.get().inner.node_key()
 }
 
-pub fn get_node_info() -> NodeInfo {
-    NODE.get().inner.info
+pub fn get_node_info() -> Result<NodeInfo> {
+    Ok(NODE
+        .try_get()
```
luckysori, line 105: 🤔 It's kind of arbitrary to have this one be a `try_get` and not the others. It's circumstantial because this one just happens to be called earlier than the others at the moment.

I think we should investigate why the `Node` is not initialised when we expect it to be rather than papering over it. Looking at the code it does not seem possible that this would be a timing issue, because the `NODE` should be `set` before this is called. There's something strange going on.

klochowicz: I absolutely would not mind if *nothing* panics, ie. we use `try_get()` everywhere. IMHO it's unacceptable to panic in any case, especially when we can't get the logs as a result.
this is just not a good UX.

holzeis:
> We now log an error if for whatever reason we couldn't display the node id at
startup (e.g. node wasn't ready) - this should not be a showstopper by any means, and it gets
logged at later stage anyway.

I am wondering how we should deal with an app where the backend crashed. It is basically not usable to the user isn't it? 
Maybe we should be able to handle any unexpected crash / panic and ask the user to share the logs if that happens. The user shouldn't be able to do anything else in that state.

klochowicz:
> > We now log an error if for whatever reason we couldn't display the node id at
> > startup (e.g. node wasn't ready) - this should not be a showstopper by any means, and it gets
> > logged at later stage anyway.
> 
> I am wondering how we should deal with an app where the backend crashed. It is basically not usable to the user isn't it? Maybe we should be able to handle any unexpected crash / panic and ask the user to share the logs if that happens. The user shouldn't be able to do anything else in that state.

Depending *what* in the backend has crashed, I guess different things are possible.
For the sake of simplicity, we've got a giant monolithic 'node' that does everything right now (a lot of the things are unrelated). I am *not* proposing to change this, because it's not our highest priority and would take quite a lot of our time for questionable gain. One could argue that trades history and quite a lot of functionality should be still accessible, but I treat that as a very low priority.

I think in an ideal world we would at least display a message *what happened* and allow sharing the logs with 10101 before we exit the app. 
The reason for displaying a message is that sometimes the user can fix the error on their side! (e.g. issues with the internet). Empowering the user to fix this things would give us more happy users and less debugging crashlogs about which we can't do anything.

Eventually (the holy grail of crash recovery) would be if the user can export their seed words / cloud backup of their seed (+ VSS-style channel state eventually). Because you don't want them to lose money, if the issue is serious.

klochowicz:
closing this in favour of #1301

PR #1213, opened 2023-09-04
feat: model lightning payment fees

Restioson:
Here is how this looks for a payment from the faucet:
image

I'm just not really sure if this is true: is the fee for faucet -> wallet really 0?

This brings us closer to fully implementing #872


holzeis (approved):
LGTM 👍 (styling could be nicer 🙈)

> I'm just not really sure if this is true: is the fee for faucet -> wallet really 0?

That's correct, only the payer is paying fees.

Restioson:
> LGTM 👍 (styling could be nicer 🙈)

I've got a branch for this! PR will be opened shortly :) This is just modelling the data and making it available to the frontend

Restioson:
> That's correct, only the payer is paying fees.

In that case, should we disable displaying the fee if it's an inbound payment?

holzeis:
> > That's correct, only the payer is paying fees.
> 
> 
> 
> In that case, should we disable displaying the fee if it's an inbound payment?

It will always be 0 so we could as well hide it.

bonomat:
> > That's correct, only the payer is paying fees.
> 
> In that case, should we disable displaying the fee if it's an inbound payment?

I guess no, if it is an inbound payment we don't have to display the fee.

PR #1214, opened 2023-09-04
Refactor showSnackbar to fix lint on latest flutter

Restioson:
When I updated my flutter, I started getting a lint failure for using BuildContext across an async gap. This PR fixes that as well as refactoring showSnackbar so it can be used more widely


klochowicz (approved):

klochowicz (approved):
LGTM!

PR #1215, opened 2023-09-04
feat: swap side of dollar sign in fiat text (val$ -> $val)

Restioson:
image

The dollar sign should go in front of the amount (e.g `$10`, not `10$`, see [here](https://english.stackexchange.com/a/11327))


klochowicz (approved):
thanks for spotting this!

holzeis:
@Restioson can you apply the same change to the stable screen? 🙂

Restioson:
It is automatically applied since that is using FiatText too :)

Issue #227, opened 2023-03-07
Persist limit orders


bonomat:
The maker is currently not persisting anything, which means, on a restart, it won't remember anything. 

This ticket is about storing 
- orders and their states

Once we have positions, we should also store them

holzeis:
@bonomat this ticket looks relevant for milestone 1.0.22, but lacks some description. May I ask you to add details or close the ticket?

luckysori:
> Once we have positions, we should also store them

We can do this as part of https://github.com/get10101/10101/issues/973, or as a follow-up.

PR #1216, opened 2023-09-05
Simulate position between maker and coordinator

### LINKED ISSUE #973: Simulate position with coordinator
The maker does not actually setup a dlc with the coordinator as it is a trusted setup. Accordingly the maker will simply create a simulated position in his database once the coordinator setup the dlc with the taker.

This position will always be updated contrary to how the coordinator manages its positions which are always inserted newly on open trade.

luckysori:
Fixes #973.


coordinator/src/db/trades.rs
```diff
@@ -51,6 +51,57 @@ pub fn insert(
...
+    pub timestamp: OffsetDateTime,
+}
+
+/// Find all the trades for which we have a `temporary_contract_id`.
+///
+/// Some might not have one due to backwards compatibility.
+pub fn get_all_with_temporary_contract_id(
+    conn: &mut PgConnection,
+) -> Result<Vec<TradeForContract>> {
+    let trades: Vec<(
```
holzeis, line 69: 🔧 With #1246 the coordinator will hedge expired positions to the maker, so this query should filter expired positions, as they should be hedged by the coordinator. (that should also work in retrospect for positions that have expired before the change.)

coordinator/src/position/trade_position.rs
```diff
@@ -0,0 +1,432 @@
...
+///
+/// Basically, coordinator-apps (total) is equal to maker-coordinator.
+///
+/// It's important to note that this is the position _caused by trading_. That is, it
+/// excludes force-closures, position expiries, etc.
+fn compute_trade_position_maker(
+    trades: Vec<TradeForContract>,
+    contract_states: HashMap<[u8; 32], ContractState>,
+) -> ContractsPerContractSymbol {
+    let grouped_trades = &trades
```
holzeis, line 136: ❓ How will this logic behave when we are using rollovers, which is also changing the `temporary_contract_id`.

coordinator/src/cli.rs
```diff
@@ -64,6 +64,10 @@ pub struct Opts {
         default_value = "16f88cf7d21e6c0f46bcbc983a4e3b19726c6c98858cc31c83551a88fde171c0"
     )]
     oracle_pubkey: String,
+
+    /// The endpoint of the maker.
+    #[clap(long, default_value = "http://localhost:18000")]
```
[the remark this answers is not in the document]
holzeis: wouldn't it be nicer the maker calls an coordinator api instead of the coordinator? You are anyways only doing the sync in an interval. That could as well also happen on the maker.

holzeis (commented):
It's hard to review, the code is very complex! I added some remarks after the first round going through the code, but I'd like to review it a second time again before approving.

I feel like we are making this very complicated. I think we wanted to add the maker so that the design is correct from the beginning, but introducing this shortcut kind of fails that purpose. Not sure if it would have been easier to do everything in the coordinator.

Cargo.toml
```diff
@@ -20,3 +20,6 @@ lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/
 lightning-persister = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
 
 rust-bitcoin-coin-selection = { git = "https://github.com/p2pderivatives/rust-bitcoin-coin-selection" }
+
+# Waiting for the next release
+xtra = { git = "https://github.com/Restioson/xtra/", rev = "d98393a" }
```
bonomat, line 25: Do we really want to introduce xtra?

bonomat (commented):
I'm not a fan of this solution. 
Can't we use the websocket approach where the coordinator tells the maker if a match was found and then simply ignore the process of setting up a DLC? 
The downside of this is that the maker wouldn't know if the position was correctly setup with the taker or not but I guess we just need to use some optimism here?

Cargo.toml
luckysori: Since `xtra` is such a small dependency and the actor pattern is useful here, I am keen to introduce it.

luckysori:
@bonomat @holzeis Please review this again.

I have not been able to test it locally because my new environment is not letting me start the app locally atm. I will only merge once I am able to resolve those problems and verify that the order matches are propagated correctly.

Cargo.toml
Restioson: That reminds me to cut a release from master on xtra

holzeis (commented):
The change is definitely simpler than before. I like it. However it is certainly not as powerful as your other proposal.

While before the coordinator was relying entirely on the dlc states (which is the most reliable source of truth) with this change the maker solely relies on the updates from the coordinator / orderbook.

I think that is the way to go, but it will also put us at the risk that the position reflected to the maker may not be entirely correct. However, I think we need to find those scenarios and have to update the maker accordingly.

@luckysori @bonomat what about existing positions? Would it make sense to add an api to the coordinator returning all orders where the maker has been matched / is matched to an active position? That way the maker is not relying solely on getting updates in deltas.

maker/src/position.rs
```diff
@@ -0,0 +1,203 @@
...
+        } = update;
+
+        self.position.update_tentenone(contract_symbol, contracts)
+    }
+}
+
+/// The overall position of the maker.
+#[derive(Debug)]
+struct Position {
+    tentenone: HashSet<tentenone::Position>,
```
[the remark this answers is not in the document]
luckysori: Yep.

maker/src/orderbook_ws.rs
```diff
@@ -0,0 +1,128 @@
...
+                let url = url.clone();
+                let authenticate = auth_fn;
+                let mut stream = spawn_blocking(move || {
+                    orderbook_client::subscribe_with_authentication(url, authenticate)
+                })
+                .await
+                .expect("task to complete");
+
+                while let Ok(Some(msg)) = stream.try_next().await {
+                    {
```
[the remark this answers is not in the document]
luckysori: Nope, good catch!

luckysori:
@holzeis: I've updated the PR with a proposal to ensure that the maker's simulated position is resilient to restarts.

Unfortunately I still cannot test this e2e locally because my app won't build.

maker/src/orderbook_ws.rs
```diff
@@ -87,35 +95,61 @@ impl Client {
...
     msg: String,
     position_manager: &xtra::Address<position::Manager>,
+    maker_trader_id: &PublicKey,
 ) -> Result<()> {
     tracing::trace!(%msg, "New message from orderbook");
 
     let msg = serde_json::from_str::<OrderbookMsg>(&msg).context("Deserialization failed")?;
 
     match msg {
+        OrderbookMsg::AllOrders(orders) => {
```
holzeis, line 105: 🚨You will only received non-expired orders here. You won't get consistently the taken orders here over time as the makers orders expire after 60 seconds. (unless configured differently)

holzeis (changes requested):
I don't think we can rely on the `AllOrders` event here, as they won't include expired orders.

maker/src/orderbook_ws.rs
luckysori: Thanks for the tip. The name definitely does not convey this.

I'm going to add an API to the WebSocket to get all the matched/taken orders for a trader.

luckysori: It turns out none of the current orderbook messages work well for this purpose, so I just ended up adding a new one and letting the maker ask for it via a new orderbook request.

It would be much nicer if the orderbook could drive these updates instead, but it's not easy to implement, specially since we want to avoid adding maker-specific logic to the orderbook/coordinator.

maker/src/orderbook_ws.rs
```diff
@@ -0,0 +1,178 @@
...
+                    )
+                })
+                .collect::<Vec<_>>();
+
+            tracing::info!(n = %orders.len(), "Received limit order filled matches");
+
+            let _ = position_manager.send(PositionUpdateTenTenOne(orders)).await;
+        }
+        OrderbookMsg::Match(FilledWith { order_id, .. }) => {
+            // We cannot rely directly on this message because the match does not specify the
```
bonomat, line 156: > We cannot rely directly on this message because the match does not specify the direction.

Why not? We know the `order_id` and given we created the order, we know the direction.

coordinator/src/orderbook/db/orders.rs
```diff
@@ -306,6 +308,52 @@ pub fn get_by_trader_id_and_state(
...
+
+    let filled_matches = orders
+        .into_iter()
+        .map(|(order_id, quantity, direction_maker)| {
+            let quantity = Decimal::from_f32(quantity).expect("to fit into Decimal");
+
+            let quantity = match direction_maker {
+                Direction::Long => quantity,
+                Direction::Short => -quantity,
+            };
```
bonomat, line 348: I like this logic, maybe we should start using this accross the board.

maker/src/position.rs
```diff
@@ -0,0 +1,220 @@
...
+            .cloned()
+            .unwrap_or(tentenone::Position::new(ContractSymbol::BtcUsd));
+
+        position.update(order_id, contracts);
+
+        self.tentenone.replace(position);
+    }
+
+    #[cfg(test)]
+    fn get_tentenone(&self, contract_symbol: &ContractSymbol) -> Decimal {
```
bonomat, line 110: Should we add an HTTP Api to query this ? Could come handy in our production setup to know what is going on .

maker/src/position.rs
```diff
@@ -0,0 +1,220 @@
+use async_trait::async_trait;
```
bonomat, line 1: Honest rust question: why is this not in `./position/mod.rs`?

bonomat (approved):
I have a hard time imagine how this evolves once we have a bitmex position as well. But let's get this in if it works :)

maker/src/orderbook_ws.rs
luckysori: I left the comment because I tried it and it didn't work. The problem stems from the fact that the `quantity` in the `Match` is absolute, so the sign is always positive and there is no way to figure out the direction.

If we remembered the orders we created as the maker we could figure it out, but that information is not persisted on the maker.

maker/src/position.rs
luckysori: It's optional to use `mod.rs` since Rust 2018, although I prefer not using it: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs.

At some point @holzeis wanted to try out introducing `mod.rs`, but it's a bit annoying when you have multiple files open with the same `mod.rs` name. I would be keen to align it and remove all `mod.rs` files in favour of the newer standard.

maker/src/orderbook_ws.rs
bonomat: > If we remembered the orders we created as the maker we could figure it out, but that information is not persisted on the maker.

Right ... there is this ticket for that https://github.com/get10101/10101/issues/227

luckysori: Good point!

I think it wouldn't hurt to persist the limit orders and it might be handy in this scenario, but it's still pretty limited as we only learn about the match and we cannot guarantee that the trade is executed. At least once the order is filled we know that the protocol was successfully started, so it's very likely that the trade will be executed.

The issue also mentions persisting the position, but in the current version where the coordinator is the only source of truth for this I actually thing relying solely on the orderbook/coordinator is best, as there can't be disagreements. Once we introduce maker-coordinator DLCs we can get rid of all this, and then persisting some position metadata will be more interesting.

holzeis (approved):

coordinator/src/orderbook/db/orders.rs
luckysori: How do you think we should action your comment? I assume we are already mapping like this elsewhere.

bonomat: Oh, maybe we are already doing this :)

Issue #947, opened 2023-07-13
Release for Android


bonomat:
We should release our apk for Android as well. There should be a similar think as Apple's Testflight for Android. Alternatively (or additionally) we can ship the APK file directly.

DennisDyallo:
Hi! I'm not an Android dev, but been using Android for over a decade. There is a way to release into a smaller crowd.

As a developer:
https://support.google.com/googleplay/android-developer/answer/9845334?hl=en

As a user:
https://support.google.com/googleplay/answer/7003180?hl=en

Anyway, I hope Android could be prioritized more. We're probably a lot more Android users ready to test your app than iPhone users. Android users are also more tech savvy. Greater addressable market!

PR #1219, opened 2023-09-05
CI Pipeline to publish to Google Play Store

bonomat:
Resolves https://github.com/get10101/10101/issues/947

Tested on my own fork: https://github.com/bonomat/10101/actions/runs/6085887769/job/16511009829

Note: only regtest version has been tested, mainnet version is under review and cannot yet be installed but it should work™️


.github/workflows/android-fastlane.yml
```diff
@@ -0,0 +1,187 @@
+name: "Fastlane: Build and release ios app to TestFlight"
```
holzeis, line 1: typo: iOS - android to Google Play Store

holzeis (approved):
Looks good, let's give it a go! 🚀

justfile
```diff
@@ -15,8 +15,11 @@ public_coordinator_http_port := "80"
 # command to get the local IP of this machine
 get_local_ip := if os() == "linux" {
  "ip -o route get to 1 | sed -n 's/.*src \\([0-9.]\\+\\).*/\\1/p'"
+} else if os() == "macos" {
+ "ipconfig getifaddr en0"
```
klochowicz, line 19: I'm not sure whether this will work on all the machines - does the current solution not work for you? 
are you using wifi or ethernet?

klochowicz: there's little detail in your commit message - it would have been easier if you described what didn't work for you.

klochowicz (approved):
exciting.
the only thing I'm not sure whether it's needed / correct / belongs in this PR is the change of the local_ip derivation.

justfile
bonomat: Does the old solution work for you?

`ifconfig | grep -Eo 'inet (addr:)?([0-9]*\\.){3}[0-9]*'` does not return anything for me.

I'm happy to drop this commit.

PR #1227, opened 2023-09-06
Add plumbing for sending push notifications

klochowicz:
- store the FCM tokens sent from the app in the coordinator
- add a notification service that sends push notifications to users (provide a sender as a means to send messages from different parts of the system).

Next step would be actually using this new NotificationService to send notifications.


coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+        tokio::spawn({
+            let fcm_api_key = fcm_api_key;
+            let client = fcm::Client::new();
+            async move {
+                while let Some(Notification {
+                    user_fcm_token,
+                    notification_kind,
+                }) = notification_receiver.recv().await
+                {
+                    tracing::info!("Sending notification: {notification_kind} to {user_fcm_token}");
```
bonomat, line 69: Imho it's nicer to _annotate_  (not sure if this is the right term) the log message with the variables, i.e. 

```
tracing::info!(notification_kind, user_fcm_token, "Sending notification");
```

coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+/// Prepares the notification text
+fn build_notification<'a>(kind: NotificationKind) -> fcm::Notification<'a> {
+    let mut notification_builder = fcm::NotificationBuilder::new();
+    match kind {
+        NotificationKind::ChannelClose => {
+            notification_builder.body("Close channel request.");
+            notification_builder.title("Someone wants to close a position with you! 🌻");
```
bonomat, line 101: Is this message meant to be for closing a channel (LN), a subchannel (DLC-LN) or a position (similar to DLC-LN but n a different domain-level)?

coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+        NotificationKind::PositionSoonToExpire => {
+            notification_builder.title("Your Position is about to expire");
+            notification_builder.body("Open the app to react.");
+        }
+        NotificationKind::PositionExpire => {
+            notification_builder.title("Your position has expired");
```
bonomat, line 108: Here we go, it's already expired :) I would call the variant then `NotificationKind::PositionExpired`

coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+
+async fn send_notification<'a>(
+    client: &fcm::Client,
+    api_key: &str,
+    fcm_token: &str,
+    notification: fcm::Notification<'a>,
+) -> Result<()> {
+    anyhow::ensure!(!api_key.is_empty(), "FCM API key is empty");
+
+    let mut message_builder = fcm::MessageBuilder::new(api_key, fcm_token);
```
bonomat, line 123: ah, here is `fcm` used. I would add the dependency in the same commit to make it atomic. 
For example, if we were to undo this commit only we would miss removing `fcm` from cargo.toml.

mobile/lib/main.dart
```diff
@@ -63,12 +63,22 @@ import 'features/stable/stable_value_change_notifier.dart';
 final GlobalKey<NavigatorState> _rootNavigatorKey = GlobalKey<NavigatorState>(debugLabel: 'root');
 final GlobalKey<NavigatorState> _shellNavigatorKey = GlobalKey<NavigatorState>(debugLabel: 'shell');
 
-void main() {
+void main() async {
```
bonomat, line 66: Does this work? My limited flutter knowledge says you'd need

```suggestion
Future<void> main() async {
```

bonomat: Also, can the main be async?

mobile/pubspec.lock
```diff
@@ -1073,26 +1073,26 @@ packages:
...
   test_core:
     dependency: transitive
     description:
       name: test_core
-      sha256: "99806e9e6d95c7b059b7a0fc08f07fc53fabe54a829497f0d9676299f1e8637e"
+      sha256: "5138dbffb77b2289ecb12b81c11ba46036590b72a64a7a90d6ffb880f1a29e93"
       url: "https://pub.dev"
     source: hosted
-    version: "0.5.3"
+    version: "0.5.1"
```
bonomat, line 1095: why all the downgrades?

coordinator/src/db/user.rs
```diff
@@ -24,6 +28,8 @@ impl From<RegisterParams> for User {
             email: value.email.unwrap_or("".to_owned()),
             nostr: value.nostr.unwrap_or("".to_owned()),
             timestamp: OffsetDateTime::now_utc(),
+            // TODO: Consider allowing passing the FCM token in the registration request.
+            fcm_token: "".to_owned(),
```
bonomat, line 32: Is a user's fcm token final or will it change? I guess if the token is final you could pass it in the registration.

mobile/lib/main.dart
```diff
@@ -560,12 +561,39 @@ Future<void> configureFirebase(FlutterLocalNotificationsPlugin localNotification
...
+      updateFcmToken(token).then((_) {
+        FLog.info(text: "Firebase token updated");
+      });
+    } else {
+      FLog.warning(text: "Firebase token is null");
+    }
+  });
+
+  // Firebase sometimes updates tokens at runtime, make sure we handle that case
+  messaging.onTokenRefresh.listen((String token) {
```
bonomat, line 578: Ah, so it is not final, then I would not put it into the registration request

bonomat (commented):
A few nits but overall I think this goes into the right direction.

coordinator/Cargo.toml
```diff
@@ -11,6 +11,7 @@ diesel_migrations = "2.0.0"
 dlc = "0.4.0"
 dlc-messages = "0.4.0"
 dlc-trie = "0.4.0"
+fcm = "0.9.2"
```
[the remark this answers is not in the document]
bonomat: It's being used here: https://github.com/get10101/10101/pull/1227#discussion_r1316920855

coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+    PositionSoonToExpire,
+    PositionExpire,
+}
+
+impl Display for NotificationKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NotificationKind::ChannelClose => write!(f, "ChannelClose"),
+            NotificationKind::PositionSoonToExpire => write!(f, "PositionSoonToExpire"),
+            NotificationKind::PositionExpire => write!(f, "PositionExpire"),
```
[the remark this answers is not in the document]
bonomat: Seems like your intention was to say the position already expired. See here: https://github.com/get10101/10101/pull/1227#discussion_r1316918945

coordinator/src/db/user.rs
bonomat: Looks like it is not final, then I would not put it here: https://github.com/get10101/10101/pull/1227#discussion_r1316940179

coordinator/src/routes.rs
```diff
@@ -117,6 +118,7 @@ pub fn router(
         .route("/api/trade", post(post_trade))
         .route("/api/rollover/:dlc_channel_id", post(rollover))
         .route("/api/register", post(post_register))
+        .route("/api/fcm_token", post(post_fcm_token))
```
holzeis, line 121: We really should start authenticating the user.

https://github.com/get10101/10101/issues/1255

mobile/lib/main.dart
```diff
@@ -63,12 +63,23 @@ import 'features/stable/stable_value_change_notifier.dart';
...
-void main() {
+void main() async {
   WidgetsBinding widgetsBinding = WidgetsFlutterBinding.ensureInitialized();
   FlutterNativeSplash.preserve(widgetsBinding: widgetsBinding);
 
   setupFlutterLogs();
 
+  try {
+    FLog.info(text: "Initialising Firebase");
+    await initFirebase();
```
holzeis, line 74: 🔧 we should put all the firebase specific code into a service instead of having it in the main.dart

mobile/native/src/api.rs
```diff
@@ -328,6 +328,12 @@ pub async fn register_beta(email: String) -> Result<()> {
     order::register_beta(email).await
 }
 
+/// Send the Firebase token to the LSP for push notifications
+#[tokio::main(flavor = "current_thread")]
+pub async fn update_fcm_token(fcm_token: String) -> Result<()> {
+    order::update_fcm_token(fcm_token).await
```
holzeis, line 334: 🔧 the orders module does not seem to be the right one for the `update_fcm_token` function. How about creating a `users` module?

holzeis (commented):
It would be nice if we could combine the notifications logic for the websocket with the push notifications into a dedicated module.

coordinator/src/notification_service.rs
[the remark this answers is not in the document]
klochowicz: Yes, a typo, fixed it. :)

mobile/lib/main.dart
klochowicz: yes, it works and all the examples in the official Dart documentation show it like this:

https://dart.dev/codelabs/async-await#working-with-futures-async-and-await

mobile/lib/main.dart
klochowicz: yes, it's not final; it can change at any arbitrary point in time (it doesn't change often though).

coordinator/src/routes.rs
klochowicz: I don't disagree with your statement, but how is this related to this particular ticket? 🤔

are you suggesting that this PR makes the security situation any worse?

if yes, I'd like to explain that, unline some of the other endpoints, this one has *minimal* security implications. if someone happens to spoof an FCM token they'd get a push notification not meant for them in the worst case.

holzeis: Just referring the ticket so we don't forget to tackle this as well.

mobile/lib/main.dart
klochowicz: moved into a dedicated module.

mobile/native/src/api.rs
klochowicz: good point. moved to `users` module, along with registration part, which didn't fit there either 😊

klochowicz:
> It would be nice if we could combine the notifications logic for the websocket with the push notifications into a dedicated module.

I'm afraid I don't follow you. would you mind to elaborate?

coordinator/Cargo.toml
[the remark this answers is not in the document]
klochowicz: I've squashed these 2 commits so it's easier if you review commit-by-commit.

mobile/pubspec.lock
klochowicz: I'm actually not sure why they ended up downgrading 🥺
I've reverted the change and they didn't appear again.

klochowicz: I've reverted it from here, but it should probably go back in; I suspect that @Restioson had a toolchain that was above our supported flutter version and that bumped the packages. My Dart is `3.0` and it always brings it down. 
I'll have a look whether newer Dart can coexist with the same Flutter version we officially use.

holzeis:
> > It would be nice if we could combine the notifications logic for the websocket with the push notifications into a dedicated module.
> 
> I'm afraid I don't follow you. would you mind to elaborate?

No need to action in this PR, but we currently have an api for notifying the user using the websocket and I guess now we get a second for notifying via push notification. It would be nice if we could have a single api to notify the user, e.g. with a combination try first websocket if not possible notify on push notification.

holzeis (approved):
Let's give it a go if it works 👍

coordinator/src/notification_service.rs
```diff
@@ -0,0 +1,132 @@
...
+    ChannelClose,
+    PositionSoonToExpire,
+    PositionExpired,
+}
+
+impl Display for NotificationKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NotificationKind::ChannelClose => write!(f, "ChannelClose"),
+            NotificationKind::PositionSoonToExpire => write!(f, "PositionSoonToExpire"),
```
holzeis, line 20: nit: maybe that's the `rollover` notification?

bonomat (approved):
LGTM

coordinator/src/notification_service.rs
klochowicz: I thought about it for a while, but the it's *before* rollover, and the user can decide whether to rollover or not.
I wonder whether we should also allow notifications *after* a successful rollover etc?

klochowicz:
> > > It would be nice if we could combine the notifications logic for the websocket with the push notifications into a dedicated module.
> > 
> > 
> > I'm afraid I don't follow you. would you mind to elaborate?
> 
> No need to action in this PR, but we currently have an api for notifying the user using the websocket and I guess now we get a second for notifying via push notification. It would be nice if we could have a single api to notify the user, e.g. with a combination try first websocket if not possible notify on push notification.

Oh, I see. I'm not sure whether this can be achieved, because they're constraints are different.

1. Websocket stream is *current* info to online & authenticated users.
2. Push notifications is information for users that are can be offline.

It seems that it's easy to draw the line between the content of this information, e.g. there's no reason to send the same info on both channels. 
I imagine that the push notification subsystem can be achieved by having their own tokio tasks that monitor for things, blissfully unaware that there is such a thing as websocket stream from the orderbook (and therefore having loose coupling).

Another thing that IMHO makes them logically separate is that the websocket stream comes from *orderbook*, whereas the push notifications are sent from the *coordinator* (our LSP).

PR #1228, opened 2023-09-06
fix: get local ip on mac

bonomat:
the old command does not return anything for me


Restioson:
This gives me deja vu... didn't we already do this?

Restioson (approved):

klochowicz (approved):
doublechecked that the new one works on intel macs too 👍

Issue #1069, opened 2023-08-04
Pay funding rate


bonomat:
https://github.com/get10101/10101/issues/1068 introduces perpetual futures. 
A funding rate needs to be paid for keeping a position open. 

In contrast to what we did in the past I propose 2 changes: 

1. we only charge the funding rate at the funding intervals (TBD), e.g. every 8 hours. 
Meaning, if a user has a position open from 0:00-7:59 he does not pay a funding rate, but if he has the position open from 7:59-8:01 he pays one. 
2. The funding rate is defined by the orderbook - this allows us to have a "real" funding rate depending on the current orders in the orderbook.

bonomat:
To pay funding rate we update the margin of the app/coordinator.

PR #1233, opened 2023-09-06
Propose a solution for the funding rate

luckysori:
I decided to write an ADR because I came to the conclusion that the original solution reflected in #1069 does not work.

Because none of the solutions discussed in the document are ideal, I want to use this PR as an opportunity to reach consensus before actually implementing a solution to #1069.

### DIFF
--- a/docs/007-funding-rate.adoc
+++ b/docs/007-funding-rate.adoc
@@ -8,9 +8,8 @@ Status: Proposed
 
 == Decision
 
-The funding rate should be charged through a Lightning payment.
-The funding rate *must* be paid before renewing the contract.
-No funding rate needs to paid for the period of time _before_ the first renewal.
+Use the `ContractInput` argument when calling `dlc_manager::Manager::renew_offer`.
+By setting new values of `offer_collateral` and `accept_collateral` on `ContractInput` we can rebalance the margins between the two parties, which will have the desired effect on the payouts.
 
 == Context
 
@@ -22,6 +21,23 @@ In the 10101 ecosystem, the funding rate must come from the 10101 orderbook.
 This value will be updated at a fixed interval e.g. every 8 hours.
 
 == Options considered
+
+=== Use the `ContractInput` argument in the `rust-dlc` DLC channel renew protocol
+
+By setting a new collateral distribution between trader and coordinator based on the funding fee, we will recreate the DLC with new margins.
+The updated margin will have an effect on how the payout curve should look like.
+
+If the contract is _refunded_ (catastrophic error), the returned collateral will take into account all the funding fees exchanged.
+
+==== Pros
+
+- Atomic update to the DLC and funding fee exchange.
+- Easy implementation.
+- Funding fee exchange is reflected on the `refund_transaction`.
+
+==== Cons
+
+- Fee is only available after the DLC channel is settled.
 
 === Lightning payment before the DLC is renewed
 
@@ -94,3 +110,4 @@ Furthermore, atomicity can be achieved once PTLCs are supported by `rust-lightni
 == Advice
 
 - Philipp on Sep 6, 2023: Adjusting the CET payouts depending on the funding fee is preferred over the Lightning payment before the DLC is renewed.
+- Lucas on Sep 7, 2023: The renew protocol _can_ be used to solve this problem, just not in the way that we originally imagined.
--- a/docs/007-funding-rate.adoc
+++ b/docs/007-funding-rate.adoc
@@ -1,0 +1,96 @@
+= ADR 007 - Funding rate
+Lucas Soriano <lucas@10101.finance>
+:toc:
+:icons: font
+:attributes: 2023-09-06
+
+Status: Proposed
+
+== Decision
+
+The funding rate should be charged through a Lightning payment.
+The funding rate *must* be paid before renewing the contract.
+No funding rate needs to paid for the period of time _before_ the first renewal.
+
+== Context
+
+The funding rate is the cost of holding a perpetual swap position that is on the side of the market sentiment.
+There needs to be a funding rate because that is what motivates traders to keep their perpetual swaps positions open _against_ the market sentiment.
+The sign of the funding rate determines which of the two parties involved must pay the funding fee to keep their position open.
+
+In the 10101 ecosystem, the funding rate must come from the 10101 orderbook.
+This value will be updated at a fixed interval e.g. every 8 hours.
+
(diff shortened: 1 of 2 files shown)


docs/007-funding-rate.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+They would still be able to close their position and open a new one.
+
+=== Adjust the CET payouts depending on the funding fee
+
+If the funding fee is `X` paid from `A` to `B`, for every `o_A`, CET output amount belonging to `A`, we would subtract `X`: `o_A - X`; and for every `o_B`, CET output belonging to `B`, we would add `X`: `o_B + X`.
+
+Unfortunately this doesn't work in all scenarios:
+
+- If a CET already only pays to `B` then the payout cannot be increased!
```
bonomat, line 63: If this is the case the user (`A`) was basically liquidated.

docs/007-funding-rate.adoc
```diff
@@ -0,0 +1,96 @@
...
+- If a CET has `o_A < X`, then it will not be possible to increase `o_B` sufficiently.
+
+As such, in some scenarios the fee will not be paid.
+This problem might compound as the same position might be rolled over multiple times.
+If the advantaged party ends up getting liquidated, the winning party will not see any funding fee.
```
bonomat, line 68: Imho this is a good design: the longer the position is open, the earlier the user gets liquidated because its margin is being reduced by the funding rate. 

Expressed differently: 

- Alice has margin `m_A`
- Bob has margin `m_B`
- Funding rate is `f`.

After the first rollover, e.g. after 8h, we have 
- Alice margin `m_A - f`
- Bob margin `m_B + f`

The liquidation for the party going long is calculated as: 
`= (price_opening * leverage)/(leverage+1)`
 
Now, the important thing is that the `leverage` is only constant at the beginning when the position was opened, later the `leverage` is a function of the `margin` and the `price`, i.e. 
- initially we had: `margin = quantity / (price_opening * leverage)`
- after the opening we have `leverage = quantity / (price_opening * margin)`

Which means, the liquidation price will change but each party will always get the funding rate until either of the party gets liquidated, i.e. does not have enough money to pay for the funding rate.

docs/007-funding-rate.adoc
```diff
@@ -0,0 +1,96 @@
...
+:icons: font
+:attributes: 2023-09-06
+
+Status: Proposed
+
+== Decision
+
+The funding rate should be charged through a Lightning payment.
+The funding rate *must* be paid before renewing the contract.
+No funding rate needs to paid for the period of time _before_ the first renewal.
```
bonomat, line 13: One important thing we should consider is that funding rate is charged retrospectively and not in advance, e.g. 

- the funding rate should be charged every 8h (00:00, 8:00, 16:00 - aka `funding event`)
- a position which is open at these times should require a funding rate to remain open
- the funding is paid at the `funding event`
- a position which was only opened between two `funding events`, may not be charged, e.g. a position open between 00:01 and 07:59
  - this is ok, because you can't _close_ a position, you can only sell it to someone else in the market who then has the position open past the next funding event. 

The problem I want to solve is the following: 
- you might have multiple positions before a funding event, but only one during the funding event, e.g. a user who constantly buys and sells opened and closed his position multiple times. 
  - If we were to charge the user the funding rate of 8h in advance, we would need to charge him for each trade, 
  - while if we charge him retrospectively at the funding event, we only charge him once

bonomat (commented):
Thanks for writing this up, I would go with `Adjust the CET payouts depending on the funding fee`

docs/007-funding-rate.adoc
luckysori: This is true. But that does not mean that the funding fees were paid.

Basic example:

- 1st rollover makes `A` pay to `B`. The "A gets liquidated" CET cannot get adjusted because the output already pays everything to `B`.
- At the updated expiry the oracle implicitly unlocks the "A gets liquidated" CET. `B` should have earned the full DLC payout _plus the funding fee_, but they only get the DLC payout.

In a more extreme scenario, if the contract is rolled over 20 times and the funding fee is paid 15 times from `A` to `B`, if `A` gets liquidated with the CET at the boundary of the curve, `B` misses out on `15 - 5 = 10` trading fees.

Obviously this only needs to apply to contracts that are closed on-chain. If I were keeping track of the funding fees outside of the CETs I would want to do the accounting without capping gains/losses based on the collateral. But this would mean there would be a discrepancy between on-chain and off-chain settlement.

docs/007-funding-rate.adoc
luckysori: There are a couple of reasons why I don't like this idea so much:

- Makes liquidation more likely, which is the worst case scenario for traders and is expensive and cumbersome to enforce.
- Traders can miss out on funding fees (see [comment above](https://github.com/get10101/10101/pull/1233#discussion_r1317840973)).

docs/007-funding-rate.adoc
luckysori: > the funding rate should be charged every 8h

I was thinking about this yesterday and this means the user needs to come online every 8 hours. I guess this might be solved by our work on background tasks though. Long expiries (currently set to 7 days, although pretty arbitrary) might not work well with funding rates.

Updating the expiry and charging the funding rate have to happen together because the funding rate is paid precisely to have the right to keep the position open.

> The problem I want to solve is the following:
>
>- you might have multiple positions before a funding event, but only one during the funding event, e.g. a user who constantly buys and sells opened and closed his position multiple times.
>     - If we were to charge the user the funding rate of 8h in advance, we would need to charge him for each trade,
>     - while if we charge him retrospectively at the funding event, we only charge him once

I think the proposed solution does respect this, but we cannot ensure atomicity (for now) if something goes awry.

docs/007-funding-rate.adoc
bonomat: > Makes liquidation more likely, which is the worst case scenario for traders and is expensive and cumbersome to enforce.

Yes, but I don't think it's that much more likely, e.g. current funding rate on BitMEX is 0.01% for 8h, meaning, his margin gets reduced by 0.01% and it's 0.01% more likely that you get liquidated. That's like nothing :D

luckysori:
After investigating the differences between (1) adapting the margin and (2) adapting the payouts, I have concluded that the effect is equivalent. (1) is obviously much simpler so I think we can merge this ADR in its current state as discussed offline.

PR #1234, opened 2023-09-06
fix: use string for url instead of URL

bonomat:
The reason for that is that URL is putting a trailing / at the file which becomes an issue in downstream libraries such as esplora-client. This is a temp fix, a proper fix will follow.

Note: make sure to pass in ESPLORA_ENDPOINT always without a trailing `/`, i.e. `ESPLORA_ENDPOINT=http://46.17.98.29:3000`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -617,7 +617,7 @@ dependencies = [
 
 [[package]]
 name = "coordinator"
-version = "1.2.4"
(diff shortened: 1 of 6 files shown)


bonomat:
> Thanks for fixing! 👍

Thanks for helping debugging it.

Issue #1205, opened 2023-09-01
No pending funding transaction found!


holzeis:
I just opened a new JIT channel on mainnet and did not pay JIT channel opening fees, because the transaction could not be found upon the channel ready event.

The following error happened when trying to register the funding transaction to get payed.

```
Failed to fetch transaction: 5430bccdf192cf1df8475e9d72bb6403cd8d714e2fad064fcdc6795bffb7c44a from esplora. Error: reqwest::Error { kind: Decode, source: Error("expected value", line: 1, column: 1) } 
```

This results in no funding transaction getting registered and thus when trying to pay for it, it fails because we do not know how much fees we have to pay.

holzeis:
I think @luckysori suggestion to generally setup an async task that is regularly checking if fees have to be paid may be the less failure prune approach here.

holzeis:
Maybe as well fixed, when running our own esplora instance.

bonomat:
I assume the problem was that we checked too fast for the transaction. It was eventually confirmed. 
https://mempool.space/tx/5430bccdf192cf1df8475e9d72bb6403cd8d714e2fad064fcdc6795bffb7c44a


I wonder if we should bother with this or try to build out the proper LSP flow https://github.com/get10101/10101/issues/843?

bonomat:
Unasigning myself again. I believe this requires a different fix than just using our self-hosted esplora instance. 
I propose to 
- make the fetching more robust, i.e. retry
- look into if we can use the LSP flow from #843

holzeis:
> I assume the problem was that we checked too fast for the transaction. It was eventually confirmed. 

The broadcast should have happened over the same explora endpoint and it was definitely broadcasted before we make the query. Why do you think that we checked too fast? Shouldn't we also be able to fetch unconfirmed transactions?

holzeis:
My guess is that we run into a 429 too many requests error.

bonomat:
I don't think so, we would have seen this in the log as shown here: https://github.com/get10101/10101/issues/658#issue-1717224782

holzeis:
> I don't think so, we would have seen this in the log as shown here: https://github.com/get10101/10101/issues/658#issue-1717224782

I don't think so look at the error

Error: reqwest::Error { kind: Decode, source: Error("expected value", line: 1, column: 1) }

The response couldn't be parsed. We do not check if the http request was actually successful. My guess is that if we would check that we would have seen a 429 error.

bonomat:
[...]

PR #1236, opened 2023-09-06
chore: make fetching tx fee more resilliant

bonomat:
it looks like we sometimes have a timing issue when fetching a tx because the publishment was not yet finalized, with this we simply retry after one second

resolves https://github.com/get10101/10101/issues/1205


holzeis (approved):
I don't believe it's a timing issue, but that change doesn't hurt either.

PR #1239, opened 2023-09-07
Add user registration timestamp to `dlc_channels` endpoint

luckysori:
Often times we get multiple entries with the same (or similar) `user_email`s as users reopen their channels.

Adding the context of when the user was registered to this HTTP response helps identify the relevant channel quickly when debugging.


holzeis (approved):

luckysori:
Bypassing because we have a couple of flaky tests unrelated to this PR.

PR #1246, opened 2023-09-07
Deal with expired position the right way

### LINKED ISSUE #1071: Show loading screen when app starts with an expired position
So far we collaboratively close a position from the coordinator whenever a user comes online. 

This has the disadvantage that a user might not be aware of a protocol being executed and might close the app too fast which could break things. 

To overcome this we envision a loading screen when the user comes online which blocks the app for a short time and executes the protocol in the background.

### LINKED ISSUE #1173: Ensure that expired positions are updated on the maker
If a position is expired the coordinator should reduce the maker's position so that he can update his hedged position. 

Basically the coordinator needs to take an order of the opposite direction from the maker.

holzeis:
This PR got quite large and addresses several issues we had with closing expired positions.

1. Expired positions are now automatically matched against the orderbook

Since we currently have only one maker running which is publishing short and long orders this means the order for the expired position will get matched with the maker again, basically reducing his position with the coordinator.

2. Once the user comes online (connects to our websocket) we check for any pending matches (e.g. due to an expired position) and send a `AsyncMatch` message which the app is immediately executing.

For the following video, I reduced the position expiry to 30 seconds and the regular interval check for expired positions to 5 seconds. What happened in the background is that the orderbook found that the position expired found a match and saved that for later when the user is coming back online. Once online the match is executed with the user.

https://github.com/get10101/10101/assets/382048/ed0855bb-a732-4978-b24e-fc31ab8ab96e

resolves #1173 
resolves #1071


crates/orderbook-commons/src/lib.rs
```diff
@@ -172,31 +172,42 @@ pub struct FilledWith {
...
+            })
+            .collect();
+        average_execution_price(amounts)
+    }
+}
 
-        let nominal_prices: Decimal = self.matches.iter().fold(Decimal::ZERO, |acc, m| {
-            acc + (m.quantity / m.execution_price)
-        });
+pub struct QuantityPrice {
```
bonomat, line 220: Nit: 

That's a weird struct 🙈. Maybe make two out of them and pass them in separately? 
Alternatively, pass in the match and spare the creation of the new struct?

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
```diff
@@ -0,0 +1,12 @@
...
+    'Open',
+    'Match',
+    'Taken',
+    'Failed'
+);
+
+ALTER TABLE "orders" ADD COLUMN "order_state" "OrderState_Type" NOT NULL DEFAULT 'Open';
+
+UPDATE orders SET order_state = 'Taken' WHERE taken = true;
+UPDATE orders SET order_state = 'Failed' WHERE taken = false;
```
bonomat, line 24: Note: when running this the first time, all open not taken orders will be set to 'failed'. Not an issue because our maker will create new orders immediately, but I wanted to point it out :)

coordinator/src/node/expired_orders.rs
```diff
@@ -0,0 +1,17 @@
...
+use orderbook_commons::OrderType;
+
+pub fn sync_expired_orders(node: Node) -> Result<()> {
+    let mut conn = node.pool.get()?;
+    let expired_orders =
+        orders::get_all_orders(&mut conn, OrderType::Limit, OrderState::Open, true)?;
+
+    for expired_order in expired_orders {
+        orders::set_order_state(&mut conn, expired_order.id, OrderState::Failed)?;
+    }
```
bonomat, line 14: Nit: 

this could be a single db call: 

```
UPDATE orders SET order_state = 'Failed' WHERE taken = false and  expiry < now();;
```

coordinator/src/orderbook/db/orders.rs
```diff
@@ -169,9 +195,30 @@ pub fn set_is_taken(
...
+/// Returns the number of affected rows: 1.
+pub fn set_order_state(
+    conn: &mut PgConnection,
+    id: Uuid,
+    order_state: orderbook_commons::OrderState,
+) -> QueryResult<OrderbookOrder> {
+    let mut is_taken = false;
+    if order_state == orderbook_commons::OrderState::Taken {
+        is_taken = true;
+    }
```
bonomat: Nit:

```suggestion 
let is_taken = order_state == orderbook_commons::OrderState::Taken;
```

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
```diff
@@ -10,3 +10,22 @@ ALTER TABLE "orders" ADD COLUMN "order_state" "OrderState_Type" NOT NULL DEFAULT
 
 UPDATE orders SET order_state = 'Taken' WHERE taken = true;
 UPDATE orders SET order_state = 'Failed' WHERE taken = false;
+
+CREATE TYPE "MatchState_Type" AS ENUM (
```
bonomat, line 29: Changes in this file should be in an separate migration file or all in the same file.
If we end up deploying the commit which creates this file before this commit here, we will not re-run the migration and will get an error.

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
```diff
@@ -10,3 +10,22 @@ ALTER TABLE "orders" ADD COLUMN "order_state" "OrderState_Type" NOT NULL DEFAULT
...
+    'Pending',
+    'Filled',
+    'Failed'
+);
+
+CREATE TABLE "matches" (
+    id UUID PRIMARY KEY,
+    match_state "MatchState_Type" NOT NULL,
+    order_id UUID REFERENCES orders (trader_order_id) NOT NULL,
+    trader_id TEXT NOT NULL,
```
bonomat, line 39: Nit: 

could be a reference to user table

coordinator/src/node.rs
```diff
@@ -140,6 +143,19 @@ impl Node {
         let (fee_payment_hash, invoice) = self.fee_invoice_taker(trade_params).await?;
         let mut connection = self.pool.get()?;
 
+        let order_id = trade_params.filled_with.order_id;
+        let order = orders::get_with_id(&mut connection, order_id)?
+            .with_context(|| "Could not find order with id {order_id}")?;
+
+        ensure!(
+            order.expiry > OffsetDateTime::now_utc(),
+            "Can't execute expired match"
```
bonomat, line 152: Nit: 
```suggestion
            "Can't execute a trade on an expired order"
```

coordinator/src/node.rs
```diff
@@ -140,6 +143,19 @@ impl Node {
...
+        );
+        ensure!(
+            order.order_state == OrderState::Match,
+            "Can't execute trade with in invalid state"
```
bonomat, line 156: Could be helpful to find future bugs if we print in what state the order was instead.

coordinator/src/node.rs
```diff
@@ -151,12 +167,18 @@ impl Node {
...
                         &mut connection,
                         trade_params.filled_with.order_id,
                         OrderState::Failed,
                     )?;
 
+                    matches::set_match_state(
+                        &mut connection,
+                        trade_params.filled_with.order_id,
+                        MatchState::Failed,
+                    )?;
```
bonomat: You might want to handle the error of ` orders::set_order_state(` here instead of bubling it up or update both tables atomically, otherwise it may happen that one query is run and the other one isn't.

coordinator/src/node.rs
```diff
@@ -174,11 +196,45 @@ impl Node {
...
+                        &mut connection,
+                        trade_params.filled_with.order_id,
+                        OrderState::Failed,
+                    )?;
+
+                    matches::set_match_state(
+                        &mut connection,
+                        trade_params.filled_with.order_id,
+                        MatchState::Failed,
+                    )?;
```
bonomat, line 213: Same comment as above. It's my understanding that these db updates should run atomically.

coordinator/src/node.rs
```diff
@@ -174,11 +196,45 @@ impl Node {
...
+            &mut connection,
+            trade_params.filled_with.order_id,
+            MatchState::Filled,
+        )?;
+
+        orders::set_order_state(
+            &mut connection,
+            trade_params.filled_with.order_id,
+            OrderState::Taken,
+        )?;
```
bonomat, line 236: Same again.

coordinator/src/node/expired_positions.rs
```diff
@@ -33,70 +36,88 @@ pub async fn close(node: Node) {
...
-                Err(e) => {
-                    tracing::warn!(
-                        "Failed to get quote from bitmex for {} at {}. Error: {e:?}",
-                        position.trader,
-                        position.expiry_timestamp
-                    );
-                    continue;
-                }
-            };
+        tracing::debug!(trader_pk=%position.trader, %position.expiry_timestamp, "Attempting to close expired position");
```
bonomat, line 82: is `%position.trader` printed nicely?

coordinator/src/node/expired_positions.rs
```diff
@@ -33,70 +36,88 @@ pub async fn close(node: Node) {
...
+            // need separate models for a limit and a market order.
+            price: Decimal::ZERO,
+            quantity: Decimal::try_from(position.quantity).expect("to fit into decimal"),
+            trader_id: position.trader,
+            direction: position.direction.opposite(),
+            order_type: OrderType::Market,
+            // This order can basically not expire, but if the user does not come back online within
+            // 8 weeks I guess we can safely assume that this channel is abandoned and we should
+            // consider force closing.
+            expiry: OffsetDateTime::now_utc().add(Duration::weeks(8)),
```
bonomat, line 98: 8 weeks is probably too long, we should force close sooner because of the CET's expiry but I understand where you want to get with this.

coordinator/src/orderbook/trading.rs
```diff
@@ -133,39 +156,75 @@ impl Trading {
...
-                .with_context(|| format!("Could not match order {}", order.id))
-                .map_err(|e| TradingError::NoMatchFound(format!("{e:#}")))?;
+            let matched_orders = match match_order(&order, opposite_direction_orders) {
+                Ok(Some(matched_orders)) => matched_orders,
+                Ok(None) => {
+                    // todo(holzeis): Currently we still respond to the user immediately if there
+                    // has been a match or not, that's the reason why we also
+                    // have to set the order to failed here. But actually we
+                    // could keep the order until either expired or a
+                    // match has been found and then update the state correspondingly.
```
bonomat, line 202: image

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
```diff
@@ -6,7 +6,10 @@ CREATE TYPE "OrderState_Type" AS ENUM (
     'Failed'
 );
 
-ALTER TABLE "orders" ADD COLUMN "order_state" "OrderState_Type" NOT NULL DEFAULT 'Open';
+ALTER TABLE "orders"
```
bonomat, line 17: Same as in the other commit, please do not update old migration files. This can lead to annoying errors.

crates/orderbook-commons/src/order_matching_fee.rs
```diff
@@ -14,11 +14,15 @@ fn order_matching_fee(quantity: f32, price: Decimal, fee_per_cent: Decimal) -> b
...
-        .to_f64()
-        .expect("fee to fit in f64");
+    let fee: f64 = match price != Decimal::ZERO {
+        true => {
+            let fee = quantity * (Decimal::ONE / price) * fee_per_cent;
+            fee.round_dp_with_strategy(8, RoundingStrategy::MidpointAwayFromZero)
+                .to_f64()
+                .expect("fee to fit in f64")
+        }
+        false => 0.0,
```
bonomat, line 24: MHmm... I don't like this. It should not happen that we get a match with `price=0`.

crates/orderbook-commons/src/order_matching_fee.rs
```diff
@@ -14,11 +14,15 @@ fn order_matching_fee(quantity: f32, price: Decimal, fee_per_cent: Decimal) -> b
     let quantity = Decimal::from_f32(quantity).expect("quantity to fit in Decimal");
     let price = price;
 
-    let fee = quantity * (Decimal::ONE / price) * fee_per_cent;
-    let fee = fee
-        .round_dp_with_strategy(8, RoundingStrategy::MidpointAwayFromZero)
-        .to_f64()
-        .expect("fee to fit in f64");
+    let fee: f64 = match price != Decimal::ZERO {
```
bonomat, line 17: Can you say when this can happen that the price is 0?

coordinator/src/orderbook/db/custom_types.rs
```diff
@@ -130,6 +131,51 @@ impl FromSql<OrderStateType, Pg> for OrderState {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, FromSqlRow, AsExpression)]
+#[diesel(sql_type = OrderReasonType)]
+pub(crate) enum OrderReason {
+    Manual,
+    Expired,
+    Liquidated,
+    TakeProfit,
+    StopLoss,
```
bonomat, line 141: Maybe we should only introduce these variants when we need/use them?

coordinator/src/orderbook/trading.rs
```diff
@@ -119,12 +121,24 @@ impl Trading {
...
                         // this is taking some time.
-                        let result = self.process_new_order(new_order_msg.new_order).await;
+                        let new_order = new_order_msg.new_order;
+                        let result = self
+                            .process_new_order(new_order, new_order_msg.order_reason)
+                            .await;
                         new_order_msg.sender.send(result).await?;
                     }
                     TradingMessage::NewUser(new_user_msg) => {
+                        // todo(holzeis): shouldn't we also remove users from that set if they are
```
bonomat, line 131: Yes.

coordinator/src/orderbook/db/custom_types.rs
```diff
@@ -130,6 +131,51 @@ impl FromSql<OrderStateType, Pg> for OrderState {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, FromSqlRow, AsExpression)]
+#[diesel(sql_type = OrderReasonType)]
+pub(crate) enum OrderReason {
```
bonomat, line 136: `OrderReason` sounds weird to me. Is it more an `OrderType`?

coordinator/src/orderbook/trading.rs
```diff
@@ -230,6 +268,30 @@ impl Trading {
...
+            orders::get_by_trader_id_and_state(&mut conn, trader_id, OrderState::Match)?
+        {
+            tracing::debug!(%trader_id, order_id=%order.id, "Notifying trader about pending match");
+
+            let matches = matches::get_matches_by_order_id(&mut conn, order.id)?;
+            let filled_with = FilledWith::try_from(matches)?;
+
+            let message = match order.order_reason {
+                OrderReason::Manual => OrderbookMsg::Match(filled_with),
+                _ => OrderbookMsg::ForceMatch { order, filled_with },
```
bonomat, line 284: This might mess with us in the future: every non-manual order will be a `ForcedMatch`. Can you say this for sure? If not, maybe better to write all combinations out so that we see if a new variant was added.

coordinator/src/orderbook/db/custom_types.rs
```diff
@@ -130,6 +131,51 @@ impl FromSql<OrderStateType, Pg> for OrderState {
...
+    Manual,
```
bonomat, line 137: What's an `manual` order? 

Maybe add a description to what each variant means please?

mobile/lib/features/trade/force_order_change_notifier.dart
```diff
@@ -0,0 +1,81 @@
...
+                content = const Text("Your position has been closed due to expiry.");
+              case OrderReason.liquidated:
+                content = const Text("Your position has been closed due to liquidation.");
+              case OrderReason.takeProfit:
+                content = const Text(
+                    "Your position has been closed as the take profit order found a match.");
+              case OrderReason.stopLoss:
+                content = const Text(
+                    "Your position has been closed as the stop loss order found a match.");
+              // a manual order should not appear as a force trade - that should not happen.
```
bonomat, line 62: If this is the case, maybe add a log message if it happens?

mobile/native/src/db/models.rs
```diff
@@ -221,6 +222,19 @@ impl Order {
             .load(conn)
     }
 
+    /// Gets any force order in the database. A force order is defined by any order with an reason
+    /// not equal to `Manual`
```
bonomat, line 226: I'm not buying this concept of a _forced_ order yet... The naming does not align in my brain 🙈

bonomat (commented):
Overall good work. 
It's not a good idea to change migration files, this makes commit not atomic and we could run into weird problems. 

I don't like the concept of a _forced_ order yet. You seem to define every order as _forced_ which was not _manual_. Maybe we can describe the concept differently? 

Otherwise, I think this looks good.

coordinator/src/orderbook/db/custom_types.rs
holzeis: We already have an `OrderType` 🙈

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
holzeis: I'll exclude limit orders

coordinator/src/node/expired_positions.rs
holzeis: > The % sigil operates similarly, but indicates that the value should be recorded using its [fmt::Display](https://doc.rust-lang.org/nightly/core/fmt/trait.Display.html) implementation:

https://docs.rs/tracing/latest/tracing/

coordinator/src/node/expired_positions.rs
holzeis: what would be a better value?

coordinator/src/node/expired_positions.rs
bonomat: That's not helpful 😅 will it be printed as `8b143` or as `[08,11,20,03,...` ? The latter one is not really helpful and we have heaps of them already, hence I'm asking.

coordinator/src/node/expired_positions.rs
bonomat: I'm not sure what our refund timeout is atm, but probably something like `refund_timeout-x` where x is a small number.

mobile/native/src/db/models.rs
holzeis: renamed to async order

holzeis:
> It's not a good idea to change migration files, this makes commit not atomic and we could run into weird problems. 

I will check how hard it is to adapt the commits to have a dedicated migration script per commit.

> I don't like the concept of a _forced_ order yet. You seem to define every order as _forced_ which was not _manual_. Maybe we can describe the concept differently? 

As discussed offline I renamed it to _async_ order.

crates/orderbook-commons/src/order_matching_fee.rs
holzeis: It definitely happened to me while developing, but I can drop that commit if you like.

crates/orderbook-commons/src/order_matching_fee.rs
holzeis: I think it happens when we only have a ask or bid price.

coordinator/src/orderbook/trading.rs
holzeis: I made it explicit, so we do not run into pitfalls when adding a new variant.

coordinator/src/node/expired_positions.rs
holzeis: @luckysori maybe you can help us out here? what is the refund timeout atm?

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
holzeis: I've split up the migration scripts into dedicated folders per commit.

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
holzeis: I've split up the migration scripts into dedicated folders per commit.

coordinator/src/orderbook/db/custom_types.rs
holzeis: How about SubOrderType?

crates/orderbook-commons/src/lib.rs
holzeis: Unfortunately, that's not easily done. I need the relationship of the quantity to execution price. If I would split them up into two vecs I'd have to guess on the index that they relate, which is also not very nice.

Maybe improving the name of the struct helps? any suggestions?

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
holzeis: I though about it, but it we unfortunately do not register the maker as user, so we can't reference here the user table as it would otherwise fail to create the matches for the maker.

coordinator/src/orderbook/trading.rs
holzeis: I've created a ticket for that todo

https://github.com/get10101/10101/issues/1268

coordinator/migrations/2023-09-04-090152_add_order_state_and_matches/up.sql
holzeis: I created a ticket for it

https://github.com/get10101/10101/issues/1269

bonomat (approved):
Looks good, but please wait for a second pair of eyes

luckysori (commented):
Thanks for working on this long and complicated task!

Whilst I do like the concept of letting the coordinator post an order on behalf of the trader, who would most likely choose this over settling their position on-chain, the implementation is quite complicated and I am not convinced we are modelling things correctly with the introduction of `Matches` and `OrderReason`.

At this stage I don't really want to block the merge since other teammates have already approved this. My main worry is that the orderbook is becoming harder to understand and maintain, and I think it was already in a bad spot before this PR which adds more features.

coordinator/src/orderbook/trading.rs
```diff
@@ -1,24 +1,177 @@
...
+    /// Processes a new limit and market order
+    ///
+    /// Limit order: update price feed
+    /// Market order: find match and notify traders
+    async fn process_new_order(&self, new_order: NewOrder) -> Result<Order> {
+        tracing::info!(trader_id=%new_order.trader_id, "Received a new {:?} order", new_order.order_type);
+
+        if new_order.order_type == OrderType::Limit && new_order.price == Decimal::ZERO {
+            return Err(TradingError::InvalidOrder(
+                "Limit order with zero price are not allowed".to_string(),
```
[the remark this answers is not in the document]
holzeis: Market orders never have a price.

coordinator/src/node/expired_positions.rs
```diff
@@ -33,70 +36,83 @@ pub async fn close(node: Node) {
...
+            let order_id = order.id.to_string();
 
-        if !node.is_connected(&position.trader) {
-            tracing::debug!(
-                "Could not close expired position with {} as trader is not connected.",
-                position.trader
-            );
-            continue;
-        }
+            if order.expiry < OffsetDateTime::now_utc() {
```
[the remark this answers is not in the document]
holzeis: How about adding the expiry to the matches table?

crates/orderbook-commons/src/lib.rs
```diff
@@ -278,6 +282,25 @@ impl From<RoutingFees> for lightning::routing::gossip::RoutingFees {
     }
 }
 
+pub enum MatchState {
+    Pending,
+    Filled,
+    Failed,
+}
+
+pub struct Matches {
```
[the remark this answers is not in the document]
holzeis: That's the reason why I called it Matches! The issue is that those models should actually aligned. Unfortunately I couldn't map the model used as is to the database, that's the reason for the weird `matches` model.

coordinator/src/orderbook/trading.rs
```diff
@@ -133,39 +156,75 @@ impl Trading {
...
-
-            for order_id in orders_to_set_taken {
-                tracing::info!(%order_id, "Setting order to taken.");
-                if let Err(err) = orderbook::db::orders::set_is_taken(&mut conn, order_id, true) {
-                    let order_id = order_id.to_string();
-                    tracing::error!(order_id, "Could not set order to taken. Error: {err:#}");
-                }
+            let match_params = matched_orders.matches();
+            for match_param in match_params {
+                matches::insert(&mut conn, match_param)?;
```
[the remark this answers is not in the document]
holzeis: Because I want to track if a match got executed on per trader. The model would have gotten to complex otherwise.

Also considering that we could have multiple matches per order. By having a single match record we can easily reflect 1:n and m:n relations.

coordinator/migrations/2023-09-08-090152_add_order_state/up.sql
```diff
@@ -0,0 +1,12 @@
...
+CREATE TYPE "OrderState_Type" AS ENUM (
+    'Open',
+    'Match',
+    'Taken',
+    'Failed'
+);
+
+ALTER TABLE "orders" ADD COLUMN "order_state" "OrderState_Type" NOT NULL DEFAULT 'Open';
+
+UPDATE orders SET order_state = 'Taken' WHERE taken = true;
```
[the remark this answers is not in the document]
holzeis: already did in a later commit.

coordinator/src/node.rs
```diff
@@ -173,11 +196,35 @@ impl Node {
...
+                if let Err(e) = self
+                    .close_position(&position, closing_price, channel_id, fee_payment_hash)
+                    .await
+                {
+                    update_order_and_match(
+                        &mut connection,
+                        trade_params.filled_with.order_id,
+                        MatchState::Failed,
+                        OrderState::Failed,
+                    )?;
```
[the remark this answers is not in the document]
holzeis: I do agree that we should use transactions more consistently. I am opting to put that out of scope for that PR as the change is already quite big as is.

coordinator/src/node/expired_positions.rs
```diff
@@ -1,29 +1,32 @@
...
-pub async fn close(node: Node) {
-    let mut conn = match node.pool.get() {
-        Ok(conn) => conn,
-        Err(e) => {
-            tracing::error!("Failed to get pool connection. Error: {e:?}");
-            return;
-        }
-    };
+pub async fn close(node: Node, trading_sender: mpsc::Sender<TradingMessage>) -> Result<()> {
+    let mut conn = node.pool.get().map_err(|e| anyhow!("{e:#}"))?;
```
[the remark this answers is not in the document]
holzeis: Indeed its not needed :)

coordinator/src/bin/coordinator.rs
```diff
@@ -189,6 +191,9 @@ async fn main() -> Result<()> {
         }
     });
 
+    let (tx_price_feed, _rx) = broadcast::channel(100);
+    let (_handle, trading_sender) = trading::Trading::start(pool.clone(), tx_price_feed.clone());
```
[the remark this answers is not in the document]
holzeis: don't you have to use it as it will otherwise get canceled, because it lost its scope?

coordinator/src/bin/coordinator.rs
```diff
@@ -189,6 +191,9 @@ async fn main() -> Result<()> {
         }
     });
 
+    let (tx_price_feed, _rx) = broadcast::channel(100);
```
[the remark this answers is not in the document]
holzeis: Maybe I am mistaken, but the `tx_price_feed` is received in the `websocket.rs` and is useful because with it I can broadcast the updated price to all connected users instead of having to send it to everyone individually.

```
// We subscribe *before* sending the "joined" message, so that we will also
// display it to our client.
let mut rx = state.tx_price_feed.subscribe();
```

cc: @bonomat

coordinator/src/orderbook/trading.rs
```diff
@@ -1,24 +1,177 @@
...
+
+            Ok(())
+        })
+    }
+
+    /// Processes a new limit and market order
+    ///
+    /// Limit order: update price feed
+    /// Market order: find match and notify traders
+    async fn process_new_order(&self, new_order: NewOrder) -> Result<Order> {
```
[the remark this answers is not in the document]
holzeis: Agreed, I am opting to not put that into scope of that PR.

coordinator/src/orderbook/trading.rs
```diff
@@ -1,24 +1,177 @@
...
+
+        let mut conn = self.pool.get()?;
+        let order = orderbook::db::orders::insert(&mut conn, new_order.clone())
+            .map_err(|e| anyhow!("Failed to insert new order into db: {e:#}"))?;
+
+        if new_order.order_type == OrderType::Limit {
+            // we only tell everyone about new limit orders
+            self.tx_price_feed
+                .send(OrderbookMsg::NewOrder(order.clone()))
+                .map_err(|error| anyhow!("Could not update price feed due to '{error}'"))?;
```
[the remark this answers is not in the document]
holzeis: I think we still need it.. see comment on other discussion.

coordinator/src/orderbook/db/matches.rs
```diff
@@ -0,0 +1,171 @@
...
+use rust_decimal::prelude::FromPrimitive;
+use rust_decimal::prelude::ToPrimitive;
+use rust_decimal::Decimal;
+use std::str::FromStr;
+use time::OffsetDateTime;
+use uuid::Uuid;
+
+#[derive(Insertable, QueryableByName, Queryable, Debug, Clone, PartialEq)]
+#[diesel(table_name = matches)]
+struct Matches {
```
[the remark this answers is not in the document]
holzeis: I would have loved to name it `Match` unfortunately there is already a model named `Match`. We should have probably completely refactored the entire model, but instead I opted to simply go with a weird model name for now.

The refactor could come afterwards at any time I guess.

coordinator/migrations/2023-09-08-090152_add_order_state/up.sql
```diff
@@ -0,0 +1,12 @@
...
+UPDATE orders SET order_state = 'Failed' WHERE taken = false and order_type = 'market';
```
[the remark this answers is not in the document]
holzeis: Yes, that's the assumption! I think it's fair to do so as we are currently matching any market order immediately or set it to `Failed`. if there is a market order open which is taken false, means there must have been an error or the market order was submitted just about the time when we restarted and the order was not filled yet. In any case that would also mean that the order has `Failed` 🙃

crates/orderbook-commons/src/lib.rs
```diff
@@ -131,6 +131,18 @@ pub struct Match {
     pub execution_price: Decimal,
 }
 
+impl From<Matches> for Match {
```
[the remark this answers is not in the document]
holzeis: Unfortunately we have already a `Match` model and I did not want to refactor this model entirely. So I opted for the lesser effort, but definitely a bit uglier solution to name the model `Matches`. I think we can refactor that at a later point and then rename `Matches` to `Match`.

coordinator/src/bin/coordinator.rs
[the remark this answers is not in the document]
luckysori: That's the `RemoteHandle`.

PR #1248, opened 2023-09-07
chore: mainnet candle stick chart

bonomat:
Mainnet looks better than the testnet version which comes handy when doing demo videos or screenshots.

I keep changing this manually when I need quickly some screenshots so I thought I'd just change this permanently. 
Downside: the price will now be different to what the chart shows.

### DIFF
--- a/mobile/lib/features/trade/application/candlestick_service.dart
+++ b/mobile/lib/features/trade/application/candlestick_service.dart
@@ -6,14 +6,9 @@ import 'dart:convert';
 class CandlestickService {
   const CandlestickService();
 
-  Future<List<Candle>> fetchCandles(String network, int amount) async {
-    var url = "testnet.bitmex.com";
-    if (network == "mainnet") {
-      url = "www.bitmex.com";
-    }
-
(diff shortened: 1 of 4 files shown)


holzeis:
> Downside: the price will now be different to what the chart shows.

Why not also create the orders from the mainnet bitmex stream?

bonomat:
> > Downside: the price will now be different to what the chart shows.
> 
> Why not also create the orders from the mainnet bitmex stream?

it wil eventually mess with our maker and hedging strategy.

Restioson (approved):

holzeis:
> it wil eventually mess with our maker and hedging strategy.

That's confusing to me, can you elaborate on that. How will it mess with the hedging strategy?

bonomat:
> > it wil eventually mess with our maker and hedging strategy.
> 
> That's confusing to me, can you elaborate on that. How will it mess with the hedging strategy?

If we want to track that or testnet maker makes money we need to take the prices what are currently being traded on testnet.bitmex.com and not the prices from mainnet bitmex.com

holzeis (approved):
Lets give it a go!

PR #1261, opened 2023-09-08
fix: GoError: There is nothing to pop when paying with faucet

### LINKED ISSUE #1253: GoError: There is nothing to pop
## What did I do?
Hit done after hitting the "Pay the invoice with 10101 faucet"

IMG_619CBAE3AA7D-1

Restioson:
Fixes #1253. This was happening because the pay with faucet button also tries to pop twice, as does done. But, this is done in the background, so they can overlap and do the same thing twice. The fix is just to disable the done button while the invoice is being paid by the faucet.


Restioson:
Huh, why did the justfile change

Restioson:
What the hell?

PR #1293, opened 2023-09-13
Wait for 3 confirmations to show on-chain payments as successful

Restioson:
I'm not sure if this is considered as enough. Thoughts?

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -66,6 +66,7 @@ mod node;
 
 pub mod channel_status;
 
+use crate::api::Status;
 pub use channel_status::ChannelStatus;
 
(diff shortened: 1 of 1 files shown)


holzeis (approved):
I think its good enough for now!

We could potentially also let the user configure that for themselves at some point.

klochowicz (approved):

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -315,14 +315,19 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
...
                 (
-                    api::Status::Pending,
                     // Unconfirmed transactions should appear towards the top of the history
                     OffsetDateTime::now_utc().unix_timestamp() as u64,
                     0,
                 )
             }
         };
 
+        let status = if n_confirmations >= 3 {
```
bonomat, line 325: Imho 1 conf would have been enough :)

Issue #1258, opened 2023-09-08
Sync should not block the whole app


holzeis:
Currently we show the app as syncing after the startup and thus hiding the transaction history and also rendering the usable balance as 0 (preventing the user from trading).

The only thing that we have to check upon startup is if the channel is still open (meaning that the funding transaction hasn't been spend) if so the user should be free to use the app. We don't have to wait for the whole on-chain sync to finish.

bonomat:
Not 100% if this is the case, but we can syn the on-chain wallet separately from the off-chain wallet and display it respectively. We should be able to continue once the off-chain wallet is synced which is usually just a couple of requests.

bonomat:
Assumption is this is an easy win: timbox to 2h or move into future iteration.

PR #1294, opened 2023-09-13
Refresh on-chain txs in background

### LINKED ISSUE #1288: Live refresh on-chain transactions
Currently, the on-chain transactions are only refreshed at app startup. They should be refreshed continuously as with lightning payments.

Restioson:
Fixes #1288

### DIFF
--- a/mobile/lib/features/wallet/wallet_change_notifier.dart
+++ b/mobile/lib/features/wallet/wallet_change_notifier.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart' hide Flow;
 import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
@@ -9,6 +11,8 @@ import 'domain/wallet_info.dart';
 
 class WalletChangeNotifier extends ChangeNotifier implements Subscriber {
   final WalletService service;
+  late final Timer timer;
+
   WalletInfo walletInfo = WalletInfo(
     balances: WalletBalances(onChain: Amount(0), lightning: Amount(0)),
     history: List.empty(),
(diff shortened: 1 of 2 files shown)


holzeis (changes requested):
Lets spawn a task on the rust side instead of blocking a frb thread (we have only 4) every 30 seconds for about 5-7 seconds.

holzeis:
Note, if we'd also remove the initial sync on startup this should also fix https://github.com/get10101/10101/issues/1258

mobile/lib/features/wallet/wallet_change_notifier.dart
```diff
@@ -29,6 +33,14 @@ class WalletChangeNotifier extends ChangeNotifier implements Subscriber {
     super.notifyListeners();
   }
 
+  Future<void> initialize() async {
+    await refreshWalletInfo();
+
+    timer = Timer.periodic(const Duration(seconds: 30), (Timer t) async {
+      await refreshWalletInfo();
```
bonomat, line 40: I'm not sure if it does, but if `refreshWalletInfo` triggers an on-chain sync, than this would be too often.

Restioson:
> Lets spawn a task on the rust side instead of blocking a frb thread (we have only 4) every 30 seconds for about 5-7 seconds.

Ah, I didn't realised it blocked a FRB thread, can fix that too

luckysori (commented):
I think this is unnecessary, as we already have a periodic task managed by the `ln_dlc_node::Node`. See https://github.com/get10101/10101/issues/1288#issuecomment-1725887321.

holzeis:
Can we close that after also #1332 has been merged?

Restioson:
Indeed it can be closed as @luckysori mentioned this isn't needed

PR #1295, opened 2023-09-13
[FEEDBACK NEEDED] Expand the wallet balance rows properly

Restioson:
Since updating my flutter, I've been experiencing #1129. This PR restores the expected behaviour, but I don't know how it interacts with older flutter versions. Could someone with an older version (i.e, current behaviour of main is as expected) of flutter please test and make sure this doesn't break anything?

### DIFF
--- a/mobile/lib/features/wallet/wallet_screen.dart
+++ b/mobile/lib/features/wallet/wallet_screen.dart
@@ -183,7 +183,7 @@ class _WalletScreenState extends State<WalletScreen> {
                     isExpanded: _isBalanceBreakdownOpen,
                   )
                 ],
-                expansionCallback: (i, isOpen) => setState(() => _isBalanceBreakdownOpen = !isOpen),
+                expansionCallback: (i, isOpen) => setState(() => _isBalanceBreakdownOpen = isOpen),
               ),
               const SizedBox(height: 10.0),
               if (walletChangeNotifier.lightning().sats == 0)


klochowicz:
hey, if it stops you from contributing, it might be worthwhile to try `fvm` with a locked flutter version? we would all benefit from that!

klochowicz:
I haven't tried yet; what I actually noticed is that some other earlier commit of yours bumped the packages and my dart/flutter env tries hard to downgrade them every time 😅 so perhaps `fvm` would be a good idea to make sure you're on the same version as everyone else!

Restioson:
> I haven't tried yet; what I actually noticed is that some other earlier commit of yours bumped the packages and my dart/flutter env tries hard to downgrade them every time 😅 so perhaps fvm would be a good idea to make sure you're on the same version as everyone else!

Oh, I see - I'll do that in another PR then. Sorry, I hadn't realised that :sweat_smile: I just listened to the cheerful "an update is available!" message

klochowicz:
> > I haven't tried yet; what I actually noticed is that some other earlier commit of yours bumped the packages and my dart/flutter env tries hard to downgrade them every time 😅 so perhaps fvm would be a good idea to make sure you're on the same version as everyone else!
> 
> Oh, I see - I'll do that in another PR then. Sorry, I hadn't realised that 😅 I just listened to the cheerful "an update is available!" message

all good, I am still puzzled why flutter doesn't have that by default 🤦🏻‍♂️ it's so much easier to deal with it in for rust with `rust-toolchain.toml`

bonomat:
Unfortunately on Flutter `Flutter 3.10.6 • channel stable` it breaks the expected behavior.

klochowicz:
Closing in favour of #1333

Issue #1211, opened 2023-09-02
Send a push notification to the app when LSP wants to collab close the expired position


klochowicz:

klochowicz:
blocked on #1227

holzeis:
When do we want to collab close a channel?

PR #1297, opened 2023-09-14
feat: Send push notifications for expired and expiring positions

klochowicz:
- Send notification to the user if their position is about to expire.
- Add unit tests for the logic determining which notification to send.
- periodically load all relevant positions with their fcm_tokens from the DB and
- attempt to deliver notifications


coordinator/src/bin/coordinator.rs
```diff
@@ -228,7 +229,24 @@ async fn main() -> Result<()> {
...
+                if let Ok(mut conn) = pool.get() {
+                    if let Err(e) = query_and_send_position_notifications(&mut conn, &sender).await
+                    {
+                        tracing::error!("Failed to send notifications: {e:#}");
+                    }
+                } else {
+                    tracing::error!("Failed to get connection from pool");
+                }
+
+                tokio::time::sleep(Duration::from_secs(30 * 60)).await;
```
holzeis, line 246: 🔧 Please put that into a constant

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+
+/// Load all recent positions with the DB with associated fcm tokens and send push
+/// notification about expiring/expired positions if needed.
+pub async fn query_and_send_position_notifications(
+    conn: &mut PgConnection,
+    notification_sender: &mpsc::Sender<Notification>,
+) -> Result<()> {
+    let users = user::all(conn)?;
+
+    let positions_with_fcm_tokens = positions::Position::get_all_positions_with_expiry_after(
```
holzeis, line 321: 🔧 I this could be nicely done with an inner join!

At least we should only load users which have an expired position.

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+            notification.notification_kind,
+            NotificationKind::PositionExpired
+        );
+        assert_eq!(notification.user_fcm_token, position_just_expired.1);
+    }
+}
+
+/// Load all recent positions with the DB with associated fcm tokens and send push
+/// notification about expiring/expired positions if needed.
+pub async fn query_and_send_position_notifications(
```
holzeis, line 315: 🔧 I don't see the notification service doing anything else than to send notifications. This send_expiry_notifications_if_applicable should probably be its own component or maybe added to the expired_positions component?

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
     tracing::debug!("Sent: {:?}", response);
     Ok(())
 }
+
+/// Send notifications to users with positions that are about to expire or have
+/// just expired
+pub async fn send_expiry_notifications_if_applicable(
```
holzeis, line 150: nit: reading the code from top to down I would have expected that function to come after `query_and_send_position_notifications` and certainly the tests at the very end of the file. 🙃

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+    positions: &[(Position, FcmToken)],
+    notification_sender: &mpsc::Sender<Notification>,
+) {
+    let now = OffsetDateTime::now_utc();
+
+    for (position, fcm_token) in positions {
+        if position.expiry_timestamp <= now
+            && now < position.expiry_timestamp + time::Duration::hours(1)
+        {
+            if let Err(e) = notification_sender
```
holzeis, line 160: 🔧 With #1246 we are creating an `AsyncOrder` for expired positions and try to send them when processing the new order. Currently we only try through the websocket, but if the user is not online we should send the push notification. 

Please add this push notification there as well, once #1246 is merged.

https://github.com/get10101/10101/pull/1246/files#diff-069f682364e37633796656003287aac144d258002b4f2f2d7933b6a94dcc5010R249

holzeis (changes requested):

holzeis (dismissed):
~📣 With that logic we are sending the user every 30 minutes a push notification. That's too much!~

I propose the following:

### If the position is about to expire and we are in the rollover weekend
(see also https://github.com/get10101/10101/pull/1287/files#diff-f36ed55db10ebb7b64adaf1ba2895679c0254ebd0ce016492f283d4001ac4f84R373-R383) 

- Friday 3pm UTC: Send a notification that the rollover weekend has started and the user should come online to rollover to the next weekend (Sunday 3pm UTC)
- Saturday 3pm UTC: Reminder that the position is about to expire if the user is not coming online to rollover.
- Sunday 12am UTC: Reminder that the position will expire in 3 hours unless the user is coming online to rollover.

### If the position has expired and was not online to close the position.

- At trying to notify about the `AsyncOrder` and the user is not online: Send notification that the position has been expired and the user should come online to close the position.
- For the next week everyday at 3pm UTC: Send a notification that the position has expired and the user should come online or we have to force close the dlc_channel.

### If the position is still expired after a week

Send last notification to the user that we will force close the dlc_channel in due time

klochowicz:
> 📣 With that logic we are sending the user every 30 minutes a push notification. That's too much!

have you read the code and seen the unit tests? this is certainly **not** the case!

also, if you had something specific in mind, it would be helpful to put it *on the ticket*.

holzeis:
> > 📣 With that logic we are sending the user every 30 minutes a push notification. That's too much!
> 
> have you read the code and seen the unit tests? this is certainly **not** the case!

Fair enough I miss read the condition, but wouldn't it still send 2 notifications as you are checking for 1 hour after but run it ever 30 minutes? Anyways it's surely not as critical as sending a notification non-stop.

> If you had something specific in mind, it would be helpful to put it *on the ticket*, to avoid impedance mismatch at this stage.

I hadn't something specific in mind it was just a proposal after looking over your PR. However we should align it with the rest of the code. We don't need another task that regularly checks if a position has expired we already have that. See `expired_positions`.


> I would actually probably prefer moving with this forward *and* adding the additional ideas that you've described. They're not either-or, and they can be done in a follow-up. 
> 
> 
> 
> I've convert the above into a ticket so we can discuss it freely there.

I am not blocking that PR as the the amount of push notifications are less than assumed when asking for changes, however I don't see a rush in bringing this out in anyway possible, we should take the time to get this to a point where we are all happy with it.

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+            .iter()
+            .find(|u| u.pubkey == p.trader.to_string())
+            .map(|u| FcmToken(u.fcm_token.clone()));
+
+        if let Some(fcm_token) = maybe_fcm_token {
+            Some((p, fcm_token))
+        } else {
+            tracing::warn!(?p, "No FCM token for position");
+            None
+        }
```
luckysori, line 341: :wrench: This can just be a `map` (or a `find_map` altogether).

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
     tracing::debug!("Sent: {:?}", response);
     Ok(())
 }
+
+/// Send notifications to users with positions that are about to expire or have
+/// just expired
+pub async fn send_expiry_notifications_if_applicable(
+    positions: &[(Position, FcmToken)],
+    notification_sender: &mpsc::Sender<Notification>,
```
luckysori, line 152: :upside_down_face: It would be nicer to test this logic without the `Sender`. You could define a pure function like

```rust
fn expiry_state(position: &Position) -> ExpiryState {
    // ...
}

enum ExpiryState {
    NotExpired,
    SoonToExpire,
    Expired,
}
```
and just test that one.

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use crate::logger::init_tracing_for_test;
+    use time::Duration;
+
+    fn soon_expiring_position() -> (Position, FcmToken) {
+        let mut position = Position::dummy();
+        position.creation_timestamp = OffsetDateTime::now_utc() - Duration::days(1);
```
luckysori, line 193: :thinking: Does the `creation_timestamp` matter at all?

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+/// notification about expiring/expired positions if needed.
+pub async fn query_and_send_position_notifications(
+    conn: &mut PgConnection,
+    notification_sender: &mpsc::Sender<Notification>,
+) -> Result<()> {
+    let users = user::all(conn)?;
+
+    let positions_with_fcm_tokens = positions::Position::get_all_positions_with_expiry_after(
+        conn,
+        OffsetDateTime::now_utc() - time::Duration::hours(1),
```
luckysori, line 323: :wrench: We should document this logic somehow (variable, constant, comments). I think we are choosing to load positions in the range `[expired an hour ago, not yet expired]`, but this is not obvious looking at the code.

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+/// just expired
+pub async fn send_expiry_notifications_if_applicable(
+    positions: &[(Position, FcmToken)],
+    notification_sender: &mpsc::Sender<Notification>,
+) {
+    let now = OffsetDateTime::now_utc();
+
+    for (position, fcm_token) in positions {
+        if position.expiry_timestamp <= now
+            && now < position.expiry_timestamp + time::Duration::hours(1)
```
luckysori, line 158: :question: Can we use a `Range` to define this predicate? It should be more readable if possible.

coordinator/src/notification_service.rs
klochowicz: but it's just as easy with the sender? 🤔

coordinator/src/notification_service.rs
klochowicz: the reason I unfolded is that I wanted to print out the warning instead of just ignoring the ones that didn't have a FCM token.

if you have a solution that does it even neater with that requirement, please share, as I always am keen on improving my iterator-fu 🙏☺️

coordinator/src/notification_service.rs
klochowicz: cool idea! I'll look into it!

coordinator/src/notification_service.rs
klochowicz: you're totally right. it was chosen somewhat arbitrarily, and I admit that it doesn't stand out enough. I'll improve it!

coordinator/src/notification_service.rs
klochowicz: not really, I could remove them. I only set them as it's nicer to have more realistic data :) would you prefer if I removed this part?

coordinator/src/notification_service.rs
klochowicz: I mentioned that in the comment, but I don't think the optimisation is worthwhile at this stage?
🤔 especially if we're doing it very infrequently? (potentially, if we rework according to your other suggestions, we would only do this during the rollover weekend)

coordinator/src/notification_service.rs
klochowicz: I've renamed the module to `notifications.rs` to avoid confusion. I don't think that a separate module for other free functions would add anything in this case.

it can house the struct and free functions, they're related.

coordinator/src/notification_service.rs
klochowicz: moved `query_and_send_position_notifications` above this function.

coordinator/src/notification_service.rs
klochowicz: this will be covered by #1211

klochowicz:
> > > 📣 With that logic we are sending the user every 30 minutes a push notification. That's too much!
> > 
> > 
> > have you read the code and seen the unit tests? this is certainly **not** the case!
> 
> Fair enough I miss read the condition, but wouldn't it still send 2 notifications as you are checking for 1 hour after but run it ever 30 minutes? Anyways it's surely not as critical as sending a notification non-stop.

Yes, you're correct -  it would do it twice. I wanted to avoid the situation when e.g. people get it 1 min before the expiry, and can't react.

> 
> > If you had something specific in mind, it would be helpful to put it _on the ticket_, to avoid impedance mismatch at this stage.
> 
> I hadn't something specific in mind it was just a proposal after looking over your PR. However we should align it with the rest of the code. We don't need another task that regularly checks if a position has expired we already have that. See `expired_positions`.

This one only checks within particular time window, and I prefer that for separating concerns and concentrating logic of when to send notifications in one place. An infrequent read from a DB is not that costly, readability and testability is preferred IMHO.

> 
> > I would actually probably prefer moving with this forward _and_ adding the additional ideas that you've described. They're not either-or, and they can be done in a follow-up.
> > I've convert the above into a ticket so we can discuss it freely there.
> 
> I am not blocking that PR as the the amount of push notifications are less than assumed when asking for changes, however I don't see a rush in bringing this out in anyway possible, we should take the time to get this to a point where we are all happy with it.

isn't this the default situation with every review we have that doesn't solve a critical bug in production? 🤓

FYI: The only drawback is that #1211, which I'm currently working on, is depending on it.

coordinator/src/notification_service.rs
klochowicz: there's no ranges support for `OffsetDateTime` unfortunately :(

coordinator/src/notification_service.rs
klochowicz: Added it to the same PR, as it turned out to be a small task.

coordinator/src/db/positions.rs
```diff
@@ -67,6 +67,24 @@ impl Position {
...
     }
 
+    pub fn get_all_positions_with_expiry_within(
+        conn: &mut PgConnection,
+        start: OffsetDateTime,
+        end: OffsetDateTime,
+    ) -> QueryResult<Vec<crate::position::models::Position>> {
+        let positions = positions::table
+            .filter(positions::expiry_timestamp.gt(start))
+            .filter(positions::expiry_timestamp.lt(end))
```
bonomat, line 77: Just to confirm, you are using here `>` and not `>=`, so explicitly exclusive the boundaries.

coordinator/src/notifications.rs
```diff
@@ -0,0 +1,356 @@
...
+pub struct NotificationService {
+    notification_sender: mpsc::Sender<Notification>,
+}
+
+impl NotificationService {
+    /// Start the notification service
+    ///
+    /// If an empty string is passed in the constructor, the service will not send any notification.
+    /// It will only log the notification that it would have sent.
+    pub fn new(fcm_api_key: String) -> Self {
```
bonomat, line 62: I know it wasn't introduced here, but wouldn't it be nicer to use an option? 
The advantage of using an option is that we move the check into the type system that the value is set while if it is `empty` you as the dev have to check it.

coordinator/src/notifications.rs
```diff
@@ -0,0 +1,356 @@
...
+    anyhow::ensure!(!api_key.is_empty(), "FCM API key is empty");
+
+    let mut message_builder = fcm::MessageBuilder::new(api_key, &fcm_token.0);
+    message_builder.notification(notification);
+    let message = message_builder.finalize();
+    let response = client
+        .send(message)
+        .await
+        .context("could not send FCM notification")?;
+    tracing::debug!("Sent: {:?}", response);
```
bonomat, line 161: Nit: log message says `sent` but you are printing the `response`.

coordinator/src/notifications.rs
```diff
@@ -0,0 +1,356 @@
...
+    .collect::<Vec<_>>();
+
+    send_expiry_notifications_if_applicable(&positions_with_fcm_tokens, notification_sender).await;
+
+    Ok(())
+}
+
+/// Send notifications to users with positions that are about to expire or have
+/// just expired
+pub async fn send_expiry_notifications_if_applicable(
```
bonomat, line 205: Can this be private? It's only being called from the function above.

bonomat (approved):
Looks good.

coordinator/src/notification_service.rs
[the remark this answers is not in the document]
klochowicz: yes, I tried to limit the amount of notifications to not overwhelm the user (cutoff time arbitrarily chosen, it's now with the constant).
the notification itself does not expire on the phone, unless user willingly dismisses it.

coordinator/src/notification_service.rs
```diff
@@ -130,3 +144,205 @@ async fn send_notification<'a>(
...
+                .send(Notification::new(
+                    fcm_token.clone(),
+                    NotificationKind::PositionExpired,
+                ))
+                .await
+            {
+                tracing::error!("Failed to send PositionExpired notification: {:?}", e);
+            }
+        } else if position.expiry_timestamp > now
+            && position.expiry_timestamp <= now + time::Duration::hours(1)
```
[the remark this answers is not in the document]
klochowicz: yes, that's the limitation currently (perhaps a 1h window of delivery e.g. ~12h before it expires would be better!), ie. at the start at the "rollover window".

coordinator/src/notifications.rs
klochowicz: this one mentioned here  is only the API key (the one that it is likely an option, e.g. in local testing). 

In this PR I've introduced `FcmToken` type that checks during initialisation if it's not an empty string, so that we can take advantage of the type system (these ones should be populated, unless a person has disabled push notifications - haven't tested that theory, but I assume you would not get a firebase token if you said 'no' to notification permission).

coordinator/src/notifications.rs
klochowicz: good spotting!

klochowicz:
@holzeis I've rebased on your recent work, trying to improve the naming somewhat, so I'd like to hear your opinion on this. I think my naming doesn't cut it perhaps either - so I was thinking maybe `feed` or `user_feed` would be best? these are messages that people get when they're connected, in contrast to the notifications that arrive at any time, even if they've have their position closed.
thoughts?

Also, I've moved the window to [13h,12h] before expiry, and adjusted polling interval to be just under 1h (to minimise the doubling of the notifications).

coordinator/src/message.rs
```diff
@@ -3,19 +3,19 @@ use bitcoin::secp256k1::PublicKey;
...
-/// This value is arbitrarily set to 100 and defines the message accepted in the notification
+/// This value is arbitrarily set to 100 and defines the message accepted in the message
 /// channel buffer.
 const NOTIFICATION_BUFFER_SIZE: usize = 100;
 
-// TODO(holzeis): This enum should be extended to allow for sending push notifications.
-pub enum Notification {
-    Message {
+/// Message sent to users via the websocket.
+pub enum CoordinatorMessage {
```
holzeis, line 17: 🔧 `CoordinatorMessage` is not correct here as this also includes messages from the orderbook. Actually the majority of messages are coming from the orderbook.

coordinator/src/notifications.rs
```diff
@@ -0,0 +1,355 @@
...
+    pub fn new(fcm_api_key: String) -> Self {
+        if fcm_api_key.is_empty() {
+            // Log it as error, as in production it should always be set
+            tracing::error!("FCM API key is empty. No notifications will not be sent.");
+        }
+
+        let (notification_sender, mut notification_receiver) = mpsc::channel(100);
+
+        // TODO: use RAII here
+        tokio::spawn({
```
holzeis, line 74: 🔧 I don't think we need that task at all if `fcm_api_key.is_empty`. You could move the `fcm_api_key.is_empty` condition outside of this function and only create a notification service if the `fcm_api_key` is set.

coordinator/src/notifications.rs
```diff
@@ -0,0 +1,355 @@
...
+    pub fn get_sender(&self) -> mpsc::Sender<Notification> {
+        self.notification_sender.clone()
+    }
+}
+
+/// Prepares the notification text
+fn build_notification<'a>(kind: NotificationKind) -> fcm::Notification<'a> {
+    let mut notification_builder = fcm::NotificationBuilder::new();
+    match kind {
+        NotificationKind::ChannelClose => {
```
holzeis, line 113: 🔧 Can we please rename that notification kind to something like `PositionClose`, with `ChannelClose` I am automatically thinking of the LN Cannel.

coordinator/src/orderbook/trading.rs
```diff
@@ -224,7 +232,21 @@ async fn process_new_order(
                 }
                 Err(e) => {
                     tracing::warn!(%trader_id, order_id, "{e:#}");
-                    // TODO(holzeis): send push notification to user
+
+                    if let Some(user) = user::by_id(conn, trader_id.to_string())? {
```
holzeis, line 236: 🔧 It would be nicer to use the push notification service if I wouldn't have to query the user here. I suggest moving the query to the notification_service and keep the api to only provide the `trader_id`.

holzeis (approved):
LGTM! I've left some remarks, but they are not blocking! We can also address them in a follow up PR.

coordinator/src/bin/coordinator.rs
```diff
@@ -261,7 +266,28 @@ async fn main() -> Result<()> {
 
     let notification_service = NotificationService::new(opts.fcm_api_key);
 
-    let _sender = notification_service.get_sender();
+    tokio::spawn({
+        let sender = notification_service.get_sender();
+        let pool = pool.clone();
+        async move {
+            loop {
+                match pool.get() {
```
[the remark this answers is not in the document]
klochowicz: Keeping the connection up and utilising it once per hour IMHO is more wasteful. 
It should be more optimal to reach for it when actually using it, if it's done so infrequently, so that connection can be used somewhere else in the meantime.

that's my understanding at least, please correct me if I'm wrong!

coordinator/src/notifications.rs
klochowicz: the reason I kept it (as stipulated in the previous commit) is to ensure we can at least see the correct behaviour in the logs, whilst running locally etc.

coordinator/src/message.rs
klochowicz: I'll change it to `WebsocketMessage`, thank you fro the suggestion.

I'm not sure whether it would make sense to couple them together (in my mind push notifications are completely separate thing, that user can opt-in into; and also needs FCM key to work locally).

coordinator/src/notifications.rs
klochowicz: I've removed it completely! the idea to try to prompt the user when we would like to close the lightning LN channel collaboratively to avoid having locked up liquidity, but alas we're not using that, so there was little point in prematurely adding an enum variant!

coordinator/src/orderbook/trading.rs
klochowicz: interesting idea; I'll experiment with such a thing in the follow-up!

coordinator/src/message.rs
klochowicz: edit: there was already `WebsocketMessage` somewhere else, so I ended up with `OrderbookMessage`.

coordinator/src/bin/coordinator.rs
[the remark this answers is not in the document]
holzeis: I think you are right. Keeping a connection idle for a longer period of time is less efficient than requesting a new connection every time we need it. 👍

coordinator/src/notification_service.rs
luckysori: I prefer to remove everything that's not strictly necessary so that the test is simpler, but not a big deal.

coordinator/src/notification_service.rs
luckysori: Fair enough. I think you could have done that within the `find_map`, but this code has diverged so I can't really give it a go myself anymore!

PR #1298, opened 2023-09-14
chore: Sync position with dlc channel state

### LINKED ISSUE #1271: Show order filling dialog when app is restarted in an intermediate dlc channel state
At the moment the app tries to recover the dlc channel state if it is in some intermediate state. However the user is not aware of that and may close the app too early.

The user should be shown that the app tries to recover the dlc channel state.

Note, there should be a timeout, but that's subject of another ticket already.

### LINKED ISSUE #1273: Sync position with dlc channel state
Due to an error I am having an open position but a closed dlc. It would be great if the app would be smart enough to sync these states, while the DLC is the source of truth.

if DLC is in `SubChannelState::Signed`: Create position (take data from order) if not existing
if DLC is in `SubChannelState::OffChainClosed`: Delete position if still existing

holzeis:
This change will sync the position to the dlc channel state in the following scenarios

1. DLC Channel in state `Signed` but no position: Create position from `filling` order.
2. DLC Channel in state `OffChainClosed` and a position exists. Delete the position.
3. DLC Channel in state `CloseOffered` or `CloseAccepted`: Inform the UI that the dlc channel is recovering.
4. DLC Channel in state `Offered`, `Accepted` or `Finalized`: Inform the UI that the dlc channel is recovering.
5. DLC Channel in any other state but with position: Delete position the channel might have been force closed.

<details>
<summary>Recover from scenarios 1 and 2. </summary>

I have commented the part of the code that should update the position state, so that the spinner will go forever when opening or closing a position.

https://github.com/get10101/10101/assets/382048/bfd80362-da78-46e5-8580-7b171ab03c9e
</details>

<details>
<summary>Recover from scenarios 3 and 4.</summary>

https://github.com/get10101/10101/assets/382048/b4d60db7-ca2a-40e4-b19c-936f4817ecb1

Interestingly this takes quite a long time to complete. I did at a 5 seconds sleep for the channel to get ready with https://github.com/get10101/10101/commit/4a5e66d63aaa12a07a43c52a2972dc56a3e3b9f5, but it still takes far too long.

Manually calling the `period_check` to process any pending actions immediately "fixed" that issue, but should be fixed in `rust-dlc`. There is no reason why we should check every 30 seconds if pending actions need to be processed as pending action can only ever happen after channel reestablishment.

</details>

<details>
<summary>Recover from scenario 5 (force closure)</summary>

https://github.com/get10101/10101/assets/382048/fd8f8009-dd05-47c4-bc80-993ad8832665

</details>

fixes #1273
fixes #1271


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -513,6 +513,14 @@ where
             }
         }
     }
+
+    pub async fn periodic_check(&self) -> Result<()> {
```
klochowicz, line 517: it would be helpful to keep the name to see what we're actually periodically checking

```suggestion
    pub async fn sub_channel_periodic_check(&self) -> Result<()> {
```

mobile/lib/common/recover_dlc_change_notifier.dart
```diff
@@ -0,0 +1,55 @@
...
+      taskStatus = recoverDlc.taskStatus;
+
+      if (taskStatus == TaskStatus.pending) {
+        // initialize dialog for the pending task
+        showDialog(
+          context: shellNavigatorKey.currentContext!,
+          builder: (context) {
+            TaskStatus status = context.watch<RecoverDlcChangeNotifier>().taskStatus;
+
+            // todo(holzeis): Reusing the order submission status dialog is not nice, but it's actually suitable for any task execution that has pending,
```
klochowicz, line 32: is there a particular reason you're not just keeping this as normal `TODO`s ? e.g. do you have a commit hook that checks for these specifically? 

on my end is worse than normal `TODO` as it's harder to spot it both with the naked eye when reading the code and also doesn't get picked up by my IDE which e.g. automatically shows me TODOs that are not on the main branch.

also, one can see the author of a TODO with regular `git blame`.

if you don't mind either way, please consider the points above 🙏

mobile/lib/common/recover_dlc_change_notifier.dart
```diff
@@ -0,0 +1,55 @@
...
+            // failed and success states. We may should consider renaming this dialog for its more generic purpose.
```
klochowicz, line 33: 🙃 if you're reusing and repurposing, might as well rename it now to encourage more reuse in the future! 

totally agree with the reasoning.

mobile/native/src/event/api.rs
```diff
@@ -34,6 +34,7 @@ pub enum Event {
 pub enum BackgroundTask {
     AsyncTrade(OrderReason),
     Rollover(TaskStatus),
+    RecoverDlc(TaskStatus),
```
klochowicz, line 37: 🙃 this is a great spot to write some commets (or in the classes). 
e.g. I was away at a conference, and some of the names are pretty cryptic. 
more docs would be great!

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+    // periodically checking if a pending action needs to be sent. Note, pending actions can only
+    // get created on channel reestablishment.
+    tokio::time::sleep(Duration::from_secs(5)).await;
+    let node = NODE.try_get().context("failed to get ln dlc node")?;
+    if let Err(e) = node.inner.periodic_check().await {
+        tracing::error!("Failed to process periodic check! Error: {e:#}");
+    }
+
+    // Sending a failed event with a delay. This event will be ignored by the UI if the task has
+    // been successful in the meantime.
```
klochowicz, line 654: why is the delay needed? it would be helpful to have that information.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+
+    Ok(())
+}
+
+/// Sends a pending RecoverDlc background task notification to the UI, allowing the UI to show a
+/// dialog with a spinner that the DLC protocol is still in progress.
+/// Also triggers the `periodic_check` to process any actions that might have been created after the
+/// channel reestablishment.
+///
+/// fixme(holzeis): We currently use different events for show the recovery of a dlc and the waiting
```
klochowicz, line 632: 🙏 again, please consider using standard uppercase `FIXME` instead of custom one (with your name added to it). see other comment.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+/// for an order execution in the happy case (without an restart in between). Those events and
+/// dialogs should be aligned.
+async fn recover_dlc() -> Result<()> {
+    tracing::warn!("It looks like the app was closed while the protocol was not finished yet.");
```
klochowicz, line 636:
```suggestion
    tracing::warn!("It looks like the app was closed while the protocol was still executing.");
```

nit: the other version was not quite right gramatically, `yet` is mostly used with Present Perfect tense.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+    event::publish(&EventInternal::BackgroundNotification(
+        BackgroundTask::RecoverDlc(TaskStatus::Pending),
+    ));
+
+    // fixme(holzeis): We are manually calling periodic check here to speed up the processing of
```
klochowicz, line 641: pretty please, prefer `FIXME` 🙃

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+    // pending actions. Note, this might also not speed up the process, as the coordinator might
+    // have to resend a message to continue the protocol. This should be fixed in `rust-dlc` and any
+    // pending actions should be processed immediately once the channel is ready instead of
+    // periodically checking if a pending action needs to be sent. Note, pending actions can only
+    // get created on channel reestablishment.
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
klochowicz, line 647: why are we waiting 5s here?
the comment is about later part of the code, but this is confusing.

🤔 is this only for the UI?

mobile/native/src/ln_dlc/node.rs
```diff
@@ -298,6 +258,19 @@ impl Node {
             )
             .context("Failed to update position after DLC creation")?;
 
+            // Sending always a recover dlc background notification success message here as we do
+            // not know if we might have reached this state after a restart. This event is only
+            // received by the UI at the moment indicating that the dialog can be closed.
+            // If the dialog is not open, this event would be simply ignored by the UI.
+            //
+            // fixme(holzeis): We should not require that event and align the UI handling with
```
klochowicz, line 266:
```suggestion
           // FIXME: We should not require that event and align the UI handling with
```

mobile/native/src/ln_dlc/node.rs
```diff
@@ -329,10 +302,22 @@ impl Node {
...
-            position::handler::update_position_after_dlc_closure()
+            position::handler::update_position_after_dlc_closure(Some(filled_order))
                 .context("Failed to update position after DLC closure")?;
 
+            // Sending always a recover dlc background notification success message here as we do
+            // not know if we might have reached this state after a restart. This event is only
+            // received by the UI at the moment indicating that the dialog can be closed.
+            // If the dialog is not open, this event would be simply ignored by the UI.
+            //
+            // fixme(holzeis): We should not require that event and align the UI handling with
```
klochowicz, line 313:
```suggestion
// FIXME We should not require that event and align the UI handling with
```

mobile/lib/common/domain/background_task.dart
```diff
@@ -49,3 +49,17 @@ class Rollover {
     return bridge.BackgroundTask_Rollover(TaskStatus.apiDummy());
   }
 }
+
+class RecoverDlc {
```
klochowicz, line 53: is this only used for the purpose of opening / closing a dialog? 

perhaps we could call it more specifically, like `RecoverDlcStatusDialog` or something like that?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -525,6 +532,136 @@ pub fn get_usable_channel_details() -> Result<Vec<ChannelDetails>> {
...
+///
+/// - DLC Channel in state `Signed` but no position: Create position from `filling` order.
+/// - DLC Channel in state `OffChainClosed` and a position exists. Delete the position.
+/// - DLC Channel in state `CloseOffered` or `CloseAccepted`: Inform the UI that the dlc channel is
+///   recovering.
+/// - DLC Channel in state `Offered`, `Accepted` or `Finalized`: Inform the UI that the dlc channel
+///   is recovering.
+/// - DLC Channel in any other state but with position: Delete position the channel might have been
+///   force closed.
+async fn sync_position_with_dlc_channel_state() -> Result<()> {
```
klochowicz, line 545: 🤔 is there any way to cover this functionality with tests?

mobile/native/src/orderbook.rs
```diff
@@ -109,6 +110,15 @@ pub fn subscribe(
...
 
+                                // Sending a failed event with a delay. This event will be ignored by the UI if the task has
+                                // been successful in the meantime.
+                                tokio::spawn(async {
+                                    tokio::time::sleep(ROLLOVER_PROTOCOL_TIMEOUT).await;
+                                    event::publish(&EventInternal::BackgroundNotification(
+                                        BackgroundTask::RecoverDlc(TaskStatus::Failed),
+                                    ));
+                                });
+
```
klochowicz, line 121: 🚨 what's the purpose of this one? 

we've just got a request from orderbook about doing rollover, and suddenly we're sending a *failure* about a different task? 

this seems incredibly hacky and counter-intuitive. Did you try to send information about `BackgroundTask::Rollover` perhaps? 

Also, seeing that the closing a dialog window development panned out from the backend side, I'm quite concerned with the direction of it - the backend should absolutely not be in the position to decide on the presentation layer for 
to be honest, I hoped for having that information sent when we *know* we failed etc. 😒
this does not seem better than just a timeout on the window in flutter, but it is an awful lot more complicated.

mobile/native/src/orderbook.rs
```diff
@@ -109,6 +110,15 @@ pub fn subscribe(
                                 tracing::info!("Received a rollover request from orderbook.");
                                 event::publish(&EventInternal::BackgroundNotification(BackgroundTask::Rollover(TaskStatus::Pending)));
 
+                                // Sending a failed event with a delay. This event will be ignored by the UI if the task has
+                                // been successful in the meantime.
+                                tokio::spawn(async {
+                                    tokio::time::sleep(ROLLOVER_PROTOCOL_TIMEOUT).await;
+                                    event::publish(&EventInternal::BackgroundNotification(
+                                        BackgroundTask::RecoverDlc(TaskStatus::Failed),
```
klochowicz, line 118:
```suggestion
BackgroundTask::Rollover(TaskStatus::Falied)
```

perhaps this is what  you wanted?
I honestly can't tell because the commit message is unrelated 😞

mobile/lib/features/trade/async_order_change_notifier.dart
```diff
@@ -27,8 +27,11 @@ class AsyncOrderChangeNotifier extends ChangeNotifier implements Subscriber {
 
   @override
   void notify(bridge.Event event) {
-    if (event is bridge.Event_BackgroundNotification &&
-        event.field0 is bridge.BackgroundTask_AsyncTrade) {
+    if (event is bridge.Event_BackgroundNotification) {
+      if (event.field0 is! bridge.BackgroundTask_AsyncTrade) {
+        // ignoring other kinds of background tasks
+        return;
+      }
```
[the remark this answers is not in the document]
holzeis: The background notification can be anything. In this scenario we are only interested in background notifications to the async trade, but I can only subscribe to the event itself. In order to prevent printing confusing "received unexpected event" messages, I have changed the condition to this.

It wasn't a problem with combining `&&` and `is`;)

mobile/native/src/trade/position/handler.rs
```diff
@@ -209,7 +209,7 @@ pub fn update_position_after_dlc_creation(
 }
 
 /// Delete a position after closing a DLC channel.
-pub fn update_position_after_dlc_closure(filled_order: Order) -> Result<()> {
+pub fn update_position_after_dlc_closure(filled_order: Option<Order>) -> Result<()> {
```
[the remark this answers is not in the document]
holzeis: yeah, that happens when you are fixing up too much 🙈 I am going to squash it with the rest of the changes 👍

mobile/lib/common/recover_dlc_change_notifier.dart
holzeis: I think I've seen it on other projects (`rust-dlc`?) done like that - and I like it. It shows ownership of the todo. `git blame` is not always trustworthy as moving code will easily change that.

I like doing it like that and I don't think it hurts, but if you are feeling strongly about it I can change it to a regular todo.

fyi: my IDE is highlighting these todos the same way its highlighting any other todo.

mobile/native/src/ln_dlc/mod.rs
holzeis: can you elaborate on the custom one? 😅 Is there a best practice that you are referring to 🙂. Happy to change it, but as describe above to me it only adds value seeing who added the `todo` / `fixme`

mobile/native/src/ln_dlc/mod.rs
holzeis: I guess the comment wasn't clear enough 🙈. I am going to adapt it.

The 5 seconds wait is pretty arbitrary, but necessary to wait for the channel reestablishment to finish. However that code shouldn't even be here, but handled in `rust-dlc`. I'll probably also wrap that into a separate commit so we can drop it once its fixed in `rust-dlc.

Note, the channel reestablishment identifies where the protocol stopped and creates the required actions to continue / recover the protocol. Since we are only calling the `periodic_check` every 30 seconds I figured I am speeding that up by immediately calling the `periodic_check` here and thus continue the protocol earlier.

mobile/native/src/orderbook.rs
holzeis: > 🚨 what's the purpose of this one?

To show a failed (timed out) rollover in the frontend, if it hasn't received a success before the timeout.

> we've just got a request from orderbook about doing rollover, and suddenly we're sending a failure about a different task?

yes it would get ignored by the frontend if it has already gotten a success message! 

> Also, seeing that the closing a dialog window development panned out from the backend side, I'm quite concerned with the direction of it - the backend should absolutely not be in the position to decide on the presentation layer for

I do agree, that this timeout code should not be in the backend. I'll move it to the frontend. 👍 

> to be honest, I hoped for having that information sent when we know we failed etc. 😒
this does not seem better than just a timeout on the window in flutter, but it is an awful lot more complicated.

It's unfortunately not that easy to know when we've failed. Running into a timeout is just the easiest thing to do first. There is actually an open ticket to address this properly. This was not the intention of that change, but rather to add a simple safeguard that the user can always use the app eventually. (even if the protocol failed / or did not finish)

mobile/native/src/orderbook.rs
holzeis: oh yeah.. good catch 👍  but as discussed above, I think that code should be anyways moved to the frontend.

mobile/native/src/ln_dlc/mod.rs
holzeis: Moved that piece to the frontend, I think its cleaner now. Thanks for highlighting that short coming. 👍

mobile/lib/common/domain/background_task.dart
holzeis: At the moment, yes there is only one consumer being the UI. The event still has a general purpose though. The sender of an event should not care about the receiver.

mobile/native/src/ln_dlc/mod.rs
holzeis: oh yeah, I think thats a very good idea 🙈. I will try to write some tests for that! 👍

mobile/lib/common/recover_dlc_change_notifier.dart
holzeis: if you are using this plugin https://github.com/alphapapa/magit-todos, I guess the only issue was that I did not capitalized the keyword.

> Items in KEYWORD(username): format are also read

I have capitalized all fixmes and todos, let me know if this is working for you now.

mobile/native/src/ln_dlc/mod.rs
holzeis: I had a look into it and I fear adding regression tests for this functionality will bring a lot more effort and code changes to this PR than I want to invest at the moment.

I'll make sure that I have everything manually tested though.

luckysori (commented):
LGTM, but I would really appreciate some tests. I think we can fairly simply define a pure function that can be unit tested. Then it wouldn't be so important to review the tricky logic.

mobile/native/src/orderbook.rs
klochowicz: also:  sorry for the typo with `Falied` in my suggestion 😭

mobile/native/src/event/api.rs
```diff
@@ -32,8 +32,14 @@ pub enum Event {
 #[frb]
 #[derive(Clone)]
 pub enum BackgroundTask {
```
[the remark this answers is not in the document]
holzeis: well, they are messages tasks executed in the background 😅. Any suggestion for a better name? 🙂

mobile/native/src/ln_dlc/recover.rs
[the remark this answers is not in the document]
holzeis: I don't like this long file names, but I can't think of anything better. 👍

mobile/native/src/ln_dlc/recover.rs
```diff
@@ -0,0 +1,134 @@
...
+        let dlc_channel = dlc_channels
+            .iter()
+            .find(|dlc_channel| dlc_channel.channel_id == channel_details.channel_id);
+
+        let positions = db::get_positions()?;
+
+        match (positions.first(), dlc_channel) {
+            (Some(_), Some(dlc_channel)) => {
+                if matches!(dlc_channel.state, SubChannelState::Signed(_)) {
+                    tracing::info!("DLC channel and position are in sync");
```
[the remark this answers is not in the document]
holzeis: I put it to debug, but I don't mind those log messages here, as they are only printed once at startup, and do not really pollute anything.

holzeis:
@luckysori @klochowicz I have added unit tests. https://github.com/get10101/10101/pull/1298/commits/de37a349573b9f4947fd56e7ca70f74ba8308ef1#diff-817ef51e9e3a8d899433cfad50ab3e46d729def3e1bd663f5d5614a8a468e174

Please have another look.

klochowicz (approved):
LGTM!

mobile/native/src/event/api.rs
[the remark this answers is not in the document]
luckysori: I think it's fine looking at it again.

mobile/native/src/event/api.rs
```diff
@@ -32,8 +32,14 @@ pub enum Event {
...
+    /// The order book submitted an trade which was matched asynchronously while the app was
+    /// offline.
     AsyncTrade(OrderReason),
+    /// The order book submitted its intention to rollover the about to expire position.
     Rollover(TaskStatus),
+    /// The app was started with a dlc channel in an intermediate state. This task is in pending
+    /// until the dlc protocol reaches a final state.
+    RecoverDlc(TaskStatus),
```
luckysori, line 42: :upside_down_face: I find the term "recover" a bit overloaded. How about `ContinueSubchannelProtocol`?

luckysori (approved):
Thank you for adding the tests!

PR #1301, opened 2023-09-15
Avoid panics at startup

klochowicz:
- ensure the app doesn't crash even if the backend cannot be started (e.g. no Internet, coordinator/esplora is down)
- clearly communicate limited functionality by a alert dialog
- show red 'Offline' status in the corner of the screen (thermostat)

Even if the backend cannot be started, the user can:
- export seed (!)
- share logs
- see most of the UI (although most of the functionality is disabled)

I also cleaned some unnecessary `expect` calls whilst investigating this issue.
flutter_05

flutter_06

### DIFF
--- a/crates/ln-dlc-node/src/node/storage.rs
+++ b/crates/ln-dlc-node/src/node/storage.rs
@@ -14,10 +14,9 @@ use lightning::chain::transaction::OutPoint;
 use lightning::ln::PaymentHash;
 use lightning::ln::PaymentPreimage;
 use lightning::ln::PaymentSecret;
+use parking_lot::Mutex;
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::sync::Mutex;
-use std::sync::MutexGuard;
 use time::OffsetDateTime;
 
 /// Storage layer interface.
@@ -108,7 +107,7 @@ impl Storage for InMemoryStore {
     // Payments
 
     fn insert_payment(&self, payment_hash: PaymentHash, info: PaymentInfo) -> Result<()> {
-        self.payments_lock().insert(payment_hash, info);
+        self.payments.lock().insert(payment_hash, info);
 
         Ok(())
     }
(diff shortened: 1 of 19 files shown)


crates/ln-dlc-node/src/node/storage.rs
```diff
@@ -14,10 +14,9 @@ use lightning::chain::transaction::OutPoint;
 use lightning::ln::PaymentHash;
 use lightning::ln::PaymentPreimage;
 use lightning::ln::PaymentSecret;
+use parking_lot::Mutex;
```
bonomat, line 17: Can you explain why `parking_lot::Mutex` is better than `std::sync::Mutex`?

bonomat (approved):
LGTM

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -95,7 +96,9 @@ pub async fn refresh_wallet_info() -> Result<()> {
 }
 
 pub fn get_seed_phrase() -> Vec<String> {
-    NODE.get().get_seed_phrase()
+    SEED.try_get()
+        .expect("SEED to be initialised")
```
luckysori, line 100: :upside_down_face: May as well use `get` here then.

luckysori (approved):
LGTM!

crates/ln-dlc-node/src/node/storage.rs
klochowicz: A few reasons (arranged in order of relevance to our project):

- as some time ago we concluded that we don't care about the state of poisoning enough to warrant crashing the app, having a mutex that does not crash at such unlikely event is better for us
- it supports deadlock detection if we ever want to use it (I believe we turned it off recently, but we *can* turn it on)
- it's faster than the `std` counterpart

mobile/native/src/ln_dlc/mod.rs
klochowicz: I did this on purpose to have a more clear error message to readability if it ever happens.
 the `"used set() before get()"` with a line number is IMHO far worse than this when we have more than one thing that can print this ☺️

PR #1303, opened 2023-09-15
chore: Add logs around periodic check

holzeis:
Adding logs around the periodic check to debug whether the periodic check interferes with the processing of DLC messages.


klochowicz (approved):
LGTM! (I've fixed the formatting, as it was just `dprint` run)

Issue #1309, opened 2023-09-18
Android build fails


bonomat:
Our CI build for Android failed: https://github.com/get10101/10101/actions/runs/6220173722/job/16879658066

```
 │ [!] Your project requires a newer version of the Kotlin Gradle plugin.                     │
│ Find the latest version on https://kotlinlang.org/docs/releases.html#release-details, then │
│ update /Users/runner/work/10101/10101/mobile/android/build.gradle:                         │
│ ext.kotlin_version = '<latest-version>'              
```

Looks like we downgraded/upgraded a version.

holzeis:
Should be fixed with https://github.com/get10101/10101/pull/1310/commits/ee22c0de2022104fd878c7e4f058cb4bc76d4574

bonomat:
Was not fixed unfortunately. 

This ticket is related: https://github.com/fastlane/fastlane/issues/21091

bonomat:
https://github.com/get10101/10101/pull/1311 did not fix it neither, trying the second recommendation from https://github.com/fastlane/fastlane/issues/21091 and bump fastlane

bonomat:
With the latest fix it seems to be mitigated by simply retrying up to 5 times.

PR #1311, opened 2023-09-18
fix: allow uploading to google playstore again

bonomat:
fixes https://github.com/get10101/10101/issues/1309

Tested locally successfully


holzeis (approved):
Weird that this has to be added to make it work again.

Issue #1244, opened 2023-09-07
Flaky test `maker_can_open_channel_to_coordinator_and_send_payment`


luckysori:
CI repro [here](https://github.com/get10101/10101/actions/runs/6105399183/job/16568856595#step:14:2058).

klochowicz:
is it actually that test? 
looking at the CI output, the previous test has run into some issues with the state of the lightning channel:

`2023-09-07T04:40:00.154503Z ERROR dlc_manager::sub_channel_manager: Got close notification from LDK in a state that we don't handle yet    `

probably that's why the subsequent test couldn't pay the invoice.

🤔 we might need to add a sanity check for the LN and DLC states at the beginning and end of tests to catch these errors better!

bonomat:
@klochowicz : this should not have an impact on the test @luckysori mentioned as in this test we are opening the channel from the maker using on-chain funds.

bonomat:
The issue in this test seems to be esplora: the maker was not able to broadcast the fund transaction and the channel was force-closed immediately. 

One assumption I have is that the maker might have been funded using a coinbase tx and it was not able to spend the money yet. I'll give it a try here: https://github.com/get10101/10101/pull/1252

bonomat:
> One assumption I have is that the maker might have been funded using a coinbase tx and it was not able to spend the money yet. I'll give it a try here: #1252

Note: this was not the case. This ticket is still up for grabs.

bonomat:
Parked until this happens again.

Issue #1243, opened 2023-09-07
Flaky test `force_close_ln_dlc_channel`


luckysori:
CI repro [here](https://github.com/get10101/10101/actions/runs/6091875463/job/16533953052?pr=1216#step:8:4931).

bonomat:
My assumption is that this was due to the machine being overloaded and bitcoin was not done with processing all mined blocks. In this particular test we are generating quite a few blocks.

I propose to avoid this particular problem simply checking and reacting to the response from bitcoind.

klochowicz:
I've added container logs: https://github.com/get10101/10101/pull/1389 .
Meanwhile, I could not reproduce it locally, so I'm closing the issue until it appears again, as it's not directly actionable (hopefully next time we'll get more detailed logs with the PR linked above).

PR #1312, opened 2023-09-19
chore(ci): Print services logs upon e2e test suite error

klochowicz:
It's hard to determine the cause of test flakiness without all the auxiliary
logs from other services.

Print the logs of maker, coordinator and docker containers in case e2e tests fail.


klochowicz:
related to #1244 #1243

bonomat (approved):

holzeis (approved):

PR #1314, opened 2023-09-19
Small UI fixes

klochowicz:
# fix(ui): Rename buy & sell buttons to also mention long & short
flutter_08

# chore: Style the values in the StatusScreen
flutter_09

# fix: Display Settings icon in Stable screen

flutter_07


mobile/lib/features/trade/trade_screen.dart
```diff
@@ -286,7 +286,7 @@ class TradeScreen extends StatelessWidget {
                         tradeBottomSheet(context: context, direction: Direction.long);
                       }
                     },
-                    label: const Text("Buy"),
+                    label: const Text("Buy / Long"),
```
holzeis, line 289: Obviously subjective, but I like the "Buy" and "Sell" labels better than "Buy / Long" and "Sell / Short".

holzeis (commented):
I am fine with the other changes, but not on board on the "Buy / Long" and "Sell / Short" labels. However I will not block that PR if others think the new labels look better.

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -83,7 +83,7 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
     WalletInfo walletInfo = context.watch<WalletChangeNotifier>().walletInfo;
 
     Direction direction = widget.direction;
-    String label = direction == Direction.long ? "Buy" : "Sell";
+    String label = direction == Direction.long ? "Buy / Long" : "Sell / Short";
```
bonomat, line 86: I'm against this change. Can we leave just "Buy" and "Sell" please?

bonomat (approved):
I agree with [Richie](https://github.com/get10101/10101/pull/1314#pullrequestreview-1632678180). All good but please leave the `Buy`/`Sell` labels :)

mobile/lib/features/trade/trade_screen.dart
klochowicz: sure, I can remove that change from this changeset! 
It was one of the feedback from the conference - people are *really* confused by only 'buy' and 'sell' (and asked questions like "can I *buy* Bitcoin here? what it is this?".
I took the new naming after Kollider (which has it like that), and thought that it is only an improvement - if less people get confused, but still understandable for pros. (especially that we're targetting *retail* traders).

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
klochowicz: copying over from the other comment:

> sure, I can remove that change from this changeset! 
It was one of the feedback from the conference - people are *really* confused by only 'buy' and 'sell' (and asked questions like "can I *buy* Bitcoin here? what it is this?".
I took the new naming after Kollider (which has it like that), and thought that it is only an improvement - if less people get confused, but still understandable for pros. (especially that we're targetting *retail* traders).

klochowicz:
> I agree with [Richie](https://github.com/get10101/10101/pull/1314#pullrequestreview-1632678180). All good but please leave the `Buy`/`Sell` labels :)

cool, I'll remove it from here, but I'd be keen on still exploring that avenue based on user feedback I've received during Builder Days. :)

PR #1332, opened 2023-09-19
Ensure app wallet balance and history appear ready ASAP

### LINKED ISSUE #1258: Sync should not block the whole app

### LINKED ISSUE #1331: Starting the app takes a long time
Starting the app now takes a long time. 
It looks like we are stuck at the loading screen until the wallet is fully synced. 

- iOS
- Commit `b543f431efc531d288ecd32e2a5c4d719944fb2c`
- Build `2122`
- Version `1.2.6`

Declaring this as a bug because it wasn't the case in the past.

luckysori:
Fixes #1258.
Fixes #1331.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -105,6 +105,11 @@ pub async fn refresh_wallet_info() -> Result<()> {
         if let Err(e) = wallet.sync() {
             tracing::error!("Manually triggered on-chain sync failed: {e:#}");
         }
+
+        if let Err(e) = node.inner.sync_lightning_wallet() {
+            tracing::error!("Manually triggered Lightning wallet sync failed: {e:#}");
+        }
+
         if let Err(e) = keep_wallet_balance_and_history_up_to_date(node) {
             tracing::error!("Failed to keep wallet history up to date: {e:#}");
         }
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -245,25 +245,44 @@ pub fn run(data_dir: String, seed_dir: String, runtime: &Runtime) -> Result<()> 
         let _running = node.start(event_handler)?;
         let node = Arc::new(Node::new(node, _running));
 
+        // Refresh the wallet balance and history eagerly so that it can complete before the
+        // triggering the first on-chain sync. This ensures that the UI appears ready as soon as
+        // possible.
+        //
+        // TODO: This might not be necessary once we rewrite the on-chain wallet with bdk:1.0.0.
+        spawn_blocking({
+            let node = node.clone();
+            move || keep_wallet_balance_and_history_up_to_date(&node)
+        })
+        .await
+        .expect("task to complete")?;
+
+        runtime.spawn({
+            let node = node.clone();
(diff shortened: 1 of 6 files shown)


klochowicz (approved):

bonomat (commented):
You could run it locally against regtest and see if it fixes the startup bug.

holzeis (approved):

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -89,12 +90,32 @@ pub const FUNDING_TX_WEIGHT_ESTIMATE: u64 = 220;
...
+/// Trigger an on-chain sync followed by an update to the wallet balance and history.
+///
+/// We do not wait for the triggered task to finish, because the effect will be reflected
+/// asynchronously on the UI.
 pub async fn refresh_wallet_info() -> Result<()> {
     let node = NODE.try_get().context("failed to get ln dlc node")?;
     let wallet = node.inner.wallet();
 
-    spawn_blocking(move || wallet.sync()).await??;
-    keep_wallet_balance_and_history_up_to_date(node)?;
```
holzeis, line 97: 🤔 Maybe it will be faster if we run `keep_wallet_balance_and_history_up_to_date` before everything else? It may use outdated data, but that should be updated after the sync immediately.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -224,6 +245,28 @@ pub fn run(data_dir: String, seed_dir: String, runtime: &Runtime) -> Result<()>
...
+            let node = node.clone();
+            move || {
+                // We wait a minute to not interfere with the app's startup. We want to be able to
+                // quickly refresh the wallet _before_ the first on-chain sync, as this one can take
+                // a long time and blocks the wallet refresh since they both need to acquire the
+                // same mutex.
+                //
+                // TODO: This should not be necessary as soon as we rewrite the on-chain wallet with
+                // bdk:1.0.0.
+                std::thread::sleep(Duration::from_secs(60));
```
holzeis, line 258: 🤔 calling the `keep_wallet_balance_and_history_up_to_date` function first, may prevent us from having to sleep here?

holzeis (approved):
Tested locally, it still feels like its doing more at startup as I would expect. But its definitely faster. About 2 seconds until the app is ready to use 👍

mobile/native/src/ln_dlc/mod.rs
luckysori: I considered that too, but now this is only called if the user pulls down on the wallet screen. In such a scenario, I think it would be okay to wait for the syncs before updating the UI.

On startup this is no longer called.

mobile/native/src/ln_dlc/mod.rs
luckysori: I like your suggestion! I'll give it a go.

holzeis (approved):

PR #1333, opened 2023-09-20
Bump flutter to 3.13.4

klochowicz:
After more people on the team started bumping into this issue, @luckysori and myself agreed that the best way forward would be to just bump flutter version, as it looks like they've fixed a bug that we were depending on 😅

it is recommended to update it locally to the latest version.
I've tested it locally and it seemed to work.


bonomat (approved):
I trust that you tested all features, not just the account expander.

luckysori (approved):
LGTM, thank you!

PR #1334, opened 2023-09-20
feat: also return commit hash and number in version api

bonomat:
Useful to figure out what exact version the coordinator is running. 

Note: breaking change for old clients but I guess we can live with that. It will show a warning that the coordinator is not reachable until they upgrade


holzeis (approved):
LGTM!

PR #1335, opened 2023-09-20
Enrich logic to derive wallet history trades based on filled orders

luckysori:
We've talked about removing the hacky logic of using the order index to derive the trades for a while.

This proposal can be relatively easily extended to support position resizing.

We still assume that filled orders are fetched in the correct order, and that all orders have the same contract symbol.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -1,5 +1,9 @@
 use self::node::WalletHistories;
 use crate::api;
+use crate::api::PaymentFlow;
+use crate::api::Status;
+use crate::api::WalletHistoryItem;
+use crate::api::WalletHistoryItemType;
 use crate::calculations;
 use crate::channel_fee::ChannelFeePaymentSubscriber;
 use crate::commons::reqwest_client;
@@ -63,6 +67,7 @@ use time::OffsetDateTime;
 use tokio::runtime::Runtime;
 use tokio::sync::watch;
 use tokio::task::spawn_blocking;
+use trade::Direction;
 
 mod lightning_subscriber;
 mod node;
@@ -70,7 +75,6 @@ mod sync_position_to_dlc;
 
 pub mod channel_status;
 
-use crate::api::Status;
 pub use channel_status::ChannelStatus;
 
 const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_millis(200);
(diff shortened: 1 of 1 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -560,6 +504,128 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
...
+                        order
+                            .execution_price()
+                            .expect("execution price to be set on a filled order"),
+                    )?;
+
+                    let opening_price = open_order
+                        .execution_price()
+                        .expect("initial execution price to be set on a filled order");
+
+                    let pnl = calculations::calculate_pnl(
```
bonomat, line 560: Why do we need to re-calculate this every time? Shouldn't this be final if the order was filled?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -560,6 +504,128 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
...
+                    Short => -order.quantity,
+                };
+                let updated_total_contracts = total_contracts + new_contracts;
+
+                // Closing the position.
+                if updated_total_contracts == 0.0 {
+                    let open_order = previous_order;
+                    let trader_margin = open_order
+                        .trader_margin()
+                        .expect("Filled order to have a margin");
```
bonomat, line 549: Should we handle this error ? Otherwise we panic and 💥

bonomat (commented):
It's hard to see what's new and what's old, so I just added some comments 😅. Will finish the review later.

mobile/native/src/ln_dlc/mod.rs
luckysori: Hmm, this one didn't change and we just know that a filled order must have a margin. It sucks to not use the type system for this, but we chose to create an `orders` table with nullable values.

bonomat: Got ya

mobile/native/src/ln_dlc/mod.rs
luckysori: The thing is the PNL is not persisted anywhere in the app (other than when closing a subchannel, but that's also limiting once we allow position resizing via subchannel renewal).

It's definitely not very efficient to have to process this stuff so often, although we were doing the same thing before this PR. We might want to record trades with the relevant metadata as new orders are filled (and executed), so that we can just iterate through those rows here.

luckysori:
> It's hard to see what's new and what's old, so I just added some comments 😅. Will finish the review later.

It might be easier to just ignore the previous version altogether. What matters is that the behaviour should be the same as before. Maybe I can just extract the original, add some tests and ensure that the new version does the same thing. That should be pretty easy.

bonomat (approved):
I guess this is ok :)

PR #1336, opened 2023-09-20
Get maker position information via HTTP API

### LINKED ISSUE #1328: Query maker position via HTTP API
_Originally posted by @bonomat in https://github.com/get10101/10101/pull/1216#discussion_r1329772534_

luckysori:
Fixes #1328.

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -1,5 +1,8 @@
 use crate::health::Health;
 use crate::health::OverallMakerHealth;
+use crate::position;
+use crate::position::ContractSymbol;
+use crate::position::GetPosition;
 use axum::extract::Path;
 use axum::extract::State;
 use axum::http::StatusCode;
@@ -10,9 +13,6 @@ use axum::routing::post;
 use axum::Json;
 use axum::Router;
 use bitcoin::secp256k1::PublicKey;
-use diesel::r2d2::ConnectionManager;
-use diesel::r2d2::Pool;
-use diesel::PgConnection;
 use ln_dlc_node::node::InMemoryStore;
 use ln_dlc_node::node::Node;
 use ln_dlc_node::node::NodeInfo;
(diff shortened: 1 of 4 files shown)


maker/src/routes.rs
```diff
@@ -242,6 +245,23 @@ pub async fn sync_on_chain(State(state): State<Arc<AppState>>) -> Result<(), App
     Ok(())
 }
 
+#[derive(Serialize)]
+pub struct Position {
+    tentenone: HashMap<ContractSymbol, Decimal>,
```
bonomat, line 250: What do you think of making this a custom struct with 2 fields instead of map? That would allow us to extend it easily, e.g. 

```
pub struct Position {
    positions: Vec<TenTenOnePosition>
}

pub struct TenTenOnePosition{
  symbol: String,
  quantity: Decimal
}

```

luckysori: Sure, can make it a HashSet instead, with the custom struct as the item, like you suggest.

klochowicz: I guess the downside of that that the type system will not guarantee that we won't be  having multiple positions with the same ContractSymbel

klochowicz (approved):
LGTM!

maker/src/routes.rs
luckysori: You can get the same behaviour by using a `HashSet` and implementing `Hash`, `PartialEq` and `Eq` on the custom struct only using the `ContractSymbol`. Obviously that's not as foolproof as deriving those traits on the key of a `HashMap`, but it's close enough!

luckysori: @bonomat: I did what you suggested, but the diff sucks because I force-pushed after rebasing as well. Sorry!

bonomat (approved):
Thanks for incorporating my feedback

maker/src/routes.rs
bonomat: No worries, the changeset is relatively small :)

PR #1339, opened 2023-09-21
fix: Log error instead of panicking in rust-dlc upon subchannel force-closure attempt

klochowicz:
We've run into `unimplemented!` block in rust-dlc in our regtest setup.
In order to unblock the regtest environment, this quickfix logs an error and
continues instead of panicking.

See: https://github.com/p2pderivatives/rust-dlc/pull/151


klochowicz:
I'm merging this as the quickfix got approved by Tibo.

holzeis (approved):
Nice, thanks! Fortunately we do not automatically force close if a position expires :)

PR #1340, opened 2023-09-21
Make maker more resilient against stream closures

klochowicz:
- reconnect price feed stream if it closes
- don't crash upon start if the orderbook stream is unavailable
- convey status of orderbook stream into maker health status

### DIFF
--- a/maker/src/orderbook_ws.rs
+++ b/maker/src/orderbook_ws.rs
@@ -1,3 +1,4 @@
+use crate::health::ServiceStatus;
 use crate::position;
 use crate::position::OrderTenTenOne;
 use crate::position::PositionUpdateTenTenOne;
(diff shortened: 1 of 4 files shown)


holzeis (approved):
LGTM 👍

maker/src/orderbook_ws.rs
```diff
@@ -79,6 +86,7 @@ impl Client {
                 let authenticate = auth_fn;
                 match orderbook_client::subscribe_with_authentication(url, authenticate).await {
                     Ok((mut sink, mut stream)) => {
+                        let _ = orderbook_status.send(ServiceStatus::Online);
```
luckysori, line 89: :upside_down_face: We could also send this once we are authenticated. There's a relevant message handled below.

luckysori (approved):

maker/src/orderbook_ws.rs
klochowicz: good idea! it looks perhaps a bit more convoluted now, but it's certainly more correct. :-)

klochowicz:
@luckysori I've done the 2-step update especially for you ❤️‍🔥thank you for coming up with the idea, it's a good habit!

PR #1341, opened 2023-09-21
fix: Add dedicated message if payment failed

### LINKED ISSUE #1121: "Send Payment Failure" with contradictory message
When trying to send sats on lightning from my 10101 wallet to an external wallet, it fails.

Can have several reasons BUT, the error message seems contradictory, here's a screenshot
IMG_6860

holzeis:
Note, this is just a quick patch for that scenario. We need to rework that dialog to wait for the payment to be successful or not, and not just depend whether we were able to submit the payment.

fixes #1121

### DIFF
--- a/mobile/lib/features/wallet/wallet_screen.dart
+++ b/mobile/lib/features/wallet/wallet_screen.dart
@@ -5,17 +5,17 @@ import 'package:get_10101/common/amount_text.dart';
 import 'package:get_10101/common/fiat_text.dart';
 import 'package:get_10101/common/submission_status_dialog.dart';
 import 'package:get_10101/common/value_data_row.dart';
-import 'package:get_10101/features/wallet/domain/wallet_type.dart';
 import 'package:get_10101/features/trade/position_change_notifier.dart';
-import 'package:get_10101/features/wallet/seed_screen.dart';
-import 'package:get_10101/features/wallet/send_payment_change_notifier.dart';
-import 'package:get_10101/util/preferences.dart';
 import 'package:get_10101/features/wallet/balance_row.dart';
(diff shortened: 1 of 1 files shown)


mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -84,7 +84,9 @@ class _WalletScreenState extends State<WalletScreen> {
                       Padding(
                         padding: const EdgeInsets.only(top: 20, left: 10, right: 10, bottom: 5),
                         child: Text(
-                            "Your Payment will be shown up in the wallet history automatically once it has been processed!",
+                            PendingPaymentState.failed == state
+                                ? "Failed to submit your payment. Please make sure that you have sufficient funds that the receiver is able to receive."
```
bonomat, line 88: Is there an `and` missing?

```suggestion
                                ? "Failed to submit your payment. Please make sure that you have sufficient funds and that the receiver is able to receive."
```

bonomat (approved):

mobile/lib/features/wallet/wallet_screen.dart
holzeis: Maybe? I thought the sentence makes sense that way. You must have sufficient funds that the receiver is able to receive.

Maybe to complicated? To be honest I haven't spent much time on it, but figured that's such an easy thing ticket, let's fix it very quickly, knowing that we have to rework that entire screen anyways 🙂

bonomat: My english is not the yellow from the egg. 
I think it makes sense as you had it.

SXBT69:
Thx!

PR #1343, opened 2023-09-21
fix: Show dialog only when submitting an order

### LINKED ISSUE #518: Success popup is shown twice
I couldn't reproduce it yet, but every once in a while the success popup after submitting an order is shown twice. Overlapping, meaning that I have to hit close twice to get back to the order / positions overview.

holzeis:
fixes #518


luckysori (approved):
Concept ACK.

PR #1351, opened 2023-09-22
fix: Do not rollover already rolled over positions

holzeis:
We need to use the current time instead of the expiry time, as we would otherwise rollover everytime.

### DIFF
--- a/coordinator/src/node/rollover.rs
+++ b/coordinator/src/node/rollover.rs
@@ -32,7 +32,6 @@ use trade::ContractSymbol;
 struct Rollover {
     counterparty_pubkey: PublicKey,
     contract_descriptor: ContractDescriptor,
-    expiry_timestamp: OffsetDateTime,
     margin_coordinator: u64,
     margin_trader: u64,
     contract_symbol: ContractSymbol,
@@ -146,7 +145,6 @@ impl Rollover {
         Ok(Rollover {
             counterparty_pubkey: offered_contract.counter_party,
             contract_descriptor: contract_info.clone().contract_descriptor,
-            expiry_timestamp,
             margin_coordinator,
             margin_trader,
             oracle_pk: oracle_announcement.oracle_public_key,
@@ -165,7 +163,7 @@ impl Rollover {
 
     /// Calculates the maturity time based on the current expiry timestamp.
     pub fn maturity_time(&self) -> OffsetDateTime {
-        coordinator_commons::calculate_next_expiry(self.expiry_timestamp, self.network)
+        coordinator_commons::calculate_next_expiry(OffsetDateTime::now_utc(), self.network)
     }
 }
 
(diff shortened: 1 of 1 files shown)


coordinator/src/node/rollover.rs
```diff
@@ -165,7 +163,7 @@ impl Rollover {
 
     /// Calculates the maturity time based on the current expiry timestamp.
     pub fn maturity_time(&self) -> OffsetDateTime {
-        coordinator_commons::calculate_next_expiry(self.expiry_timestamp, self.network)
+        coordinator_commons::calculate_next_expiry(OffsetDateTime::now_utc(), self.network)
```
bonomat, line 166: I'm not sure I follow this change. We should not always rollover, but only allow rollover from the old expiry to the next. Otherwise a user could extend an already expired contract. 

Also, looking at the whole logic: shouldn't be there a check somewhere in the `propose_rollover()` function if a rollover is still possible or if the position has already expired?

holzeis: > I'm not sure I follow this change. We should not always rollover, but only allow rollover from the old expiry to the next. Otherwise a user could extend an already expired contract. 

The commit message may be cryptic, but we are calculating the next expiry time here. Obviously we need to take the current date for calculating that, because if we use the current expiry we would always get the next expiry and not the closest to the current time.

> Also, looking at the whole logic: shouldn't be there a check somewhere in the `propose_rollover()` function if a rollover is still possible or if the position has already expired?


Yes, that check is basically comparing if the current expiry is the same on the position, If not propose rollover if it is then we have nothing to do and the position was just created in the rollover window or has already been rolled over.

bonomat: > The commit message may be cryptic, but we are calculating the next expiry time here. Obviously we need to take the current date for calculating that, because if we use the current expiry we would always get the next expiry and not the closest to the current time.

> Or in other words always get newer expiry and we only want the next from the current time.

Ah, that makes sense: we always want to get the _next_ expiry relative to _now_.

bonomat: > Yes, that check is basically comparing if the current expiry is the same on the position, If not propose rollover if it is then we have nothing to do and the position was just created in the rollover window or has already been rolled over.

But what if the position has already been expired? Am I missing this check?

holzeis: Look into the check if we need to rollover 😉 (on `rollover.rs`)

bonomat: Oh, I'm a blind 🦄. Thanks. 

```
 if expiry_timestamp < OffsetDateTime::now_utc() {
            bail!("Cannot rollover an expired position");
        }
```

bonomat (approved):
Thanks for the clarifications.

PR #1352, opened 2023-09-22
fix: Update order and match in all trade scenarios

holzeis:
We ended up in several instances with a pending match if for any other reason than the ones covered the trade failed.

The trade result is now handled at a single point for every err and ok scenario.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -143,12 +143,55 @@ impl Node {
     }
 
     pub async fn trade(&self, trade_params: &TradeParams) -> Result<Invoice> {
-        let (fee_payment_hash, invoice) = self.fee_invoice_taker(trade_params).await?;
         let mut connection = self.pool.get()?;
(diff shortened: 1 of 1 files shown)


coordinator/src/node.rs
```diff
@@ -143,12 +143,55 @@ impl Node {
...
+                    MatchState::Filled,
+                    OrderState::Taken,
+                )?;
+                Ok(invoice)
+            }
+            Err(e) => {
+                tracing::error!(
+                    %trader_id,
+                    %order_id,
+                    "Failed to execute trade"
```
bonomat, line 169:
```suggestion
                    "Failed to execute trade: {e:#}"
```

bonomat (approved):
That's better. Thank you.

coordinator/src/node.rs
bonomat: Note: I did not try if this compiles.

PR #1354, opened 2023-09-22
fix: Do not delete orders

### LINKED ISSUE #1347: Could not delete `order`
Logs from the database: 

- setup: regtest VM

`2023-09-21 19:41:14.059 UTC [123] ERROR:  update or delete on table "orders" violates foreign key constraint "matches_order_id_fkey" on table "matches"`

Could find relevant logs in the coordinator yet, maybe we are swallowing the error?

holzeis:
Instead of deleting orders, I propose to keep them in state `Failed`. That might blow up our database, but we could at any time delete `Failed` orders if we want to.

To be consistent with the existing implementation I am also sending a `DeleteOrder` event to update the prices for the apps.

fixes #1347

### DIFF
--- a/coordinator/src/orderbook/db/orders.rs
+++ b/coordinator/src/orderbook/db/orders.rs
@@ -276,13 +276,20 @@ pub fn set_order_state(
     Ok(OrderbookOrder::from(order))
 }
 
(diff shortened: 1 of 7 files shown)


bonomat (approved):
Works for me. Alternatively we could introduce a new state `Deleted`.

PR #1355, opened 2023-09-22
fix: Ensure the fcm token is stored to the users table

### LINKED ISSUE #1348: User has no fcm token
I re-installed the regtest app and registered a new user but the user db does not show me having an fcm token.

- regtest
- iPhone
- build 2152
- commit  8fa4e95bf25d3287d2a95e1d2e569058c33abc33
- node  id: 036a564b61ee0e74c34751a42f910beded04cdb65cad67c9104eb0fee5ab96ce03

logs: 

[2023-09-22_064343.log](https://github.com/get10101/10101/files/12696858/2023-09-22_064343.log)

holzeis:
This change addresses a couple of things

1. The fcm token update required the node to be up, but did not wait for the backend to start. Thus it was very likely that we were running into a race condition with the node not yet being initialized and the fcm token not being updated to the database.
2. Upon initial registration we call the register endpoint to create the user. In the background the fcm token update api was called. It was very likely that the user was not faster than the fcm token update, thus the user record did not exist at the time the fcm token is being updated. (this would have been fixed at the second start of the app)

I opted to remove the fcm token update api alltogether and rather send the `fcm token` with our initial `OrderbookMsg::Authenticate`. This ensures firstly, that the node has been successfully started before sending the `fcm_token` and secondly protects the api from being misused by an unauthenticated request. (note, we may want to switch other apis to a similar interaction pattern)

I also opted to ignore the scenario where the `fcm_token` could change during runtime. Judging from https://firebase.google.com/docs/cloud-messaging/manage-tokens the freshness threshold is two months, so I guess we should be good given that the `fcm_token` will be updated on every start of the app.

Lastly I adapted the user email insert and fcm token update to upserts to deal with a potential conflict that either update could come first. This way we ensure we will not loose the email nor the `fcm_token` in any race condition.

fixes #1348

### DIFF
--- a/coordinator/src/db/user.rs
+++ b/coordinator/src/db/user.rs
@@ -1,8 +1,9 @@
+use crate::schema;
 use crate::schema::users;
 use anyhow::bail;
 use anyhow::Result;
+use bitcoin::secp256k1::PublicKey;
 use coordinator_commons::RegisterParams;
-use coordinator_commons::TokenUpdateParams;
 use diesel::prelude::*;
 use serde::Deserialize;
 use serde::Serialize;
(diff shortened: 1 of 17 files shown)


coordinator/src/db/user.rs
```diff
@@ -45,27 +46,54 @@ pub fn by_id(conn: &mut PgConnection, id: String) -> QueryResult<Option<User>> {
...
-        .values(&user)
+        .values(User {
+            id: None,
+            pubkey: trader_id.to_string(),
+            email: email.clone(),
+            nostr: "".to_owned(),
+            timestamp: OffsetDateTime::now_utc(),
+            fcm_token: "".to_owned(),
+        })
+        .on_conflict(schema::users::pubkey)
```
bonomat, line 63: Nice!

coordinator/src/db/user.rs
```diff
@@ -45,27 +46,54 @@ pub fn by_id(conn: &mut PgConnection, id: String) -> QueryResult<Option<User>> {
...
-    let affected_rows = diesel::update(users::table)
-        .filter(users::pubkey.eq(&pubkey))
-        .set(users::fcm_token.eq(fcm_token))
+pub fn upsert_fcm_token(
+    conn: &mut PgConnection,
+    trader_id: PublicKey,
+    token: String,
+) -> Result<()> {
+    tracing::debug!(%trader_id, "Updating token for client.");
+    // TODO(holzeis): We should store a last login or freshness timestamp of the token into the
```
bonomat, line 76: Good idea. That way we get an easy overview of when users logged in the last time.

coordinator/src/db/user.rs
```diff
@@ -45,27 +46,54 @@ pub fn by_id(conn: &mut PgConnection, id: String) -> QueryResult<Option<User>> {
...
+    // database. The recommendation is to assume any token not being updated after 2 months to be
+    // stale.
+    let affected_rows = diesel::insert_into(users::table)
+        .values(User {
+            id: None,
+            pubkey: trader_id.to_string(),
+            email: "".to_owned(),
```
bonomat, line 83: Are you overwriting the email with an empty field ?

bonomat: Answering my own question: No! :D

holzeis: It would run into a conflict and update only the `fcm_token` 😉

coordinator/src/db/user.rs
```diff
@@ -45,27 +46,54 @@ pub fn by_id(conn: &mut PgConnection, id: String) -> QueryResult<Option<User>> {
...
-
-    let affected_rows = diesel::update(users::table)
-        .filter(users::pubkey.eq(&pubkey))
-        .set(users::fcm_token.eq(fcm_token))
+pub fn upsert_fcm_token(
+    conn: &mut PgConnection,
+    trader_id: PublicKey,
+    token: String,
+) -> Result<()> {
+    tracing::debug!(%trader_id, "Updating token for client.");
```
bonomat, line 75: Would it be handy to see the token here as well? Or did you not add it due to privacy concerns in the logs?

```suggestion
     tracing::debug!(%trader_id, %token, "Updating token for client.");
``

bonomat (approved):
Looks good. Thanks.

coordinator/src/db/user.rs
holzeis: Haha, I also added it for testing 😅. I don't know if there are any privacy concerns. I left it, because it wasn't added before either.

Happy to add it though.

PR #1356, opened 2023-09-22
fix: Send push notification for expired position if user is not online

holzeis:
Before we were only sending a push notification if the message was not successfully send via the channel. Which is basically never the case.

### DIFF
--- a/coordinator/src/message.rs
+++ b/coordinator/src/message.rs
@@ -1,5 +1,12 @@
+use crate::db::user;
+use crate::notifications::FcmToken;
+use crate::notifications::Notification;
+use crate::notifications::NotificationKind;
 use anyhow::Result;
 use bitcoin::secp256k1::PublicKey;
+use diesel::r2d2::ConnectionManager;
+use diesel::r2d2::Pool;
+use diesel::PgConnection;
 use futures::future::RemoteHandle;
 use futures::FutureExt;
 use orderbook_commons::Message;
@@ -18,6 +25,7 @@ pub enum OrderbookMessage {
     TraderMessage {
         trader_id: PublicKey,
         message: Message,
+        notification: Option<NotificationKind>,
     },
 }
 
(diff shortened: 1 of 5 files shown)


coordinator/src/message.rs
```diff
@@ -59,9 +70,29 @@ pub fn spawn_delivering_messages_to_authenticated_users(
                             Some(sender) => {
                                 if let Err(e) = sender.send(message).await {
                                     tracing::warn!("Connection lost to trader {e:#}");
+                                } else {
+                                    // Skip optional push notifications if the user was successfully notified via the websocket.
```
bonomat, line 74: We could add log message on `trace` here.

coordinator/src/message.rs
```diff
@@ -59,9 +70,29 @@ pub fn spawn_delivering_messages_to_authenticated_users(
...
+                        if let Some(notification_kind) = notification {
+                            if let Some(user) = user::by_id(&mut conn, trader_id.to_string())? {
+                                tracing::debug!(%trader_id, "Sending push notification to user");
+
+                                if let Ok(fcm_token) = FcmToken::new(user.fcm_token) {
+                                    if let Err(e) = notification_sender
+                                        .send(Notification {
+                                            user_fcm_token: fcm_token,
+                                            notification_kind,
+                                        })
```
bonomat: I've tried to simplify the logic a bit and add another error message, feel free to drop my commit: f43fb608db145c3974c7e1b1ba0abdb25e82ae90

coordinator/src/orderbook/trading.rs
```diff
@@ -224,29 +217,22 @@ async fn process_new_order(
                 },
             };
 
-            let msg = OrderbookMessage::TraderMessage { trader_id, message };
+            let notification = match &order.order_reason {
+                OrderReason::Expired => Some(NotificationKind::PositionExpired),
+                OrderReason::Manual => None,
```
bonomat, line 222: What's a notification of `None`?

bonomat (approved):
Thanks. One question and I tried to simplify the logic a bit. Feel free to drop my commit if you disagree.

coordinator/src/message.rs
holzeis: I thought about simplifying that, but I guess the best solution would be to get rid of the FcmToken wrapper, not sure why this was added in the first place.

Happy to squash in your commit though 😉

coordinator/src/orderbook/trading.rs
holzeis: Maybe too implicit? But the idea is if you don't set a notification Kind it means there should be not push notification sent if the user is not online.

The idea is to have optional notification kinds that will only get sent if the user is offline.

bonomat: Right, that was the other code I've refactored. Thank you for the clarification. Can we add this to the message so that it's clear what behavior to expect?

PR #1358, opened 2023-09-22
Fund maker and open channel

bonomat:
This comes handy if LND doesn't work as expected.

You can then use `curl -X POST http://localhost:18000/api/pay-invoice/putinvoicehere` to pay any invoices and fund the app. 

Or you can use the new `fund with maker button` :) 

image

### DIFF
--- a/crates/tests-e2e/examples/fund.rs
+++ b/crates/tests-e2e/examples/fund.rs
@@ -11,6 +11,7 @@ use serde::Deserialize;
 use std::time::Duration;
 use tests_e2e::coordinator::Coordinator;
 use tests_e2e::http::init_reqwest;
+use tests_e2e::maker::Maker;
 use tracing::metadata::LevelFilter;
 use tracing_subscriber::filter::Directive;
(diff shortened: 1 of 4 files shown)


holzeis (approved):
Makes sense, I am not happy to see that lnd just doesn't want to work reliably on regtest, but I guess that's the way to go to unblock us in many situations.

luckysori:
Might be nice to add another button on regtest builds. And maybe to fund the maker with `just fund` too.

holzeis:
> This comes handy if LND doesn't work as expected.
> 
> 
> 
> You can then use `curl -X POST http://localhost:18000/api/pay-invoice/putinvoicehere` to pay any invoices and fund the app. 

As far as I can tell the maker is funded, otherwise it couldn't open a channel to the coordinator.

I agree on the button although what do we keep lnd then around for? 🙈

bonomat:
> I agree on the button although what do we keep lnd then around for? 🙈

I opted for keeping it around ;)

PR #1362, opened 2023-09-23
fix: Wait for firebase to initialise before geting the fcm token

holzeis:
It can happen that we run into a race condition where we are trying to get a firebase token but firebase is not yet initialised. Awaiting the firebase initialisation should fix the issue.


bonomat (approved):
Let’s go

PR #1365, opened 2023-09-25
feat: Add close button to dialog if it remains in pending for more than 30 seconds

### LINKED ISSUE #1108: `OrderSubmissionStatusDialog` can hang forever

holzeis:
fixes #1108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Sync position with dlc channel state
 - Extend coordinator's `/api/version` with commit hash and number
 - When pulling down wallet sync it waits until the sync is finished
+- Add close button to dialog if it remains on pending for more than 30 seconds
 
 ## [1.2.6] - 2023-09-06
 
--- a/mobile/lib/common/recover_dlc_change_notifier.dart
+++ b/mobile/lib/common/recover_dlc_change_notifier.dart
@@ -32,12 +32,6 @@ class RecoverDlcChangeNotifier extends ChangeNotifier implements Subscriber {
             return TaskStatusDialog(title: "Catching up!", status: status, content: content);
           },
         );
-
-        // setting the task status to failed after a timeout of 30 seconds.
-        Future.delayed(const Duration(seconds: 30), () {
-          taskStatus = TaskStatus.failed;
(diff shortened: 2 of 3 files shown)


bonomat (commented):
Am I understanding it right, that the dialog will simply show `failed` after `30` seconds?
If so, what about the task in the backend? It might still succeed nevertheless .

holzeis:
> Am I understanding it right, that the dialog will simply show `failed` after `30` seconds? If so, what about the task in the backend? It might still succeed nevertheless .

If it succeeds, it succeeds and we are happy. It still took a very long time and the user should be free to close the dialog after some time.

Do you prefer a dedicated icon / status for timeout? I was think to also add the share log button if we run into a dialog. So the user can directly report an issue.

Alternatively we could as well simply show a close button while we keep the pending animation.

bonomat:
> If it succeeds, it succeeds and we are happy. It still took a very long time and the user should be free to close the dialog after some time.

Just to clarify: it will not show as failed, but the user can just close the dialog instead?

holzeis:
> > If it succeeds, it succeeds and we are happy. It still took a very long time and the user should be free to close the dialog after some time.
> 
> Just to clarify: it will not show as failed, but the user can just close the dialog instead?

If the task succeeds before the timeout, it will not show failed if thats what you are concerned about. The timer gets cancelled if we receive a `TaskStatus` other than `Pending`.

bonomat:
> > > If it succeeds, it succeeds and we are happy. It still took a very long time and the user should be free to close the dialog after some time.
> > 
> > 
> > Just to clarify: it will not show as failed, but the user can just close the dialog instead?
> 
> If the task succeeds before the timeout, it will not show failed if thats what you are concerned about. The timer gets cancelled if we receive a `TaskStatus` other than `Pending`.

No, I'm concerned about the situation where it succeeds after the timeout.

holzeis:
> No, I'm concerned about the situation where it succeeds after the timeout.

Any suggestion on how you would deal with a timeout then?

bonomat (approved):
Looks good. Let's get this in.

Issue #1259, opened 2023-09-08
Integrate in-app waiting list


bonomat:
We want to be able to control who can use our app (meaning, who can acquire in-bound liquidity)
1. With the end of this month, we will move away from ViralLoops stopping us from receiving more sign-ups
2. We want to launch Android which can be heavily simplified by simply publishing an APK
3. It would be cool to simply let users download our app and only get access once there is space - this is also in preparation for our public launch

To do this, we want an in-app waiting list. 
A simple version could look like this: 

1. user starts the app and see the well-known `enter email` field.
2. user enters his email address
  1. if user is on the waiting list and enabled for beta, the user can proceed
  2. if the user is not on the waiting list, we add him to the waiting list
  3. if the user is on the waiting list but not yet enabled for beta, the user cannot proceed (or may not fund the wallet etc.)

Ideally we have an email verification process which enforces real emails aswell as ensures that the user is in control of the email, e.g. we could send a magic link to the user which he has to click to open the app. 
Otherwise we have two challenges: 

1. a user could enter an invalid email defeating the purpose of having an email in the first place
2. a user could share his email or use one of someone else to enter the app

klochowicz:
@bonomat how about adding one more table into our database?
We could prepopulate it it off by copying all the emails from the current users table, with a POST request to add new email to the beta users.

bonomat:
Yes, I'd imagine a simple table like this: 

<table>
<tr>
 <td>email
 <td>is_enabled_for_beta
 <td>timestamp
</table>

and then add all existing users with `is_enabled_for_beta=true` and all our waiting list users with `is_enabled_for_beta=false`

holzeis:
@bonomat how will the flag be displayed in the app if the user is not enabled?

bonomat:
Do you want me to give you a design or a rough idea? 

The flow idea is mentioned in the ticket and it could be a simple message as: 

```
Sorry mate, you are not part of the beta yet, come back tomorrow.
```

Note: this ticket was created before our meeting yesterday and I hope @klochowicz took that into account.

holzeis:
A rough idea should be enough.

I was wondering if the user would be blocked out directly at the welcome screen or can browse around but not fund the wallet.

klochowicz:
> Note: this ticket was created before our meeting yesterday and I hope @klochowicz took that into account.

Could you be more clear please? Does this change anything in terms of scope of the ticket or whether we'd want to do it or not?

[...]

PR #1374, opened 2023-09-26
Add in-app waiting list

klochowicz:
Check access to the waitlist during in-app registration and prevent from using
the app if they're not allowed.

- If the user is not on the waitlist, add them and display a message that they
  do not have access yet, pointing to our Telegram as well.
- if the user is on the waitlist, check whether they're allowed to access.

Only after successful admission store the email and allow

Add a coordinator setting to allow auto-adding new users (to allow throttling
users during open beta, allowing us to stay within our liquidity constraints).

Add an API to the coordinator to query the waitlist and to add a user to the
waitlist (with options to enable/disable access).

The waitlist is stored as a `waitlist` table in coordinator database, initially
populated from the `users` table to ensure all our current users have immediate access.

Note: I didn't try to go too fancy initially. In the futurue, we could e.g.
     unsubscribe from the orderbook stream if we just want to disable the
     trading functionality.

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -1,4 +1,6 @@
 use crate::db;
+use crate::db::waitlist;
+use crate::db::waitlist::WaitlistEntry;
 use crate::routes::AppState;
 use crate::AppError;
 use anyhow::Context;
@@ -300,4 +302,48 @@ pub async fn is_connected(
         AppError::BadRequest(format!("Invalid public key {target_pubkey}. Error: {err}"))
     })?;
     Ok(Json(state.node.is_connected(&target)))
+}
+
+#[derive(Deserialize)]
+pub struct AddWaitlistRequest {
+    email: String,
+    /// Is the user allowed to access the app. Defaults to true. User access can
+    /// be revoked at a later point.
+    allowed: Option<bool>,
+}
+
+pub async fn add_to_waitlist(
+    State(state): State<Arc<AppState>>,
+    Json(body): Json<AddWaitlistRequest>,
+) -> Result<(), AppError> {
+    let mut conn =
+        state.pool.clone().get().map_err(|e| {
+            AppError::InternalServerError(format!("Failed to acquire db lock: {e:#}"))
+        })?;
+
+    // We allow access by default, because the sole fact of putting an email in
+    // the app signs up the user to the waitlist.
+    let new_entry = WaitlistEntry::new(body.email, body.allowed.unwrap_or(true));
+    tracing::debug!(?new_entry, "Adding user to waitlist");
+
+    waitlist::upsert(&mut conn, new_entry).map_err(|e| {
+        AppError::InternalServerError(format!("Failed to insert into waitlist: {e:#}"))
+    })?;
+
+    Ok(())
+}
+
+/// Return the whole waitlist
+pub async fn get_waitlist(
+    State(state): State<Arc<AppState>>,
+) -> Result<Json<Vec<WaitlistEntry>>, AppError> {
+    let mut conn =
+        state.pool.clone().get().map_err(|e| {
+            AppError::InternalServerError(format!("Failed to acquire db lock: {e:#}"))
+        })?;
+
+    let waitlist = waitlist::all(&mut conn)
+        .map_err(|e| AppError::InternalServerError(format!("Failed to get waitlist: {e:#}")))?;
+
+    Ok(Json(waitlist))
 }
--- a/coordinator/src/db/waitlist.rs
+++ b/coordinator/src/db/waitlist.rs
@@ -1,0 +1,64 @@
+use crate::schema::waitlist;
+use diesel::prelude::*;
+use serde::Serialize;
+use time::OffsetDateTime;
+
+#[derive(Queryable, Insertable, Identifiable, Debug, PartialEq, Clone, Serialize)]
+#[diesel(table_name = waitlist)]
+#[diesel(primary_key(email))]
+pub struct WaitlistEntry {
+    pub email: String,
+    /// Timestamp when the user was added to the waitlist
+    pub created_timestamp: OffsetDateTime,
+    /// Is the user allowed to access the app
+    pub allowed: bool,
+    /// Timestamp when the user was allowed access
+    pub allowed_timestamp: Option<OffsetDateTime>,
+}
+
+impl WaitlistEntry {
+    pub fn new(email: String, allowed: bool) -> Self {
+        let allowed_timestamp = if allowed {
+            Some(OffsetDateTime::now_utc())
+        } else {
+            None
+        };
+
+        Self {
+            email,
+            created_timestamp: OffsetDateTime::now_utc(),
+            allowed,
+            allowed_timestamp,
+        }
+    }
+
+    pub fn allow(&mut self) {
+        self.allowed = true;
+        self.allowed_timestamp = Some(OffsetDateTime::now_utc());
+    }
+}
+
+/// Insert or modify a waitlist entry
+pub fn upsert(conn: &mut PgConnection, entry: WaitlistEntry) -> QueryResult<WaitlistEntry> {
+    diesel::insert_into(waitlist::table)
+        .values(entry.clone())
+        .on_conflict(waitlist::email)
+        .do_update()
+        .set((
+            waitlist::allowed.eq(entry.allowed),
+            waitlist::allowed_timestamp.eq(entry.allowed_timestamp),
+        ))
+        .get_result(conn)
+}
+
+pub fn with_email(conn: &mut PgConnection, email: &str) -> QueryResult<Option<WaitlistEntry>> {
+    Ok(waitlist::table
+        .filter(waitlist::email.eq(email))
+        .load::<WaitlistEntry>(conn)?
+        .into_iter()
+        .next())
+}
+
+pub fn all(conn: &mut PgConnection) -> QueryResult<Vec<WaitlistEntry>> {
+    waitlist::table.load::<WaitlistEntry>(conn)
+}
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -1,6 +1,8 @@
+use crate::admin::add_to_waitlist;
 use crate::admin::close_channel;
 use crate::admin::connect_to_peer;
 use crate::admin::get_balance;
+use crate::admin::get_waitlist;
 use crate::admin::is_connected;
 use crate::admin::list_channels;
 use crate::admin::list_dlc_channels;
@@ -10,6 +12,8 @@ use crate::admin::open_channel;
 use crate::admin::send_payment;
 use crate::admin::sign_message;
 use crate::db::user;
+use crate::db::waitlist;
+use crate::db::waitlist::WaitlistEntry;
 use crate::message::NewUserMessage;
 use crate::node::Node;
 use crate::orderbook::routes::get_order;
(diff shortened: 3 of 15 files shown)


coordinator/src/admin.rs
```diff
@@ -301,3 +303,47 @@ pub async fn is_connected(
...
+        AppError::InternalServerError(format!("Failed to insert into waitlist: {e:#}"))
+    })?;
+
+    Ok(())
+}
+
+/// Return the whole waitlist
+pub async fn get_waitlist(
+    State(state): State<Arc<AppState>>,
+) -> Result<Json<Vec<WaitlistEntry>>, AppError> {
```
bonomat, line 339: Nit: we probably should introduce an own type for the API and not leak the db model here.

coordinator/src/routes.rs
```diff
@@ -138,6 +142,8 @@ pub fn router(
             get(get_settings).put(update_settings),
         )
         .route("/api/admin/sync", post(post_sync))
+        .route("/api/admin/waitlist/add", post(add_to_waitlist))
```
bonomat, line 145: There is no need to add `add`. Just put both, the `post` and `get` onto `/api/admin/waitlist`

coordinator/src/db/waitlist.rs
```diff
@@ -0,0 +1,64 @@
...
+        .on_conflict(waitlist::email)
+        .do_update()
+        .set((
+            waitlist::allowed.eq(entry.allowed),
+            waitlist::allowed_timestamp.eq(entry.allowed_timestamp),
+        ))
+        .get_result(conn)
+}
+
+pub fn with_email(conn: &mut PgConnection, email: &str) -> QueryResult<Option<WaitlistEntry>> {
```
bonomat, line 54: Nit: `by_email` might be more suitable because you are filtering by `email`.

coordinator/src/routes.rs
```diff
@@ -324,16 +330,37 @@ pub async fn post_register(
...
         .get()
         .map_err(|e| AppError::InternalServerError(format!("Could not get connection: {e:#}")))?;
 
-    if let Some(email) = register_params.email {
+    if let Some(email) = register_params.email.clone() {
+        let waitlist_entry = if let Some(waitlist_entry) = waitlist::with_email(&mut conn, &email)
+            .map_err(|e| {
+            AppError::InternalServerError(format!("could not retrieve waitlist: {e:#}"))
+        })? {
+            tracing::info!("User is on the waitlist");
```
bonomat, line 344: This log is not that useful unless you say which user is in the waitlist. Might be a debugging left over to see if the API works?

coordinator/src/routes.rs
```diff
@@ -324,16 +330,37 @@ pub async fn post_register(
...
+            waitlist_entry
+        } else {
+            tracing::info!("Signing up user onto the waitlist");
```
bonomat, line 347: Same as above.

coordinator/src/routes.rs
```diff
@@ -324,16 +330,37 @@ pub async fn post_register(
...
+
+            let allowed = state.settings.read().await.auto_allow_new_users;
+            waitlist::upsert(&mut conn, WaitlistEntry::new(email.clone(), allowed)).map_err(
+                |e| AppError::InternalServerError(format!("could not insert into waitlist: {e:#}")),
+            )?
+        };
+
+        if !waitlist_entry.allowed {
+            return Err(AppError::BadRequest(
```
bonomat, line 356: Should we log here that a user tried to access but access was denied?

coordinator/src/routes.rs
```diff
@@ -324,16 +330,37 @@ pub async fn post_register(
...
-    tracing::info!(?register_params, "Registered new user");
 
     let mut conn = state
         .pool
         .get()
         .map_err(|e| AppError::InternalServerError(format!("Could not get connection: {e:#}")))?;
 
-    if let Some(email) = register_params.email {
+    if let Some(email) = register_params.email.clone() {
+        let waitlist_entry = if let Some(waitlist_entry) = waitlist::with_email(&mut conn, &email)
```
bonomat, line 340: What's the reason you are using `if let some ... else` over `match`?

coordinator/src/settings.rs
```diff
@@ -15,6 +15,9 @@ const SETTINGS_FILE_NAME: &str = "coordinator-settings.toml";
 /// Top-level settings.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Settings {
+    /// If set to true, new users are allow to trade on the platform without
+    /// previously being on the waitlist.
+    pub auto_allow_new_users: bool,
```
bonomat, line 20: Should we add a cli arg to overwrite this? Otherwise it might get cumbersome with testing.

holzeis:
I guess I have missed a decision, but didn't we had an open discussion in one of our last calls, that we don't need a waiting list, if we simply let users fail to fund their app if we do not have enough liquidity?

mobile/lib/common/settings_screen.dart
```diff
@@ -72,6 +74,13 @@ class _SettingsScreenState extends State<SettingsScreen> {
                     rust.api.forceCloseChannel();
                   },
                   child: const Text("Force-close channel")),
+              ElevatedButton(
+                  onPressed: () async {
+                    final messenger = ScaffoldMessenger.of(context);
+                    await Preferences.instance.clear();
+                    showSnackBar(messenger, "Shared preferences cleared");
+                  },
+                  child: const Text("Clear shared preferences")),
```
bonomat, line 83: What's `shared preferences`?

bonomat (commented):
Thank you. A couple of nits, nothing critical.
Only one main point as also mentioned in the ticket: 

I wonder if this feature is useful if we don't solve these two problems: 

> Otherwise we have two challenges:
> - a user could enter an invalid email defeating the purpose of having an email in the first place
> - a user could share his email or use one of someone else to enter the app

bonomat:
> I guess I have missed a decision, but didn't we had an open discussion in one of our last calls, that we don't need a waiting list, if we simply let users fail to fund their app if we do not have enough liquidity?

You stand correctly and you did not miss this [decision](https://github.com/get10101/10101/issues/1259#issuecomment-1731869460).

holzeis:
> > I guess I have missed a decision, but didn't we had an open discussion in one of our last calls, that we don't need a waiting list, if we simply let users fail to fund their app if we do not have enough liquidity?
> 
> You stand correctly and you did not miss this [decision](https://github.com/get10101/10101/issues/1259#issuecomment-1731869460).

Ok, thanks for referencing the ticket. So that's why I was confused of that commit, I thought there had been another decision that I have missed 🙈 

> Yes, it would change the scope of this ticket. In the planning call we said that we don't want an in-app waiting list but simply tell the user that there is not enough liquidity available at the moment and they should try again later.

I am still in favor of not having to introduce an in-app waiting list, but rather fail the user if they try to fund the wallet if we do not have enough liquidity. I think it's a simpler solution.

klochowicz:
> > > I guess I have missed a decision, but didn't we had an open discussion in one of our last calls, that we don't need a waiting list, if we simply let users fail to fund their app if we do not have enough liquidity?
> > 
> > 
> > You stand correctly and you did not miss this [decision](https://github.com/get10101/10101/issues/1259#issuecomment-1731869460).
> 
> Ok, thanks for referencing the comment. So that's why I was confused of that PR, I thought there had been another decision that I have missed 🙈
> 
> > Yes, it would change the scope of this ticket. In the planning call we said that we don't want an in-app waiting list but simply tell the user that there is not enough liquidity available at the moment and they should try again later.
> 
> I am still in favor of not having to introduce an in-app waiting list, but rather fail the user if they try to fund the wallet if we do not have enough liquidity. I think it's a simpler solution.

I find the current PR in-app waiting list a much better solution, because of the following reasons:
- we can gauge the interest in the app, even when we don't have the liquidity
- we can send an email to people when we *add* liquidity. I strongly suspect that if we fail to notify them proactively they will not be converted into users.
- we can *control* access to the beta (e.g. keep the waitlist if we'd like to, by whitelisting emails if we wish so. This solves the old problem of people registering with different email than they were given access via TestFlight)

I have made amendments into the scope of the ticket to ensure we can just toggle *all-in* approach if we deem we have sufficient liquidity, which was to me the only appeal of not having an in-app waiting list (ie. making it *simpler* to open it up further if we want).  Now we have best of both worlds, allowing flexibility. If we change the coordinator setting we can literally **not have** waiting list, at least perceived one, at an instant. 

The code is simple enough to also not to worry about the added technical complexity here. 

I guess the bottom line is that it was scheduled into the iteration, and the name of the ticket literally meant "Add in-app waiting list", so I am puzzled why there is pushback here 🤔. If it was deemed unneeded, the correct solution would have to close the ticket.

holzeis:
> we can gauge the interest in the app, even when we don't have the liquidity

I don't see why we would need a waiting list for that as a new user needs to enter their email address anyways when they install the app. Our users table would serve the same purpose, doesn't it?
 
> we can send an email to people when we _add_ liquidity.

We can do that without the waiting list. We have the users email in the users table + we know what users have an announced channel. The only thing missing (that I would add) is updating the announced channel on a rejected creation if we do not have enough liquidity. Then we could simply query all users with a "rejected" channel and send them a notification / email when we added new liquidity. Although we might not even have to target those users directly.

> I strongly suspect that if we fail to notify them proactively they will not be converted into users.

I kind of doubt that, but I guess if that's true it would mean a more complex workflow + potentially more unused liquidity.

e.g. To ensure every user with access will not run into a liquidity issue we will have to reserve the maximum channel capacity amount for every user without a channel. We've already seen that not every user opens a channel (at least not right away) so that would already be unused liquidity. Now following that thought, we could 
- remove their access again and replace those users with somebody else from the waiting list. 
- grant another one access and risk that the other user may run into the liquidity issue? (but that wouldn't address your initial concern)

Now this example is very hypothetical as I don't think we actually run into a scenario where we have a fixed liquidity where we are slowing matching the users to exactly match that liquidity. I think the way it will turn out is that we will have to monitor the coordinators liquidity and add more funds if we see we are running out of liquidity.

Bottom line, I dislike having to deal with a waiting list and manually manage who gets access. Granted, auto allowing would fix that, but why having that feature then in the first place?

> we can _control_ access to the beta (e.g. keep the waitlist if we'd like to, by whitelisting emails if we wish so.

Yes, we can _control_ access, but if we do that you can't call it a public beta and if we are not going public we can stay with our TestFlight and GooglePlay invites to _control_ access.

> This solves the old problem of people registering with different email than they were given access via TestFlight)

I guess when we are going public, there won't be a dedicated invite per email anymore. We would be in the regular android stores + apk and publish a public TestFlight link without limit. So we end up with whatever email they enter when installing the app.

> I have made amendments into the scope of the ticket to ensure we can just toggle all-in approach if we deem we have sufficient liquidity, which was to me the only appeal of not having an in-app waiting list (ie. making it simpler to open it up further if we want). Now we have best of both worlds, allowing flexibility. If we change the coordinator setting we can literally not have waiting list, at least perceived one, at an instant.
>
> The code is simple enough to also not to worry about the added technical complexity here.

The auto allow would simply render that feature unused and why would we add code that we know we are not going to use. Also, the proposed solution would still not inform the user properly that the coordinator currently doesn't have enough liquidity to onboard this new user. To be more concrete on what I would have liked to see in this PR is the following.

When the user registers an interceptable invoice at the coordinator (for funding the app). The coordinator checks if he would have enough liquidity to open that channel. If not, inform the user that we can't onboard at the moment.

> I guess the bottom line is that it was scheduled into the iteration, and the name of the ticket literally meant "Add in-app waiting list", so I am puzzled why there is pushback here 🤔.

I am aware that all of this hasn't been clearly stated on the ticket nor in the title of the ticket, however, it has been discussed in objective alignment and in the comments on the ticket as well https://github.com/get10101/10101/issues/1259#issuecomment-1731869460.

> If it was deemed unneeded, the correct solution would have to close the ticket.

It's not deemed unneeded, its just the solution design was discussed to be different.

Issue #1286, opened 2023-09-12
Design Update UI to open bigger channels


bonomat:
This ticket is about updating Figma mocks for allowing users open a bigger channel. 

See here: https://github.com/get10101/10101/issues/1280

holzeis:
![open-channel](https://github.com/get10101/10101/assets/382048/9d223d19-aa72-4efa-8ce2-374f2b10f4b5)
![Receive Funding](https://github.com/get10101/10101/assets/382048/0569fc87-cada-415f-bea0-bee90e0625b9)
![Receive Funding-1](https://github.com/get10101/10101/assets/382048/d0b41a7f-cf4b-47f3-b4ee-ede833255625)

PR #1384, opened 2023-09-27
feat: Bigger Channels

holzeis:
This PR introduces the following changes

- Splits up the onboarding and create invoice flow. The onboarding flow already aligns with the design inputs from https://github.com/get10101/10101/issues/1286#issuecomment-1733323492.
- Introduces liquidity options that can be configured in the database
- Configurable leverage based on liquidity option.
- Adds the coordinator_leverage to the coordinator position
- Formats amounts in input fields
- Splits up creating an interceptable invoice (for onboarding without scid) and a regular invoice (short channel id is known)

Everything is configurable so we can decide on what values to use separately. However this PR proposes the following configurations.

<table>
<tr>
 <td>Rank</td>
 <td>Title</td>
 <td>Trade up to</td>
 <td>Max Deposit</td>
 <td>Min Deposit</td>
 <td>Min Fee</td>
 <td>Min Percentage</td>
 <td>Coordinator Leverage*</td>
<tr>
 <td>1</td>
 <td>Large</td>
 <td>3,000,000</td>
 <td>3,000,000</td>
 <td>750,000</td>
 <td>10,000</td>
 <td>1%</td>
 <td>2.0</td>
<tr>
 <td>2</td>
 <td>Medium</td>
 <td>1,500,000</td>
 <td>1,500,000</td>
 <td>375,000</td>
 <td>10,000</td>
 <td>1%</td>
 <td>2.0</td>
<tr>
 <td>3</td>
 <td>Small</td>
 <td>200,000</td>
 <td>200,000</td>
 <td>50,000</td>
 <td>10,000</td>
 <td>1%</td>
 <td>2.0</td>
</table>
* Coordinator leverage impacts the amount that the cooridnator has to bring in to allow for the `trade_up_to` value. We decided to go for a higher leverage to allow both a higher trading volume + less locked up capital on the coordinator side.

------

https://github.com/get10101/10101/assets/382048/2219329a-b759-4c89-82ed-f5977dd129f6

TODO
- [x] Fetch configuration data (liquidity options, min and max amounts) from coordinator.
- [x] Adapt channel capacity policy for the selected liquidity option.
- [x] Adapt open channel fee calculation.
- [x] Create a regular invoice when not onboarding.
- [x] Align tests


coordinator/migrations/2023-09-28-163500_liquidity_options/up.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE "liquidity_options" (
+       id SERIAL PRIMARY KEY NOT NULL,
+       rank SMALLINT NOT NULL,
+       title TEXT NOT NULL,
+       trade_up_to BIGINT NOT NULL,
```
bonomat, line 5: Might be worth adding `_sats` or a comment so that we know the unit of those fields.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -10,30 +12,33 @@ use crate::node::Node;
...
+use tokio::task::block_in_place;
 
 /// Event handler for the coordinator node.
 // TODO: Move it out of this crate
 pub struct CoordinatorEventHandler<S> {
     pub(crate) node: Arc<Node<S>>,
     pub(crate) pending_intercepted_htlcs: PendingInterceptedHtlcs,
     pub(crate) event_sender: Option<EventSender>,
-    pub(crate) max_channel_size_sats: u64,
+    pub(crate) max_app_fund_amount: u64,
```
bonomat, line 41: I'd like to have the unit of this in the name or at least as a comment.

coordinator/src/db/liquidity_options.rs
```diff
@@ -0,0 +1,66 @@
...
+use diesel::RunQueryDsl;
+use time::OffsetDateTime;
+
+#[derive(Queryable, Debug, Clone, PartialEq)]
+#[diesel(table_name = liquidity_options)]
+pub(crate) struct LiquidityOption {
+    pub id: i32,
+    pub rank: i16,
+    pub title: String,
+    // amount the user can trade up to in sats
```
bonomat, line 16:
```suggestion
/// amount the user can trade up to in sats
```

coordinator/src/node.rs
```diff
@@ -213,7 +210,29 @@ impl Node {
...
+                    db::channels::get(&user_channel_id, connection)?.with_context(|| {
+                        format!(
+                            "Couldnt find shadow channel. trader_id={}, user_channel_id={}",
+                            trade_params.pubkey, channel_details.user_channel_id
+                        )
+                    })?;
+
+                let liquidity_option =
+                    db::liquidity_options::get(connection, channel.liquidity_option_id)?;
+                let coordinator_leverage = liquidity_option.coordinator_leverage;
```
bonomat, line 227: Alternatively you could do a `join` on db level and spare one db request.

coordinator/src/node.rs
```diff
@@ -206,14 +203,22 @@ impl Node {
         let order_id = trade_params.filled_with.order_id.to_string();
         tracing::info!(trader_id, order_id, "Executing match");
 
+        //TODO(holzeis): The coordinator leverage should come from the selected liquidity option.
+        let coordinator_leverage = 2.0;
```
luckysori: :thinking: Surely it is leverage 1?! Or have we decided to change this already? We should probably change it in a dedicated patch if so.

coordinator/src/node.rs
```diff
@@ -240,17 +251,27 @@ impl Node {
...
         trade_params: &TradeParams,
         fee_payment_hash: PaymentHash,
+        coordinator_leverage: f32,
     ) -> Result<()> {
         let peer_id = trade_params.pubkey;
         tracing::info!(%peer_id, ?trade_params, "Opening position");
 
         let margin_trader = margin_trader(trade_params);
-        let margin_coordinator = margin_coordinator(trade_params);
+        let margin_coordinator = margin_coordinator(trade_params, coordinator_leverage);
```
luckysori, line 277: :wrench: Shouldn't the coordinator leverage be added to the trade params then?

coordinator/src/node/unrealized_pnl.rs
```diff
@@ -32,7 +34,7 @@ fn sync_position(
     position: &Position,
     quote: Quote,
 ) -> Result<()> {
-    let pnl = position.calculate_coordinator_pnl(quote)?;
+    let pnl = position.calculate_coordinator_pnl(quote, position.coordinator_leverage)?;
```
luckysori, line 37: :thinking: It's already in the `Position` that you are calling the method on! No need to pass it from the outside.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -99,12 +102,37 @@ where
...
+
+                let user_channel_id = match channel {
+                    Some(mut channel) => {
+                        channel.channel_state = ChannelState::Pending;
+
+                        if let Err(e) = self.node.storage.upsert_channel(channel.clone()) {
+                            tracing::error!("Failed to update channel. Error: {e:#}");
+                        }
+
+                        channel.user_channel_id
```
bonomat, line 124: Nit: you could `clone` the `channel.user_channel_id` before and then _move_ the `channel` into `upsert_channel` function. This might be more efficient resource wise.

coordinator/src/position/models.rs
```diff
@@ -61,6 +60,7 @@ pub struct Position {
     pub expiry_timestamp: OffsetDateTime,
     pub update_timestamp: OffsetDateTime,
     pub trader: PublicKey,
+    pub coordinator_leverage: f32,
```
luckysori, line 63: We should probably reorganise and re-document this struct to reflect the fact that it is no longer just the trader's position.

coordinator/src/position/models.rs
```diff
@@ -80,7 +80,11 @@ impl Position {
     }
 
     /// Calculates the profit and loss for the coordinator in satoshis
-    pub fn calculate_coordinator_pnl(&self, quote: Quote) -> Result<i64> {
+    pub fn calculate_coordinator_pnl(
+        &self,
+        quote: Quote,
+        coordinator_leverage: f32,
```
luckysori, line 86: :thinking: It's in `Self` already.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -172,13 +200,27 @@ where
...
-                    &self.node,
-                    &self.pending_intercepted_htlcs,
+                let user_channel_id = UserChannelId::from(user_channel_id).to_string();
+
+                tracing::info!(
                     user_channel_id,
-                    channel_id,
-                    counterparty_node_id,
-                )?;
+                    channel_id = %channel_id.to_hex(),
```
bonomat, line 207: I think it doesn't need `%` here because it's already a string (same the line below).

coordinator/src/position/models.rs
```diff
@@ -112,11 +116,15 @@ impl Position {
         Ok(pnl)
     }
 
-    pub fn calculate_settlement_amount(&self, closing_price: Decimal) -> Result<u64> {
+    pub fn calculate_settlement_amount(
+        &self,
+        closing_price: Decimal,
+        coordinator_leverage: f32,
```
luckysori, line 122: :thinking: It's in `Self` already.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -354,10 +356,10 @@ where
                 .peer_manager
                 .get_peer_node_ids()
                 .iter()
-                .any(|(id, _)| *id == target_node_id)
+                .any(|(id, _)| *id == trader_id)
             {
                 tracing::info!(
-                    %target_node_id,
+                    %trader_id,
```
luckysori, line 438: :wrench: Prefer consistent naming i.e. `peer_id`.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -366,7 +368,7 @@ where
             }
 
             tracing::debug!(
-                %target_node_id,
+                %trader_id,
```
luckysori, line 447: :wrench: Prefer consistent naming i.e. `peer_id`.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -245,6 +269,59 @@ where
...
+    counterparty_node_id: PublicKey,
+) -> Result<()>
+where
+    S: Storage,
+{
+    let user_channel_id = UserChannelId::from(user_channel_id).to_string();
+
+    tracing::info!(
+        user_channel_id,
+        channel_id = %channel_id.to_hex(),
```
bonomat, line 286: Same as before: `%` is redundant.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -381,11 +383,15 @@ where
     // in mind, if the coordinator (the only party that can handle this event) has a channel
     // with the target of this payment we must treat this interceptable HTLC as a regular
     // payment.
+    //
+    // Note, that scenario should not be happening, as we only create an fake scid when creating an
+    // onboarding channel. Any other invoice should create the invoice hint with the actual short
+    // channel id.
```
luckysori, line 465: :wrench: I agree with your changes and the comment, but I think this just means we need to remove this lengthy comment.

We can just log a warning if we hit this path instead, as it is unexpected but also not necessarily a problem.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -422,10 +428,15 @@ where
...
-    let channel_value = expected_outbound_amount_msat / 1000 * crate::LIQUIDITY_MULTIPLIER;
+    let expected_outbound_amount =
+        Decimal::from(expected_outbound_amount_msat) / Decimal::from(1000);
+    let trade_up_to = Decimal::from(liquidity_request.trade_up_to);
+    let coordinator_leverage =
+        Decimal::try_from(liquidity_request.coordinator_leverage).expect("to fit into decimal");
 
     let fake_scid = FakeScid::new(requested_next_hop_scid);
+    let channel_value = expected_outbound_amount + (trade_up_to / coordinator_leverage);
+    let channel_value = channel_value.to_u64().expect("to fit into u64");
```
luckysori: :wrench: It would be good to extract these mathsy bits into functions that can be documented and even tested.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -378,11 +459,15 @@ where
     // in mind, if the coordinator (the only party that can handle this event) has a channel
     // with the target of this payment we must treat this interceptable HTLC as a regular
     // payment.
+    //
+    // Note, that scenario should not be happening, as we only create an fake scid when creating an
+    // onboarding channel. Any other invoice should create the invoice hint with the actual short
```
bonomat, line 464: Isn't the face scid being used as long as the channel has 0 confirmations?

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -413,20 +504,26 @@ where
...
-    let channel_value = expected_outbound_amount_msat / 1000 * LIQUIDITY_MULTIPLIER;
-    ensure!(
-        channel_value <= max_channel_size_sats,
-        "Failed to open channel because maximum channel value exceeded"
-    );
+    let expected_outbound_amount =
+        Decimal::from(expected_outbound_amount_msat) / Decimal::from(1000);
+    let trade_up_to = Decimal::from(liquidity_request.trade_up_to);
+    let coordinator_leverage =
+        Decimal::try_from(liquidity_request.coordinator_leverage).expect("to fit into decimal");
```
bonomat, line 511: Not sure if it makes much difference, but I believe you could use `from_f32` instead of `try_from`.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -422,10 +428,15 @@ where
...
-    // TODO(holzeis): check liquidity option here
-    let channel_value = expected_outbound_amount_msat / 1000 * crate::LIQUIDITY_MULTIPLIER;
+    let expected_outbound_amount =
+        Decimal::from(expected_outbound_amount_msat) / Decimal::from(1000);
+    let trade_up_to = Decimal::from(liquidity_request.trade_up_to);
+    let coordinator_leverage =
+        Decimal::try_from(liquidity_request.coordinator_leverage).expect("to fit into decimal");
 
     let fake_scid = FakeScid::new(requested_next_hop_scid);
+    let channel_value = expected_outbound_amount + (trade_up_to / coordinator_leverage);
```
luckysori, line 513: I guess this assumes that the coordinator wants to have the same leverage for all channels. Might be a bit limiting, but we can start with this.

luckysori (commented):
LGTM, but 3 patches left to review.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -159,23 +179,67 @@ where
...
+                    .forwarding_fee_proportional_millionths,
+            },
+            cltv_expiry_delta: MIN_CLTV_EXPIRY_DELTA,
+            htlc_minimum_msat: None,
+            htlc_maximum_msat: None,
+        };
+
+        tracing::info!(
+            peer_id = %target_node,
+            route_hint_hop = ?route_hint_hop,
```
bonomat, line 240: Is there a particular reason why you want `route_hint_hop` in the logs? I assume it look terrible :)

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
```diff
@@ -302,6 +324,88 @@ pub(crate) async fn send_interceptable_payment(
     Ok(())
 }
 
+/// Sends a regular payment assuming, assuming all channels on the path exist.
```
bonomat, line 327:
```suggestion
/// Sends a regular payment assuming all channels on the path exist.
```

crates/ln-dlc-node/src/tests/just_in_time_channel/fail_intercepted_htlc.rs
```diff
@@ -93,12 +102,19 @@ async fn fail_intercepted_htlc_if_connection_lost_after_funding_tx_generated() {
     let invoice_amount = 10_000;
     setup_coordinator_payer_channel(invoice_amount, &coordinator, &payer).await;
 
+    let liquidity_request = LiquidityRequest {
```
bonomat, line 105: Given we use the same values all over the place you could add a helper function creating a default only for tests.

mobile/assets/10101_logo_icon_white_background.png
bonomat, line 1: Nit: is the logo centered? :D

mobile/lib/common/application/channel_info_service.dart
```diff
@@ -15,21 +16,24 @@ class ChannelInfoService {
...
     return Amount(maxCapacity);
   }
 
+  Future<Amount> getMaxAppFundAmount() async {
+    int maxAppFundAmount = await rust.api.maxAppFundAmount();
+    return Amount(maxAppFundAmount);
+  }
+
+  Future<List<LiquidityOption>> getLiquidityOptions(bool activeOnly) async {
+    final liquidityOptions = await rust.api.liquidityOptions();
```
bonomat, line 30: This returns a result right? Don't you have to handle the error?

mobile/native/src/api.rs
```diff
@@ -318,12 +329,56 @@ pub fn contract_tx_fee_rate() -> Result<u64> {
...
+pub fn max_app_fund_amount() -> Result<u64> {
+    ln_dlc::max_app_fund_amount().map(|amount| amount.to_sat())
+}
+
+#[derive(Debug, Clone)]
+pub struct LiquidityOption {
+    pub id: i32,
+    pub rank: i16,
+    pub title: String,
+    // the amount the trader can trade up to
```
bonomat, line 341: `///`

mobile/native/src/channel_fee.rs
```diff
@@ -23,142 +25,141 @@ const WAIT_FOR_OUTBOUND_CAPACITY_TIMEOUT: Duration = Duration::from_secs(60);
...
-                tracing::info!("Successfully triggered funding transaction fees payment of {funding_tx_fees_msats} msats to {}", config::get_coordinator_info().pubkey);
-            }
-            Err(e) => {
-                tracing::error!("Failed to pay funding transaction fees of {funding_tx_fees_msats} msats to {}. Error: {e:#}", config::get_coordinator_info().pubkey);
-            }
-        };
+    /// Attempts to pay the fees for opening an inbound channel.
+    fn pay_jit_channel_open_fee(&self, amount_msats: u64) -> Result<()> {
+        let channels = self.channel_manager.list_channels();
+        // Assuming the user ever only has one channel. Needs to be changed when we are supporting
```
bonomat, line 61: Should we add a check here if there are more channels and print a warning? Could be useful to find weird bugs in production.

mobile/native/src/channel_fee.rs
```diff
@@ -23,142 +25,141 @@ const WAIT_FOR_OUTBOUND_CAPACITY_TIMEOUT: Duration = Duration::from_secs(60);
...
+                    })?;
+
+                if channel.channel_state != ChannelState::OpenUnpaid {
+                    tracing::debug!("Channel inbound fees have already been paid. Skipping.");
+                    return Ok(());
+                }
 
-        Ok(())
-    }
+                let liquidity_options = tokio::task::block_in_place(ln_dlc::liquidity_options)?;
```
bonomat, line 77: Why does this have to be inside a `tokio::task::block_in_place`?

mobile/native/src/channel_fee.rs
```diff
@@ -23,142 +25,141 @@ const WAIT_FOR_OUTBOUND_CAPACITY_TIMEOUT: Duration = Duration::from_secs(60);
...
+                    .iter()
+                    .find(|l| l.id == channel.liquidity_option_id)
+                    .with_context(|| {
+                        format!(
+                            "Couldn't find liquidity option for id {}",
+                            channel.liquidity_option_id
+                        )
+                    })?;
+
+                let amount = Decimal::from(amount_msats) / Decimal::from(100);
```
bonomat, line 88: Why divided by `100`? Shouldn't it be `1000` since it's msats?

bonomat (approved):
It probably works™️

mobile/lib/features/wallet/onboarding/fund_wallet_modal.dart
```diff
@@ -0,0 +1,193 @@
...
+                        : QrImageView(
+                            data: widget.invoice,
+                            embeddedImage:
+                                const AssetImage('assets/10101_logo_icon_white_background.png'),
+                            embeddedImageStyle: const QrEmbeddedImageStyle(
+                              size: Size(50, 50),
+                            ),
+                            version: QrVersions.auto,
+                            padding: const EdgeInsets.all(5),
+                          ),
```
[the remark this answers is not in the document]
holzeis: yes, we tested it.. The icon must not be too big though.

mobile/lib/features/wallet/onboarding/onboarding_screen.dart
```diff
@@ -0,0 +1,153 @@
...
+  @override
+  Widget build(BuildContext context) {
+    // TODO(holzeis): Fetch those options from the coordinator
+    Amount maxReceiveAmount = Amount(2000000);
+    Amount minReceiveAmount = Amount(50000);
+    final options = [
+      LiquidityOption(
+          title: "Trader",
+          liquidity: Amount(1200000),
+          minDeposit: Amount(100000),
```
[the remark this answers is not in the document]
holzeis: The clicks are not visible in the video, but if you'd click on one option you get the immediate response on your input. Give it a go in the simulator and let me know if its still unclear.

coordinator/src/routes.rs
```diff
@@ -344,10 +409,20 @@ pub async fn post_register(
...
 ) -> Result<Json<LspConfig>, AppError> {
+    let mut conn = state
+        .pool
+        .get()
+        .map_err(|e| AppError::InternalServerError(format!("Could not get connection: {e:#}")))?;
+
+    let liquidity_options = db::liquidity_options::get_all(&mut conn).map_err(|e| {
+        AppError::InternalServerError(format!("Failed to get all liquidity options: {e:#}"))
+    })?;
+
```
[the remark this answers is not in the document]
holzeis: liquidity options aren't really configurations are they?

crates/ln-dlc-node/src/channel.rs
```diff
@@ -62,7 +62,7 @@ impl Channel {
     ) -> Self {
         Channel {
             user_channel_id,
-            channel_state: ChannelState::Pending,
+            channel_state: ChannelState::Announced,
```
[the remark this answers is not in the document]
holzeis: > 👀 Seems like a bit of a sneaky change. Do we want to introduce separate constructors for Announced and Pending? I worry that there might be a place where we are using new for a different role (coordinator, app, maker) and this change is unwanted.

Well, the initial state of any channel is `Announced` if we upsert a channel which wasn't announced, we created a new one (Announced) and immediately update the channel state there.

> It's quite unfortunate that we're using the same model for app, coordinator and maker, because they all want different things out of the shadow representation of the channel.

Agreed

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -33,7 +34,7 @@ pub struct CoordinatorEventHandler<S> {
     pub(crate) node: Arc<Node<S>>,
     pub(crate) pending_intercepted_htlcs: PendingInterceptedHtlcs,
     pub(crate) event_sender: Option<EventSender>,
-    pub(crate) max_channel_size_sats: u64,
+    pub(crate) max_app_fund_amount: u64,
```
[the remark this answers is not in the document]
holzeis: It means something different. We have now a variable `max_channel_size` calculated from the liquidity options. Based on our current liquidity options that would be 1,000,000 (`max_app_fund_amount`) + 1,500,000 (`trade_up_to`) = 2,500,000

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -43,13 +44,13 @@ where
     pub fn new(
         node: Arc<Node<S>>,
         event_sender: Option<EventSender>,
-        max_channel_size_sats: u64,
+        max_app_fund_amount: u64,
```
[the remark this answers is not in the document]
holzeis: Not really, see comment above.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -396,6 +399,12 @@ where
         return Ok(());
     }
 
+    let max_app_fund_amount_msat = max_app_fund_amount * 1000;
+    ensure!(
+        expected_outbound_amount_msat <= (max_app_fund_amount * 1000),
+        "Failed to open channel because maximum fund amount exceeded, {expected_outbound_amount_msat} > {max_app_fund_amount_msat}"
+    );
```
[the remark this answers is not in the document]
holzeis: I am confused. This is exactly what we are doing now 😅 what do you refer to?

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -127,17 +130,35 @@ where
...
+        let has_channel = self
+            .channel_manager
+            .list_channels()
+            .iter()
+            .any(|channel| channel.counterparty.node_id == trader_id);
+
+        ensure!(
+            !has_channel,
+            "Cannot create onboarding invoice if user has already a channel!"
+        );
```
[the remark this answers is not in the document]
holzeis: It's a safeguard, but I don't mind removing it, as it shouldn't happen anyways.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -159,23 +180,62 @@ where
...
+                    "Failed to insert shadow JIT channel for counterparty {trader_id} \
+                             with user channel id {user_channel_id}"
                 )
             })?;
         }
 
         tracing::info!(
-            peer_id = %target_node,
+            %user_channel_id,
+            %trader_id,
```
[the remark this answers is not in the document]
holzeis: 😅 I guess we had the same idea in mind, but I was always using `trader_id` 🙈

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -14,15 +17,20 @@ use crate::ChainMonitor;
 use crate::EventHandlerTrait;
 use crate::NetworkGraph;
 use crate::PeerManager;
+pub use ::dlc_manager as rust_dlc_manager;
```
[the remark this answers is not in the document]
holzeis: that wasn't my change, the `dprint fmt` just changed the order of imports.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -99,7 +99,16 @@ type NodeGossipSync =
 type NodeEsploraClient = EsploraSyncClient<Arc<TracingLogger>>;
 
 type RequestedScid = u64;
-type FakeChannelPaymentRequests = Arc<parking_lot::Mutex<HashMap<RequestedScid, PublicKey>>>;
+type FakeChannelPaymentRequests = Arc<parking_lot::Mutex<HashMap<RequestedScid, LiquidityRequest>>>;
```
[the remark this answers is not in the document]
holzeis: Agreed, I am leaving a todo.

crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs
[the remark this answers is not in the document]
holzeis: can stress it enough that I am only assuming here 🤪

crates/ln-dlc-node/src/tests/multi_hop_payment.rs
```diff
@@ -39,6 +40,21 @@ async fn multi_hop_payment() {
...
+    // channels with short channel ids.
+    wait_until(Duration::from_secs(20), || async {
+        let useable_channels_length = payer.channel_manager.list_usable_channels().len();
+        let scids_set = payer
+            .channel_manager
+            .list_usable_channels()
+            .iter()
+            .all(|c| c.short_channel_id.is_some());
+
+        Ok((useable_channels_length == 1 && scids_set).then_some(0))
```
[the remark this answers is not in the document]
holzeis: nice catch, thanks!

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -31,13 +31,18 @@ use bdk::BlockTime;
 use bdk::FeeRate;
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::Amount;
+pub use channel_status::ChannelStatus;
```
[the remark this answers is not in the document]
holzeis: I didn't `dprint fmt` was doing it all the time.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -732,38 +742,107 @@ pub fn contract_tx_fee_rate() -> Result<u64> {
...
+                format!(
+                    "Failed to insert shadow JIT channel with user channel id {user_channel_id}"
+                )
+            })?;
 
         tracing::info!(
-            ?final_route_hint_hop,
-            "Registered interest to open JIT channel with coordinator"
+            %user_channel_id,
+            "Registering interest to open JIT channel with coordinator"
```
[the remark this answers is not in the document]
holzeis: code has been moved now.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -732,38 +742,107 @@ pub fn contract_tx_fee_rate() -> Result<u64> {
...
+            .storage
+            .upsert_channel(channel)
+            .with_context(|| {
+                format!(
+                    "Failed to insert shadow JIT channel with user channel id {user_channel_id}"
+                )
+            })?;
 
         tracing::info!(
-            ?final_route_hint_hop,
```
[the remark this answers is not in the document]
holzeis: yeah, I think it was very noisy.

coordinator/src/node.rs
holzeis: it was leverage 1.0 but with the new liquidity options its always 2.0.

coordinator/src/node/unrealized_pnl.rs
holzeis: good catch, thanks :)

coordinator/src/position/models.rs
holzeis: thanks, nice catch 👍

crates/ln-dlc-node/src/channel.rs
```diff
@@ -37,10 +36,7 @@ pub struct Channel {
     /// the funding transaction. We use the `user_channel_id` as identifier over the entirety
     /// of the channel lifecycle.
     pub channel_id: Option<ChannelId>,
-    /// The fake short channel id (generated for intercept invoice)
-    ///
-    /// This is only `Some` for JIT channels.
-    pub fake_scid: Option<FakeScid>,
+    pub liquidity_option_id: i32,
```
luckysori, line 39: :question: Can this be filled for non-JIT channels?

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
holzeis: ok, changed it.

crates/ln-dlc-node/src/channel.rs
```diff
@@ -148,7 +128,8 @@ impl Channel {
                     channel_details.inbound_capacity_msat / 1000,
                     0,
                     channel_details.counterparty.node_id,
-                    None,
+                    // legacy record
+                    1,
```
luckysori, line 132: :thinking: I think for some channels having a liquidity option ID makes no sense e.g. public channels.

An `Option` might be a better choice in terms of modelling.

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -182,8 +181,13 @@ where
...
         if is_for_jit_channel {
-            let fake_scid = FakeScid::new(intercept_scid);
-            let channel = Channel::new_jit_channel(trader_id, fake_scid);
+            let channel = Channel::new(
+                user_channel_id,
+                0,
+                0,
+                trader_id,
+                liquidity_request.liquidity_option_id,
+            );
```
luckysori, line 190: :wrench: I highly recommend to adapt the `new_jit_channel` constructor instead.

crates/ln-dlc-node/src/node/storage.rs
```diff
@@ -78,12 +78,10 @@ pub trait Storage {
     fn upsert_channel(&self, channel: Channel) -> Result<()>;
     /// Get channel by `user_channel_id`
     fn get_channel(&self, user_channel_id: &str) -> Result<Option<Channel>>;
-
-    /// Get channel by `user_channel_id`
-    fn get_channel_by_fake_scid(&self, fake_scid: FakeScid) -> Result<Option<Channel>>;
-
     /// Get all non pending channels.
     fn all_non_pending_channels(&self) -> Result<Vec<Channel>>;
+    /// Get announce channel with counterparty
```
luckysori, line 83:
```suggestion
    /// Get announced channel with counterparty.
```

Also, this is weird because it only applies to the coordinator (I hate the `Storage` trait).

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
holzeis: not for all channels, but for all positions of that channel. I think it needs to be that way otherwise we can not hold up to the trade up to value selected in the liquidity option. 

maybe we could increase it.

coordinator/src/position/models.rs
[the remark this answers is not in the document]
holzeis: yes, it is. good catch 👍

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -760,7 +760,40 @@ pub fn create_onboarding_invoice(amount_sats: u64, liquidity_option_id: i32) ->
...
+            Some(channel) => channel.user_channel_id,
+            None => {
+                let user_channel_id = UserChannelId::new();
+                let channel = Channel::new(
+                    user_channel_id,
+                    0,
+                    0,
+                    config::get_coordinator_info().pubkey,
+                    liquidity_option_id,
+                );
```
luckysori, line 781: :thinking: I'm confused: why are we doing this twice (also below) now?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add liquidity options to the onboaring flow
+- Pay fixed channel opening fees instead of onchain transaction fees
+- Increased maximum channel size
+- Add minimum channel size
+- Split onboading and create invoice flow
+- Format amounts in input fields
```
luckysori, line 15: :wrench:

```suggestion
- Add liquidity options to the on-boarding flow.
- Pay fixed channel-opening fees instead of on-chain transaction fees.
- Increase maximum channel size to X.
- Add minimum channel size.
- Split on-boarding flow from create-invoice flow.
- Format amounts in input fields.
- Set coordinator leverage to 2?
```

crates/coordinator-commons/src/lib.rs
```diff
@@ -68,14 +68,47 @@ impl RegisterParams {
...
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LiquidityOption {
+    pub id: i32,
+    pub rank: i16,
+    pub title: String,
+    // amount the trader can trade up to in sats
+    pub trade_up_to: i64,
+    // min deposit in sats
+    pub min_deposit: i64,
```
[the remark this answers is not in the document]
holzeis: I think @luckysori was right when saying that this value can't be negative.. So I am going to change that to `u64`

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
holzeis: ok, changed it.

crates/ln-dlc-node/src/node/invoice.rs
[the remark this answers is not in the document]
holzeis: changed to `peer_id` - I don't care too much 🙂

coordinator/src/node.rs
holzeis: I am opting to skip that suggestion 🙈

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
holzeis: does it your formatting do it differently?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -679,15 +685,19 @@ pub fn max_channel_value() -> Result<Amount> {
...
-            channel_value_sats = lsp_config.max_channel_value_satoshi,
-            "Received channel config from LSP"
-        );
-        Ok(Amount::from_sat(lsp_config.max_channel_value_satoshi))
+        Ok(Amount::ZERO)
     }
 }
 
+pub fn max_app_fund_amount() -> Result<Amount> {
+    let lsp_config = poll_lsp_config()?;
```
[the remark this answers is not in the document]
holzeis: I did not come up with the name, but I am happy to change it.

luckysori (approved):
Very nice! Mostly everything seems good to go.

coordinator/src/routes.rs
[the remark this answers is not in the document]
luckysori: Well, we can change our config at runtime too. And the endpoint is called `/api/lsp/config`. The lines are blurry here.

It's definitely nice that it is stored in the DB, so I'm not complaining too much.

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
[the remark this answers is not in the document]
luckysori: I suppose it's pretty close. When I left this comment I didn't yet realise that you had now diverted regular payments away from the `HTLCIntercepted` event too.

I think it would still be cleaner to branch at the very top, before dealing with anything related to the JIT channel case, but we can leave it as is since the payment via `HTLCIntercepted` is no longer supported.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
luckysori: Hmm, I hope so!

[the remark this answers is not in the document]
luckysori: `dprint` doesn't do this for me. I'm using `dprint 0.41.0` and `rustfmt 1.5.2-nightly (a368898 2023-05-02)` (which is very old).

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -732,38 +742,107 @@ pub fn contract_tx_fee_rate() -> Result<u64> {
     }
 }
 
-pub fn create_invoice(amount_sats: Option<u64>) -> Result<Invoice> {
+pub fn liquidity_options() -> Result<Vec<LiquidityOption>> {
+    let lsp_config = poll_lsp_config()?;
+    tracing::trace!(liquidity_options=?lsp_config.liquidity_options, "Received liquidity options");
```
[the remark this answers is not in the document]
luckysori: But does this get called that often?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -732,38 +742,107 @@ pub fn contract_tx_fee_rate() -> Result<u64> {
...
+    liquidity_option_id: i32,
+}
+
+pub fn create_onboarding_invoice(amount_sats: u64, liquidity_option_id: i32) -> Result<Invoice> {
     let runtime = get_or_create_tokio_runtime()?;
 
     runtime.block_on(async {
         let node = NODE.get();
         let client = reqwest_client();
+        let user_channel_id = UserChannelId::new();
```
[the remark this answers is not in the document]
luckysori: Interesting that we use the same `UserChannelId` for both. I don't think it should be a problem, but it's not designed to be synced across peers AFAIK.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -679,15 +685,19 @@ pub fn max_channel_value() -> Result<Amount> {
     {
         Ok(Amount::from_sat(existing_channel))
     } else {
-        let lsp_config = poll_lsp_config()?;
-        tracing::info!(
-            channel_value_sats = lsp_config.max_channel_value_satoshi,
-            "Received channel config from LSP"
-        );
-        Ok(Amount::from_sat(lsp_config.max_channel_value_satoshi))
+        Ok(Amount::ZERO)
```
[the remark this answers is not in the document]
holzeis: updated the docs.

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -71,18 +83,6 @@ pub(crate) mod invoice;
...
-pub use crate::node::oracle::OracleInfo;
-pub use ::dlc_manager as rust_dlc_manager;
-pub use channel_manager::ChannelManager;
-pub use dlc_channel::dlc_message_name;
-pub use dlc_channel::sub_channel_message_name;
-pub use invoice::HTLCStatus;
-pub use storage::InMemoryStore;
-pub use storage::Storage;
-pub use sub_channel_manager::SubChannelManager;
-pub use wallet::PaymentDetails;
```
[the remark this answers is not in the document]
holzeis: don't know what my `dprint fmt`, but I am not actively reordering those imports.. 🙈

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
holzeis: I am using `dprint 0.40.2`.. I'll update and see if it fixes things.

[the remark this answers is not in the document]
holzeis: upgrade to `dprint 0.41.0` did not change anything for me. It is still reordered every time 🙁

mobile/native/src/channel_fee.rs
holzeis: what @luckysori said 🙂

mobile/native/src/channel_fee.rs
holzeis: ouch.. nice catch. 👍

crates/ln-dlc-node/src/channel.rs
holzeis: I guess not.. I'll change it to an option.

crates/ln-dlc-node/src/channel.rs
holzeis: Agreed

crates/ln-dlc-node/src/channel.rs
holzeis: but it will mean we can't define a foreign key. :/

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
luckysori: I think it's based one `rustfmt` though. Which version of `rustfmt` are you using?

crates/ln-dlc-node/src/node/invoice.rs
holzeis: you convinced me.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
holzeis: using the `new_jit_channel` constructor now.

holzeis: that was some rebase error. thanks for spotting it.. I removed the unnecessary second time.

CHANGELOG.md
holzeis: >  Increase maximum channel size to X.

This is configurable, I wouldn't put a number here.

> Set coordinator leverage to 2?

It's not really set to 2 its made configurable on the selected liquidity option.

coordinator/src/db/channels.rs
```diff
@@ -144,6 +144,9 @@ impl From<ln_dlc_node::channel::ChannelState> for ChannelState {
             ln_dlc_node::channel::ChannelState::Announced => ChannelState::Announced,
             ln_dlc_node::channel::ChannelState::Pending => ChannelState::Pending,
             ln_dlc_node::channel::ChannelState::Open => ChannelState::Open,
+            // mapping `ChannelState::OpenUnpaid` to open as the coordinator references the payment
+            // through the payment hash. Hence no need to reflect that state twice.
+            ln_dlc_node::channel::ChannelState::OpenUnpaid => ChannelState::Open,
```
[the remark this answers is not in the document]
holzeis: But then I would need to make sure that the channel gets from `OpenUnpaid` to `Open`. I wanted to avoid that as it would be additional potential failure prune code, that has no real value does it?

crates/ln-dlc-node/src/channel.rs
```diff
@@ -151,13 +155,10 @@ impl Channel {
 
 #[derive(PartialEq, Debug, Clone)]
 pub enum ChannelState {
-    /// Corresponds to a JIT channel which an app user has registered interest in opening.
```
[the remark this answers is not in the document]
holzeis: it does again :)

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -698,6 +698,8 @@ pub fn max_app_fund_amount() -> Result<Amount> {
     Ok(Amount::from_sat(lsp_config.max_app_fund_amount))
 }
 
+// TODO(holzeis): We might want to consider caching the lsp config, as this shouldn't change too
+// often and even if, I guess we can live with the user having to restart to get the newest configs?
```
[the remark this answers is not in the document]
holzeis: we need it every time we open a position as well as we need to fetch the coordinators leverage. I really don't like that part of the design.

coordinator/src/node.rs
holzeis: I tend to agree and gave it a try. But it got very quickly very complicated, so I opted to leave it as is for now. But the whole `filled_with` and `trade_params` model probably need some rework anyways.

mobile/native/src/ln_dlc/mod.rs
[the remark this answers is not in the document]
holzeis: probably not.

PR #1385, opened 2023-09-27
fix: fallback fee too high

bonomat:
Sometimes we have a ridiculous high fee estimate. This is due to the fallback fee being in **BTC/kvB**. Setting it to 0.00001 should be 1000sat/kvb or 1sat/vbyte

😅


holzeis (approved):
Whoops 😅. Thanks for the fix 👍

PR #1388, opened 2023-09-27
Store app logs on disk during development

klochowicz:
Also, fix broken `run-regtest` configuration.

### DIFF
--- a/justfile
+++ b/justfile
@@ -126,6 +126,21 @@ run-regtest-android args="":
         --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}" \
         --dart-define="ORACLE_ENDPOINT={{public_regtest_oracle_endpoint}}" \
         --dart-define="ORACLE_PUBKEY={{public_regtest_oracle_pk}}" \
+        | tee  ../data/app_regtest.log
+
+
+# Specify correct Android flavor to run against our public regtest server
+run-regtest-android args="":
+    #!/usr/bin/env bash
+    cd mobile && \
+      flutter run {{args}} \
+        --dart-define="COMMIT=$(git rev-parse HEAD)" \
+        --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
+        --dart-define="ESPLORA_ENDPOINT={{public_regtest_esplora}}" \
+        --dart-define="COORDINATOR_P2P_ENDPOINT={{public_regtest_coordinator}}" \
+        --dart-define="COORDINATOR_PORT_HTTP={{public_coordinator_http_port}}" \
+        --dart-define="ORACLE_ENDPOINT={{public_regtest_oracle_endpoint}}" \
+        --dart-define="ORACLE_PUBKEY={{public_regtest_oracle_pk}}" \
         --flavor test \
         | tee  ../data/app_regtest_android.log
 
--- a/justfile
+++ b/justfile
@@ -97,7 +97,8 @@ run args="":
       --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
       --dart-define="REGTEST_FAUCET=http://localhost:8080" \
       --dart-define="REGTEST_MAKER_FAUCET=http://localhost:18000/api/pay-invoice" \
-      --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2"
+      --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2" \
+      | tee  ../data/app.log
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -97,7 +97,8 @@ run args="":
       --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
       --dart-define="REGTEST_FAUCET=http://localhost:8080" \
       --dart-define="REGTEST_MAKER_FAUCET=http://localhost:18000/api/pay-invoice" \
-      --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2"
+      --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2" \
+      | tee -a ../data/app.log
```
bonomat, line 101: I'm against this change. I like reading the logs in my terminal. Can't you just add this when you call `just run`?

holzeis: +1

Restioson: This should provide a superset of the previous functionality, though, as tee will redirect the output to `../data/app.log` as well as stdout, i.e you can still read it from stdout

Restioson, line 101: This will only redirect stdout and not stderr: maybe could do something like https://stackoverflow.com/a/53051506 ?

Restioson: left is tailing the logfile, right is running `just run` itself image

Restioson: Also, is always appending what we want?

klochowicz: Yes, it *is* still giving us the terminal as it was. Note that this is super useful, because we often want to be in control (e.g. by doing hot restart, taking a screenshot or whatever).

So there's no downside for the terminal lovers :-)

klochowicz: > This will only copy stdout to the logfile and not stderr: maybe could do something like https://stackoverflow.com/a/53051506 ?

indeed, I forgot to redirect stderr to stdout with `2>&1` before the pipe . I'll try it out to see if it works correctly (with our interactive terminal etc.)

klochowicz: > Also, is always appending what we want?

I'd say yes, it's a laughably small amount of space compared to our build. I can't imagine this becoming too unwieldy. I can also ensure I wipe it when we do `just wipe` if you think it would be a good idea.

klochowicz: > Also, is always appending what we want?

I removed the appending. We typically would be interested in the last run.

klochowicz: > This will only copy stdout to the logfile and not stderr: maybe could do something like https://stackoverflow.com/a/53051506 ?

I tried with using stderr as well, but I decided against it. all the logs we're outputting are on stdout, and we get ~800 lines of warnings from Flutter using deprecated code en stderr in the begininng that really make it not the preferred choice 😅 (but if we did it, it would be the way I mentioned above).

Restioson: Ah, I see. No problem then! Are we sure that Flutter's exception logs will be printed to stdout?

klochowicz: to be honest, I can't invest more time into it. we'll see when we get there. it's still an improvement IMHO ;D

bonomat (commented):
I still don't understand why you not just add `| tee ../data/app.log` when you run a command from the cli?

justfile
Restioson: Yep agree!

Restioson:
> I still don't understand why you not just add | tee ../data/app.log when you run a command from the cli?

I guess it still lowers friction for a common usecase

klochowicz:
> I still don't understand why you not just add `| tee ../data/app.log` when you run a command from the cli?

it's the reasonable default to have that log stored!

also: this PR fixes a really annoying bug that recently crept in - I need to apply fix locally to test anything against our regtest now ;/

bonomat:
> it's the reasonable default to have that log stored!

I'm sorry, but I'm not convinced. With this change I have to reconfiger my IDE to not index the new log files which is annoying on my side. 

If you want this, what about you create shell aliases for you which adds the `| tee ..` automatically?

klochowicz:
> > it's the reasonable default to have that log stored!
> 
> I'm sorry, but I'm not convinced. With this change I have to reconfiger my IDE to not index the new log files which is annoying on my side.
> 
> If you want this, what about you create shell aliases for you which adds the `| tee ..` automatically?

The log file is located in `data` directory, which already is ignored by the IDE, so it does not change anything on your side.

The upside would be that in the event of something unexpected happening in the app, the logs would be there. 

a concrete example: I just had this happen yesterday, when the wallet refresh ran into a deadlock. due to amount of spam, I could not gather the logs 😭

However, seeing that you deem this as not important, I will close this PR.

PR #1392, opened 2023-09-28
Two small fixes

bonomat:


maker/src/orderbook_ws.rs
```diff
@@ -39,16 +39,17 @@ pub struct Client {
...
         auth_sk: SecretKey,
         position_manager: xtra::Address<position::Manager>,
         orderbook_status: watch::Sender<ServiceStatus>,
     ) -> Self {
-        let domain = endpoint.domain().expect("domain");
-        let port = endpoint.port().expect("port");
-
-        let url = format!("ws://{domain}:{port}/api/orderbook/websocket");
+        endpoint
+            .set_scheme("ws")
```
holzeis, line 49: We will need to distinguish between `ws` and `wss` at some point here. I guess this logic is assuming that we are using http.

holzeis (approved):
Thanks for fixing. Once this lands in production we should also get rid of the port 81 workaround there.

klochowicz (approved):

PR #1395, opened 2023-09-28
chore: print storage so that we can monitor our disk storage

bonomat:


klochowicz:
FYI: with #1402 we might save a lot of space; a lot of unneeded dependencies - tooling for `flutter_rust_bridge` and whole Flutter SDK disappears from most of the jobs!

PR #1400, opened 2023-09-29
Move ln-dlc-node tests out of the crate

klochowicz:
- don't include the tests in the workspace by default, unless we're testing them (this should speed up the build a bit and cause it to use less space)
- action long overdue refactor where app and coordinator specific code was located in the ln_dlc_node crate due to the test requirement.


crates/ln-dlc-node-tests/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(test)]
+mod tests;
```
luckysori, line 2: It would be interesting to benchmark the differences between the original, this version and this version but with a single integration test crate i.e.

```
tests/
  integration/
    main.rs
    foo.rs
    bar.rs
```

I always go back to this blog post when thinking about this topic: https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html

crates/ln-dlc-node-tests/src/tests/dlc/collaborative_settlement.rs
```diff
@@ -1,9 +1,8 @@
-use crate::node::InMemoryStore;
-use crate::node::Node;
 use crate::tests::dlc::create::create_dlc_channel;
 use crate::tests::init_tracing;
 use crate::tests::wait_until_dlc_channel_state;
 use crate::tests::SubChannelStateName;
+use crate::tests::TestNode;
```
luckysori, line 5: :question: Can we just call it `Node`?

luckysori (commented):
In terms of af257e59eeaacd2c9a70b5893f31bc033476b47a, can you run some simple benchmarks with `-Z unstable-options --timings=html` to compare before and after?

klochowicz:
> In terms of [af257e5](https://github.com/get10101/10101/commit/af257e59eeaacd2c9a70b5893f31bc033476b47a), can you run some simple benchmarks with `-Z unstable-options --timings=html` to compare before and after?

Unfortunately, I don't have capacity to do so as I'm busy doing feature work.

I went with this change as it was discussed beforehand - mainly to remove the architectural wrinkles that were very odd (e.g. coordinator and app conflict in a **low-level** crate.

I find side-effect of removing linking of some libraries a nice addition to this change, but not the driver for merging.

crates/ln-dlc-node-tests/src/tests/dlc/collaborative_settlement.rs
klochowicz: we could, but because we have many `Node`s , I would need to start qualifying it in many places; it was easier to use a different name.

Cargo.toml
```diff
@@ -8,8 +8,10 @@ default-members = [
   "crates/bitmex-stream",
   "crates/coordinator-commons",
   "crates/ln-dlc-node",
+  # "crates/ln-dlc-node-tests",
   "crates/orderbook-client",
   "crates/orderbook-commons",
+  # "crates/tests-e2e",
```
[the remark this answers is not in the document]
klochowicz: we are not statically linking them in our binaries.

this means faster compile time (less things to link) and space saving. 

of course leaving them commented out is not needed, I just wanted to be explicit about these exclusions to draw attention to that 🙃

luckysori:
> I went with this change as it was discussed beforehand - mainly to remove the architectural wrinkles that were very odd (e.g. coordinator and app conflict in a low-level crate.

I probably wasn't part of that discussion. What's the conflict?

From the PR I assumed that this patch was motivated by speeding up compile times.

klochowicz:
> > I went with this change as it was discussed beforehand - mainly to remove the architectural wrinkles that were very odd (e.g. coordinator and app conflict in a low-level crate.
> 
> I probably wasn't part of that discussion. What's the conflict?
> 
> From the PR I assumed that this patch was motivated by speeding up compile times.

As far as I remember, we both had a discussion where you recommended moving the tests out of the crate so that `EventHandler` implementations can live in their respective crates. 
At the time the `EventHandler` work was such a big endeavour that I just left TODOs, but wanted to revisit it.

Am I misremembering things? 🤔

note: I actually got motivated to finally address it by trimming the disk space need, as at the time of submitting this PR we still had failing jobs on CI due to lack of disk space (in the end, I fixed that another way).

Cargo.toml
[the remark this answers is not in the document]
luckysori: > we are not statically linking them in our binaries.
>
> this means faster compile time (less things to link) and space saving.

Wait, really? Just because we declare them in the workspace we are linking test crates into our binaries? That doesn't seem right since they should not be in the dependency tree of the binaries 🤔

luckysori:
> As far as I remember, we both had a discussion where you recommended moving the tests out of the crate so that `EventHandler` implementations can live in their respective crates. At the time the `EventHandler` work was such a big endeavour that I just left TODOs, but wanted to revisit it.
> 
> Am I misremembering things? 🤔

I can't say I remember the specific conversation, but it doesn't mean it didn't happen!

Looking at this PR I didn't realise that moving the tests out of the crate was a requirement for moving the `EventHandler`s out. If that's the case then I'm okay with it.

> note: I actually got motivated to finally address it by trimming the disk space need, as at the time of submitting this PR we still had failing jobs on CI due to lack of disk space (in the end, I fixed that another way).

👍

luckysori (approved):
The only mildly surprising thing is keeping certain crates outside of the workspace, but if you're sure that it's beneficial then we can give it a go!

klochowicz:
> The only mildly surprising thing is keeping certain crates outside of the workspace, but if you're sure that it's beneficial then we can give it a go!

I've attempted to merge it, but the rebase was too time consuming. I'm afraid that this patch has aged very quickly over the 3 weeks since it got submitted ; if you still like it you can take it over @luckysori  .

In the meantime, I'll close it as I don't have the capacity to bring it up to `main`.

PR #1402, opened 2023-09-29
chore(ci): Generate FFI bindings only once in the CI

klochowicz:
New job `generate-ffi` is triggered straight away, when it finishes it uploads
both generated files for Rust and Dart.

Upload its result once and utilise it in other jobs, instead of wasting minutes
achieving the same result and introducing unneeded dependencies, e.g. flutter
into jobs that can get away without having it. This should in turn limit our
disk usage in these jobs, causing less errors on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,26 +23,8 @@ jobs:
           components: rustfmt
       - uses: dprint/check@v2.1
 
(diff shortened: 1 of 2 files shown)


holzeis (approved):
LGTM 👍

bonomat (approved):
Nice

PR #1403, opened 2023-09-30
Improve Close channel buttons

klochowicz:
- always show close channel button
- catch exceptions and show errors if any
- prevent accidental abuse of the API by repeated clicks

### DIFF
--- a/mobile/lib/common/settings_screen.dart
+++ b/mobile/lib/common/settings_screen.dart
@@ -5,6 +5,7 @@ import 'package:f_logs/f_logs.dart';
 import 'package:flutter/material.dart';
 import 'package:get_10101/common/scrollable_safe_area.dart';
 import 'package:get_10101/bridge_generated/bridge_definitions.dart' as bridge;
+import 'package:get_10101/common/snack_bar.dart';
 import 'package:intl/intl.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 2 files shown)


mobile/lib/common/settings_screen.dart
```diff
@@ -58,15 +58,15 @@ class _SettingsScreenState extends State<SettingsScreen> {
       appBar: AppBar(title: const Text("Settings")),
       body: ScrollableSafeArea(
           child: Column(children: [
+        ElevatedButton(
+            onPressed: () {
+              rust.api.closeChannel();
```
bonomat: Imho we should not just expose `closeChannel` but do a smart check if there is still a DLC, if there is one, collab close the position first and then collab close the channel.

bonomat (approved):

mobile/lib/common/settings_screen.dart
luckysori: In such a scenario I would prefer to instruct the user to close their position first.

Collab closing automatically is a bit weird because:

- The user likely wants to explicitly control their trades.
- We cannot even guarantee that the order will be executed any time soon.

bonomat: If we don't want this feature we should prevent him from clicking this button at least :)

klochowicz: I like the idea with preventing the action when there's an open position.

Instead of disabling the button,  as it's hard for discoverability, I will display the snackbar saying "you have to close your position before closing the channel" and skip the footgun API call

PR #1405, opened 2023-10-01
Reduce some log spam

### LINKED ISSUE #1394: Reduce log spam
Right now we print all orders we receive from the orderbook. I think we can reduce this: 

```New message from orderbook....```

bonomat:
Resolve #1394


holzeis (approved):
LGTM 👍

PR #1412, opened 2023-10-03
fix: release differently for regtest and mainnet

bonomat:
Otherwise we have to manually "promote" each release for our regtest testers.


luckysori (approved):
LGTM, but what do I know.

PR #1413, opened 2023-10-03
Ensure match between opening and closing trades in app trade history

### LINKED ISSUE #1379: Wallet history does not show position openings and closings correctly
Only one opening position is shown although I have multiple orders.

image

image

luckysori:
Fixes #1379.

By rounding the order quantities (contracts) to 2 decimal places it is _much less likely_ that opening and closing trades do not match.

A mismatch can break the trades in the wallet and trade history view, because a closing trade can  end up being ignored as an unsupported (extending, reducing, direction-changing) trade. This error can compound, as the error is carried forward an all subsequent trades will almost certainly be missing too.

Once we add support for all other forms of trading, this error can mostly go away as all cases will be handled.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -15,6 +15,7 @@ use crate::ln_dlc::node::Node;
 use crate::ln_dlc::node::NodeStorage;
 use crate::trade::order;
 use crate::trade::order::FailureReason;
+use crate::trade::order::Order;
 use crate::trade::position;
 use anyhow::anyhow;
 use anyhow::bail;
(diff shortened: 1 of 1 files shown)


klochowicz (approved):

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -626,6 +622,21 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
     Ok(trades)
 }
 
+/// Compute the number of contracts for the [`Order`] relative to its [`Direction`].
+fn compute_relative_contracts(order: &Order) -> Decimal {
```
klochowicz, line 626: thank you for extracting this into a dedicated method, it's more clear now! :)

PR #1415, opened 2023-10-03
Catch possible exception during registration

klochowicz:
Catch a hole in the registration process, where the app would only store the email locally if the coordinator was unreachable during app registration.

### DIFF
--- a/crates/ln-dlc-node/src/node/storage.rs
+++ b/crates/ln-dlc-node/src/node/storage.rs
@@ -124,7 +124,7 @@ impl Storage for InMemoryStore {
     ) -> Result<()> {
         let mut payments = self.payments.lock();
         match payments.get_mut(payment_hash) {
-            Some(mut payment) => {
+            Some(payment) => {
                 payment.status = htlc_status;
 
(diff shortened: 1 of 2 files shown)


bonomat (approved):
LGTM

crates/ln-dlc-node/src/node/storage.rs
```diff
@@ -124,7 +124,7 @@ impl Storage for InMemoryStore {
     ) -> Result<()> {
         let mut payments = self.payments.lock();
         match payments.get_mut(payment_hash) {
```
holzeis, line 126: 🤔 If we don't need it mutable, I guess we can also change that to a simple `get`?

holzeis (approved):

crates/ln-dlc-node/src/node/storage.rs
luckysori: We are still mutating it, it's just that the returned type is already `&mut`, so the `mut` keyword is superfluous.

luckysori (approved):

PR #1418, opened 2023-10-04
chore: Return to wallet dashboard on claimed payment

### LINKED ISSUE #180: Pending receive / success handling after "receive payment" flow
Follow up of #71  and #178

When receiving we cannot really refresh the wallet besides actually picking up the payment in the backend and then triggering a refresh of balance and transaction history. 

One thing that we could add is some form of `Success` screen, that tells the user that the payment will appear once it was sent and picked up by the Lightning node. We have that in the [Figma prototype](https://www.figma.com/file/Sr3SfZ46sZstKjSwftmTtc/10101-v1-(incl-bitcoin.design)?node-id=5652%3A92292&t=55DjyHsH66ykrAcQ-1). 

I think this is not optimal, because that success diakog might just come out of nowhere if the user navigates somewhere else and only does the payment later. Instead I propose this more guided UX:

Once the user hits `Done` in the "receive paymnet" flow we display the`SubmissionStatusDialog` in `Pending` state; and transition to `Success` state once we detect the payment coming in (which will show the payment in the wallet history and reflect the balance change). The user can just close that dialog and it will be gone (e.g. if the user does not want to receive anymore).
Note: This means we would have to update the state of the dialog by the payment. With the current implementation of the dialog this is not easily possible, so we would have to adapt that.

holzeis:
I did not wasted time on showing some kind of success dialog. I still think we should show something, but I guess this is something to do by somebody with a better eye for design than me :)

resolves #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add minimum channel size
 - Split onboading and create invoice flow
 - Format amounts in input fields
+- Return to wallet dashboard on claimed payment
 
 ## [1.3.0] - 2023-09-27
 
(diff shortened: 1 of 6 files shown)


luckysori (approved):
Concept ACK.

PR #1428, opened 2023-10-05
chore: Notify only users that haven't rolled over yet

### LINKED ISSUE #1421: Don't send notifications to already rolled over positions
https://github.com/get10101/10101/pull/1419 introduces a push notification to users if the rollover window is open. 
We also notify users if they have allready rolled over their position which is not nice :)

holzeis:
fixes #1421


bonomat (approved):
LGTM

PR #1430, opened 2023-10-05
feat: Remind user to close expired position

holzeis:
Reminder is sent daily for 7 days after the position has expired


bonomat (approved):
Works for me.

PR #1432, opened 2023-10-05
Hedge maker's 10101 position on BitMEX

### LINKED ISSUE #1172: Actually hedge position
After https://github.com/get10101/10101/issues/973 lands we know the position size of the maker. We now should make use of our bitmex client and actually hedge the position. 

Note: bitmex has a min of 100 contracts per trade, hence, we need to be a bit smarter with hedging

luckysori:
Fixes #1172.

I have verified that it works using my own account on `testnet.bitmex.com`.

Mosf of the diff is replacing `maker/src/trading/bitmex_client.rs` with `maker/src/trading/bitmex_ws_client.rs` which is mostly lifted from the [ItchySats automaker](https://github.com/get10101/automaker/blob/57ede841af378a4e738df0e8988baf2a9c166293/src/bitmex/subscriber.rs).


maker/src/position.rs
```diff
@@ -1,30 +1,109 @@
...
+    type Stop = ();
+
+    async fn started(&mut self, mailbox: &mut Mailbox<Self>) -> Result<(), Self::Stop> {
+        tokio::spawn({
+            let mailbox = mailbox.clone();
+            async move {
+                loop {
+                    // We sleep first to allow the 10101 and BitMEX positions to be up-to-date
+                    // before we start hedging.
+                    tokio::time::sleep(Duration::from_secs(60)).await;
```
holzeis, line 37: 🤔 Is this safe? what if the BitMEX positions are not up-to-date?

holzeis (approved):
LGTM 👍

maker/src/position.rs
luckysori: One thing we could do is to only hedge if we know that the positions are up-to-date by at least a minute.

holzeis: Shouldn't the first hedge event after we have synchronised the tentenone position?

bonomat (approved):
Tested locally, there is a bit of a lag between opening a position and the maker hedging, we will see how much this will cost us :D

luckysori:
> Tested locally, there is a bit of a lag between opening a position and the maker hedging, we will see how much this will cost us :D

Yep, we should tighten this! I think this implementation should be similar to what we were doing in ItchySats, where we were hedging every minute.

Issue #1427, opened 2023-10-05
Opening positions with larger margins takes forever


holzeis:
Just opened a position with 800 contracts (margin: 2,900,000 sats) and it took about 35 seconds to setup the DLC.

bonomat:
related to #250

holzeis:
closing this ticket as https://github.com/get10101/10101/pull/1436 has been merged.

PR #1436, opened 2023-10-06
fix: reduce amount of CETs to be generated

bonomat:
This is a simple quick fix to fix the max amounts of payouts to 100. This means, the higher the trader amount is, the higher the steps are. We should go for a proper discretized curve to fox this.

quick fix for https://github.com/get10101/10101/issues/1427


holzeis (approved):
Nice 🚀

Issue #1390, opened 2023-09-28
Dev environment is unusable - can't fund.


klochowicz:
**Describe the bug**

doing `just all` fails *every single time* on my mac.

```
2023-09-28T03:48:39.742974Z  INFO fund: Manually broadcasting node announcement and waiting for a few seconds...
thread 'main' panicked at 'to be able to fund: Could not verify channel is open. Please wipe and try again', crates/tests-e2e/examples/fund.rs:44:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**To Reproduce**

1. Run `just all`

**Expected behavior**

Funding should work, it did work for me *all the time* until recently.

bonomat:
Might be worth pointing out that this is coming from `just fund`. 
The issue is that LND does not find/accept the node announcement of our coordinator. 

@holzeis : made it work on his machine and testnet setup. Maybe he can describe his (manual) steps to get it work?

klochowicz:
it got fixed for me with #1387 - can you try if it works for you @bonomat @holzeis ?

klochowicz:
I'm going to close it, as I am unblocked. if anyone still experiences this on master, please feel free to re-open :-)

bonomat:
> it got fixed for me with #1387 - can you try if it works for you @bonomat @holzeis ?

just worked for me as well 🤷‍♂️

klochowicz:
it started failing for me again, and after looking into it a bit, it was only due to the peer_alias not being populated 🤔 which has no impact on the ability to use the channel.

holzeis:
Documenting some pitfalls that will fail your faucet payment, but isn't a bug

- lnd will set the flag `sync_to_graph` to `false` if the hasn't been a new block for quite some time. Mine a few new blocks or use the auto mining script to fix that issue.
- lnd will quickly give up on a channel if a payment failed, ensure your are resetting the mission control, before retrying payments.

klochowicz:
I'm closing this, as it works on my machine. If it fails for anyone else, please re-open.

PR #1438, opened 2023-10-06
chore: Create maker channel last

holzeis:
This is not a fix, but for me this significantly improves the chances of the funding script to succeed.

related to https://github.com/get10101/10101/issues/1390


holzeis:
@klochowicz this is what I am doing locally to get the fund script working again.

bonomat (approved):
This also does not work for me all the time but I'm happy to get it in if this makes things better for you.

Issue #1435, opened 2023-10-06
Exception when funding the wallet


bonomat:
1. open wallet
2. click through funding screen
3. select biggest one
4. payment is shown as pending and the below error is logged
5. channel was opened succesfully though.

```
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following GoError was thrown during a scheduler callback:
There is nothing to pop

When the exception was thrown, this was the stack:
#0      GoRouterDelegate.pop (package:go_router/src/delegate.dart:94:5)
#1      GoRouter.pop (package:go_router/src/router.dart:468:20)
#2      _FundWalletModalState.build.<anonymous closure> (package:get_10101/features/wallet/onboarding/fund_wallet_modal.dart:71:30)
#3      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1297:15)
#4      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1236:9)
#5      SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1085:5)
#6      _invoke (dart:ui/hooks.dart:170:13)
#7      PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:401:5)
#8      _drawFrame (dart:ui/hooks.dart:140:31)
```

Local dev environenemt.

PR #1444, opened 2023-10-06
fix: Use go to wallet screen instead of pop

holzeis:
It's weird but if you pop only once you don't get back to the wallet screen and if you pop twice you get back to the wallet screen but also get the exception.

fixes: #1435


bonomat (approved):
LGTM

PR #1448, opened 2023-10-06
feat: Store last fcm token update time

klochowicz:
This can give us an idea about the FCM token freshness


coordinator/src/db/user.rs
```diff
@@ -30,6 +31,7 @@ impl From<RegisterParams> for User {
             nostr: value.nostr.unwrap_or("".to_owned()),
             timestamp: OffsetDateTime::now_utc(),
             fcm_token: "".to_owned(),
+            fcm_token_update: Some(OffsetDateTime::now_utc()),
```
holzeis, line 34: 🤔 Shouldn't this be `None` since we are not setting a `fcm_token` here?

coordinator/src/db/user.rs
```diff
@@ -51,18 +53,21 @@ pub fn upsert_email(
...
     let user: User = diesel::insert_into(users::table)
         .values(User {
             id: None,
             pubkey: trader_id.to_string(),
             email: email.clone(),
             nostr: "".to_owned(),
-            timestamp: OffsetDateTime::now_utc(),
+            timestamp,
             fcm_token: "".to_owned(),
+            fcm_token_update: Some(OffsetDateTime::now_utc()),
```
holzeis, line 66: 🤔 same here

coordinator/src/schema.rs
```diff
@@ -239,6 +239,7 @@ diesel::table! {
         nostr -> Text,
         timestamp -> Timestamptz,
         fcm_token -> Text,
+        fcm_token_update -> Nullable<Timestamptz>,
```
holzeis, line 242: It looks like the `fcm_token_update` is always set, also if the `fcm_token` is empty. This would then rather represent the `last_login` or `updated_at` timestamp. I fine either way, but it would be nice to align the model correspondingly.

holzeis (approved):

PR #1452, opened 2023-10-07
feat: Enter contracts in full amounts

### LINKED ISSUE #1408: Round contracts amount to 2 numbers after the comma
Screenshot 2023-10-02 at 10 45 44

### LINKED ISSUE #1409: Format market price as currency
Screenshot 2023-10-02 at 10 46 42

- show comma at 1000th position
- show currency

holzeis:
- Sets the new default amount to 10 dollars
- Margin can not be edited anymore, only contracts
- Margin is updated when price changes
- Contracts can only be entered in full amounts

This will also fix a well known old exception that was always thrown when either value (margin, quantity) was changed on the trade screen.

https://github.com/get10101/10101/assets/382048/48960ddd-e9e0-4604-beb4-ce657a2ff060

fixes #1408 
fixes #1409

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Only contracts are editable for new orders
+- Contracts can only be entered in full amounts
+- New default amount is 10 contracts
+
 ## [1.4.1] - 2023-10-09
 
 - Make trade / stable screen default if user has a position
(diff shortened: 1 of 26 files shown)


bonomat (approved):
Nice. Eventually we could let the coordinator dictate the min contract size so that we can configure it.

holzeis:
@bonomat I noticed that all validation logic was missing, since we now need to validate on the quantity. I've basically copied the validation from the margin input field with a few adaptions. e.g. I've removed the trading capacity as it didn't make any sense to me.

Please have another look https://github.com/get10101/10101/compare/8bee6c34b682a8321ee3432cb225bf7ffb5d7f81..789e027d94044bc9a99bbb77b4813d322bdb12a6

PR #1455, opened 2023-10-07
feat: Make trade / stable screen default if user has a position

### LINKED ISSUE #1318: Make trade screen default if you have an open position

holzeis:
Stores if a position is open in the preferences and reads from it at startup in a loading screen. Once loaded you will be transferred to the correct page.

I had to default the trade screen to the positions, as otherwise you would have always defaulted to the orders, since at load time the position is not there yet.

https://github.com/get10101/10101/assets/382048/58fef395-df8f-4f2d-b413-5d8531bead1d

resolves #1318

### DIFF
--- a/mobile/lib/util/preferences.dart
+++ b/mobile/lib/util/preferences.dart
@@ -1,3 +1,5 @@
+import 'package:get_10101/features/stable/stable_screen.dart';
+import 'package:get_10101/features/trade/trade_screen.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 class Preferences {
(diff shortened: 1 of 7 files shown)


bonomat (commented):
I would do it differently: 

- when opening/closing a position, store in the app shared preferences the details about the position
- loading the shared preferences is fast and doesn't have to wait for anything in the backend.

holzeis:
> I would do it differently: 
> 
> 
> 
> - when opening/closing a position, store in the app shared preferences the details about the position
> 
> - loading the shared preferences is fast and doesn't have to wait for anything in the backend. 

That's a great idea. This way we can also easily differentiate between a stable and a regular position 👍

mobile/lib/util/preferences.dart
```diff
@@ -7,6 +9,27 @@ class Preferences {
 
   static const userSeedBackupConfirmed = "userSeedBackupConfirmed";
   static const emailAddress = "emailAddress";
+  static const openPosition = "openPosition";
+
+  getOpenPosition() async {
+    SharedPreferences preferences = await SharedPreferences.getInstance();
+    return preferences.getString(openPosition);
```
bonomat, line 16: Do you need to add `?? "";` same as in `getEmailAddress`?

bonomat (approved):
LGTM, just one question.

mobile/lib/util/preferences.dart
holzeis: Doesn't matter as it returns null in case it's not set and that would run into the default path on the loading screen.

PR #1462, opened 2023-10-09
fix: add missing input for ios release

bonomat:


.github/workflows/ios-fastlane.yml
```diff
@@ -69,6 +69,10 @@ on:
         required: true
         description: "The public key of the oracle. Must match with the public key returned at http://${oracle_endpoint}/oracle/publickey"
         type: string
+      app_scheme:
+        required: true
+        description: "For regtest enter 'test' or mainnet enter 'Runner'"
```
holzeis, line 74: 🙃 that's weird.

holzeis (approved):

.github/workflows/ios-fastlane.yml
bonomat: c'est la vie

PR #1468, opened 2023-10-10
feat: new metric for counting dlc channels

bonomat:
cc @holzeis you love this one :)

Signed-off-by: Philipp Hoenisch <philipp@coblox.tech>

### DIFF
--- a/coordinator/src/metrics.rs
+++ b/coordinator/src/metrics.rs
@@ -38,6 +38,10 @@ lazy_static! {
         .u64_observable_gauge("channel_is_usable")
         .with_description("If a channel is usable")
         .init();
+    pub static ref DLC_CHANNELS_AMOUNT: ObservableGauge<u64> = METER
+        .u64_observable_gauge("dlc_channel_amount")
+        .with_description("Number of DLC channels")
+        .init();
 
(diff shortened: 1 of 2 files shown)


coordinator/src/metrics.rs
```diff
@@ -76,6 +80,9 @@ pub fn collect(node: Node) {
     position_metrics(&cx, &node);
 
     let inner_node = node.inner;
+    if let Ok(dlc_channels) = inner_node.list_dlc_channels() {
+        DLC_CHANNELS_AMOUNT.observe(&cx, dlc_channels.len() as u64, &[]);
```
holzeis, line 84: It would be nice if we could count the dlc channels in different states as well e.g. healthy dlc channels are Signed and OffchainClosed

OnChainClosed should not be part of the metric I guess

And all others are unhealthy.

holzeis (approved):

coordinator/src/metrics.rs
bonomat: Can do, I would want to avoid recording all combinations though because this explodes in total metrics. 
Can you state which states you want to track?

PR #1469, opened 2023-10-10
fix: fail e2e tests if there was an error

### LINKED ISSUE #1473: Fix: e2e tests are currently not working
Due to a change in the CI script we didn't notice that our e2e tests were not working. 
The problem is that we do not register the user and hence fail when generating an interceptable invoice.

bonomat:
fixes #1473


holzeis (approved):
👍

holzeis:
Seems like the e2e tests are also very flake 😕

PR #1477, opened 2023-10-11
feat: Recover from stuck rollover state

holzeis:
- Rollback channel state to the last stable state if in intermediate state at reconnect
- Adds the rollover contract to the rollover proposal

Upon starting the app it checks if the channel is in an intermediate state. If so it rolls back to the last "stable" state. The coordinator does the same upon connecting with the app and will the propose another rollover.

Note, in case the app broke before processing the `RenewRevoke` message, the coordinator will not be able to identify that the rollover failed for the user and remains the renewed contract. Meaning the rollover will not be retried. That puts the coordinator at an advantage and is just another reason why this bug should be fixed properly in rust-dlc eventually. However, the channel and subchannel is usable for both parties and a collaborative close is possible. If the renewed contract on the cooridinator side were to expire again, the rollover would be rejected by the app, as it is in the context of a contract unkown to the app. So the user can either close the position or wait for it to expire.

Note, this is merely a quick fix to unblock users that are in a stuck state after rollover. A proper fix should be implemented in rust-dlc, rendering this fix obsolete.


coordinator/src/db/positions.rs
```diff
@@ -41,13 +42,14 @@ pub struct Position {
...
         conn: &mut PgConnection,
-        trader_pubkey: String,
+        trader_pubkey: PublicKey,
+        states: Vec<crate::position::models::PositionState>,
     ) -> QueryResult<Option<crate::position::models::Position>> {
         let x = positions::table
-            .filter(positions::trader_pubkey.eq(trader_pubkey))
-            .filter(positions::position_state.eq(PositionState::Open))
+            .filter(positions::trader_pubkey.eq(trader_pubkey.to_string()))
+            .filter(positions::position_state.eq_any(states.into_iter().map(PositionState::from)))
```
bonomat, line 52: That's cool :)

coordinator/src/node/rollover.rs
```diff
@@ -172,10 +141,77 @@ impl Rollover {
...
+            };
+
+            let (retry_rollover, contract_id) = match position.position_state {
+                PositionState::Rollover => self.rollback_channel_if_needed(conn, signed_channel)?,
+                PositionState::Open => (false, signed_channel.get_contract_id()),
+                _ => bail!("Unexpected position state {:?}", position.position_state),
+            };
+
+            if coordinator_commons::is_eligible_for_rollover(OffsetDateTime::now_utc(), network)
+                && !position.is_expired()
```
bonomat, line 182: You could do an early return at the beginning of the function and check if the position has expired. If it is, we wouldn't need to do any of this.

coordinator/src/node/rollover.rs
```diff
@@ -215,6 +252,44 @@ impl Node {
             contract.get_temporary_id(),
         )
     }
+
+    pub fn rollback_channel_if_needed(
```
bonomat, line 256: I guess this doesn't have to be public.

coordinator/src/node/rollover.rs
```diff
@@ -215,6 +252,44 @@ impl Node {
...
+        &self,
+        connection: &mut PgConnection,
+        signed_channel: SignedChannel,
+    ) -> Result<(bool, Option<ContractId>)> {
+        if is_channel_in_intermediate_state(&signed_channel) {
+            let trader_id = signed_channel.counter_party;
+            tracing::warn!(%trader_id, "Found signed channel contract in an intermediate state. Trying to rollback channel to the last stable state!");
```
bonomat, line 263: We could log the channel state as well - could be useful for debugging purposes to see in what stage users get stuck.

coordinator/src/node/rollover.rs
```diff
@@ -215,6 +252,44 @@ impl Node {
...
+            self.inner
+                .rollback_channel(&signed_channel)
+                .map_err(|e| anyhow!("{e:#}"))?;
+
+            let contract = self
+                .inner
+                .get_contract_by_dlc_channel_id(&signed_channel.channel_id)?;
+            db::positions::Position::set_position_to_open(
```
bonomat, line 271: Why do you set the position to open, shouldn't it be open already?

crates/orderbook-commons/src/lib.rs
```diff
@@ -149,7 +149,7 @@ pub enum Message {
         order: Order,
         filled_with: FilledWith,
     },
-    Rollover,
+    Rollover(Option<String>),
```
bonomat, line 152: Why don't you use the proper type for contract_id?

mobile/native/src/ln_dlc/recover_rollover.rs
```diff
@@ -0,0 +1,72 @@
...
+use ln_dlc_node::node::rust_dlc_manager::Storage;
+
+impl Node {
+    /// Checks and recovers from a potential stuck rollover.
+    pub async fn recover_rollover(&self) -> Result<()> {
+        tracing::debug!("Checking if dlc channel got stuck in rollover.");
+        let channels = self.inner.channel_manager.list_channels();
+        let channel_details = match channels.first() {
+            Some(channel_details) => channel_details,
+            None => return Ok(()),
```
bonomat, line 14: Might be worth logging that everything is Ok here, or why an we return early and don't continue

bonomat (commented):
Concept ACK, will need to have another look later but here are some preliminary comments

bonomat (approved):
I've nothing else to say about this :)

holzeis:
I'd like to add one more thing and that is to allow recovering a stuck rollover state even if the rollover window isn't open.

coordinator/src/node/rollover.rs
holzeis: I could do an early return if the position is `Open` but IMHO that would make it less readable.

coordinator/src/node/rollover.rs
holzeis: in case we are in a stuck intermediate rollover state the position is in state `Rollover`. For a proper rollback we have to set the position back to `Open`.

crates/orderbook-commons/src/lib.rs
holzeis: Because the type `ContractId` is not known in the `orderbook_commons` and I prefer sending a string over a byte array :)

Issue #1085, opened 2023-08-08
Collaborative closure of the LN-DLC channel (when DLC subchannel is open)


luckysori:
### Context

We currently only allow collaboratively closing the Lightning channel if there is no active subchannel associated with it. If there is a subchannel, we instruct the parties to first collaboratively close the subchannel.

This is acceptable, unless the subchannel is stuck in an intermediate state[^1] (it's neither open nor closed). If the subchannel is stuck, the parties will not be able to close the subchannel collaboratively, so they will have to force-close instead. Force-closing is much less convenient as it involves long timelocks and higher transaction fees.

### Solution

Let's permit the collaborative closure of the LN-DLC channel via LDK under the right circumstances i.e. if the subchannel is stuck.

Things to consider:

- Both parties need to be online. This suggests to me that it should be initiated by the app, but that could lead to it being abused! As a solution I think we should let the app call a coordinator endpoint so that the coordinator can propose the collaborative closure of the LN-DLC channel. The coordinator would then have final say on whether to do so or not. We could even let the coordinator decide then and there if the channel should be force-closed instead.

- We need to patch `rust-lightning`, as we currently build an invalid `closing_transaction` if we try to collaboratively close when the subchannel is open. This is because we wrongly use the `glue_txo` as an input instead of the `funding_txo`.

- Closing collaboratively through LDK will ignore the state of the DLC channel. This would probably be desirable if the subchannel got stuck when opening a position as we can safely ignore the DLC that was never actually created, but in all other cases (currently only when closing a position) this would be bad. To deal with this we could further extend `rust-lightning` to allow collaboratively closing the channel with a custom split, but it might be simpler to only allow collaborative closure in specific scenarios.

[^1]: This should never happen, but this is still beta software so it does actually happen.

holzeis:
Still relevant and should probably be tackled sooner than later.

PR #1478, opened 2023-10-12
feat: collaborative close a (potentially broken) channel

bonomat:
resolves https://github.com/get10101/10101/issues/1085

the new endpoint can be called via curl, i.e.: 

```
curl -d '{"channel_id":"0f0afb20f66fded2bc4dfb58aa11628743b2dfca083bbc2c4a0a9b59483c6a58", "price":"30000", "fee_rate_sats_vb": 4}' -H "Content-Type: application/json" -X POST http://localhost:8000/api/admin/channels/revert
```

Next step would be: 
- [x] ~let the user confirm it: will be done in a follow-up PR.~
- [x] make it work even when the user is offline, i.e. store the message and send only when he comes online


coordinator/src/admin.rs
```diff
@@ -144,6 +156,98 @@ pub async fn list_dlc_channels(
     Ok(Json(dlc_channels))
 }
 
+pub async fn collaborative_revert(
+    State(state): State<Arc<AppState>>,
+    revert_params: Json<CollaborativeRevert>,
+) -> Result<Json<String>, AppError> {
```
holzeis, line 166: 🔧 Moving the actual logic into a separate function would simplify the error handling and improve maintainability. IMHO we shouldn't have much logic in the API layer.

coordinator/src/admin.rs
```diff
@@ -144,6 +156,98 @@ pub async fn list_dlc_channels(
...
+    let position = Position::get_position_by_channel_id(&mut conn, channel_id_string.clone())
+        .map_err(|error| {
+            tracing::error!(
+                channel_id = revert_params.channel_id.clone(),
+                "Could not get position for channel {error:#}"
+            );
+            AppError::InternalServerError(format!("Failed to load position from db: {error:#}"))
+        })?;
+
+    let settlement_amount = position
```
holzeis, line 207: why not use the `unrealized_pnl_sat` from the position?

holzeis: ok, looks like the `unrealized_pnl_sat` is never set 🙈

coordinator/src/admin.rs
```diff
@@ -144,6 +156,98 @@ pub async fn list_dlc_channels(
...
+        - settlement_amount as i64;
+    let trader_amount = subchannel.fund_value_satoshis - coordinator_amount as u64;
+
+    let fee = weight_to_fee(
+        COLLABORATIVE_REVERT_TX_WEIGHT,
+        revert_params.fee_rate_sats_vb,
+    )
+    .expect("To be able to calculate constant fee rate");
+    let address = state.node.inner.get_unused_address();
+    let coordinator_amount = Amount::from_sat(coordinator_amount as u64 - fee / 2);
```
holzeis, line 251: 🔧 what if the fee is uneven? one side should floor and the other ceil.

coordinator/src/admin.rs
```diff
@@ -144,6 +156,98 @@ pub async fn list_dlc_channels(
...
+            channel_id,
+            coordinator_address: address,
+            coordinator_amount,
+            trader_amount,
+        })
+        .map_err(|error| {
+            AppError::InternalServerError(format!("Could not get notify trader {error:#}"))
+        })?;
+
+    Ok(Json(
```
holzeis, line 298: ❓ Is wrapping it in Json really necessary?

coordinator/src/routes.rs
```diff
@@ -604,3 +617,132 @@ pub async fn version() -> Result<Json<Version>, AppError> {
...
+        .context("Could not find provided channel")
+        .map_err(|e| AppError::BadRequest(format!("Channel not found: {e:#}")))?;
+
+    let channel_manager = inner_node.channel_manager.clone();
+
+    let mut own_sig = None;
+
+    let mut revert_transaction = revert_params.transaction.clone();
+
+    let position = Position::get_position_by_channel_id(&mut conn, channel_id_string.clone())
```
holzeis, line 673: 🔧 if you would go for the `user_channel_id`, you wouldn't have to add a new database query to fetch the channel by `channel_id`

holzeis: Even better, why don't you simply provide already the counterparty pubkey from the subchannel, instead of taken the d-tour over the `channel_id`?

coordinator/src/routes.rs
```diff
@@ -604,3 +617,132 @@ pub async fn version() -> Result<Json<Version>, AppError> {
         branch: env!("BRANCH_NAME").to_string(),
     }))
 }
+
+pub async fn collaborative_revert_confirm(
```
holzeis, line 625: 🔧 Moving the actual logic into a separate function would simplify the error handling and improve maintainability. IMHO we shouldn't have much logic in the API layer.

crates/tests-e2e/tests/collaborative_revert_position.rs
```diff
@@ -0,0 +1,67 @@
+use coordinator::admin::Balance;
+use native::api;
+use tests_e2e::bitcoind::Bitcoind;
+use tests_e2e::coordinator::Coordinator;
+use tests_e2e::setup;
+use tests_e2e::wait_until;
+
+#[tokio::test]
+#[ignore = "need to be run with 'just e2e' command"]
+async fn can_revert_channel() {
```
holzeis, line 10: Nice :)

mobile/native/native.xcodeproj/xcuserdata/bonomat.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -7,12 +7,12 @@
 		<key>native-cdylib.xcscheme_^#shared#^_</key>
 		<dict>
 			<key>orderHint</key>
-			<integer>10</integer>
```
holzeis, line 10: 🔧 I guess the bonomat.xcuserdatad shouldn't be committed.

holzeis (commented):
Will have another look when the last todo is done.

coordinator/src/admin.rs
bonomat: 😅 I didn't trust these values. Hence I didn't go for it

coordinator/src/admin.rs
bonomat: We will pay 1 sat more to the miners which I think is ok

mobile/native/native.xcodeproj/xcuserdata/bonomat.xcuserdatad/xcschemes/xcschememanagement.plist
bonomat: Sure, it's needed to run the tests 🤪

coordinator/src/admin.rs
bonomat: Maybe not, but I like it if everything is JSON.

coordinator/src/routes.rs
bonomat: good point. Thanks.

mobile/native/native.xcodeproj/xcuserdata/bonomat.xcuserdatad/xcschemes/xcschememanagement.plist
bonomat: [`4e12b91` (#1478)](https://github.com/get10101/10101/pull/1478/commits/4e12b9195227607fb452c21c6705fcd9bd49b0a6)

coordinator/src/admin.rs
bonomat: [`7f32d67` (#1478)](https://github.com/get10101/10101/pull/1478/commits/7f32d6783948fa50602024f897ad4168250d31c9)

coordinator/src/routes.rs
bonomat: [`a6e2598` (#1478)](https://github.com/get10101/10101/pull/1478/commits/a6e259816dbd3f35f27ef04a904f6f3ecdbb4eed)

coordinator/src/routes.rs
bonomat: [`c505c6d` (#1478)](https://github.com/get10101/10101/pull/1478/commits/c505c6dec395e48931fe505300efe4d440991a32)

coordinator/src/orderbook/collaborative_revert.rs
```diff
@@ -0,0 +1,83 @@
...
+    remote_handle
+}
+
+/// Checks if there are any pending collaborative reverts
+async fn process_pending_collaborative_revert(
+    conn: &mut PgConnection,
+    notifier: mpsc::Sender<OrderbookMessage>,
+    trader_id: PublicKey,
+) -> Result<()> {
+    if let Some(channel) = channels::get_by_trader_pubkey(trader_id, conn)? {
```
holzeis, line 51: 🔧 I guess this will produce ugly error messages on the coordinator, as this will be run for every user opening the app. They might have an announced channel that never got funded, that would produce "Can't revert a channel ..." error. Even though we didn't even intended to do that.

I propose you turn that logic around. Firstly fetch per `trader_id` if there is a pending `collaborative_revert` then check if the `channel_id` is still correct. Not sure if the last check is actually required though.

holzeis (approved):
Just a last comment, but it can be addressed in a follow up.

coordinator/src/orderbook/collaborative_revert.rs
bonomat: Good point. 

done: [`fe4c71f` (#1478)](https://github.com/get10101/10101/pull/1478/commits/fe4c71fceb1fb056cc90104ac96acfc83c4994ce)

PR #1480, opened 2023-10-12
Reduce usage of `dynamic` keyword

Restioson:
This removes some uses of the `dynamic` keyword, thus making the code more type safe.

I've added quite a few TODOs here which must first be resolved. These are all at sites where I've explicitly added null checks (`!`) where they were previously implicit. Hence, I'm not sure if they were intentionally implicit before, or we just forgot to handle the null case. Maybe someone who is familiar with those parts can give some guidance on how we should handle nulls there?

I also need to still handle the clipping case when text is too long (force ellipsise it).

### DIFF
--- a/mobile/lib/features/wallet/wallet_screen.dart
+++ b/mobile/lib/features/wallet/wallet_screen.dart
@@ -70,14 +70,15 @@ class _WalletScreenState extends State<WalletScreen> {
                           runSpacing: 10,
                           children: [
                             ValueDataRow(
-                                type: ValueType.amount,
-                                value: sendPaymentChangeNotifier
-                                    .pendingPayment?.decodedInvoice.amountSats,
+                                value: AmountText(
+                                    amount: (sendPaymentChangeNotifier
+                                        .pendingPayment
+                                        ?.decodedInvoice
+                                        .amountSats)!), // TODO: what to do when null
                                 label: "Amount"),
                             ValueDataRow(
-                                type: ValueType.text,
-                                value:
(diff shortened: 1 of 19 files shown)


mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -70,14 +70,15 @@ class _WalletScreenState extends State<WalletScreen> {
...
                           children: [
                             ValueDataRow(
-                                type: ValueType.amount,
-                                value: sendPaymentChangeNotifier
-                                    .pendingPayment?.decodedInvoice.amountSats,
+                                value: AmountText(
+                                    amount: (sendPaymentChangeNotifier
+                                        .pendingPayment
+                                        ?.decodedInvoice
+                                        .amountSats)!), // TODO: what to do when null
```
bonomat, line 77: Maybe we can omit the whole row if the value is null?

bonomat (approved):
Looks good, thanks.

mobile/lib/features/wallet/wallet_screen.dart
klochowicz: right now the `ValueDataRow` will display the row, but show an empty space without the value. We could modify that behaviour, but it's a decent fallback (the layout doesn't change).

Restioson: > right now the ValueDataRow will display the row, but show an empty space without the value

really? to me it seems like the null value would be cast to an `Amount` which would cause an exception

PR #1483, opened 2023-10-13
feat: Send on chain funds

holzeis:
- Pay zero amount invoice
- Register payment success and failed event handlers
- Show confirm dialog on sending payment
- Send payment to onchain address or BIP21 uri
- Redesign send screen

send-receive


Restioson (approved):
🎉 Thanks for the improvements and taking it across the finish line!

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -223,18 +224,32 @@ where
         Ok(route_hint_hop)
     }
 
-    pub fn send_payment(&self, invoice: &Invoice) -> Result<()> {
-        let (status, err) = match pay_invoice(invoice, Retry::Attempts(10), &self.channel_manager) {
+    pub fn pay_invoice(&self, invoice: &Invoice, amount: Option<u64>) -> Result<()> {
+        let (result, amt_msat) = match invoice.amount_milli_satoshis() {
```
bonomat, line 228: I'm not sure I follow this logic: 
Are you saying: if there is no amount in the `invoice`, then take the `amount`, else, take the `invoice`'s amount. 
However, you let the user enter an `amount` in the frontend nevertheless?

holzeis: Yes, so if the invoice does not have an amount we expect an amount to be provided. It's optional because I do not know upfront if this is a zero amount invoice or not.

bonomat (approved):
LGTM. Shall we trigger a release once this is in and tested on our dev env?

holzeis:
> LGTM. Shall we trigger a release once this is in and tested on our dev env?

Let's go for it. 


We might as well add this https://github.com/get10101/10101/pull/1485

Just needs your confirmation that it works for android as well.

PR #1488, opened 2023-10-16
 fix: Notify users on websocket if online for rollover

holzeis:
If the app is in background it might still be connected to our websocket. In that case we don't have to notify the user via push notification, but can directly send the websocket message.

Note, it might happen that the user has the app in the background and does not open the app again at the time of notification. In that case the app would disconnect eventually, the app wouldn't have rolled over (unless we are very lucky and the app is not in the background for long) and the user would have missed that push notification. However, we are sending atm 3 push notifications in total during the rollover weekend, so that should be fine.


coordinator/src/scheduler.rs
```diff
@@ -133,17 +152,37 @@ fn build_rollover_notification_job(
...
                         {
-                            tracing::error!("Failed to send {notification:?} notification: {e:?}");
+                            Err(e) => {
+                                tracing::error!(%trader_id, "Couldn't find signed channel. Skipping proposing rollover. Error: {e:#}");
+                                continue;
+                            }
+                            Ok(signed_channel) => signed_channel,
+                        };
+
+                        tracing::debug!(%trader_id, position_id=position.id, "Proposing to rollover users position");
```
luckysori, line 175:
```suggestion
                        tracing::debug!(%trader_id, position_id=position.id, "Proposing to rollover user's position");
```

coordinator/src/scheduler.rs
```diff
@@ -133,17 +152,37 @@ fn build_rollover_notification_job(
...
+                let node = node.clone();
                 async move {
-                    for (position, fcm_token) in positions_with_token {
-                        tracing::debug!(trader_id=%position.trader, "Sending reminder to rollover position.");
-                        if let Err(e) = notification_sender
-                            .send(Notification::new(fcm_token.clone(), notification.clone()))
-                            .await
+                    for position in positions {
+                        let trader_id = position.trader;
+                        tracing::debug!(%trader_id, "Sending reminder to rollover position.");
```
luckysori, line 162:
```suggestion
                        tracing::debug!(%trader_id, "Sending rollover reminder.");
```

coordinator/src/scheduler.rs
```diff
@@ -133,17 +152,37 @@ fn build_rollover_notification_job(
...
+
+                        let message = OrderbookMessage::TraderMessage {
+                            trader_id,
+                            message: Message::Rollover(
+                                signed_channel.get_contract_id().map(hex::encode),
+                            ),
+                            notification: Some(notification.clone()),
+                        };
+                        if let Err(e) = notifier.send(message).await {
+                            tracing::error!(%trader_id, "Failed to notify user to rollover their position. Error: {e:#}");
```
luckysori, line 185:
```suggestion
                            tracing::error!(%trader_id, "Failed to send rollover reminder: {e:#}");
```

coordinator/src/scheduler.rs
```diff
@@ -133,17 +152,37 @@ fn build_rollover_notification_job(
...
+                        tracing::debug!(%trader_id, "Sending reminder to rollover position.");
+
+                        let signed_channel = match node
+                            .inner
+                            .get_signed_channel_by_trader_id(trader_id)
                         {
-                            tracing::error!("Failed to send {notification:?} notification: {e:?}");
+                            Err(e) => {
+                                tracing::error!(%trader_id, "Couldn't find signed channel. Skipping proposing rollover. Error: {e:#}");
+                                continue;
```
luckysori, line 170: 🙃 I think it would be nicer to get rid of the explicit `continue` by defining a helper function to process each `position`. That way we can use `?` and return a `Result` in the helper and simply log for all returned errors in the `for` loop.

crates/ln-dlc-node/src/node/dlc_manager.rs
```diff
@@ -64,3 +68,26 @@ pub fn signed_channel_state_name(signed_channel: &SignedChannel) -> String {
...
+    P: Send + Sync,
+{
+    pub fn get_signed_channel_by_trader_id(&self, trader_id: PublicKey) -> Result<SignedChannel> {
+        let channel = self.get_dlc_channel_signed(&trader_id)?.with_context(|| {
+            format!("Expect signed dlc channel to exist. trader_id={trader_id}")
+        })?;
+
+        let dlc_channel_id = channel
+            .get_dlc_channel_id(0)
+            .with_context(|| format!("Expected to get dlc_channel id. trader_id={trader_id}"))?;
```
luckysori, line 83: 🔧 We can `expect` here since a `Signed` channel must have a DLC channel ID.

crates/ln-dlc-node/src/node/dlc_manager.rs
```diff
@@ -64,3 +68,26 @@ pub fn signed_channel_state_name(signed_channel: &SignedChannel) -> String {
...
     name.to_string()
 }
+
+impl<P> Node<P>
+where
+    P: Send + Sync,
+{
+    pub fn get_signed_channel_by_trader_id(&self, trader_id: PublicKey) -> Result<SignedChannel> {
+        let channel = self.get_dlc_channel_signed(&trader_id)?.with_context(|| {
+            format!("Expect signed dlc channel to exist. trader_id={trader_id}")
```
luckysori, line 78: 🙃 The way you wrote this originally sounded like an `expect`. 

```suggestion
            format!("Could not find signed DLC channel. trader_id={trader_id}")
```

luckysori (approved):

PR #1489, opened 2023-10-16
feat: Add stable flag to orders and positions

### LINKED ISSUE #1441: Add flag to position indicating if its a synthetic usd position
We currently imply that information from the direction and the leverage, but we do not know if that's actually created as synthetic usd or a regular position on the trading view.

holzeis:
This allows us to know the intent under which a short leverage 1 positions has been created without guessing.

Note, we have far too many models that represent the same thing. 🙈

https://github.com/get10101/10101/assets/382048/67b11d17-4879-4e9c-ab1b-6bbc7283ef6c

resolves #1441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Receive bitcoin in wallet
 - Send bitcoin in wallet to address or BIP21 URI
 - Redesign send and receive screens
+- Add stable flag to position and order
 
 ## [1.4.1] - 2023-10-09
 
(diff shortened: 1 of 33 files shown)


Restioson (approved):
looks like a sensible way to do it!

holzeis:
@luckysori you mentioned we might not need / want that change. Waiting for your review. 😉

luckysori:
> @luckysori you mentioned we might not need / want that change. Waiting for your review. 😉

I think it's fine. It just seemed like unnecessary complexity and not that useful to me because, no matter what the user's intent is, they're the same thing.

If the rest of the team was onboard with this, it's already done and it's small, I can live with it.

klochowicz:
> > @luckysori you mentioned we might not need / want that change. Waiting for your review. 😉
> 
> I think it's fine. It just seemed like unnecessary complexity and not that useful to me because, no matter what the user's intent is, they're the same thing.
> 
> If the rest of the team was onboard with this, it's already done and it's small, I can live with it.

FWIW I think it's unnecessary complexity and could have instead by handled by analytics platform if we care about analysing whether user has clicked a particular button to open this kind of position.
that said, it's done, so if it is useful to others, I'm not opposed.

luckysori (approved):

PR #1491, opened 2023-10-16
fix: Handle null fees if no maker order

holzeis:
In order to calculate the order matching fee a price is required. If we haven't got a price from the maker yet the fee would be null.

The stable screen should probably handle missing orders more gracefully. e.g. by disabling the stabilize or bitcoinize button.

### DIFF
--- a/mobile/lib/features/stable/bitcoinize_confirmation_sheet.dart
+++ b/mobile/lib/features/stable/bitcoinize_confirmation_sheet.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:get_10101/common/domain/model.dart';
 import 'package:get_10101/common/value_data_row.dart';
 import 'package:get_10101/features/stable/stable_dialog.dart';
 import 'package:get_10101/features/stable/stable_value_change_notifier.dart';
@@ -84,7 +85,7 @@ class BitcoinizeBottomSheet extends StatelessWidget {
           const SizedBox(height: 16.0),
           ValueDataRow(
               type: ValueType.amount,
-              value: stableValues.fee,
+              value: stableValues.fee ?? Amount.zero(),
               label: 'Fees',
               valueTextStyle: const TextStyle(fontSize: 18),
               labelTextStyle: const TextStyle(fontSize: 18)),


mobile/lib/features/stable/bitcoinize_confirmation_sheet.dart
```diff
@@ -84,7 +85,7 @@ class BitcoinizeBottomSheet extends StatelessWidget {
           const SizedBox(height: 16.0),
           ValueDataRow(
               type: ValueType.amount,
-              value: stableValues.fee,
+              value: stableValues.fee ?? Amount.zero(),
```
klochowicz, line 88: maybe instead we could consider changing `ValueDataRow` to always present `-` (or some other preconfigured character) on `null`? it would be more intuitive for the user (instead of magic numbers) and much easier to use for the devs?

not sure whether `dynamic` can be asserted for `null` though, but if it can be, that makes everything so much easier :)

holzeis: I wouldn't spent much time on that as this is only happening if there isn't a order available.

Also I guess this will help to improve things https://github.com/get10101/10101/pull/1480

Restioson: Indeed, #1480 would help here as it'd force you to handle the null case. We could move a default val into ValueDataRow if we want, but it might make more sense for it to change depending on what kind of value it is.

Are we sure we don't want to just hide the fee if it's null, though? I learned recently that in dart you can just do
```rust
if (stableValues.fee != null)
    ValueDataRow(...)
```
inside of a list, and if the condition is `false` the item will be omitted.

klochowicz: you can do the `if` solution, but I find personally a bit more elegant wrapping something in `Visibility` widget, as it's clear that something's visibility is toggled. 
that might be subjective preference though, as it amounts to the same 🤷‍♂️

Restioson: >  you can do the if solution, but I find personally a bit more elegant wrapping something in Visibility widget, as it's clear that something's visibility is toggled

I agree that it's subjective - my argument for `if` is that if you use `Visibility`, you still need to provide a default value, whereas with `if` you can just assert with `!` that the value is not null

holzeis: Guys your are discussing this change longer than it took me to create it.


Without this fix we get an ugly null exception if we don't have any orders.

Let's get a fix in. @klochowicz or @Restioson feel free to simply take over that change and apply any patch that you think is necessary from your point of view. Let's just not waste any more time on that insignificant fix.

klochowicz: @holzeis I don't understand your pushback; the general fix I suggested was a **5 minute** (I timed it!) fix, and that applies to all potential places where we could have passed in a null value, and it doesn't falsely proclaim that fees are `0`. (making an empty space).

I have done it for you: https://github.com/get10101/10101/pull/1499 . feel free to close this PR and look at the other fix.

holzeis: > @holzeis I don't understand your pushback; the general fix I suggested was a **5 minute** (I timed it!) fix, and that applies to all potential places where we could have passed in a null value, and it doesn't falsely proclaim that fees are `0`. (making an empty space).
> 
> 
> 
> I have done it for you: https://github.com/get10101/10101/pull/1499 . feel free to close this PR and look at the other fix.

This is getting ridiculous!

PR #1493, opened 2023-10-16
fix: Push modal up when keyboard is shown

holzeis:
Added additional improvements to make the view nice for smaller displays.

Simulator Screenshot - iPhone 15 Pro - 2023-10-16 at 17 11 27

Simulator Screenshot - iPhone SE (3rd generation) - 2023-10-16 at 17 11 13

### DIFF
--- a/mobile/lib/features/wallet/receive_screen.dart
+++ b/mobile/lib/features/wallet/receive_screen.dart
@@ -6,6 +6,7 @@ import 'package:get_10101/common/amount_text.dart';
 import 'package:get_10101/common/amount_text_input_form_field.dart';
 import 'package:get_10101/common/color.dart';
 import 'package:get_10101/common/domain/model.dart';
+import 'package:get_10101/common/scrollable_safe_area.dart';
 import 'package:get_10101/common/snack_bar.dart';
 import 'package:get_10101/features/wallet/application/faucet_service.dart';
 import 'package:get_10101/features/wallet/domain/share_payment_request.dart';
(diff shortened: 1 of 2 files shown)


mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -78,112 +79,112 @@ class _ReceiveScreenState extends State<ReceiveScreen> {
...
-          children: [
-            Padding(
-              padding: const EdgeInsets.all(20.0),
-              child: Column(crossAxisAlignment: CrossAxisAlignment.stretch, children: [
-                const SizedBox(height: 10),
-                Row(
-                  mainAxisAlignment: MainAxisAlignment.spaceEvenly,
-                  children: [
-                    Expanded(
+        body: ScrollableSafeArea(
```
Restioson, line 82: Completely irrelevant to this change, but we might need to take another look at `ScrollableSafeArea` *if* performance becomes a concern - it is using `IntrinsicHeight`, which apparently can be quite expensive.

The reason I say *if* is because premature optimisation is still the root of all evil, and it seems perfectly fine for now

mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -78,112 +79,112 @@ class _ReceiveScreenState extends State<ReceiveScreen> {
...
-                              Text("Lightning",
-                                  style: TextStyle(color: lightningColor, fontSize: 16)),
-                              Icon(Icons.bolt, color: lightningColor),
+                              Text("Bitcoin", style: TextStyle(color: bitcoinColor, fontSize: 16)),
+                              Icon(
+                                Icons.currency_bitcoin,
+                                color: bitcoinColor,
+                              ),
                             ],
-                          )),
```
Restioson, line 107: the diff here is a bit weird; did anything actually change here? did you swap the order of the buttons?

holzeis: It looks indeed weird, if the order has changed it was not intentional.

mobile/lib/features/wallet/receive_screen.dart
klochowicz: if it ever becomes concern, we could try: 
https://github.com/fzyzcjy/flutter_smooth

(same author as FRB)

before we go into manually optimising :-)

klochowicz (approved):

Issue #250, opened 2023-03-13
Inverse payout curve


holzeis:
With https://github.com/get10101/10101/pull/228 we've introduce the close position feature leveraging the `rust-dlc` implementation of the linear payout curve. However, for our btcusd trading pair this payout curve is not correct and should be replaced with an inverse payout curve.

bonomat:
Update, we can use the `PolynomialPayoutCurvePiece` and just put our own value in. The algorithm will then ensure that the steps are linear but that's ok.

This is what I got:


## Our current payout curve

![Image](https://github.com/get10101/10101/assets/224613/563e8f7b-9f43-46db-bfad-f6b220babd80)


## How it should look like with a step function of 1



![Image](https://github.com/get10101/10101/assets/224613/10e66a0d-b811-466a-a999-efb3c134148d)

## How it looks like filled into the `PolynomialPayoutCurvePiece`



![Image](https://github.com/get10101/10101/assets/224613/5d16cee0-2b77-4d9f-b9b8-63d8bfbcde6b)

The number of steps is currently ~350 which was feasible to execute on the simulator.

PR #1497, opened 2023-10-17
feat: Charge order matching fee through the margin

holzeis:
This feature deducts the order matching fee from the margin instead of having to trust the app to pay an order matching fee invoice.

- Removes the 5 seconds delay before submitting an order
- Adds fee and pnl to the trade walltet history item

Notes
- The fees will only get paid into the lightning channel once the position is closed. In other words in a collaborative settlement of a position we are substracting the opening fees and the closing fees from the `accepted_settlement_amount`. 
- The CETs are only generated with the opening fees and do not include a closing fee.

Screenshot 2023-10-17 at 17 44 37


holzeis:
> I'll review this once you've attempted to do it by modifying the collateral in `ContractInput` instead (i.e. without changes to `rust-dlc`).
> 
> (Submitting this to get rid of the pending review.)

Thanks for the suggestion! It works just fine with simply adding the fee to the initial collateral of the accepting party. However, I am wondering if this truly has the same semantics, as with the change in `rust-dlc` we would always get the order matching fees, while in this solution we'd put the fees up into the trade.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -544,13 +537,23 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
                 .trader_margin()
                 .expect("Filled order to have a margin");
 
+            let execution_price = Decimal::try_from(
+                first
+                    .execution_price()
+                    .expect("execution price to be set on a filled order"),
```
bonomat, line 543: You could just return an error here instead of panicking as this function already returns an error.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -560,23 +563,24 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
                 let new_contracts = compute_relative_contracts(order);
                 let updated_total_contracts = total_contracts + new_contracts;
 
+                let execution_price = Decimal::try_from(
```
bonomat, line 566: Same here

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -587,37 +591,53 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
                         open_order.direction,
                     )?;
 
+                    let opening_price = Decimal::from_f32(opening_price).expect("to fit into f32");
+                    let open_position_fee =
+                        order_matching_fee_taker(open_order.quantity, opening_price).to_sat()
+                            as i64;
+                    let pnl_after_open_position_fee = pnl_before_fees + open_position_fee;
+
```
bonomat, line 599: I do not understand: why do you subtract the `open_position_fee` from the PnL?

holzeis: Because the pnl is calculated from the margin that does not include the fees, but the user actually put up the margin with fees. So if we are calculating the pnl it always includes the open position fee. Since we should that fee separately I am deducting it from the pnl.

bonomat: I'm not sure this is correct. In the pnl calculation we fill in the `opening_price`, the `closing_price` and the `leverage`. We do not put in the margin the user has to put up but we calculate the margin from the price and the leverage which does not include the fee.

bonomat: As discussed in a call. Here is a [link](https://docs.google.com/spreadsheets/d/1VD-OnAuYzTYxBerkGifUEWt-XYcXGCl58SCMXXyv-_g/edit?usp=sharing) to our calculations

bonomat (approved):
Looks good. Very nice that we get rid of the extra invoice :) Next one: Jit fee :)

coordinator/src/node.rs
```diff
@@ -325,8 +315,9 @@ impl Node {
         let event_id = format!("{contract_symbol}{maturity_time}");
         tracing::debug!(event_id, "Proposing dlc channel");
         let contract_input = ContractInput {
-            offer_collateral: margin_coordinator,
-            accept_collateral: margin_trader,
+            offer_collateral: margin_coordinator - fee,
+            // the accepting party has do bring in additional margin for the fees
+            accept_collateral: margin_trader + fee,
```
luckysori, line 320: 📣 I think this is wrong: the accepting party (the app) should see their margin _reduced_ not increased.

What we want to see is:

- Coordinator LN channel balance reduced by X.
- App LN channel balance reduced by Y.
- Coordinator DLC collateral = X + fee.
- App DLC collateral = Y - fee.

In the current version the app is earning the fee.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -544,13 +537,23 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
                 .trader_margin()
                 .expect("Filled order to have a margin");
 
+            let execution_price = Decimal::try_from(
+                first
+                    .execution_price()
+                    .context("execution price to be set on a filled order")?,
+            )?;
+            let fee = order_matching_fee_taker(first.quantity, execution_price).to_sat();
+            let amount_sats = amount_sats + fee;
```
luckysori, line 546: 🔧 I think we should leave a comment here as this is not intuitive.

If I understand it correctly, we want to convey that the amount spent in this trade is the sum of the margin + the opening fee.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -544,13 +537,23 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
                 .trader_margin()
                 .expect("Filled order to have a margin");
 
+            let execution_price = Decimal::try_from(
+                first
+                    .execution_price()
+                    .context("execution price to be set on a filled order")?,
```
luckysori, line 543: 🙃 I think you want an `expect` here. Your choice of words seems to indicate that.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -544,13 +537,23 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
...
+            let amount_sats = amount_sats + fee;
+
             trades.push(WalletHistoryItem {
                 flow,
                 amount_sats,
                 timestamp: first.creation_timestamp.unix_timestamp() as u64,
                 status: Status::Confirmed, // TODO: Support other order/trade statuses
                 wallet_type: WalletHistoryItemType::Trade {
                     order_id: first.id.to_string(),
+                    fee_sat: fee,
```
luckysori, line 555: ❓Won't it be confusing to display the opening fee and an amount that already includes the opening fee?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -587,37 +591,48 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
...
+
                     // Closing a position is an inbound "payment", because the DLC channel is closed
                     // into the Lightning channel.
                     let flow = PaymentFlow::Inbound;
                     let amount_sats = (trader_margin as i64 + pnl) as u64;
 
                     trades.push(WalletHistoryItem {
                         flow,
-                        amount_sats,
+                        amount_sats: amount_sats - close_position_fee,
```
luckysori, line 604: 🤔 I don't get why we have to subtract here.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -587,37 +591,48 @@ fn derive_trades_from_filled_orders() -> Result<Vec<WalletHistoryItem>> {
...
                             order_id: order.id.to_string(),
+                            fee_sat: close_position_fee,
+                            pnl: Some(pnl),
                         },
                     });
                 }
                 // Opening the position.
                 else if total_contracts.is_zero() && !updated_total_contracts.is_zero() {
-                    // Closing a position is an outbound "payment", since coins need to leave the
+                    // Opening a position is an outbound "payment", since coins need to leave the
```
luckysori, line 616: 👍Thanks for fixing this. I forgot to do it myself D:

mobile/native/src/ln_dlc/sync_position_to_dlc.rs
```diff
@@ -52,13 +55,20 @@ impl Node {
             Some(SyncPositionToDlcAction::CreatePosition(channel_id)) => {
                 match order::handler::order_filled() {
                     Ok(order) => {
+                        let execution_price =
+                            order.execution_price().context("expect execution price")?;
```
luckysori, line 59: 🔧 Just use `expect` then.

mobile/native/src/ln_dlc/sync_position_to_dlc.rs
```diff
@@ -52,13 +55,20 @@ impl Node {
...
+                        );
+
                         let (accept_collateral, expiry_timestamp) = self
                             .inner
                             .get_collateral_and_expiry_for_confirmed_contract(channel_id)?;
 
                         position::handler::update_position_after_dlc_creation(
                             order,
-                            accept_collateral,
+                            accept_collateral - open_position_fee.to_sat(),
```
luckysori, line 71: 🤔 I don't get why we need to do this.

coordinator/src/node.rs
holzeis: 🤔 But if we substract the fee from the collateral of the user doesn't it mean he simply needs to provide less liquidity while the coordinator has to put up the fee for the trader?

luckysori: In the current version if the DLC goes on-chain at the same price at which it was created (i.e. no price movement):

- The coordinator will get `margin_coordinator - fee`.
- The app will get `margin_trader + fee`.

Surely that's incorrect.

holzeis: As discussed offline, neither approach seems to be correct. I will revert that change and leave it as it has been before, as this should be properly addressed once we are working on #250.

mobile/native/src/ln_dlc/mod.rs
holzeis: you mean its confusing that the amount of the transaction does include the fee?

mobile/native/src/ln_dlc/mod.rs
holzeis: because the users payout is

`trader_margin + pnl - fee`

maybe it would have been clearer if I added it in line 600?

mobile/native/src/ln_dlc/mod.rs
holzeis: you are competing with the review of @bonomat 🙈 https://github.com/get10101/10101/pull/1497#discussion_r1365099744

but maybe I've miss-interpreted his remark?

mobile/native/src/ln_dlc/sync_position_to_dlc.rs
holzeis: https://github.com/get10101/10101/pull/1497#discussion_r1365099744 🙈

PR #1498, opened 2023-10-17
fix: fix android local config

### LINKED ISSUE #1250: When building for Android getting error : "Execution failed for task ':app:processLocalDebugGoogleServices'"
**Describe the bug**
When building for Android, I'm getting the error "Execution failed for task ':app:processLocalDebugGoogleServices'" and the app will not install on a device.

**To Reproduce**

Build the app for Android using 

just all-android

on the command line. 

**Expected behavior**
The app should install on an Android device

**Screenshots**

**Smartphone (please complete the following information):**

Sunmi V2s and a emulator running Android api 33

** Logs
Android app will connect to 172.21.0.241 for 10101 services
Using hardware rendering with device sdk gphone64 x86 64. If you notice graphics artifacts, consider enabling software rendering with
"--enable-software-rendering".
Launching lib/main.dart on sdk gphone64 x86 64 in debug mode...

FAILURE: Build failed with an exception.

Execution failed for task ':app:processLocalDebugGoogleServices'.
> No matching client found for package name 'finance.get10101.app.local'

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

**Additional context**

This is happening on MacOs

Restioson:
Fixes #1250.

### DIFF
--- a/mobile/android/app/src/local/google-services.json
+++ b/mobile/android/app/src/local/google-services.json
@@ -23,6 +23,25 @@
           "other_platform_oauth_client": []
         }
       }
+    },
+    {
+      "client_info": {
+        "mobilesdk_app_id": "1:958970316277:android:9a9795716eaccfdc1e0cbc",
+        "android_client_info": {
+          "package_name": "finance.get10101.app.local"
+        }
+      },
+      "oauth_client": [],
+      "api_key": [
+        {
+          "current_key": "AIzaSyC6pSZrKdL69ch2ltopWjSqmu31icNZL-U"
+        }
+      ],
+      "services": {
+        "appinvite_service": {
+          "other_platform_oauth_client": []
+        }
+      }
     }
   ],
   "configuration_version": "1"


klochowicz (approved):

mobile/android/app/src/local/google-services.json
```diff
@@ -9,7 +9,7 @@
       "client_info": {
         "mobilesdk_app_id": "1:958970316277:android:c8a1d7a58212e3111e0cbc",
         "android_client_info": {
-          "package_name": "finance.get10101.app.demo"
+          "package_name": "finance.get10101.app.local"
```
klochowicz, line 12: 🤔 are you sure this would not break any of the other things that depend on having variant named "demo"? such as our regtest app? 

I guess in the worst case we can revert this simple change if it won't deploy the demo app.

Restioson: Unfortunately, I'm not sure, as I don't know how to test this. However, this only modifies the `google-services.json` for the `local` flavour (is flavour the right word?) and doesn't modify it for the `demo` flavour.

klochowicz: I hinted that I don't mind testing it by trying it out; let's just be prepared for chance of reverting it if anything happens. (which should not be a problem with such a self-contained small commit) :-)

bonomat: Note: this would break push-notifications for local setup (provided it was working in the first place). FCM might throw an error on startup now.
We will need a special google-services.json config for local if we want to have push notification for `.local`.

PR #1499, opened 2023-10-17
fix(ui): Prevent exceptions on null values in ValueDataRow

klochowicz:
A more robust fix for #1491 ; it ensures that we don't need to be cautious whenever we use ValueDataRow; the safe fallback is an empty spot in the value column.


holzeis (approved):
Great, can we move on now?

PR #1507, opened 2023-10-19
fix: a position without a channel can be removed

bonomat:


mobile/native/src/ln_dlc/sync_position_to_dlc.rs
```diff
@@ -36,18 +36,27 @@ impl Node {
...
+
+        let positions = db::get_positions()?;
+        let first_position = positions.first();
+
         let channel_details = match channels.first() {
             Some(channel_details) => channel_details,
-            None => return Ok(()),
+            None => {
+                // if we don't have a channel, but still a position, we can safely close said
+                // position
```
holzeis, line 47: 🔧 a log message would be nice here.

holzeis (approved):

PR #1509, opened 2023-10-19
feat: Improve balance design

holzeis:
Screenshot 2023-10-20 at 11 16 40

- Total is now also adding the USDT balance in sats (amount with unrealized pnl). mid term idea is that you can choose the currency you like to see your total balance in.
- I removed the animation showing the send and receive buttons on the balance. It just didn't fit anymore. I can revert that and try to make it look nicer if somebody objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Charge order matching fee through the margin
+- Improve wallet balance design
 
 ## [1.4.2] - 2023-10-18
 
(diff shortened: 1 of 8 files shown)


luckysori (commented):
I quite like the new combined balance! I do not like how much space is dedicated to the detailed balances though. I propose to remove that section and instead open a modal with the detailed balance when the user clicks on the combined balance (or something like that).

I also propose to remove USDP from the main wallet. I think the USDP balance is already displayed in the USDP tab. I think it ends up being needlessly confusing to combine on-chain + Lightning + USDP.

holzeis:
Sounds good, I can give it a go and we can compare what looks best. 👍

bonomat (approved):
LFG 🚀

Restioson:
I'm not sure if it's this PR that did it but it seems that on latest master the pull-down refresh is no longer working for me on desktop

holzeis:
> I'm not sure if it's this PR that did it but it seems that on latest master the pull-down refresh is no longer working for me on desktop

May I ask you to open an issue for that?

Restioson:
Yep, I will do so shortly

PR #1512, opened 2023-10-20
feat(app): Support on-chain wallet recovery

klochowicz:
- Allow choice between new and imported wallet on first startup
- add screens where user can input the seed (with confirmation step)
- add backend support for seed recovery
- add unit test to doublecheck whether recovered seed is the same as exported

image

image

image

### DIFF
--- a/crates/ln-dlc-node/src/seed.rs
+++ b/crates/ln-dlc-node/src/seed.rs
@@ -43,6 +43,10 @@ impl Bip39Seed {
     /// Initialise a [`Seed`] from a path.
     /// Generates new seed if there was no seed found in the given path
     pub fn initialize(seed_file: &Path) -> Result<Self> {
+        // Ensure parent directory exists
+        if let Some(parent) = seed_file.parent() {
+            create_dir_all(parent)?;
+        }
         let seed = if !seed_file.exists() {
             tracing::info!("No seed found. Generating new seed");
             let seed = Self::new()?;
--- a/crates/ln-dlc-node/src/seed.rs
+++ b/crates/ln-dlc-node/src/seed.rs
@@ -1,4 +1,5 @@
 use anyhow::bail;
+use anyhow::Context;
 use anyhow::Result;
 use bdk::bitcoin;
 use bdk::bitcoin::util::bip32::ExtendedPrivKey;
@@ -7,9 +8,10 @@ use bip39::Mnemonic;
 use bitcoin::Network;
 use hkdf::Hkdf;
 use sha2::Sha256;
+use std::fs::create_dir_all;
 use std::path::Path;
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct Bip39Seed {
     mnemonic: Mnemonic,
 }
@@ -22,6 +24,20 @@ impl Bip39Seed {
         let mnemonic = Mnemonic::generate_in_with(&mut rng, Language::English, word_count)?;
 
         Ok(Self { mnemonic })
+    }
+
+    /// Restore a [`Seed`] from a mnemonic. Writes the seed to the given path.
+    pub fn restore_from_mnemonic(seed_words: &str, target_seed_file: &Path) -> Result<Self> {
+        let mnemonic = Mnemonic::parse(seed_words)?;
+        let seed = Self { mnemonic };
+
+        // Ensure parent directory exists
+        if let Some(parent) = target_seed_file.parent() {
+            create_dir_all(parent)?;
+        }
+        seed.write_to(target_seed_file)
+            .context("cannot write to file")?;
+        Ok(seed)
     }
 
     /// Initialise a [`Seed`] from a path.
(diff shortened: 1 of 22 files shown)


klochowicz:
I've tested recovery with some funds on regtest. I worked! I tried making a video, but sync took way too long to make it worthwhile to upload.

I'll iron out the UI (cancel / confirm placement is odd) and add some more info that  initial sync can take a while (and of course that backup/recovery with seed phrase is *only* for on-chain funds.

crates/ln-dlc-node/src/seed.rs
```diff
@@ -162,4 +177,17 @@ mod tests {
...
+
+    #[test]
+    fn restore_same_seed_from_exported_mnemonic() {
+        let seed = Bip39Seed::new().unwrap();
+        let seed_words = seed.get_seed_phrase().join(" ");
+
+        let restore_path = &temp_dir().join("seed_restored");
+        let seed_restored = Bip39Seed::restore_from_mnemonic(&seed_words, restore_path).unwrap();
+
+        assert!(seed ==  seed_restored, "Restored seed should be the same as the original seed");
```
bonomat, line 189: Any reason why you didn't go with `assert_eq!`?

mobile/lib/common/loading_screen.dart
```diff
@@ -19,11 +23,27 @@ class _LoadingScreenState extends State<LoadingScreen> {
...
+      if (isSeedFilePresent) {
+        runBackend(context).then((value) {
+          logger.i("Backend started");
+        });
+      } else {
+        // No seed file: let the user choose whether they want to create a new
+        // wallet or import their old one
+        GoRouter.of(context).go(WalletChoiceScreen.route);
+        return;
+      }
```
bonomat, line 46: I would let the user enter the email first and then check for the seed file. 
@holzeis / @luckysori : what do you think?

mobile/lib/features/wallet/seed_phrase_import_confirmation.dart
```diff
@@ -0,0 +1,66 @@
...
+    return Scaffold(
+        appBar: AppBar(title: const Text('Import Seed')),
+        body: Column(
+          children: [
+            buildSeedWordsWidget(seedWords, true),
+            const SizedBox(height: 20),
+            Row(
+              mainAxisAlignment: MainAxisAlignment.end,
+              children: <Widget>[
+                ElevatedButton(
```
bonomat, line 31: I guess we don't need the cancel button given this is wrapped in an Scaffold with an AppBar which has a back button.

bonomat (commented):
It works™️: I've checked that the generated nod eid from a recovered seed phrase is identical to the node id when we set up the wallet the first time. 

That's a good start, but it definitely needs some love for the UI before we can merge this.

crates/ln-dlc-node/src/seed.rs
klochowicz: yes, `assert_eq` would need me to derive `Debug`, which in turn would make our mnemonic trivially printable. I would prefer to avoid that.
I'll add a comment to that extent.

mobile/lib/features/wallet/seed_phrase_import_confirmation.dart
klochowicz: You technically don't need it, but it's more obvious to the user this way that they can cancel. 
Not everyone will immediately notice that there's an AppBar on the first run of the app - not all platforms use that way of navigation.

mobile/lib/common/loading_screen.dart
holzeis: I do agree, but a little bit more effort as we can only complete the registration after the node id is known.

klochowicz:
screenshots of improved UI:
- stopped embedding the start inside the app scaffold (moved LoadingScreen into main route, as it should be).

## making the limitations of current solution clear.
image

## improve the font appearance on input screen
image

## improved buttons (and fixed showing the error to the user) 
image

bonomat:
I'll take over

bonomat:
Did some UI changes

https://github.com/get10101/10101/assets/224613/3a34e112-3878-4657-b6ce-68729fe331fb


https://github.com/get10101/10101/assets/224613/4c6a6f74-88b2-4559-90ab-b905fd3ab319

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -264,6 +264,7 @@ pub fn run(data_dir: String, seed_dir: String, runtime: &Runtime) -> Result<()>
             config::get_oracle_info().into(),
         )?;
         let node = Arc::new(node);
+        tracing::info!(node_id = node.info.pubkey.to_hex(), "Node started");
```
luckysori, line 267: 🔧 I think this log is redundant: https://github.com/get10101/10101/blob/ca30d2bde4dfe19f414a16967c8fcc23e0b6232b/crates/ln-dlc-node/src/node/mod.rs#L487.

luckysori (approved):
LGTM! (Mostly concept ACK.)

luckysori:
> Did some UI changes and rewrote quite a bit. Fixed a few bugs on the way and removed unneeded things.

I like how it looks overall. Just a couple of thoughts:

- Do we need to ask for the email address again if they are importing the seed? It would be nice if we could avoid that.
- `Create wallet` and `Restore wallet` buttons are not quite the same size.
- The alignment of the buttons and the list of words is not quite right. Should we center-align the words?

bonomat:
Updated the first screen a bit: 

image

bonomat:
And the other screen: 

image

bonomat:
> > Did some UI changes and rewrote quite a bit. Fixed a few bugs on the way and removed unneeded things.
> 
> I like how it looks overall. Just a couple of thoughts:
> 
>     * Do we need to ask for the email address again if they are importing the seed? It would be nice if we could avoid that.

Ideally yes. For now I think we are fine with asking the user again. I'd add this feature once we have the full recover

>     * `Create wallet` and `Restore wallet` buttons are not quite the same size.

Fixed

> 
>     * The alignment of the buttons and the list of words is not quite right. Should we center-align the words?

I played around with it a bit but liked the current solution most.

bonomat:
Going ahead and merge this to be able to move on :)

PR #1513, opened 2023-10-20
fix: funding tx not found

bonomat:
Apparently it can be that `original_funding_outpoint` is not set. In this case, we can fall back to `funding_txo`.

Also added two more log messages to see what is happening on the coordinator.

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -711,9 +711,11 @@ pub fn collaborative_revert_channel(
     let details = channel_manager
         .get_channel_details(&subchannel.channel_id)
         .context("Could not get channel details")?;
+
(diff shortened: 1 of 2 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -711,9 +711,11 @@ pub fn collaborative_revert_channel(
     let details = channel_manager
         .get_channel_details(&subchannel.channel_id)
         .context("Could not get channel details")?;
+
     let out_point = details
         .original_funding_outpoint
-        .context("Original funding tx didn't have an outpoint")?;
+        .or(details.funding_txo)
```
luckysori, line 717: 👍 I didn't know about `or`!

luckysori (approved):

PR #1514, opened 2023-10-20
fix: Charge fees on calculated settlement amount

holzeis:
After reviewing our initial approach, it turns out there is no easy way of achieving enforced fee payment without any rust-dlc changes. To simplify things I've reverted the corresponding change impacting the payout curve and margins.

A proper solution should be tackled with #250.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -42,7 +42,6 @@ use ln_dlc_node::node::dlc_message_name;
 use ln_dlc_node::node::sub_channel_message_name;
 use ln_dlc_node::node::RunningNode;
 use ln_dlc_node::WalletSettings;
-use orderbook_commons::order_matching_fee_taker;
 use orderbook_commons::MatchState;
 use orderbook_commons::OrderState;
 use rust_decimal::prelude::ToPrimitive;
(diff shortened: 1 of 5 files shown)


coordinator/src/node.rs
```diff
@@ -692,20 +680,20 @@ fn build_payout_function(
     let lower_range = PolynomialPayoutCurvePiece::new(vec![
         PayoutPoint {
             event_outcome: 0,
-            outcome_payout: fee,
+            outcome_payout: 0,
```
bonomat, line 683: It's a shame that this approach did not work. Can you explain why please?

coordinator/src/position/models.rs
```diff
@@ -172,10 +173,10 @@ fn calculate_accept_settlement_amount(
         Direction::Short => short_leverage,
     };
 
-    let margin_trader_without_opening_fees = calculate_margin(opening_price, quantity, leverage);
+    let margin_trader = calculate_margin(opening_price, quantity, leverage);
 
-    let accept_settlement_amount = Decimal::from(margin_trader_without_opening_fees)
-        + Decimal::from(pnl)
+    let accept_settlement_amount = Decimal::from(margin_trader) + Decimal::from(pnl)
```
bonomat, line 178: I didn't notice this in the last PR but instead of doing normal math and then `max(Decimal::Zero)` we should use `checked_sub` here because otherwise this might overflow and we get a gigantic number.

coordinator/src/position/models.rs
```diff
@@ -172,10 +173,10 @@ fn calculate_accept_settlement_amount(
...
+        - Decimal::from(open_position_fee)
```
bonomat, line 179: 🔧
Note that with this approach the confirmation screen in the app is not correct: 

image

The fee will be deducted from the `margin`+`pnl`... so the `Total` amount is wrong as the user only has to provide `margin` initially`.

holzeis: Hmm.. that's true. Should we keep the other approach then?

bonomat: I guess we should get the old "invoice-approach" again in the meantime.

holzeis:
The payout function is incorrect either way. Closing this PR as it doesn't make it better. #250 should get prioritized!

PR #1517, opened 2023-10-20
fix: Only add stable position to balance

holzeis:


luckysori (approved):
It's hard for me to decide what should be in the combined balance, but I am okay with not including positions in general.

holzeis:
> It's hard for me to decide what should be in the combined balance, but I am okay with not including positions in general.

I am fine either way, but before we had only "spendable" balances shown in the wallet.

PR #1518, opened 2023-10-20
Revert "chore: Use 10101 purple instead of green"

holzeis:
This reverts commit 7c7d331cb327bb1bb1207ca2c6bf780c1bbc609a.

Lets maybe digest this a bit, but after looking at the app this evening, I kind of feel like the color green looks better than everything in purple. Happy to close that PR if you are convinced of how it currently looks :)

Screenshot 2023-10-20 at 21 40 14


bonomat (commented):
I have no hard feelings either way.

luckysori (approved):
I like them both, but I'll approve because I don't mind.

PR #1519, opened 2023-10-21
fix: Create route hint on the app side

holzeis:
Gets the current SCID which should be used to identify this channel for outbound payments. This should be used in Routes to describe the first hop or in other contexts where we're sending or forwarding a payment outbound over this channel.
This is either the ChannelDetails::short_channel_id, if set, or the ChannelDetails::outbound_scid_alias. See those for more information.

Obviously, the `inbound_scid_alias` was wrong since it is outbound from the coordinator point of view.

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -120,10 +120,6 @@ pub fn router(
             "/api/prepare_onboarding_payment",
             post(prepare_onboarding_payment),
         )
-        .route(
-            "/api/prepare_regular_payment/:peer_id",
-            post(prepare_regular_payment),
-        )
         .route("/api/newaddress", get(get_unused_address))
         .route("/api/node", get(get_node_info))
         .route("/api/invoice", get(get_invoice))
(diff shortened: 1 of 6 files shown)


bonomat (approved):
Maybe this was the source of our `no_route found` errors all along? 🤔

luckysori (commented):
Cool in theory, but are the [tests failing](https://github.com/get10101/10101/actions/runs/6602287133/job/17941616327?pr=1519#step:8:12260) because of it?

holzeis:
> Cool in theory, but are the [tests failing](https://github.com/get10101/10101/actions/runs/6602287133/job/17941616327?pr=1519#step:8:12260) because of it?

I was using the inbound instead of the outbound alias. Obviously, the inbound doesn't make sense when the coordinator needs to forward the payment on the payment channel.

holzeis:
Even better, we actually do not have to ping the coordinator to create that routing hint. We can simply create the invoice on the app side and use the `inbound_scid_alias`.

bonomat (approved):
LGTM

PR #1522, opened 2023-10-24
chore: Improving collaborative revert

holzeis:
- Collaborative revert is not retried once confirmed.
- The dlc channel and order state is cleaned up after a collab revert, allowing the user to continue using the app without having to re-install the app.
- Fixed calculation issue where the amounts were wrong if the dlc got stuck in a state where the ln channel balances have already been updated.
- Considered `unspendable_punishment_reserve` when calculating the trader and coordinators amount.
- Add dialog showing the user that the collab revert was (not) successful
- Fixed minor typos
- Removed dedicated collaborative revert message as it was redundant to the trader message

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -706,6 +706,7 @@ pub fn collaborative_revert_channel(
     coordinator_address: Address,
     coordinator_amount: Amount,
     trader_amount: Amount,
+    execution_price: Decimal,
 ) -> Result<()> {
     let node = NODE.try_get().context("failed to get ln dlc node")?;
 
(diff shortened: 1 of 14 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -821,6 +825,79 @@ pub fn collaborative_revert_channel(
...
+        }
+        Err(error) => {
+            tracing::error!("Could not delete position : {error:#}");
+        }
+    }
+
+    let mut sub_channel = sub_channel;
+    sub_channel.state = SubChannelState::OnChainClosed;
+
+    let node = NODE.try_get().context("failed to get ln dlc node")?;
```
bonomat, line 892: Might be best to do this at the very beginning of the function to no do any of the other updates if we really can't get the node.

bonomat (approved):
I trust you did not change the calculation since we confirmed the math is correct in the call. Otherwise it looks good.

holzeis:
> I trust you did not change the calculation since we confirmed the math is correct in the call. Otherwise it looks good.

Not sure what call you are referring to, but I did change the calculation as the `unspendable_punishment_reserve` was missing.

PR #1528, opened 2023-10-26
feat: Charge channel opening fee through lsp flow

### LINKED ISSUE #843: JIT - Channel Fee: LSP Flow

### LINKED ISSUE #1147: Invoice descriptions should not be trusted
Currently, we trust the invoice description to correctly identify the type of transaction as well as any associated data, such as an order id (an example of [CWE-502](https://cwe.mitre.org/data/definitions/502.html)).

https://github.com/get10101/10101/blob/65f4a2fb766885ffa9e5ec528d9b5305d7fe8c85/mobile/native/src/ln_dlc/mod.rs#L351-L356

What happens if the user creates their own invoice with the description `taker-fee-NotAnOrderId?!/.sidfj`? Or, worse, what if the user pays an invoice without looking at the description? These would both show up as trade-related payments in the history. This is bad UX at best, and a security vulnerability at worst.

Imagined social engineering attack:
> Scammer: hey, do you want to make a bet on the price of bitcoin with me?
> User: sure! Shall I open a position on 10101 to bet against you?
> Scammer: I'm actually running my own maker! The software is still in beta, so you need to just transfer me the money directly and the app will set up a DLC for us.
> User: okay, I'm not sure about this, so can we try with 1000 sats first?
> Scammer: sure, here is the invoice!
> [User pays invoice]
> Scammer: check your wallet history and see the matching fee has been paid! The open position will show in the history soon. _(ed: it will not)_
> User: okay, I'll transfer you $1000 now for the bet :)

This is slightly contrived, as we do not expect any of our users to fall for this especially since the beta users should have some idea of how the app works. Note that this scam can still progress to the point of paying 1000 sats in the above example even if this is fixed.

Furthermore, in general, parsing untrusted data is not recommended to be done lightly (see [OWASP - Deserialization of untrusted data](https://owasp.org/www-community/vulnerabilities/Deserialization_of_untrusted_data)). Most parsers are not designed to handle untrusted data. What we are doing is probably fine for now (just stripping the prefix), but this could change and introduce a vulnerability.

There should be another way to figure out the type of payment from a trusted source. Perhaps this could be stored in the database? Presumably order info is already stored there.

If we can authenticate that the invoice was created by the user themselves, then we can also trust the description. This might be a bit brittle, though, since it might seem as though we can trust the description in all cases (when in reality we can only trust it if we know the user generated the payment). Maybe, in that case, it is best to also store the associated data in the database.

### LINKED ISSUE #1366: flaky tests::just_in_time_channel::create::open_jit_channel failed without payed jit channel opening fees on CI
Since we update the transaction fee costs in an async task I suspect this test failed, because the task hasn't picked up that transaction yet, leaving the fees at 0 at the time of assertion.

https://github.com/get10101/10101/actions/runs/6296191933/job/17090712299

```
thread 'tests::just_in_time_channel::create::open_jit_channel' panicked at 'assertion failed: transaction.fee() > 0', crates/ln-dlc-node/src/tests/just_in_time_channel/create.rs:97:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:117:5
   3: ln_dlc_node::tests::just_in_time_channel::create::open_jit_channel::{{closure}}
             at ./src/tests/just_in_time_channel/create.rs:97:5
   4: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/future/future.rs:125:9
   5: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-0.1.37/src/instrument.rs:272:9
   6: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/park.rs:283:63
   7: tokio::runtime::coop::with_budget
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/coop.rs:107:5
   8: tokio::runtime::coop::budget
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/coop.rs:73:5
   9: tokio::runtime::park::CachedParkThread::block_on
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/park.rs:283:31
  10: tokio::runtime::context::BlockingRegionGuard::block_on
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/context.rs:315:13
  11: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/scheduler/multi_thread/mod.rs:66:9
  12: tokio::runtime::runtime::Runtime::block_on
             at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.28.2/src/runtime/runtime.rs:304:45
  13: ln_dlc_node::tests::just_in_time_channel::create::open_jit_channel
             at ./src/tests/just_in_time_channel/create.rs:97:5
  14: ln_dlc_node::tests::just_in_time_channel::create::open_jit_channel::{{closure}}
             at ./src/tests/just_in_time_channel/create.rs:23:29
  15: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
  16: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
FAILED
```

### LINKED ISSUE #1445: Route not found when paying JIT fee
That's new: we are paying directly our counter party. 
```
2023-10-06 13:52:09.974  LogLevel.TRACE  lightning::routing::router: Searching for a route from payer 03e2fca3bc26e4e3552c5eacb009e8e365b8e23627bbaa54400713bd98ab4f5e69 to payee 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802 with MPP and 0 first hops overriding the network graph   
2023-10-06 13:52:09.975  LogLevel.ERROR  Failed to send payment err: RouteNotFound  
```

### LINKED ISSUE #1521: Pnl is not shown fully
image

holzeis:
This change was fairly simple with the new channel config flag `accept_underpaying_htlcs` 🎉. The main change is in this commit abb2feb62f44c5ef38baa200f73c34fb986089df

- open channel fee + funding transaction is now shown on the payment which triggered the inbound channel creation
- cleanup of all occurrences where we were using the order matching fee or jit channel fee invoice

resolves #843 
fixes #1147
fixes #1366
fixes #1445

https://github.com/get10101/10101/assets/382048/349187ff-cf94-401b-a53d-cae9fd9d0a5b

---
Unrelated minor fixes that could get separated into a single PRs.

- c89154cfc4318c5d441408582862f87f1bc4e2f4: The app does not need to maintain a pending intercepted htlc map.
- 114fd527f176849818dbe937caddb3f1c6a7a930: We have to fail the payment backwards in case we do not get a pre-image
- 643b32998c94319900db80e2ab440442d7bc4fa6: Add an argument to optionally do not truncate the provided value. fixes #1521 
- 82a712f27a950ae8432dd9ae3188d8209de634c7: Fix an invalid sql file


crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -180,9 +174,10 @@ where
                 reason,
                 user_channel_id,
             } => {
+                let pending_intercepted_htlcs = Arc::new(Mutex::new(HashMap::new()));
```
bonomat, line 194:
```suggestion
                // we won't intercept htlcs in the app, hence we can pass an empty map
                let pending_intercepted_htlcs = Arc::new(Mutex::new(HashMap::new()));
```

crates/ln-dlc-node/src/config.rs
```diff
@@ -53,6 +53,7 @@ pub fn app_config() -> UserConfig {
         },
         channel_config: ChannelConfig {
             cltv_expiry_delta: MIN_CLTV_EXPIRY_DELTA,
+            accept_underpaying_htlcs: true,
```
bonomat, line 56: I assume you didn't want this change in the commit: 4af7474602044b343783d7dd49689f0ad722085e but in the next

crates/ln-dlc-node/src/ln/coordinator_event_handler.rs
```diff
@@ -328,12 +328,13 @@ where
             "Pending intercepted HTLC found, forwarding payment"
         );
 
+        let fee_msat = channel.fee_sats.map(|fee| fee * 1000).unwrap_or(0);
```
bonomat, line 333: Same as above, this shouldn't be in this commit.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -233,12 +250,66 @@ where
...
+                    let user_channel_id = UserChannelId::from(user_channel_id);
+                    let channel = self
+                        .node
+                        .storage
+                        .get_channel(&user_channel_id.to_string())?
+                        .with_context(|| {
+                            format!("Couldn't find channel for user_channel_id {user_channel_id}",)
+                        })?;
+
+                    if channel.channel_state == ChannelState::Open {
```
bonomat, line 271: ❓ 
I don't understand, why would you fail if the channel is already open? 
Wouldn't that prevent you from receiving payments?

bonomat (approved):
Awesome, looking forward to having this in production :)

bonomat:
> Unrelated minor fixes that could get separated into a single PRs.
> 
>     * [43d4999](https://github.com/get10101/10101/commit/43d49997f0b31ac2958398d7bf90492e8d365c0c): The app does not need to maintain a pending intercepted htlc map.
> 
>     * [97f8db4](https://github.com/get10101/10101/commit/97f8db4ad61bcd94b09a86eff71e97467510d8d3): We have to fail the payment backwards in case we do not get a pre-image
> 
>     * [4af7474](https://github.com/get10101/10101/commit/4af7474602044b343783d7dd49689f0ad722085e): Add an argument to optionally do not truncate the provided value. fixes


You might want to top getting some of the fixes in separate PRs straight into `main` and not into the `116`-upgrade. I assume 116 might take a bit longer to land.

crates/ln-dlc-node/src/ln/app_event_handler.rs
holzeis: Because then it is already been paid for.

ChannelReady -> OpenUnpaid
PaymentClaimable -> Open

bonomat: I still don't understand: isn't the event `PaymentClaimable` called for _all_ payments? Meaning, even after the channel was opened paid successfully, if the user tries to receive something, we would fail here?

holzeis: Only if `counterparty_skimmed_fee_msat > 0`, or in other words only if the counterparty is skimming fees from the payment.

coordinator/migrations/2023-10-25-120015_add_fee_to_channel/down.sql
```diff
@@ -0,0 +1,8 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE channels DROP COLUMN "fee_sats";
+
+ALTER TABLE
+    channels
+    ADD
+        COLUMN open_channel_fee_payment_hash TEXT REFERENCES payments(payment_hash);
+CREATE INDEX IF NOT EXISTS channels_funding_txid ON channels(funding_txid);
```
luckysori, line 8: 🔧 Missing newlines here and below.

coordinator/src/node/storage.rs
```diff
@@ -161,6 +164,11 @@ impl node::Storage for NodeStorage {
         Ok(channel)
     }
 
+    fn get_channel_by_payment_hash(&self, _payment_hash: String) -> Result<Option<Channel>> {
+        // the payment hash is not stored on the coordinator side.
+        Ok(None)
```
luckysori, line 169: 🔧 `unimplemented!` is saner if this is never called. Or are we actually calling it? 😆

crates/ln-dlc-node/src/config.rs
luckysori: 🔧 Documenting this config was always really helpful.

```suggestion
            // Allows the coordinator to charge us a channel-opening fee after intercepting the app's funding HTLC.
            accept_underpaying_htlcs: true,
```

crates/ln-dlc-node/src/lib.rs
```diff
@@ -106,6 +107,9 @@ pub struct PaymentInfo {
     pub timestamp: OffsetDateTime,
     pub description: String,
     pub invoice: Option<String>,
+    /// If the payment was used to open an inbound channel, this tx id refers the funding
+    /// transaction for opening the channel.
+    pub funding_txid: Option<Txid>,
```
luckysori, line 112: 🙃 Perhaps not worth it here, but it's nice to think about ways to avoid optional fields.

For instance, we could extend the `PaymentFlow` enum to something like:

```rust
enum PaymentType {
    Inbound,
    Outbound,
    Funding {
        funding_txid: Txid,
    }
}
```

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -234,12 +251,66 @@ where
...
                 amount_msat,
-                counterparty_skimmed_fee_msat: _,
+                counterparty_skimmed_fee_msat,
                 purpose,
                 via_channel_id: _,
-                via_user_channel_id: _,
+                via_user_channel_id,
                 claim_deadline: _,
             } => {
+                if counterparty_skimmed_fee_msat > 0 {
```
luckysori, line 260: 🔧 A [`match` guard](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html) would be more idiomatic and clearer, and you would avoid some indentation.

crates/ln-dlc-node/src/ln/app_event_handler.rs
luckysori: I guess this scenario means that the coordinator is trying to take extra fees from random payments _after_ the JIT channel has already been created and the opening fee has been paid.

So the app should reject any payment that is taking extra fees arbitrarily. Perhaps eventually it would make sense to show this to the user, so that they can decide whether they want to pay for the special fee or not.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -234,12 +251,66 @@ where
...
+                    }
+
+                    let channel_fee_msats = channel.fee_sats.unwrap_or(0) * 1000;
+                    if channel_fee_msats < counterparty_skimmed_fee_msat {
+                        tracing::error!("Counterparty skimmed too much fee. Expected fee {channel_fee_msats} msats, Got fee {counterparty_skimmed_fee_msat} msats Rejecting payment!");
+                        self.node.channel_manager.fail_htlc_backwards_with_reason(
+                            &payment_hash,
+                            FailureCode::IncorrectOrUnknownPaymentDetails,
+                        );
+                        return Ok(());
```
luckysori, line 290: 🔧 Extracting the new handler into a function and returning a `Result` will mean that we don't have to duplicate failing the HTLC backwards.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -234,12 +251,66 @@ where
...
+                    }
+
+                    let channel_fee_msats = channel.fee_sats.unwrap_or(0) * 1000;
+                    if channel_fee_msats < counterparty_skimmed_fee_msat {
+                        tracing::error!("Counterparty skimmed too much fee. Expected fee {channel_fee_msats} msats, Got fee {counterparty_skimmed_fee_msat} msats Rejecting payment!");
```
luckysori, line 285: 🔧 Prefer event fields.

crates/ln-dlc-node/src/ln/app_event_handler.rs
```diff
@@ -234,12 +251,66 @@ where
...
+                            FailureCode::IncorrectOrUnknownPaymentDetails,
+                        );
+                        return Ok(());
+                    }
+
+                    tracing::info!(
+                        funding_txid=?channel.funding_txid,
+                        "Paying channel opening fee of {}",
+                        counterparty_skimmed_fee_msat / 1000
+                    );
```
luckysori, line 297: 🙃 I would reword this because it seems to indicate that we are making a payment, but we are really accepting a payment with a reduced amount.

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
```diff
@@ -33,10 +33,12 @@ async fn ln_collab_close() {
     .await
     .unwrap();
 
+    let fee_sats = 10_000;
```
luckysori, line 36: 🔧 Should this not come from the liquidity options or something like that?

mobile/native/migrations/2023-10-25-120015_add_fee_to_channel/down.sql
```diff
@@ -0,0 +1,5 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE "channels" DROP COLUMN "fee_sats";
+ALTER TABLE "channels" DROP COLUMN "open_channel_payment_hash";
+
+ALTER TABLE "payments" DROP COLUMN "funding_txid";
```
luckysori, line 5: 🔧 Missing newline.

luckysori (approved):
LGTM.

coordinator/src/node/storage.rs
holzeis: it shouldn't be called :)

crates/ln-dlc-node/src/lib.rs
holzeis: I like it 👍

crates/ln-dlc-node/src/tests/just_in_time_channel/channel_close.rs
holzeis: technically you are right, but on these tests we do not know about the liquidity options. This is only known on the application layer of the coordinator and the app. I am using a little hack to inject the fees in the liquidity request, but for that I have to do it manually.

see also https://github.com/get10101/10101/pull/1528/files#diff-a6f251cebbb64254ec28d79e916edc0275557625800fae9588e2cdc2598a545dR270-L276

PR #1530, opened 2023-10-26
 fix: allow collab revert even if ln channel is closed

bonomat:
By retrieving the channel signer through the channel monitor we are able to access the fund key of the original ln channel even though ldk closed and removed this channel.

Note: this patch includes a temporary dependency on rust-lightning. This will not be needed anymore once we upgrade to rust-lightning 116 because we can then access the channel keys directly.
Differently said, we might be able to drop this commit completely once we upgrade to rust-lightning 116

Note 2: the api to collab revert has changed: it now needs the funding tx id and vout, e.g. 

```bash
curl -d '{"channel_id":"1be2ef992aaae712cfb74a70da3f5295eb7fcfa711b13483d3c48a3767c446bd", "price":"28394", "fee_rate_sats_vb": 4, "txid": "bc46c467378ac4d38334b111a7cf7feb95523fda704ab7cf12e7aa2a99efe21b", "vout": 1}' -H "Content-Type: application/json" -X POST http://localhost:8000/api/admin/channels/revert
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Improve collab revert
+
 ## [1.4.3] - 2023-10-23
 
 - Charge order matching fee through the margin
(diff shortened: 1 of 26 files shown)


bonomat:
closed in favor of: https://github.com/get10101/10101/pull/1531

PR #1531, opened 2023-10-26
fix: allow collab revert even if ln channel is closed

bonomat:
By retrieving the channel signer through the channel monitor we are able to access the fund key of the original ln channel even though ldk closed and removed this channel.

Note: this patch includes a temporary dependency on rust-lightning. This will not be needed anymore once we upgrade to rust-lightning 116 because we can then access the channel keys directly.
Differently said, we might be able to drop this commit completely once we upgrade to rust-lightning 116

Note 2: the api to collab revert has changed: it now needs the funding tx id and vout, e.g. 

```bash
curl -d '{"channel_id":"1be2ef992aaae712cfb74a70da3f5295eb7fcfa711b13483d3c48a3767c446bd", "price":"28394", "fee_rate_sats_vb": 4, "txid": "bc46c467378ac4d38334b111a7cf7feb95523fda704ab7cf12e7aa2a99efe21b", "vout": 1}' -H "Content-Type: application/json" -X POST http://localhost:8000/api/admin/channels/revert
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,12 @@ dlc-trie = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "9815582"
 simple-wallet = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "9815582" }
 
 # We should usually track the `p2pderivatives/split-tx-experiment[-10101]` branch
-lightning = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
-lightning-background-processor = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
-lightning-transaction-sync = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
-lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
-lightning-persister = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "3b69cecf" }
+# Temporary fix: this fork allows us to access the channel monitor's channel_key_ids
+lightning = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "698be7fb" }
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -26,13 +26,20 @@ dlc-trie = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "9815582"
...
+lightning-transaction-sync = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "698be7fb" }
+lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "698be7fb" }
+lightning-persister = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "698be7fb" }
 
 rust-bitcoin-coin-selection = { git = "https://github.com/p2pderivatives/rust-bitcoin-coin-selection" }
 
 # Waiting for the next release
 xtra = { git = "https://github.com/Restioson/xtra/", rev = "d98393a" }
+
+# lightning = {path = "/Users/bonomat/src/github/p2pderivatives/rust-lightning/lightning"}
```
holzeis, line 41: 🔧 I guess you forgot to delete that 🙂

coordinator/src/admin.rs
```diff
@@ -160,13 +161,18 @@ pub async fn collaborative_revert(
         AppError::BadRequest("Invalid channel id provided".to_string())
     })?;
 
+    let out_point = OutPoint {
```
holzeis, line 164: 🤔 can't we fetch that from the shadow channel?

holzeis (approved):
LGTM 👍 

Would be nice if we wouldn't have to provide the outpoint in the params.

The funding_txid is on the channel, but I guess the vout would have to be added.

coordinator/src/admin.rs
bonomat: in theory we could fetch it from a block explorer. But I'm not sure if it is worth the hassle.

PR #1535, opened 2023-10-27
chore: Set background of scaffold on new wallet

holzeis:
- set background of scaffold to white to match the rest of the screen
- use dark status bar, needs to be done here again since it is not in the shell route.
- reduced the button border radius from 18 to 10.

Screenshot 2023-10-27 at 09 43 09


bonomat (approved):
Go for it :)

Issue #1538, opened 2023-10-27
Add max amount button to send out on-chain funds


holzeis:
A max button that sets the amount to the maximum the user can send out.

bonomat:
Afaik if you put in `0` it will send out max.

holzeis:
🙈

![image](https://github.com/get10101/10101/assets/382048/379f5989-2ac4-458f-8210-b219811cde1a)

bonomat:
The logic is implemented though 😅

https://github.com/get10101/10101/blob/68958755e16f8cb3eec4a08de507ff2e4059b74c/crates/ln-dlc-node/src/ldk_node_wallet.rs#L199-L210

bonomat:
Although, I believe it is flawed, it should be `None` or `0`

PR #1541, opened 2023-10-29
feat: drain on-chain wallet

bonomat:
Resolves https://github.com/get10101/10101/issues/1538

### DIFF
--- a/mobile/lib/features/wallet/send/send_screen.dart
+++ b/mobile/lib/features/wallet/send/send_screen.dart
@@ -38,7 +38,7 @@ class _SendScreenState extends State<SendScreen> {
   ChannelInfo? channelInfo;
 
   Destination? _destination;
-  Amount? _amount;
+  Amount _amount = Amount.zero();
 
   final TextEditingController _controller = TextEditingController();
 
@@ -64,7 +64,7 @@ class _SendScreenState extends State<SendScreen> {
         if (destination != null) {
           _destination = destination;
           _amount = destination.amount;
-          _controller.text = _amount!.formatted();
+          _controller.text = _amount.formatted();
 
           _invalidDestination = false;
           _valid = _formKey.currentState?.validate() ?? false;
@@ -103,7 +103,7 @@ class _SendScreenState extends State<SendScreen> {
                           setState(() {
                             _destination = destination;
                             _amount = destination.amount;
-                            _controller.text = _amount!.formatted();
+                            _controller.text = _amount.formatted();
 
                             _invalidDestination = false;
                             _valid = _formKey.currentState?.validate() ?? false;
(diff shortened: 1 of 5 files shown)


holzeis (approved):
I guess it works, although I am not a fan of misusing the amount input field for that.

Maybe a drain button, that fetches the max amount that can be spent and sets it into the amount field would be better?

bonomat:
> I guess it works, although I am not a fan of misusing the amount input field for that.

Imho that's a defacto standard in Bitcoin wallets.

> Maybe a drain button, that fetches the max amount that can be spent and sets it into the amount field would be better?

This is very hard to calculate in advance, you'd need to use the drain function first, then create the transaction, and then get the output's amount and return it back to the UI.

mobile/lib/features/wallet/send/send_screen.dart
```diff
@@ -196,8 +197,8 @@ class _SendScreenState extends State<SendScreen> {
 
                   final amount = Amount.parseAmount(value);
 
-                  if (amount.sats <= 0) {
-                    return "Amount is mandatory";
+                  if (amount.sats < 0) {
```
holzeis, line 200: 🔧 this will also impact paying a lightning invoice.

We need to check if the we are paying on-chain to relax that validation rule.

mobile/lib/features/wallet/send/send_screen.dart
```diff
@@ -176,13 +176,14 @@ class _SendScreenState extends State<SendScreen> {
...
+              const Text("Amount in sats (0 to drain the wallet)",
+                  style: TextStyle(fontWeight: FontWeight.bold)),
               const SizedBox(height: 2),
               AmountInputField(
                 controller: _controller,
                 label: "",
-                value: _amount ?? Amount.zero(),
-                enabled: _destination != null && _destination!.amount.sats == 0,
+                value: _amount,
+                enabled: _destination != null,
```
holzeis, line 186: 🔧 this shouldn't be editable if the amount is set on the invoice or bip21 uri.

holzeis (changes requested):
@bonomat I forgot that this screen is also used for sending lightning. I think your changes will not work in that case.

mobile/lib/features/wallet/send/send_screen.dart
bonomat: fixed

mobile/lib/features/wallet/send/send_screen.dart
bonomat: fixed

holzeis: This is still editable if the amount is set. Isn't it?

bonomat: no, it's not editable, the logic is unchanged to before.

holzeis (approved):

PR #1542, opened 2023-10-30
feat: payout curve

bonomat:
Took me longer than expected. 

image

Also plotting the computed points: 

image

A PR will follow once this is in which tackles:
- Apply fee in margin
- Cut-off payout points which are below dust

see here: https://github.com/get10101/10101/issues/1548

### DIFF
--- a/coordinator/src/payout_curve.rs
+++ b/coordinator/src/payout_curve.rs
@@ -1,91 +1,100 @@
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+use dlc_manager::contract::numerical_descriptor::NumericalDescriptor;
+use dlc_manager::contract::ContractDescriptor;
 use dlc_manager::payout_curve::PayoutFunction;
 use dlc_manager::payout_curve::PayoutFunctionPiece;
 use dlc_manager::payout_curve::PayoutPoint;
 use dlc_manager::payout_curve::PolynomialPayoutCurvePiece;
-use rust_decimal::prelude::ToPrimitive;
+use dlc_manager::payout_curve::RoundingIntervals;
 use rust_decimal::Decimal;
-use trade::cfd::calculate_long_liquidation_price;
-use trade::cfd::calculate_short_liquidation_price;
-use trade::cfd::BTCUSD_MAX_PRICE;
+use trade::ContractSymbol;
+use trade::Direction;
 
-/// Builds a [`PayoutFunction`].
+/// Builds the contract descriptor from the point of view of the coordinator.
 ///
-/// TODO: We are currently building a linear payout function for
-/// simplicity. This is *wrong*. We should build an inverse payout
-/// function like we used to do in ItchySats.
-pub fn build_payout_function(
-    total_collateral: u64,
+/// It's the direction of the coordinator because the coordinator is always proposing
+#[allow(clippy::too_many_arguments)]
+pub fn build_contract_descriptor(
     initial_price: Decimal,
-    leverage_long: f32,
-    leverage_short: f32,
+    coordinator_margin: u64,
+    trader_margin: u64,
+    leverage_coordinator: f32,
+    leverage_trader: f32,
+    coordinator_direction: Direction,
     fee: u64,
-) -> anyhow::Result<PayoutFunction> {
-    let leverage_short = Decimal::try_from(leverage_short)?;
-    let liquidation_price_short = calculate_short_liquidation_price(leverage_short, initial_price);
-
-    let leverage_long = Decimal::try_from(leverage_long)?;
-    let liquidation_price_long = calculate_long_liquidation_price(leverage_long, initial_price);
-
-    let lower_limit = liquidation_price_long
-        .floor()
-        .to_u64()
-        .expect("Failed to fit floored liquidation price to u64");
-    let upper_limit = liquidation_price_short
-        .floor()
-        .to_u64()
-        .expect("Failed to fit floored liquidation price to u64");
-
-    let lower_range = PolynomialPayoutCurvePiece::new(vec![
-        PayoutPoint {
-            event_outcome: 0,
-            outcome_payout: fee,
-            extra_precision: 0,
-        },
-        PayoutPoint {
-            event_outcome: lower_limit,
-            outcome_payout: fee,
-            extra_precision: 0,
-        },
-    ])?;
-
-    let middle_range = PolynomialPayoutCurvePiece::new(vec![
-        PayoutPoint {
-            event_outcome: lower_limit,
-            outcome_payout: fee,
-            extra_precision: 0,
-        },
-        PayoutPoint {
-            event_outcome: upper_limit,
-            outcome_payout: total_collateral,
-            extra_precision: 0,
-        },
-    ])?;
-
-    let mut pieces = vec![
-        PayoutFunctionPiece::PolynomialPayoutCurvePiece(lower_range),
-        PayoutFunctionPiece::PolynomialPayoutCurvePiece(middle_range),
-    ];
-
-    // When the upper limit is greater than or equal to the
-    // `BTCUSD_MAX_PRICE`, we don't have to add another curve piece.
-    if upper_limit < BTCUSD_MAX_PRICE {
-        let upper_range = PolynomialPayoutCurvePiece::new(vec![
-            PayoutPoint {
-                event_outcome: upper_limit,
-                outcome_payout: total_collateral,
-                extra_precision: 0,
-            },
-            PayoutPoint {
-                event_outcome: BTCUSD_MAX_PRICE,
-                outcome_payout: total_collateral,
-                extra_precision: 0,
-            },
-        ])?;
-
-        pieces.push(PayoutFunctionPiece::PolynomialPayoutCurvePiece(upper_range));
+    rounding_intervals: RoundingIntervals,
+    quantity: f32,
+    symbol: ContractSymbol,
+) -> Result<ContractDescriptor> {
+    if symbol != ContractSymbol::BtcUsd {
+        bail!("We only support BTCUSD at the moment. For other symbols we will need a different payout curve");
     }
 
-    let payout_function = PayoutFunction::new(pieces)?;
+    Ok(ContractDescriptor::Numerical(NumericalDescriptor {
+        payout_function: build_inverse_payout_function(
+            coordinator_margin,
+            trader_margin,
+            initial_price,
+            leverage_trader,
+            leverage_coordinator,
+            fee,
+            coordinator_direction,
+            quantity,
+        )?,
+        rounding_intervals,
+        difference_params: None,
+        oracle_numeric_infos: dlc_trie::OracleNumericInfo {
+            base: 2,
+            nb_digits: vec![20],
+        },
+    }))
+}
+
+/// Build a [`PayoutFunction`] for an inverse perpetual future e.g. BTCUSD. Perspective is always
+/// from the person who offers, i.e. in our case from the coordinator.
+#[allow(clippy::too_many_arguments)]
+fn build_inverse_payout_function(
+    coordinator_collateral: u64,
+    trader_collateral: u64,
+    initial_price: Decimal,
+    leverage_trader: f32,
+    leverage_coordinator: f32,
+    fee: u64,
+    coordinator_direction: Direction,
+    quantity: f32,
+) -> Result<PayoutFunction> {
+    let payout_points = payout_curve::build_inverse_payout_function(
+        quantity,
+        coordinator_collateral,
+        trader_collateral,
+        initial_price,
+        leverage_trader,
+        leverage_coordinator,
+        fee,
+        coordinator_direction,
+    )?;
+
+    let mut pieces = vec![];
+    for (lower, upper) in payout_points {
+        let lower_range = PolynomialPayoutCurvePiece::new(vec![
+            PayoutPoint {
+                event_outcome: lower.event_outcome,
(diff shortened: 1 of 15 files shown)


crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,670 @@
...
+            // if the coordinator is short, his lower bound is when the trader gets liquidated and
+            // the upper bound is when the coordinator gets liquidated
+            (liquidation_price_long, liquidation_price_short)
+        }
+    };
+    Ok((liquidation_price_lower_bound, liquidation_price_upper_bound))
+}
+
+#[cfg(test)]
+mod tests {
```
luckysori, line 318: 🙃 For ItchySats we used to have property-based tests which were quite useful because there are some properties that you want all your discretised payout curves to respect, for example:

- Total payout (coordinator + trader) is the same (or very close) for all payouts.
- Payout intervals have no gaps.
- Payout intervals are monotonically increasing (this one wasn't always true in ItchySats).

luckysori (commented):
LGTM, but I haven't done a super thorough review yet.

crates/payout_curve/src/lib.rs
bonomat: Yes, I'll look into that because we get random overflow errors sometimes :/

luckysori: I wonder why they are so slow. Have you tried running with a different [config](https://altsysrq.github.io/proptest-book/proptest/tutorial/config.html) to reduce the number of test cases?

bonomat: I believe it's because we iterate over thousands of points. I might be able to optimize it by taking bigger steps, e.g. steps of 10

bonomat:
@holzeis / @luckysori : please review. From my side this is ready to go

holzeis (approved):
I guess it works 🤷‍♂️👍

coordinator/src/payout_curve.rs
```diff
@@ -0,0 +1,104 @@
...
+                event_outcome: upper.event_outcome,
+                outcome_payout: upper.outcome_payout,
+                extra_precision: upper.extra_precision,
+            },
+        ])?;
+        pieces.push(PayoutFunctionPiece::PolynomialPayoutCurvePiece(lower_range));
+    }
+
+    let payout_function =
+        PayoutFunction::new(pieces).context("could not create payout function ")?;
```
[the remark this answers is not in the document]
bonomat: Done

crates/payout_curve/examples/payout_curve_csv.rs
```diff
@@ -0,0 +1,176 @@
...
+use rust_decimal::prelude::ToPrimitive;
+use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
+use std::fs::File;
+use trade::cfd::calculate_long_liquidation_price;
+use trade::cfd::calculate_pnl;
+use trade::cfd::calculate_short_liquidation_price;
+use trade::Direction;
+
+/// The example below will export the computed payout curve and how it should like as CSV
```
luckysori, line 21:
```suggestion
/// The example below will export the computed payout curve and how it should look like as CSV.
```

crates/payout_curve/examples/payout_curve_csv.rs
```diff
@@ -0,0 +1,176 @@
...
+        Decimal::from_f32(leverage_coordinator).context("to be able to parse f32")?,
+        quantity,
+        initial_price,
+        "./crates/payout_curve/examples/should.csv",
+    )?;
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+fn payout_to_csv(
```
luckysori, line 152:
```suggestion
fn discretized_payouts_as_csv(
```

crates/payout_curve/examples/payout_curve_csv.rs
```diff
@@ -0,0 +1,176 @@
...
+            .1
+            .outcome_payout
+            .to_string(),
+    ])?;
+    wtr.flush()?;
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn should_payout(
```
luckysori, line 200:
```suggestion
pub fn actual_payouts_as_csv(
```

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+        } else {
+            (leverage_coordinator, leverage_trader)
+        }
+    };
+
+    let (lower_range_lower, lower_range_upper) = calculate_lower_range_payouts(
+        coordinator_direction,
+        total_collateral,
+        liquidation_price_lower_bound,
+    )?;
```
luckysori, line 72: 🔧 What about `short_liquidation_range_lower`? In ItchySats for the upper one we were using the word `threshold` too.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                },
+            ))
+        })
+        .collect::<Result<Vec<(_, _)>>>()?;
+
+    Ok(pieces)
+}
+
+/// Calculates the lower range payout points between 0 and the lower liquidation point
+fn calculate_lower_range_payouts(
```
luckysori, line 196:
```suggestion
fn calculate_short_liquidation_interval_payouts(
```

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+            .to_u64()
+            .expect("to fit dec into u64"),
+        &last_payout,
+        coordinator_direction,
+        quantity,
+    )?;
+
+    let (_, last_upper_range) = mid_range
+        .last()
+        .context("expect to have at least two elements in payout curve")?
```
luckysori, line 97: 🙃 Slightly weird wording if we are not `expect`ing.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+            .expect("to fit dec into u64"),
+        liquidation_price_upper_bound
+            .to_u64()
+            .expect("to fit dec into u64"),
+        &last_payout,
+        coordinator_direction,
+        quantity,
+    )?;
+
+    let (_, last_upper_range) = mid_range
```
luckysori, line 95: 🔧 It's `last_mid_range`.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+    lower_limit: u64,
+    upper_limit: u64,
+    last_payout: &PayoutPoint,
+    direction: Direction,
+    quantity: f32,
+) -> Result<Vec<(PayoutPoint, PayoutPoint)>> {
+    let total_collateral = trader_collateral + coordinator_collateral;
+
+    let pieces = (lower_limit..=upper_limit - 1)
+        .step_by(2)
```
luckysori, line 137: ❓ Is this the reason why the curve generation is so slow? We probably don't need such a small step when we are discretising.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+    initial_price: Decimal,
+    lower_limit: u64,
+    upper_limit: u64,
+    last_payout: &PayoutPoint,
+    direction: Direction,
+    quantity: f32,
+) -> Result<Vec<(PayoutPoint, PayoutPoint)>> {
+    let total_collateral = trader_collateral + coordinator_collateral;
+
+    let pieces = (lower_limit..=upper_limit - 1)
```
luckysori, line 136: 🔧 A bit simpler.
```suggestion
    let pieces = (lower_limit..upper_limit)
```

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                lower_event_outcome_payout as u64
+            };
+
+            let upper_event_outcome = current_price + 2;
+            let pnl = calculate_pnl(
+                initial_price,
+                Decimal::from(upper_event_outcome),
+                quantity,
+                long_leverage.to_f32().context("to fit into f32")?,
+                short_leverage.to_f32().context("to fit into f32")?,
```
luckysori, line 170: 🔧 These conversions don't need to happen twice. Also, these can be `expect`s.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                0
+            } else if lower_event_outcome_payout as u64 > total_collateral {
+                // we can't payout more than there is in the contract, hence, we need to add this
+                // check
+                total_collateral
+            } else {
+                lower_event_outcome_payout as u64
+            };
+
+            let upper_event_outcome = current_price + 2;
```
luckysori, line 164: 🔧 The step should be a variable or a constant.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+                    .expect("to be able to fit decimal into u64"),
+                outcome_payout: total_collateral,
+                extra_precision: 0,
+            },
+        ),
+        // if coordinator is short, he gets 0 from his liquidation point to infinity
+        Direction::Short => (
+            PayoutPoint {
+                event_outcome: last_payout_point.event_outcome,
+                outcome_payout: last_payout_point.outcome_payout,
```
luckysori, line 264: ❓ Why is this not 0?

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+    total_collateral: u64,
+    last_payout_point: PayoutPoint,
+) -> Result<(PayoutPoint, PayoutPoint)> {
+    let (lower_range_lower, lower_range_upper) = match coordinator_direction {
+        // if coordinator is long, he gets everything from the trader's liquidation point to
+        // infinity
+        Direction::Long => (
+            PayoutPoint {
+                event_outcome: last_payout_point.event_outcome,
+                outcome_payout: last_payout_point.outcome_payout,
```
luckysori, line 249: ❓ Why is this not `total_collateral`?

crates/payout_curve/src/lib.rs
luckysori: Yeah, we definitely don't need so many steps when we want to discretise to a couple hundred CETs anyway.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+        assert_eq!(lower.outcome_payout, total_collateral);
+        assert_eq!(upper.event_outcome, BTCUSD_MAX_PRICE);
+        assert_eq!(upper.outcome_payout, total_collateral);
+
+        if PRINT_CSV {
+            let file = File::create("src/payout_curve/upper_range_long.csv").unwrap();
+            let mut wtr = csv::WriterBuilder::new().delimiter(b';').from_writer(file);
+            wtr.serialize(lower).expect("to be able to write");
+            wtr.serialize(upper).expect("to be able to write");
+            wtr.flush().unwrap();
```
luckysori, line 659: 🙃 Could write a function for this snippet.

luckysori (approved):
LGTM, although I think we should:

- Generate the payout function with a wider step.
- Add more proptests.

crates/payout_curve/src/lib.rs
```diff
@@ -0,0 +1,685 @@
...
+    let total_collateral = coordinator_collateral + trader_collateral;
+
+    // 1. we need to calculate the lower and upper bounds. This depends on the direction of the
+    let (liquidation_price_lower_bound, liquidation_price_upper_bound) =
+        calculate_payout_curve_bounds(
+            initial_price,
+            leverage_trader,
+            leverage_coordinator,
+            coordinator_direction,
+        )?;
```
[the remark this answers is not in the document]
bonomat: It's actually the other way around then: 
the `lower bound` is the `long_liquidation_price` and the `upper bound` is the `short_liquidation_price`.

Happy to rename it.

crates/payout_curve/src/lib.rs
bonomat: it will be `(long_liquidation_range_lower, long_liquidation_range_upper`) because the lower range is when the long person gets liquidated. ,

crates/payout_curve/src/lib.rs
bonomat: I assume so, yes

crates/payout_curve/src/lib.rs
bonomat: What do you mean?

crates/payout_curve/src/lib.rs
bonomat: Unfortunately I found that sometimes we get a rounding error of 1 sat in the mid range's pnl. To ensure that the payout curve is continuity we just take the last point of the mid range and go until total collateral.

crates/payout_curve/src/lib.rs
bonomat: Similar to above, we sometimes get a rounding error of 1 sat in the mid range, meaning this is `1` sat and not `0`.

crates/payout_curve/src/lib.rs
luckysori: You've used a hard-coded 2 in a couple of places.

bonomat: ah, gotya

bonomat:
> Drop https://github.com/get10101/10101/commit/af34a46006b9a41f57c56db7fdd07bb478b8be4b, maybe?

we can drop this with upgrade to 116

bonomat:
Let's give this a try on regtest

PR #1549, opened 2023-11-01
fix: only show for on-chain

bonomat:
Just for you @holzeis 😘

https://github.com/get10101/10101/assets/224613/285f6c37-083e-4b98-bd30-4bfbf18fafca


holzeis (approved):
Thanks 😘

PR #1550, opened 2023-11-02
fix: Add invisible box decorator to settings clickable

### LINKED ISSUE #1534: Settings row is not clickable on empty space
Clicking on the row should also show the setting.

holzeis:
It's weird but adding an invisible box decoration makes the white space clickable.

fixes #1534


luckysori (approved):
Love a good frontend hack.

PR #1551, opened 2023-11-03
Load persisted `rust-dlc` `ChainMonitor` whenever possible 

### LINKED ISSUE #1547: Ensure that `rust-dlc`'s `ChainMonitor` is loaded on start-up
Otherwise the `ChainMonitor` will be built from scratch and important transactions can be missed on-chain.

luckysori:
Fixes #1547.

If we cannot load it because the deserialisation fails, we introduce a workaround that ensures that we can overwrite it and move on with the `DlcManager` initialisation.

This is generally not safe, but it unblocks us for the time being. We should nevertheless remove the workaround as soon as most users have updated to this patch.

### DIFF
--- a/crates/ln-dlc-node/src/node/dlc_manager.rs
+++ b/crates/ln-dlc-node/src/node/dlc_manager.rs
@@ -9,6 +9,7 @@ use dlc_manager::channel::signed_channel::SignedChannel;
 use dlc_manager::channel::signed_channel::SignedChannelState;
 use dlc_manager::channel::Channel;
 use dlc_manager::Oracle;
+use dlc_manager::Storage;
 use dlc_manager::SystemTimeProvider;
 use dlc_sled_storage_provider::SledStorageProvider;
 use p2pd_oracle_client::P2PDOracleClient;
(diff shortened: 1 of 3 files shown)


holzeis (approved):
LGTM 👍

crates/ln-dlc-node/src/node/dlc_manager.rs
```diff
@@ -39,6 +40,15 @@ pub fn build(
     let oracle_pubkey = p2pdoracle.get_public_key();
     let oracles = HashMap::from([(oracle_pubkey, p2pdoracle)]);
 
+    // FIXME: We need to do this to ensure that we can upgrade `Node`s from LDK 0.0.114 to 0.0.116.
+    // We should remove this workaround as soon as possible.
+    if let Err(e) = storage.get_chain_monitor() {
+        tracing::error!("Failed to load DLC ChainMonitor from storage: {e:#}");
```
holzeis, line 46: 🤔 shouldn't this rather be a warning, since we are able to recover from it?

luckysori: I want to be scared every time we see this message, because we really should remove the workaround ASAP.

PR #1559, opened 2023-11-07
Allow app trader to resize open position

### LINKED ISSUE #1543: Renew sub-channel manual way
To renew a position we could
- close current position
- open a new position
- this should be a single order

TBD

luckysori:
Fixes #1543.

Test is passing, but a few things left to do:

- [x] Update app position fields when we set the position state from `Resizing` back to `Open`. For instance, the number of contracts needs to be updated!
- [x] Ensure that app trader can keep trading after resizing. Write a test for this.
- [x] Probably a few more things 😁 

---

A video[^1] with a lot of minor bugs, but it "works" and the entry in the trade history at the end appears to be correct:

https://github.com/get10101/10101/assets/9418575/04c74c51-f9ef-4a91-80a6-22a5df16ea8e

[^1]: This is still the old video, I'll update it tomorrow. It looks better, so you can check out the PR if you want!

### DIFF
--- a/coordinator/src/db/trades.rs
+++ b/coordinator/src/db/trades.rs
@@ -24,6 +24,7 @@ struct Trade {
     average_price: f32,
     timestamp: OffsetDateTime,
     fee_payment_hash: String,
+    dlc_expiry_timestamp: Option<OffsetDateTime>,
 }
 
 #[derive(Insertable, Debug, Clone)]
@@ -37,6 +38,7 @@ struct NewTrade {
     collateral: i64,
     direction: Direction,
     average_price: f32,
+    dlc_expiry_timestamp: Option<OffsetDateTime>,
 }
 
 pub fn insert(
@@ -48,6 +50,19 @@ pub fn insert(
         .get_result(conn)?;
 
     Ok(trade.into())
+}
+
+pub fn get_latest_for_position(
+    conn: &mut PgConnection,
+    position_id: i32,
+) -> Result<Option<crate::trade::models::Trade>> {
+    let trade = trades::table
+        .filter(trades::position_id.eq(position_id))
+        .order_by(trades::id.desc())
+        .first::<Trade>(conn)
+        .optional()?;
+
+    Ok(trade.map(crate::trade::models::Trade::from))
 }
 
 /// Returns the position by trader pub key
@@ -76,6 +91,7 @@ impl From<crate::trade::models::NewTrade> for NewTrade {
             collateral: value.coordinator_margin,
             direction: value.direction.into(),
             average_price: value.average_price,
+            dlc_expiry_timestamp: value.dlc_expiry_timestamp,
         }
     }
 }
@@ -97,6 +113,7 @@ impl From<Trade> for crate::trade::models::Trade {
             fee_payment_hash: PaymentHash(
                 <[u8; 32]>::from_hex(value.fee_payment_hash).expect("payment hash to decode"),
             ),
+            dlc_expiry_timestamp: value.dlc_expiry_timestamp,
         }
     }
 }
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -28,6 +28,7 @@ use dlc_manager::ChannelId;
 use dlc_manager::ContractId;
 use dlc_messages::ChannelMessage;
 use dlc_messages::Message;
+use dlc_messages::SubChannelMessage;
 use lightning::ln::channelmanager::ChannelDetails;
 use lightning::util::config::UserConfig;
 use ln_dlc_node::node;
@@ -39,6 +40,7 @@ use ln_dlc_node::WalletSettings;
 use orderbook_commons::order_matching_fee_taker;
 use orderbook_commons::MatchState;
 use orderbook_commons::OrderState;
+use rust_decimal::prelude::FromPrimitive;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 use std::sync::Arc;
@@ -53,6 +55,7 @@ use uuid::Uuid;
 pub mod closed_positions;
 pub mod connection;
 pub mod expired_positions;
+pub mod resize;
 pub mod rollover;
 pub mod routing_fees;
 pub mod storage;
@@ -192,40 +195,37 @@ impl Node {
         let order_id = trade_params.filled_with.order_id.to_string();
         tracing::info!(trader_id, order_id, "Executing match");
 
-        match self.decide_trade_action(&trade_params.pubkey)? {
+        match self.decide_trade_action(connection, trade_params)? {
             TradeAction::Open => {
                 ensure!(
                     self.settings.read().await.allow_opening_positions,
                     "Opening positions is disabled"
                 );
 
-                let channel_details = self.get_counterparty_channel(trade_params.pubkey)?;
-
-                let user_channel_id = Uuid::from_u128(channel_details.user_channel_id).to_string();
-                let connection = &mut self.pool.get()?;
-                let channel =
-                    db::channels::get(&user_channel_id, connection)?.with_context(|| {
-                        format!(
-                            "Couldnt find shadow channel. trader_id={}, user_channel_id={}",
-                            trade_params.pubkey, channel_details.user_channel_id
-                        )
-                    })?;
-
-                let coordinator_leverage = match channel.liquidity_option_id {
-                    Some(liquidity_option_id) => {
-                        let liquidity_option =
-                            db::liquidity_options::get(connection, liquidity_option_id)?;
-                        liquidity_option.coordinator_leverage
-                    }
-                    None => 1.0,
-                };
+                let coordinator_leverage =
+                    self.coordinator_leverage_for_trade(&trade_params.pubkey)?;
 
                 self.open_position(connection, trade_params, coordinator_leverage, order.stable)
                     .await?;
             }
+            TradeAction::Resize(channel_id) => {
+                ensure!(
+                    self.settings.read().await.allow_opening_positions,
+                    "Resizing positions is disabled"
+                );
+
+                self.resize_position(connection, channel_id, trade_params)
+                    .await?
+            }
             TradeAction::Close(channel_id) => {
                 let peer_id = trade_params.pubkey;
-                tracing::info!(?trade_params, channel_id = %hex::encode(channel_id), %peer_id, "Closing position");
+
+                tracing::info!(
+                    ?trade_params,
+                    channel_id = %hex::encode(channel_id),
+                    %peer_id,
+                    "Closing position"
+                );
 
                 let closing_price = trade_params.average_execution_price();
 
@@ -366,6 +366,7 @@ impl Node {
             coordinator_margin: margin_coordinator as i64,
             expiry_timestamp: trade_params.filled_with.expiry_timestamp,
             temporary_contract_id,
+            coordinator_leverage,
             trader_margin: margin_trader as i64,
             stable,
         };
@@ -384,6 +385,7 @@ impl Node {
                 coordinator_margin: new_position.coordinator_margin,
                 direction: new_position.direction,
                 average_price: average_entry_price,
+                dlc_expiry_timestamp: Some(trade_params.filled_with.expiry_timestamp),
             },
         )?;
 
@@ -423,6 +425,9 @@ impl Node {
                 coordinator_margin: position.coordinator_margin,
                 direction: position.direction.opposite(),
                 average_price: closing_price.to_f32().expect("To fit into f32"),
+                // A closing trade does not require an expiry timestamp for the DLC, because the DLC
+                // is being _removed_.
+                dlc_expiry_timestamp: None,
             },
         )?;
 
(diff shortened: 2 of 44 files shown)


mobile/native/src/calculations/mod.rs
```diff
@@ -25,9 +25,12 @@ pub fn calculate_pnl(
     leverage: f32,
     direction: Direction,
 ) -> Result<i64> {
+    // FIXME: We can no longer assume that the coordinator always has the same leverage! It needs to
```
luckysori, line 28: I noticed that this was wrong so I changed it to the most correct value that we can possibly hard-code, but we need to find a way to know the current coordinator liquidity for a channel on the app so that we can correctly cap the profits of the trader.

I don't think we should bother fixing this now, but I will open an issue (later).

coordinator/src/node.rs
```diff
@@ -434,6 +436,88 @@ impl Node {
...
+        // DLC channel _as if nothing had happened_. That is, the `accept_settlement_amount` is
+        // equal to the amount the accept party originally put into the contract.
+        //
+        // TODO: Use subchannel resize protocol to ensure atomicity, simplify the implementation and
+        // improve UX.
+        {
+            // We don't charge a fee for closing because closing the channel is just an
+            // implementation detail. We will charge the fee when opening the new channel.
+            //
+            // TODO: Do we need to include the previous order-matching fee here?
```
luckysori, line 475: I think the answer is yes, but I feel like I need some help calculating the correct fee cc @holzeis.

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
             self.finalize_rollover(&r.channel_id)?;
         }
 
+        if let Message::SubChannel(SubChannelMessage::CloseFinalize(_msg)) = &msg {
```
holzeis, line 649: nit: this file is already growing quite big. It would be nice if we could combine the resize logic e.g. in a resize.rs.

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
...
+                            margin_trader as i64,
+                            f32_from_decimal(average_execution_price),
+                            f32_from_decimal(liquidation_price_trader),
+                            expiry_timestamp,
+                            temporary_contract_id,
+                            stable,
+                        ) {
+                            tracing::error!(
+                                channel_id = %channel_details.channel_id.to_hex(),
+
```
holzeis, line 836: nit: is this an extra new line or is it just badly shown on my phone 😅

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -510,7 +505,28 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
         })
     });
 
-    let trades = derive_trades_from_filled_orders()?;
+    let trades = db::get_all_trades()?;
```
holzeis, line 508: 👍 nice, but I guess that will remove all open and close position entries from the history of existing users right?

mobile/native/src/ln_dlc/node.rs
```diff
@@ -291,14 +292,28 @@ impl Node {
...
         if let Message::SubChannel(SubChannelMessage::CloseFinalize(SubChannelCloseFinalize {
             ..
         })) = msg
         {
-            let filled_order = order::handler::order_filled()?;
-            position::handler::update_position_after_dlc_closure(Some(filled_order))
-                .context("Failed to update position after DLC closure")?;
+            tracing::debug!("Checking if finalize subchannel closure is for position resize");
+
+            let positions = position::handler::get_positions()?;
```
holzeis, line 302: 🔧 this should pnl my ever return a single position, if not we should probably fail?

However, I think it's safe to simply use first here.

mobile/native/src/ln_dlc/recover_rollover.rs
```diff
@@ -1,74 +1,85 @@
...
     /// Checks and recovers from a potential stuck rollover.
     pub async fn recover_rollover(&self) -> Result<()> {
-        tracing::debug!("Checking if dlc channel got stuck in rollover.");
+        tracing::debug!("Checking if DLC channel got stuck during rollover");
+
         let channels = self.inner.channel_manager.list_channels();
         let channel_details = match channels.first() {
             Some(channel_details) => channel_details,
             None => {
-                tracing::debug!("No channel found. All good!");
```
holzeis, line 15: ❓ why did you remove that log?

mobile/native/src/ln_dlc/recover_rollover.rs
```diff
@@ -1,74 +1,85 @@
...
             .iter()
-            .find(|dlc_channel| dlc_channel.channel_id == channel_details.channel_id);
-
-        let dlc_channel = match dlc_channel {
-            Some(dlc_channel) => dlc_channel,
+            .find(|dlc_channel| dlc_channel.channel_id == channel_details.channel_id)
+        {
+            Some(subchannel) => subchannel,
             None => {
-                tracing::debug!("No dlc channel found. All good!");
```
holzeis, line 28: ❓ why did you remove this log?

mobile/native/src/ln_dlc/recover_rollover.rs
```diff
@@ -1,74 +1,85 @@
...
             .inner
             .dlc_manager
             .get_store()
             .get_channel(&dlc_channel_id)?;
 
-        let signed_channel = match channel {
+        let signed_channel = match dlc_channel {
             Some(Channel::Signed(signed_channel)) => signed_channel,
-            Some(channel) => {
-                tracing::warn!(dlc_channel_id=%hex::encode(dlc_channel_id), "Found channel in unexpected state. Expected: Signed, Found: {channel:?}");
```
holzeis, line 50: ❓ why did you remove this log?

mobile/native/src/ln_dlc/recover_rollover.rs
```diff
@@ -1,74 +1,85 @@
...
-            SignedChannelState::RenewOffered { .. }
-            | SignedChannelState::RenewAccepted { .. }
-            | SignedChannelState::RenewConfirmed { .. }
-            | SignedChannelState::RenewFinalized { .. } => {
-                let state = ln_dlc_node::node::signed_channel_state_name(&signed_channel);
-                tracing::warn!(state, "Found signed channel contract in an intermediate state. Rolling back, expecting coordinator to retry rollover!");
-                self.inner.rollback_channel(&signed_channel)?;
-            }
-            _ => {
-                tracing::debug!(signed_channel_state=%signed_channel.state, "Channel is not in an intermediate rollover state. All good.")
```
holzeis, line 69: For debugging it would be helpful to have that log message.

mobile/native/src/trade/mod.rs
```diff
@@ -1,3 +1,42 @@
...
 pub mod order;
 pub mod position;
 pub mod users;
+
+/// A trade is an event that moves funds between the Lightning wallet and a DLC channel.
+///
+/// Every trade is associated with a single market order, but an order can be associated with
+/// multiple trades.
+///
+/// If an order changes the direction of the underlying position, it must be split into _two_
```
holzeis, line 18: 🤔 do we really have to reflect the technical details on the model? I wouldn't have bothered and simply acted as if it's one atomic action. Logically it doesn't make a difference does it?

coordinator/src/db/trades.rs
```diff
@@ -24,6 +24,7 @@ struct Trade {
     average_price: f32,
     timestamp: OffsetDateTime,
     fee_payment_hash: String,
+    expiry_timestamp: Option<OffsetDateTime>,
```
holzeis, line 27: ❓ what is the expiry on the trade?

holzeis (commented):
Nice 👍. Looking forward to that change. 

I left primarily questions. Will have to look at it in more detail again tomorrow.

holzeis (approved):
Great effort! I left a couple of remarks, but overall it looks good to me 👍

Approved as it is doing the job! (tested locally)

coordinator/src/node.rs
luckysori: Good catch, didn't mean that.

mobile/native/src/ln_dlc/mod.rs
luckysori: As discussed, updating to the new version will erase the trade history.

mobile/native/src/ln_dlc/recover_rollover.rs
luckysori: I deleted these logs because I thought they were a bit noisy when debugging locally. But I can add them again.

mobile/native/src/ln_dlc/recover_rollover.rs
luckysori: I will reintroduce it although I think it should be on DEBUG, since there's nothing concerning about finding a channel that is not yet signed, right?

coordinator/src/node.rs
```diff
@@ -447,21 +531,42 @@ impl Node {
...
+        trade_params: &TradeParams,
+    ) -> Result<TradeAction> {
+        let trader_peer_id = trade_params.pubkey;
+
+        let subchannel = match self.inner.get_dlc_channel_signed(&trader_peer_id)? {
+            None => return Ok(TradeAction::Open),
+            Some(subchannel) => subchannel,
+        };
+
+        let position = db::positions::Position::get_position_by_trader(
```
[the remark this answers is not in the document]
luckysori: I see what you mean, but I don't like attempting to load the position even if the trade action is meant to be `Open`. Also optional parameters are not nice to work with.

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
...
+            )? {
+                let channel_details = self.get_counterparty_channel(node_id)?;
+
+                tracing::info!(
+                    channel_id = %channel_details.channel_id.to_hex(),
+                    peer_id = %node_id,
+                    "Found resizing position corresponding to channel that was just closed"
+                );
+
+                let trade = db::trades::get_latest_for_position(&mut conn, old_position.id)?
```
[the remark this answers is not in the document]
luckysori: I think an explicit table would be slightly superior, but I'm gonna opt to not action now because I think the risk is very low.

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
...
+                    .context("No trade for resized position")?;
+
+                // Compute absolute contracts using formula.
+                let (total_contracts, direction) = {
```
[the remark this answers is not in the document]
luckysori: I don't see the refactor potential specifically here, but I do think the new `resize.rs` file can be refactored (same goes for `position/mod.rs` in `mobile`).

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
...
+                            Err(e) => {
+                                tracing::error!(
+                                    channel_id = %channel_details.channel_id.to_hex(),
+                                    "Unable to extract temporary contract id: {e:#}"
+                                );
+                                return;
+                            }
+                        };
+
+                        if let Err(e) = db::positions::Position::update_resized_position(
```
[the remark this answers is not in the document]
luckysori: I did mean to follow the same pattern that was already there. I agree that it is not a good pattern and we should change it, but I don't think it's the right time to tackle this.

coordinator/src/node.rs
```diff
@@ -541,6 +646,233 @@ impl Node {
...
+                    }
+                });
+            }
+        }
+
+        if let Some(Message::SubChannel(SubChannelMessage::Revoke(r))) = &resp {
+            let mut conn = self.pool.get()?;
+
+            // If we have a `Resizing` position, we must have just finished the resizing protocol.
+            if (db::positions::Position::get_position_by_trader(
```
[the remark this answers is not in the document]
luckysori: Good catch. This is a remnant of a prior implementation.

coordinator/src/node.rs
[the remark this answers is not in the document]
luckysori: I left a TODO.

coordinator/src/trade/models.rs
```diff
@@ -11,9 +11,17 @@ pub struct NewTrade {
     pub trader_pubkey: PublicKey,
     pub quantity: f32,
     pub trader_leverage: f32,
+    // TODO: When opening a position it's obvious what this should be, but when closing or resizing
```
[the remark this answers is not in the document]
luckysori: I won't remove it yet in case @bonomat likes this field for some reason, but I will leave the TODO.

mobile/native/src/ln_dlc/node.rs
```diff
@@ -16,6 +16,7 @@ use dlc_messages::sub_channel::SubChannelRevoke;
 use dlc_messages::ChannelMessage;
 use dlc_messages::Message;
 use dlc_messages::SubChannelMessage;
+use itertools::Itertools;
```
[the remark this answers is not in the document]
luckysori: Not used, thanks!

mobile/native/src/trade/position/mod.rs
```diff
@@ -53,3 +73,863 @@ pub struct Position {
...
+    ///
+    /// NOTE: The order's leverage is ignored when applying an order to an existing position. It
+    /// does not seem to make much sense to allow a trader to both change the number and/or
+    /// direction of contracts (the point of creating a new order) _and_ to change the leverage.
+    /// Also, it's not so straightforward to calculate combined leverages, particularly when
+    /// reducing a position.
+    ///
+    /// TODO: This highlights the fact that orders really shouldn't have a leverage associated with
+    /// them!
+    fn apply_order_recursive(
```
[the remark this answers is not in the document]
luckysori: Haha you're welcome to refactor it, we do have tests that don't care about the recursive implementation.

mobile/native/src/trade/position/handler.rs
```diff
@@ -214,10 +224,41 @@ pub fn update_position_after_dlc_creation(
...
     tracing::debug!(?filled_order, "Removing position after DLC channel closure");
 
-    if db::get_positions()?.is_empty() {
+    let positions = db::get_positions()?;
+    if positions.is_empty() {
         tracing::warn!("No position to remove");
     }
 
+    if let Some(filled_order) = filled_order {
+        for position in positions {
```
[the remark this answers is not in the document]
luckysori: I think I did this taking into consideration the possibility that eventually more than one position will be needed to trade different contract symbols. But I can revert it.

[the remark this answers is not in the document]
luckysori: I've updated it.

mobile/native/src/trade/mod.rs
luckysori: It was simpler just to avoid having to deal with trades that change the direction of the position. But we may want to revisit this decision.

mobile/native/src/ln_dlc/recover_rollover.rs
luckysori: I have reintroduced all of them, but edited them slightly.

coordinator/src/db/trades.rs
luckysori: As discussed privately, this expiry corresponds to the expiry that needs to be used to set up the new DLC channel. It clearly does not belong in the `Trade` model, but it was a convenient thing to do to avoid creating a new table.

PR #1560, opened 2023-11-08
Update coordinator version in `Cargo.lock`

luckysori:
Building locally updates the `Cargo.lock` automatically. I think this should have happened when we updated the coordinator's `Cargo.toml`.


holzeis (approved):
Thanks 🙏

PR #1562, opened 2023-11-10
feat: Add backup and restore

holzeis:
## Backup and restore

> When you do things right, people won't be sure you've done anything at all

- Introduces the TenTenOneStorage layer encapsulating the `rust-lightning` and `rust-dlc` storage.
- Introduces a 10101 DlcStorageProvider trait with a `SledStorageProvider` and a `InMemoryDLCStoreProvider` implementation introducing an abstraction layer for storage.
- Removes the `simple-wallet` and `dlc-sled-storage-provider` crates.
- Makes the `ln_dlc_node` crate indifferent about the used storage, allowing us to use in-memory implementations of `rust-lightning` and `rust-dlc` data for the tests.
- Introduces a `RemoteBackupClient` implementation that sends backups asynchronously to the coordinator. 
- Backups are symmetrically (aes256-gcm-siv) encrypted and decrypted on the client side. The key is derived from the users secret key. The coordinator only receives and stores encrypted data.
- The users intent for creating a backup or restoring a backup is achieved by verifying a signature on the backup (value) or restore (node_id) message.

### Minor onboarding UI/UX improvements

- The is not asked for an email address when restoring from a backup.

<details><summary>Removed app bar from welcome screen</summary>
<p>

Simulator Screenshot - iPhone 14 Pro - 2023-11-10 at 08 34 52

</p>
</details> 

- Added visual improvements on the loading screen
- Wait for restore / initial backup to complete on loading screen
- Remove transitions from initial screens
- Remove splash screen once on the loading screen

### Compatibility

The change is fully compatible with older versions. Tested with an upgrade from 1.5.0. However, for the upgrade path we have to run a full backup at startup. This is achieved through a preference setting, which is by default true unless the user is creating a new wallet or restores a wallet.

The dlc-store of the actions did not adhere to the same pattern as other data stored in the dlc sled storage. (didn't use a tree, but rather stored the data directly into the sled db). I opted to ignore, that since handling that case would have made the implementation more complex, while the actions can be reconstructed in a `on_channel_reestablish` event.

### Testing

- Manually tested the upgrade path from a pre-backup version. (see video)
- Added an e2e test for restoring from a backup. Note, I had to make the storage mutable to update the states after stopping the app. This required me to introduce some unsafe code 8775a74c50e19aa09b022d1c12015a96287cf1fb. This actually only needed for testing, but I wasn't able to compile the corresponding unsafe parts only for `#[cfg(test)]` as the test are always compiled for `#[cfg(not(test))]`.
- Copied the sled database tests from `rust-dlc`, but noticed that they do not cover all scenarios. I found quite some issues during manual testing. However, my confidence in our `SledStorageProvider` is pretty high.
- Introduced `TenTenOneInMemoryStorage` for the `ln_dlc_node` crate tests, meaning that no `rust-lightning` or `rust-dlc` data is stored to disk during those tests.

### Future Work

1. There is no requirement for the coordinator to provide that API. The backup and restore API could be scaled nicely distributing the load.
2. The backups are stored into a sled database. We might want to consider using a managed nosql database for storing user backups. e.g. a managed MongoDB. 
3. Preferences are not backed up at the moment, as they are not deemed as critical at the moment. We might want to revisit that if we are storing more important data in the preferences.

https://github.com/get10101/10101/assets/382048/733753c3-8c11-48a2-a0c7-cb56eb24fec7


coordinator/src/backup.rs
```diff
@@ -0,0 +1,76 @@
...
+use coordinator_commons::Backup;
+use coordinator_commons::DeleteBackup;
+use coordinator_commons::Restore;
+use sled::Db;
+
+const BACKUPS_DIRECTORY: &str = "user_backups";
+
+/// Holds the user backups in a sled database
+///
+/// TODO(holzeis): This is fine for now, once we grow we should consider moving that into a dedicate
```
bonomat, line 13: What's the storage requirement for this?

coordinator/src/backup.rs
```diff
@@ -0,0 +1,76 @@
...
+            });
+        }
+
+        Ok(backup)
+    }
+
+    pub async fn backup(&self, node_id: PublicKey, backup: Backup) -> Result<()> {
+        backup.verify(&node_id)?;
+
+        tracing::debug!(%node_id, backup.key, "Create user backup");
```
bonomat, line 59: Is `node_id` the trader's id? If so, should we call it this here?

coordinator/src/backup.rs
```diff
@@ -0,0 +1,76 @@
...
+                .split('/')
+                .map(|key| key.to_string())
+                .collect::<Vec<String>>();
+
+            let (kind, key) = keys.split_first().expect("keys to be long enough");
+            backup.push(Restore {
+                kind: kind.as_str().try_into()?,
+                key: key.join("/"),
+                value,
+            });
```
bonomat, line 50: Nit: you could potentially do a `tree.into_iter().map(..` here.

coordinator/src/orderbook/websocket.rs
```diff
@@ -128,7 +128,7 @@ pub async fn websocket_connection(stream: WebSocket, state: Arc<AppState>) {
                     fcm_token,
                     signature,
                 }) => {
-                    let msg = create_sign_message();
+                    let msg = create_sign_message(b"Hello it's me Mario".to_vec());
```
bonomat, line 131: https://www.youtube.com/watch?v=BHPse3V1aMg

bonomat, line 131: If `create_sign_message` is not being used anymore, make sure to delete it.

coordinator/src/routes.rs
```diff
@@ -542,3 +553,55 @@ pub async fn collaborative_revert_confirm(
     })?;
     Ok(Json(serialize_hex(&raw_tx)))
 }
+
+// TODO(holzeis): There is no reason the backup and restore api has to run on the coordinator. On
+// the contrary it would be much more reasonable to have the backup and restore api run separately.
```
bonomat, line 558: Definitely. Maybe we should create from the getgo at least a different subdomain/path/port for the backup to be able to move it.

coordinator/src/routes.rs
```diff
@@ -542,3 +553,55 @@ pub async fn collaborative_revert_confirm(
...
+    state
+        .user_backup
+        .backup(node_id, backup.0)
+        .await
+        .map_err(|e| AppError::InternalServerError(e.to_string()))
+}
+
+// TODO(holzeis): There is no reason the backup and restore api has to run on the coordinator. On
+// the contrary it would be much more reasonable to have the backup and restore api run separately.
+pub async fn delete_backup(
```
bonomat, line 576: Under what circumstances do you expect `delete_backup` to be called?
Maybe we don't want to expose this endpoint?

coordinator/src/routes.rs
```diff
@@ -542,3 +553,55 @@ pub async fn collaborative_revert_confirm(
...
+    Path(node_id): Path<String>,
+    State(state): State<Arc<AppState>>,
+    backup: Json<DeleteBackup>,
+) -> Result<(), AppError> {
+    let node_id = PublicKey::from_str(&node_id).map_err(|e| {
+        AppError::BadRequest(format!("Failed to delete backup {}. {e:#}", backup.key))
```
bonomat, line 582: Nit: 
```suggestion
AppError::BadRequest(format!("Invalid node id provided {}. {e:#}", backup.key))
```

coordinator/src/routes.rs
```diff
@@ -542,3 +553,55 @@ pub async fn collaborative_revert_confirm(
...
+
+// TODO(holzeis): There is no reason the backup and restore api has to run on the coordinator. On
+// the contrary it would be much more reasonable to have the backup and restore api run separately.
+pub async fn backup(
+    Path(node_id): Path<String>,
+    State(state): State<Arc<AppState>>,
+    backup: Json<Backup>,
+) -> Result<(), AppError> {
+    let node_id = PublicKey::from_str(&node_id)
+        .map_err(|e| AppError::BadRequest(format!("Failed to backup {}. {e:#}", backup.key)))?;
```
bonomat, line 565: Nit: 

```suggestion
.map_err(|e| AppError::BadRequest(format!("Invalid node id provided {}. {e:#}", backup.key)))?;
```

coordinator/src/routes.rs
```diff
@@ -542,3 +553,55 @@ pub async fn collaborative_revert_confirm(
...
+
+// TODO(holzeis): There is no reason the backup and restore api has to run on the coordinator. On
+// the contrary it would be much more reasonable to have the backup and restore api run separately.
+async fn restore(
+    Path(node_id): Path<String>,
+    State(state): State<Arc<AppState>>,
+    signature: Json<Signature>,
+) -> Result<Json<Vec<Restore>>, AppError> {
+    let node_id = PublicKey::from_str(&node_id)
+        .map_err(|e| AppError::BadRequest(format!("Failed to restore backup. {e:#}")))?;
```
bonomat, line 599: Same as above

coordinator/src/backup.rs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        Ok(backup)
+    }
+
+    pub async fn backup(&self, node_id: PublicKey, backup: Backup) -> Result<()> {
+        backup.verify(&node_id)?;
```
bonomat, line 57: It would be good if this error would map to an `401 Unauthorized` and not an internal server error.

bonomat (commented):
Only went over the first commit so far... will continue later.

coordinator/src/backup.rs
holzeis: hmm.. hard to say, identifying the storage requirements is an exercise we haven't done yet. But for the small amounts we are currently dealing with I think it doesn't really matter.

coordinator/src/backup.rs
holzeis: it is, I am impartial on the naming, but I'd say we should go for one name. I recently got a review remark from @luckysori asking for it to be named `node_id`, but that might also depend on the context.

coordinator/src/backup.rs
holzeis: I did have it as `tree.into_iter().map(..` before, but changed it to this, as the error handling got messy in the map closure.

coordinator/src/orderbook/websocket.rs
holzeis: Note, I just changed the signature to accept the message to be signed from the caller. The message `Hello it's me Mario` was there before (just moved to the caller).

coordinator/src/routes.rs
holzeis: As we are already using a reverse proxy, we could simple direct any request to `/api/backup` or `/api/restore` to a different service once we move it. This could be completely transparent to the app.

coordinator/src/routes.rs
holzeis: This is a necessity for `rust-dlc` as it is also deleting data. More specifically, `rust-dlc` deletes contracts with a temporary contract id, once the sub channel is established. I think we could keep that data, but opted to expose the api nonetheless to stay as close to the actual state as possible.

coordinator/src/routes.rs
holzeis: Wouldn't that be redundant as it would be explained in the error anyways? `{e:#}`

coordinator/src/routes.rs
holzeis: Wouldn't that be redundant as it would be explained in the error anyways? {e:#}

coordinator/src/routes.rs
holzeis: same answer as above :)

coordinator/src/routes.rs
bonomat: The error in `{e:#}` is usually not very well formated.

coordinator/src/backup.rs
bonomat: Agreed, no hard feelings on `node_id` vs `trader_id`, I thought we are using `trader_id` more often already.

coordinator/src/routes.rs
bonomat: The problem is though that we are hardcoding the IP in the app. If we were to use `backup.10101.finance` instead, we could move it indepdently.

coordinator/src/storage.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl TenTenOneStorage for TenTenOneNodeStorage {}
+
+impl TenTenOneNodeStorage {
+    pub fn new(data_dir: String) -> TenTenOneNodeStorage {
+        let data_dir = PathBuf::from(data_dir);
+
+        if !data_dir.exists() {
+            fs::create_dir_all(data_dir.as_path())
+                .unwrap_or_else(|_| panic!("Failed to create data dir {data_dir:?}"));
```
bonomat, line 33: What's the reason for panicing here and not returning an error? (same for rest of the file).

crates/ln-dlc-storage/src/sled.rs
bonomat, line 1: I guess this file was copied from somewhere?

crates/ln-dlc-storage/src/lib.rs
```diff
@@ -0,0 +1,1191 @@
...
+    #[test]
+    fn get_empty_actions_test() {
+        let storage = DlcStorageProvider::new(InMemoryDLCStoreProvider::new());
+        storage.save_sub_channel_actions(&[]).unwrap();
+        let actions = storage
+            .get_sub_channel_actions()
+            .expect("Error getting sub channel actions");
+        assert_eq!(actions.len(), 0);
+    }
+}
```
bonomat, line 1191: I guess this file was copied from somewhere?

crates/orderbook-client/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub async fn subscribe_with_authentication(
     SplitSink<WebSocketStream<MaybeTlsStream<TcpStream>>, tungstenite::Message>,
     impl Stream<Item = Result<String, Error>> + Unpin,
 )> {
-    let signature = authenticate(create_sign_message());
+    let signature = authenticate(create_sign_message(b"Hello it's me Mario".to_vec()));
```
bonomat, line 41: Should we rather use a constant for this text somewhere?

mobile/lib/util/file.dart
```diff
@@ -28,4 +28,4 @@ Future<Directory> getActualSeedPath(bridge.Config config) async {
   final network = config.network == "mainnet" ? "bitcoin" : config.network;
   final seedPath = '$seedDir/$network';
   return Directory(seedPath);
-}
+}
```
bonomat: 🙃 missing newline

mobile/native/src/config/api.rs
```diff
@@ -18,8 +18,12 @@ pub struct Config {
     pub health_check_interval_secs: u64,
 }
 
-impl From<Config> for ConfigInternal {
-    fn from(config: Config) -> Self {
+impl From<(Config, String, String)> for ConfigInternal {
```
bonomat, line 21: I didn't see where this is being used but this from implementation looks like it's easy to confuse `data_dir` and `seed_dir`.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -156,16 +162,40 @@ pub fn get_seed_phrase() -> Vec<String> {
...
 pub fn get_node_key() -> SecretKey {
-    NODE.get().inner.node_key()
+    let seed = get_seed();
+    let time_since_unix_epoch = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("unix epos to not be earlier than now");
+    let keys_manger = KeysManager::new(
+        &seed.lightning_seed(),
+        time_since_unix_epoch.as_secs(),
+        time_since_unix_epoch.subsec_nanos(),
```
bonomat, line 186: This logic changed. Before we used the `KeysManager` from the node, now you re-initiate it. The `KeysManager` is not deterministic though (see the system time. 
Does this logic have any side effects? 
Why did you change the logic?

mobile/native/src/storage.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Creates a full backup of the lightning and dlc data.
+    pub async fn full_backup(&self) -> anyhow::Result<()> {
+        tracing::info!("Running full backup");
+        let mut handles = vec![];
+
+        let db_backup = db::backup()?;
+        let value = fs::read(db_backup)?;
+        let handle = self.client.backup("10101/db".to_string(), value);
```
bonomat, line 77: We should use constants for all those keys. I.e. `10101/db` here.

mobile/lib/common/loading_screen.dart
```diff
@@ -38,25 +39,27 @@ class _LoadingScreenState extends State<LoadingScreen> {
       if (isSeedFilePresent) {
         runBackend(context).then((value) {
           logger.i("Backend started");
-        });
+
+          if (!hasEmailAddress) {
```
bonomat: Given the size of the PR, I think these changes should be in its own PR.

mobile/lib/common/loading_screen.dart
```diff
@@ -14,58 +17,100 @@ import 'package:go_router/go_router.dart';
...
-            GoRouter.of(context).go(WelcomeScreen.route);
+        if (!hasEmailAddress) {
+          GoRouter.of(context).go(WelcomeScreen.route);
+        } else {
+          if (isFullBackupRequired) {
+            setState(() => message = "Creating initial backup!");
+            fullBackup().then((value) {
+              Preferences.instance.setFullBackupRequired(false).then((value) {
+                start(context, position);
+              });
```
bonomat, line 62: The logic in this file is getting out of hand :(

mobile/lib/common/loading_screen.dart
```diff
@@ -14,58 +17,100 @@ import 'package:go_router/go_router.dart';
...
-          logger.i("Backend started");
-
-          if (!hasEmailAddress) {
-            GoRouter.of(context).go(WelcomeScreen.route);
+        if (!hasEmailAddress) {
+          GoRouter.of(context).go(WelcomeScreen.route);
+        } else {
+          if (isFullBackupRequired) {
+            setState(() => message = "Creating initial backup!");
+            fullBackup().then((value) {
```
bonomat, line 59: There is no exception handling `.catch` on this future.

mobile/lib/common/loading_screen.dart
```diff
@@ -14,58 +17,100 @@ import 'package:go_router/go_router.dart';
...
-    Future.wait<dynamic>([
+    List<Future<dynamic>> futures = [
       Preferences.instance.hasEmailAddress(),
       Preferences.instance.getOpenPosition(),
       isSeedFilePresent(),
-    ]).then((value) {
+      Preferences.instance.isFullBackupRequired(),
+    ];
+
+    if (widget.restore != null) {
```
bonomat, line 40: What's `widget.restore`?

mobile/lib/common/routes.dart
```diff
@@ -26,6 +26,33 @@ GoRouter createRoutes() {
       navigatorKey: rootNavigatorKey,
       initialLocation: LoadingScreen.route,
       routes: <RouteBase>[
+        GoRoute(
```
bonomat, line 29: Given the size of this PR you could have spared us some of the changes in this PR (e.g. moving things around). 

Potentially, you could even split it up into two PRs. First dealing with the backup and then with the restore.

mobile/lib/common/loading_screen.dart
holzeis: That's an optional future passed as argument from the restore wallet screen to the loading screen to wait for the restore to complete before starting the backend.

crates/ln-dlc-storage/src/lib.rs
holzeis: The majority of the code was copied from the `dlc-sled-storage-provider` and adapted to allow to abstract the storage layer.

mobile/lib/common/routes.dart
holzeis: No worries I can split up the cosmetic changes.

If you really think that it will improve the review, I can also split up the backup and restore changes, but that will be more complicated.

mobile/lib/common/loading_screen.dart
holzeis: Agreed 🙈. We probably should use some kind of wizard here.

mobile/lib/common/loading_screen.dart
holzeis: Hmm.. you mean the full backup could have been separated into a separate PR?

crates/orderbook-client/src/lib.rs
holzeis: Sure why not.

mobile/native/src/ln_dlc/mod.rs
holzeis: > Why did you change the logic?

I changed it because I need the secret key for getting the public key and decrypting a backup before starting the node.

coordinator/src/storage.rs
holzeis: hmm.. I guess if we can't create the data directory, there is not much we can do - so a panic seems reasonable to me.

mobile/native/src/ln_dlc/mod.rs
holzeis: > Does this logic have any side effects?

Judging from the api docs, there shouldn't be any side effects unless the `starting_time` is unique, every time we are calling that function.

> starting_time isn't strictly required to actually be a time, but it must absolutely, without a doubt, be unique to this instance. ie if you start multiple times with the same seed, starting_time must be unique to each run. Thus, the easiest way to achieve this is to simply use the current time (with very high precision).
The seed MUST be backed up safely prior to use so that the keys can be re-created, however, obviously, starting_time should be unique every time you reload the library - it is only used to generate new ephemeral key data (which will be stored by the individual channel if necessary).

holzeis: However, I adapted the implementation so that it will always return the secret key from the node, if the node is set.

coordinator/src/routes.rs
holzeis: > The problem is though that we are hardcoding the IP in the app.

Agreed, but this is unrelated to this change. e.g. we could change the configuration to use the domain instead of the ip. e.g. `api.10101.finance` and then configure the reverse proxy to route any traffic on `api.10101.finance/api/backup` to another service.

One caveat I see in using a dedicated subdomain is that this would introduce another configuration. What benefits do you see in using a dedicated subdomain?

coordinator/src/routes.rs
holzeis: 0052b019739bcd87a3455203c16bf83246e276f6

coordinator/src/routes.rs
holzeis: 0052b019739bcd87a3455203c16bf83246e276f6

coordinator/src/routes.rs
holzeis: 0052b019739bcd87a3455203c16bf83246e276f6

coordinator/src/backup.rs
holzeis: 0052b019739bcd87a3455203c16bf83246e276f6

crates/orderbook-client/src/lib.rs
holzeis: 19b59d6c6a283d35a4f711098c66a31332fee1fb

mobile/native/src/config/api.rs
holzeis: 419342e73ffb7ce7338a7c94ca600dc52f3411b3

mobile/native/src/ln_dlc/mod.rs
holzeis: b7820829136c01832bd12a06e1aa14e35bcd23b1

mobile/native/src/storage.rs
holzeis: bffe6b0591999a6dea36683e96d65014113e3e00

mobile/lib/common/loading_screen.dart
holzeis: f8d3b7b065fdf97be524d42eea54efcc4a5066c7

holzeis:
> Added an e2e test for restoring from a backup. Note, I had to make the storage mutable to update the states after stopping the app. This required me to introduce some unsafe code https://github.com/get10101/10101/commit/e1ac6f2221d95631728aa7c3abea2632b7818ef3. This actually only needed for testing, but I wasn't able to compile the corresponding unsafe parts only for #[cfg(test)] as the test are always compiled for #[cfg(not(test))].

Addressed with 168fcf4aa34d44fc3ef89692424a13d516b0d628

coordinator/src/routes.rs
bonomat: It will be hard to roll this out at a later stage or move the backup to a new server without breaking existing apps. 

i.e. 
1. `user_a` runs version 1.6.0 and backs-up his app on 10101.finance
2. we migrate the backup service to backup.10101.finance
3. `user_a` cannot create/restore a backup before upgrading to the new version

I guess breaking changes are OK. Are we handling errors correctly, i.e. what if the backup service is not working (e.g. for an old version because we moved it already)?

mobile/lib/common/routes.dart
bonomat: I think it would have been easier. Now I'm almost through 😅

coordinator/src/routes.rs
holzeis: > we migrate the backup service to backup.10101.finance
> user_a cannot create/restore a backup before upgrading to the new version

I don't see an issue here. We can simply add an additional nginx config that points all the backup/restore requests from user_a to the new service.

But, let's not argue about that. If you feel strongly about it. I can add it as a follow up to this PR.

bonomat (approved):
Pretty awesome. I tried it and it works like magic 🪄🚀🌕

coordinator/src/routes.rs
bonomat: > I don't see an issue here. We can simply add an additional nginx config that points all the backup/restore requests from user_a to the new service.

Can we do this if the new service runs on a different machine/ip?

holzeis: Yes, we would need to configure a redirect then.

mobile/native/src/cipher.rs
```diff
@@ -0,0 +1,112 @@
...
+use bitcoin::secp256k1::SecretKey;
+use bitcoin::secp256k1::SECP256K1;
+
+#[derive(Clone)]
+pub struct AesCipher {
+    secret_key: SecretKey,
+    inner: Aes256GcmSiv,
+}
+
+impl AesCipher {
```
luckysori, line 20: What's your confidence level on our encrypted backup scheme? I'm afraid I'm not sure if this is safe.

It's always a risk to roll your own crypto, so I would like to spend some more time figuring out if what we are doing here is safe. Other than the choice of cipher which LGTM, I'm not sure about sending over the nonce in plaintext to the server, to then retrieve it and use it when decrypting. That seems incorrect.

crates/tests-e2e/tests/restore_from_backup.rs
```diff
@@ -0,0 +1,51 @@
+use anyhow::Result;
+use native::api;
+use tests_e2e::app::run_app;
+use tests_e2e::logger::init_tracing;
+use tests_e2e::setup;
+use tokio::task::spawn_blocking;
+
+#[tokio::test]
+#[ignore = "need to be run with 'just e2e' command"]
+async fn app_can_be_restored_from_a_backup() -> Result<()> {
```
luckysori, line 10: 🙃 I think the test should go with the feature so that one can check out the patch that adds the feature knowing that it was meant to work.

luckysori (commented):
LGTM except for:

- Encrypted backup scheme.
- Numerous type parameters and trait bounds added in `ln-dlc-node`. Maybe unavoidable for now, but it makes me feel uneasy.

---

I did not review some of the unrelated changes to the main feature.

coordinator/src/backup.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub async fn backup(&self, node_id: PublicKey, backup: Backup) -> Result<()> {
+        tracing::debug!(%node_id, backup.key, "Create user backup");
+        let tree = self.db.open_tree(node_id.to_string())?;
+        tree.insert(backup.key, backup.value)?;
+        tree.flush()?;
+        Ok(())
+    }
+
+    pub fn delete(&self, node_id: PublicKey, backup: DeleteBackup) -> Result<()> {
+        tracing::debug!(%node_id, key=backup.key, "Deleting user backup");
```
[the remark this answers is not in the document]
holzeis: It happens regularly, as dlc is deleting stuff everytime a position is opened.

crates/coordinator-commons/src/lib.rs
```diff
@@ -172,6 +177,66 @@ pub fn is_eligible_for_rollover(timestamp: OffsetDateTime, network: Network) ->
...
+        let kind = match value {
+            LN_BACKUP_KEY => RestoreKind::LN,
+            DLC_BACKUP_KEY => RestoreKind::DLC,
+            DB_BACKUP_KEY => RestoreKind::TenTenOne,
+            _ => bail!("Unkown restore kind"),
+        };
+
+        Ok(kind)
+    }
+}
```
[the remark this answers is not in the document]
holzeis: Because the backup api does not use the enum restore kind. The backup key is parsed from the key.

coordinator/src/storage.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct TenTenOneNodeStorage {
+    pub ln_storage: Arc<FilesystemPersister>,
+    pub dlc_storage: Arc<SledStorageProvider>,
+    pub data_dir: String,
+}
+
+impl TenTenOneStorage for TenTenOneNodeStorage {}
```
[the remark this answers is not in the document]
luckysori: Verified that this works if you implement it next to where the trait is defined:

```rust
impl<T> TenTenOneStorage for T where
    T: KVStorePersister + LDKStoreReader + DLCStoreProvider + Sync + Send + Clone
{
}
```

coordinator/src/storage.rs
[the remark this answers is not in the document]
holzeis: Clippy disagrees 🙈

crates/ln-dlc-storage/src/sled.rs
```diff
@@ -0,0 +1,320 @@
...
+        SledStorageProvider {
+            db: sled::open(path).expect("valid path"),
+        }
+    }
+
+    /// Migrates the old keys of the dlc-sled-storage-provider::SledStorageProvider to the
+    /// ln_dlc_crate::SledStorageProvider
+    pub fn migrate(&self) -> anyhow::Result<()> {
+        Ok(())
+    }
```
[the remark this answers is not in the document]
holzeis: No that is a left over that should be deleted. Thanks for spotting it. 👍

crates/orderbook-client/src/lib.rs
```diff
@@ -122,7 +123,7 @@ mod test {
     fn test_signature_get() {
         let secret_key = test_secret_key();
 
-        let message = create_sign_message();
+        let message = create_sign_message(b"Hello it's me Mario".to_vec());
```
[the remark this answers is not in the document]
holzeis: I didn't change the text so to not break the tests 🙂

mobile/native/src/ln_dlc/lightning_subscriber.rs
```diff
@@ -23,6 +23,9 @@ impl Node {
                             Event::PaymentFailed { .. } => {
                                 event::publish(&EventInternal::PaymentFailed)
                             }
+                            Event::SpendableOutputs { .. } => {
+                                event::publish(&EventInternal::SpendableOutputs)
+                            }
```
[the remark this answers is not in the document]
holzeis: Yes, we need that event to backup the spendable outputs if we receive them.

mobile/native/src/storage.rs
```diff
@@ -0,0 +1,225 @@
...
+impl TenTenOneStorage for TenTenOneNodeStorage {}
+
+impl TenTenOneNodeStorage {
+    pub fn new(data_dir: String, network: Network, secret_key: SecretKey) -> TenTenOneNodeStorage {
+        let mut data_dir = PathBuf::from(data_dir);
+        data_dir.push(network.to_string());
+
+        if !data_dir.exists() {
+            fs::create_dir_all(data_dir.as_path())
+                .unwrap_or_else(|_| panic!("Failed to create data dir {data_dir:?}"));
```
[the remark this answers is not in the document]
holzeis: Clippy disagrees 🙈

coordinator/src/storage.rs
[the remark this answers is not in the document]
luckysori: Not my Clippy!

crates/orderbook-client/src/lib.rs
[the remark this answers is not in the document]
luckysori: Hahaha just kidding.

coordinator/src/backup.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    pub fn restore(&self, node_id: PublicKey) -> Result<Vec<Restore>> {
+        tracing::debug!(%node_id, "Restoring backup");
+        let tree = self.db.open_tree(node_id.to_string())?;
+
+        let mut backup = vec![];
+        for entry in tree.into_iter() {
+            let entry = entry?;
```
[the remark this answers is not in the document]
holzeis: I think its not safe to use the app if some keys failed to get restored. Hence, yeah I meant to fail here 🙂 . However, there are definitely backups which are not as important, but that would require me to make some special handling.

mobile/native/src/cipher.rs
holzeis: > I'm not sure about sending over the nonce in plaintext to the server, to then retrieve it and use it when decrypting. That seems incorrect.

I based my implementation on that comment, suggesting that its fine to send the nonce in plaintext

> So you can either have each side keep track of a counter as the nonce, or you can send it in the clear before each message; so long as it's unique.

https://crypto.stackexchange.com/a/5818

holzeis: Here is the article I based my decision on adding the nonce in front of the message.

https://www.linkedin.com/pulse/understanding-nonces-use-aes-gcm-derek-hinch/

crates/ln-dlc-node/src/node/invoice.rs
```diff
@@ -35,10 +36,7 @@ use std::time::Duration;
 use std::time::SystemTime;
 use time::OffsetDateTime;
 
-impl<P> Node<P>
-where
-    P: Storage,
-{
+impl<S: TenTenOneStorage, N: Storage> Node<S, N> {
```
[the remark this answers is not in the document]
holzeis: Unfortunately not, because the trait bound is defined on the `Node` struct.

crates/coordinator-commons/src/lib.rs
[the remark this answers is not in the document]
holzeis: Looking at it, I think though we can get rid of the restore kind altogether

coordinator/src/storage.rs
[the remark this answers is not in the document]
holzeis:
```
warning: use of `expect` followed by a function call
  --> coordinator/src/storage.rs:30:18
   |
30 |                 .expect(&format!("Failed to create data dir {data_dir:?}"));
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|_| panic!("Failed to create data dir {data_dir:?}"))`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
   = note: `#[warn(clippy::expect_fun_call)]` on by default
```

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -121,16 +122,16 @@ pub struct LiquidityRequest {
 }
 
 /// An LN-DLC node.
-pub struct Node<S> {
+pub struct Node<S: TenTenOneStorage, N: Storage> {
```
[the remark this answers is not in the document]
holzeis: I only added it because clippy required it.

I think it stems from the fact that I switch from dyn trait arguments to a generic, which then had to be specified on all of our types. I don't think the constraints changed at all, it was just a different syntax.

coordinator/src/storage.rs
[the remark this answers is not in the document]
luckysori: You are right. I forgot that `expect` doesn't allow formatting.

I wonder if the panic would already gives you the path and appending the path to the top-level message is redundant.

crates/ln-dlc-storage/src/lib.rs
```diff
@@ -0,0 +1,1191 @@
...
+
+        self.get_data_with_prefix(&contracts, &[ContractPrefix::PreClosed.into()], None)
+    }
+
+    fn upsert_channel(&self, channel: Channel, contract: Option<Contract>) -> Result<(), Error> {
+        let serialized = serialize_channel(&channel)?;
+        let serialized_contract = match contract.as_ref() {
+            Some(c) => Some(serialize_contract(c)?),
+            None => None,
+        };
```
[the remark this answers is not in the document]
holzeis: That's code I've actually copied. I guess its not a map because the `serialize_contract` returns a result and that can't be handled equally in a map function.

crates/ln-dlc-node/src/node/mod.rs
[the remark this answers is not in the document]
luckysori: But the change leaks through the entire crate as we now have to define the trait bounds everywhere.

I do think there isn't much we can do about it. Favouring static dispatch should make sense in terms of performance, although benchmarking could be interesting.

maker/src/storage.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn write(&self, kind: u8, key: Vec<u8>, value: Vec<u8>) -> anyhow::Result<()> {
+        self.dlc_storage.write(kind, key, value)
+    }
+
+    fn delete(&self, kind: u8, key: Option<Vec<u8>>) -> anyhow::Result<()> {
+        self.dlc_storage.delete(kind, key)
+    }
+}
+
+impl KVStorePersister for TenTenOneNodeStorage {
```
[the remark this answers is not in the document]
holzeis: - `KVStoreProvider` tell that to the ldk team 😅. This trait is not defined by us 🙂 

Adapted for `DlcStoreProvider` 👍

mobile/native/src/cipher.rs
luckysori: Thanks for the info provided. This gives me much more confidence in the approach.

I know one of the sources mentions not needing the nonce to be random, but it seems safer to keep it random. And if we are gonna keep it random I think it would make sense to use something other than `thread_rng`.

luckysori (approved):
LGTM!

mobile/native/src/cipher.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn new(secret_key: SecretKey) -> Self {
```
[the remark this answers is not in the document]
holzeis: Any suggestions on how to do it? I need it as well for signing a message.

mobile/native/src/db/mod.rs
```diff
@@ -95,9 +106,35 @@ pub fn init_db(db_dir: &str, network: bitcoin::Network) -> Result<()> {
...
     Ok(())
 }
 
+/// Creates a backup of the database
+///
+/// Returns the path to the file of the database backup
+pub fn backup() -> Result<String> {
+    let connection = BACKUP_CONNECTION.get().lock();
+    let backup_dir = config::get_backup_dir();
+    let dst_path = Path::new(&backup_dir).join("trades.sqlite");
```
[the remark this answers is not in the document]
holzeis: what do you mean?

coordinator/src/backup.rs
holzeis: I've created a ticket for it.

https://github.com/get10101/10101/issues/1574

coordinator/src/storage.rs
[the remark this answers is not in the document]
holzeis: probably.. I can remove the format then and use expect!

mobile/native/src/db/mod.rs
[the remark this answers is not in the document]
luckysori: Here it says `trades.sqlite` but further app I saw `trades-X.sqlite`. Maybe it's unrelated.

mobile/native/src/cipher.rs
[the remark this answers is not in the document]
luckysori: You could pass something that implements a trait that allows you to sign stuff.

Passing around critical secret keys cannot be best practice.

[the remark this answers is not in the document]
holzeis: But wouldn't that implementation be the cipher, which then would require the secret key after all? 😅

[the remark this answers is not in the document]
holzeis: I've created a ticket for that https://github.com/get10101/10101/issues/1575

mobile/native/src/db/mod.rs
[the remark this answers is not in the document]
holzeis: oh thats unrelated.. the backup database does not include the network in its name. (because it is in the correct subfolder)

Issue #1563, opened 2023-11-12
Could not evaluate function for outcome


holzeis:
Just happened on regtest, when trying to open a new position.

```

[E] TIME: 2023-11-12T15:43:25.623099Z r: Failed to process DLC message: Failed to accept DLC channel offer for channel 5f3c3f0a4d46802c73bc9ba03e709268c0e9d2660c9ec9be7558e71d5f54f910: Invalid state: Invalid parameters were provided: Could not evaluate function for outcome 1048575, result was: -0.0000000000009094947017729282 from: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802,kind: SubChannelOffer
[D] TIME: 2023-11-12T15:43:27.064457Z r: Getting all non-pending channels
```

bonomat:
Do you happen to have more logs, eg the order with quantity , rate and amount? Then we can try to replicate it

holzeis:
Here we go

```

[D] TIME: 2023-11-12T15:43:25.081989Z r: Filling order with id: ad89d78d-52ea-44a8-85ec-50b1cb7e2425 filled: FilledWith { order_id: ad89d78d-52ea-44a8-85ec-50b1cb7e2425, expiry_timestamp: 2023-11-13 0:00:00.0 +00:00:00, oracle_pk: XOnlyPublicKey(c071e1fd881a55831cc38c85986c6c72193b4e3a98bcbc460f6c1ed2f78cf8167458a18010faf718756c69de33ee9e38a0d3f0ad48befa58e23a4b1f843e08be), matches: [Match { id: 56239227-8af1-43cf-9fd1-8266a0699d9b, order_id: 29330752-ba62-4cd2-96be-5e4bd775bcc1, quantity: 1000, pubkey: PublicKey(a9cd160312ffe21feac7e4e89b5f8b367f91a565131b007a0e8d01725cd513c71f55cd81bf1845045f812c7dd151218749ecc553a545f09feac4c788f3970ebe), execution_price: 37218 }] },order: Order { id: ad89d78d-52ea-44a8-85ec-50b1cb7e2425, leverage: 2.0, quantity: 1000.0, contract_symbol: BtcUsd, direction: Long, order_type: Market, state: Open, creation_timestamp: 2023-11-12 15:43:24.0 +00:00:00, order_expiry_timestamp: 2023-11-12 15:44:24.0 +00:00:00, reason: Manual, stable: false }
[...]

PR #1564, opened 2023-11-13
chore: ensure that payout curve cannot be negative

bonomat:
Afaik it should not happen as it is ensured in the pnl calculation. However, there might be some rounding errors happening at times. To ensure this is not happening we simply check that it's always greater 0 in a proptest.

Attempt to proof https://github.com/get10101/10101/issues/1563 but it didn't fail.


bonomat:
closed in favor of https://github.com/get10101/10101/pull/1568

bonomat:
Actually I think this is not related to https://github.com/get10101/10101/pull/1568/files

holzeis (approved):
LGTM 👍

Issue #1566, opened 2023-11-13
Payout curve with incorrect upper bound generated


luckysori:
An app panicked when trying to process a `SubChannelOffer`. 

```
[E] TIME: 2023-11-09T04:45:52.226227Z r: Aborting after panic in task info: panicked at 'assertion failed: `(left == right)`
  left: `20`,
 right: `21`', /Users/runner/.cargo/git/checkouts/rust-dlc-b5770de66c51b951/4e104b4/dlc-trie/src/digit_decomposition.rs:20:5
```

After starting the app again the channel was force-closed due to `App stale ChannelManager compared to ChannelMonitor after restart`

---

I was able to reproduce this locally by writing a test which generates a payout price range which surpasses the maximum price of `2^20-1`.

---

_Originally posted by @holzeis in https://github.com/get10101/10101/issues/1237#issuecomment-1807565813_

PR #1568, opened 2023-11-13
Ensure inverse payout curve does not go over oracle's maximum BTC price

luckysori:
Fixes to #1566.

@bonomat: Maybe you want to take this over, I just managed to reproduce the cause for the error seen in #1566 using the data from the logs. My understanding is that we just need to ensure that the price ranges can never surpass `2^20 - 1`. And that we should write a proptest to verify this too.

### DIFF
--- a/coordinator/src/payout_curve.rs
+++ b/coordinator/src/payout_curve.rs
@@ -109,3 +109,65 @@ pub fn create_rounding_interval(total_collateral: u64) -> RoundingIntervals {
         }],
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use orderbook_commons::order_matching_fee_taker;
+    use rust_decimal_macros::dec;
+    use trade::cfd::calculate_margin;
+
+    #[test]
+    fn payout_price_range_is_below_max_price() {
+        let initial_price = dec!(36780);
+        let quantity = 19.0;
+        let leverage_coordinator = 2.0;
+        let coordinator_margin = calculate_margin(initial_price, quantity, leverage_coordinator);
+
+        let leverage_trader = 1.0;
+        let trader_margin = calculate_margin(initial_price, quantity, leverage_trader);
+
+        let coordinator_direction = Direction::Long;
+
+        let fee = order_matching_fee_taker(quantity, initial_price).to_sat();
+
+        let total_collateral = coordinator_margin + trader_margin;
+        let rounding_intervals = create_rounding_interval(total_collateral);
+
+        let symbol = ContractSymbol::BtcUsd;
+
+        let descriptor = build_contract_descriptor(
+            initial_price,
+            coordinator_margin,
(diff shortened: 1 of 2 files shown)


bonomat:
Happy to take this over.

bonomat:
Are you sure that it should be `<2^20 - 1` and not `< 2^20` ? Because if so, then all we need to do is set 
`BTCUSD_MAX_PRICE` to `1_048_574;` and not `1_048_575` which it is at the moment;

Do you have the test to replicate the error in https://github.com/get10101/10101/issues/1566 to verify this?

luckysori:
> Are you sure that it should be `<2^20 - 1` and not `< 2^20` ? 

I think `2^20 - 1` is the _inclusive_ upper bound.

> Because if so, then all we need to do is set `BTCUSD_MAX_PRICE` to `1_048_574;` and not `1_048_575` which it is at the moment;

I think the test that I included in this PR generates an price range upper bound of `1_048_58X`. Does changing the constant also fix that?

> Do you have the test to replicate the error in #1566 to verify this?

It's in the first patch I pushed here.

crates/trade/src/cfd.rs
```diff
@@ -8,7 +8,9 @@ use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use std::ops::Neg;
 
-pub const BTCUSD_MAX_PRICE: u64 = 1_048_575;
+// TODO: is this correct? according to luckysori this should be 2^20 - 1 (i.e. 1_048_574) and not
+// 2^20 1_048_575) as it used to be
+pub const BTCUSD_MAX_PRICE: u64 = 1_048_574;
```
bonomat, line 13: Don't mind this change :/

bonomat:
> > Are you sure that it should be `<2^20 - 1` and not `< 2^20` ?
> 
> I think `2^20 - 1` is the _inclusive_ upper bound.

Right, not sure what I thought... `2^20 - 1=1048575` which is what we have already. 

Your test check though if all values are `<` and not `<=`. I think it should be `<=`. 

> 
> > Because if so, then all we need to do is set `BTCUSD_MAX_PRICE` to `1_048_574;` and not `1_048_575` which it is at the moment;
> 
> I think the test that I included in this PR generates an price range upper bound of `1_048_58X`. Does changing the constant also fix that?

This is what I believe I've fixed in https://github.com/get10101/10101/pull/1568/commits/3547e04e24ec7929cd42dfac4975c50e051d4fb0

> 
> > Do you have the test to replicate the error in #1566 to verify this?
> 
> It's in the [first patch](https://github.com/get10101/10101/pull/1568/files#diff-cc998bd0f27ef868e570ee1003700c0654e29674b57337885276ae3f621346daR121) I pushed here.

See comment above, I think it should be `<=` then it works

holzeis (approved):
LGTM 👍

PR #1569, opened 2023-11-13
feat: Support multiple oracles

holzeis:
This is a preparation step allowing the coordinator to accept multiple oracles as argument, to be able to force-close outdated positions, that have not been migrated to the new oracle.

The oracle_pubkey defines what oracle to be used for proposing new dlc channels

related to https://github.com/get10101/cloud-conf/issues/233

----

The oracle data is passed to the coordinator arguments like that

```
--oracle [pubkey1]@[host1] --oracle [pubkey2]@[host2] --oracle_pubkey [pubkey2]
```

This will tell the coordinator to build 2 oracle clients and use the second to for dlc channel proposals.

### DIFF
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -7,6 +7,7 @@ use crate::SubChannelManager;
 use crate::ToHex;
 use anyhow::anyhow;
 use anyhow::bail;
+use anyhow::ensure;
 use anyhow::Context;
 use anyhow::Result;
 use autometrics::autometrics;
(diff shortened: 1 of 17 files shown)


holzeis:
Note, users that upgrade with an existing position to the new version will not be able to force-close the dlc_channel, as the already use the new oracle. For this migration path only the coordinator will be able to force-close.

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -38,21 +39,31 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
...
+            let p2pd_oracles = self.oracles.clone();
             let sub_channel_manager = self.sub_channel_manager.clone();
-            let event_id = contract_input.contract_infos[0].oracles.event_id.clone();
+            let oracles = contract_input.contract_infos[0].oracles.clone();
+            let event_id = oracles.event_id;
             let dlc_message_handler = self.dlc_message_handler.clone();
             let peer_manager = self.peer_manager.clone();
             move || {
-                let announcement = oracle.get_announcement(&event_id)?;
+                let announcements: Vec<_> = p2pd_oracles
```
luckysori, line 52: 🔧 Should be singular if it's just one.

luckysori (approved):

PR #1582, opened 2023-11-16
chore: Instrument routes

holzeis:
Instrument all the things! :)


luckysori (approved):

bonomat:
If you add heaps of new traces, please monitor grafana billing for a while so that we don't get over our bounds.

holzeis:
> If you add heaps of new traces, please monitor grafana billing for a while so that we don't get over our bounds. 

This is just a single log message in case our api returns an error.

bonomat:
> > If you add heaps of new traces, please monitor grafana billing for a while so that we don't get over our bounds.
> 
> This is just a single log message in case our api returns an error.

All good then :)

PR #1585, opened 2023-11-16
fix: Do not swallow error when failing to broadcast transaction

holzeis:


crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -283,7 +284,7 @@ where
 
         self.blockchain
             .broadcast(tx)
-            .with_context(|| format!("Failed to broadcast transaction {txid}"))?;
```
luckysori, line 286: 🤔 This is weird. `anyhow::Context` should only add context, we should not be swallowing the error in the original version.

crates/tests-e2e/tests/force_close_position.rs
```diff
@@ -8,6 +8,10 @@ use tests_e2e::wait_until;
 #[tokio::test]
 #[ignore = "need to be run with 'just e2e' command"]
 async fn can_force_close_position() {
+    // TODO(holzeis): Adapt the test to be able to force close the DLC channel as well. At the
+    // moment this test is only able to force close the ln channel, as the maturity time will never
+    // be reached to get the oracles attestation to complete the dlc channel closure.
+
```
luckysori, line 14: 🤔 Seems unrelated to the patch. I would welcome a dedicated patch if you want to keep this TODO.

crates/ln-dlc-node/src/ldk_node_wallet.rs
holzeis: well, I don't know why the error is not printed, but this is printed with the old implementation.

`Error trying to close established channel: Wallet error Failed to broadcast transaction e146d889dea655b11af4d2d07da01fca730376162d858e2f09903445e3b314c5`

and with my changes it looks like that

`Error when broadcasting transaction: Failed to broadcast transaction 2d1ae764586b3fab568744e7fb7f0f3e340024a2f8cdd75e4139ae270af80a63: Esplora client error: HttpResponse(400) txid=2d1ae764586b3fab568744e7fb7f0f3e340024a2f8cdd75e4139ae270af80a63`

luckysori: Okay, the problem stems from this error being converted into a `rust-dlc` error which ignores the `std::error::Error::source`  when implementing `Display`.

We can merge it like this for now, but it would be nicer to either fix the problem in `rust-dlc` or only modifying the error that we feed to `rust-dlc` [here](https://github.com/get10101/10101/blob/f51686814131fe9f48f4f1fbe346664a8bc796c8/crates/ln-dlc-node/src/ln_dlc_wallet.rs#L145).

luckysori: This should work:

```rust
.map_err(|e| Error::WalletError(anyhow!("{e:#}").into()))?;
```

luckysori (approved):
Makes sense, although I suggested a couple of alternatives.

crates/ln-dlc-node/src/ldk_node_wallet.rs
holzeis: @luckysori thanks for the suggestions, but I feel this is not worth spending anymore time on it. Maybe you can clean it up next time you stumble over it 🙂

PR #1590, opened 2023-11-16
chore: Increase back arrow clickable space on settings

holzeis:


luckysori (approved):
These gigantic Dart diffs are awful (obviously not your fault).

PR #1602, opened 2023-11-17
fix: Spawn blocking on send payment

holzeis:
With the introduction of the backup we are spawning a tokio task on write. If this is called from a non tokio runtime we run into the following panic.

`Aborting after panic in task info: panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime'`


luckysori (approved):
This is fine, but I would prefer to use the static runtime when spawning instead.

PR #1599, opened 2023-11-17
fix: Spawn blocking on force-close and close channel

holzeis:
fixes #1598

PR #1603, opened 2023-11-17
feat: Refresh lightning wallet on received payment

holzeis:
refresh-wallet-on-received-payment

### DIFF
--- a/mobile/native/src/api.rs
+++ b/mobile/native/src/api.rs
@@ -58,12 +58,16 @@ pub struct Balances {
     pub lightning: u64,
 }
 
-/// Assembles the wallet info and publishes wallet info update event
+/// Assembles the wallet info and publishes wallet info update event.
 #[tokio::main(flavor = "current_thread")]
 pub async fn refresh_wallet_info() -> Result<()> {
     ln_dlc::refresh_wallet_info().await?;
 
     Ok(())
+}
+
+pub fn refresh_lightning_wallet() -> Result<()> {
+    ln_dlc::refresh_lightning_wallet()
 }
 
 #[derive(Clone, Debug)]
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -117,6 +117,20 @@ const ON_CHAIN_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 /// exact fee will be know.
 pub const FUNDING_TX_WEIGHT_ESTIMATE: u64 = 220;
 
+/// Triggers an update to the wallet balance and history, without an on-chain sync.
+pub fn refresh_lightning_wallet() -> Result<()> {
+    let node = crate::state::get_node();
+    if let Err(e) = node.inner.sync_lightning_wallet() {
+        tracing::error!("Manually triggered Lightning wallet sync failed: {e:#}");
+    }
+
+    if let Err(e) = keep_wallet_balance_and_history_up_to_date(&node) {
+        tracing::error!("Failed to keep wallet history up to date: {e:#}");
+    }
+
+    Ok(())
+}
+
 /// Trigger an on-chain sync followed by an update to the wallet balance and history.
 ///
 /// We do not wait for the triggered task to finish, because the effect will be reflected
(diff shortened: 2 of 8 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -121,16 +121,18 @@ pub const FUNDING_TX_WEIGHT_ESTIMATE: u64 = 220;
...
     // thread.
     let runtime = get_or_create_tokio_runtime()?;
     runtime.spawn_blocking(move || {
-        if let Err(e) = wallet.sync() {
-            tracing::error!("Manually triggered on-chain sync failed: {e:#}");
+        if full {
+            if let Err(e) = wallet.sync() {
+                tracing::error!("Manually triggered on-chain sync failed: {e:#}");
+            }
         }
```
luckysori, line 136: 🔧 Instead of introducing the flag it's nicer to simply:

```rust
sync_lightning_wallet() // fast

keep_wallet_balance_and_history_up_to_date() // fast and will update the screen asap

sync() // slow

keep_wallet_balance_and_history_up_to_date()
```

Obviously the double `keep_wallet_balance_and_history_up_to_date` is a bit silly, but the overhead is minimal and it happens in the background anyway.

Eventually we should somehow split up `keep_wallet_balance_and_history_up_to_date` anyway.

holzeis: I added a dedicated `refresh_lightning_wallet` api and got rid of the full flag. Note, `refresh_lightning_wallet` is now blocking allowing us to await the lightning wallet to be refreshed.

holzeis:
With 9d09c9a6292ab3da3e95b6f225cd664857ccf98b we are now waiting for the orders, positions and lightning wallet to be initialized before navigating to the wallet screen from the loading screen.

mobile/native/src/api.rs
```diff
@@ -58,14 +58,19 @@ pub struct Balances {
...
-/// Assembles the wallet info and publishes wallet info update event
+/// Assembles the wallet info and publishes wallet info update event.
 #[tokio::main(flavor = "current_thread")]
 pub async fn refresh_wallet_info() -> Result<()> {
     ln_dlc::refresh_wallet_info().await?;
 
     Ok(())
 }
 
+pub fn refresh_lightning_wallet() -> Result<()> {
```
luckysori, line 69: ❓Why not the `#[tokio::main(flavor = "current_thread")]` syntax?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -117,6 +117,20 @@ const ON_CHAIN_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 /// exact fee will be know.
 pub const FUNDING_TX_WEIGHT_ESTIMATE: u64 = 220;
 
+/// Triggers an update to the wallet balance and history, without an on-chain sync.
+pub async fn refresh_lightning_wallet() -> Result<()> {
```
luckysori, line 121: 🔧 I think this should not be `async`. You'll have to wrap in `spawn_blocking` afterwards.

luckysori (approved):

mobile/native/src/api.rs
holzeis: Didn't you give me a review remark on the other PRs that you'd prefer using the runtime we keep static? 😅

luckysori: I think you misunderstood my point, see [this comment](https://github.com/get10101/10101/pull/1599/files#r1398761985).

But it's fair to consider what's the best thing to do here.

Using `#[tokio::main]` just expands to creating a _new_ runtime and blocking on it. Alternatively we could use the "dedicated backend runtime" like you do here. [In the past](https://github.com/get10101/10101/blob/4d4a9a3e857c3bd3d8e49a7164826f79e4050424/mobile/native/src/ln_dlc/mod.rs#L128) we did run into problems because we were doing heavy work in the newly created runtime and that could block the UI thread. Maybe what you did here could make sense.

In any case, if you make `refresh_lightning_wallet` sync like it should be you won't need a runtime anyway.

PR #1605, opened 2023-11-17
fix: Return to the correct screen after returning from settings

holzeis:
https://github.com/get10101/10101/assets/382048/7485ee28-bbf0-424c-88f1-d94185edac7c


luckysori (approved):
Nice!

PR #1607, opened 2023-11-18
Reduce collaborative revert reliance on `ChannelManager`

### LINKED ISSUE #1546: Collaborative revert still uses the channel_manager which it shouldn't
See here: 

https://github.com/get10101/10101/blob/1f28dcce751cf74515e8bf6a40c7b8ee5bc4fc86/coordinator/src/collaborative_revert.rs#L49-L52

luckysori:
Fixes #1546.

In situations where the collaborative revert protocol might be useful, it is likely that the channel will have been erased from the `ChannelManager`.

Before this patch we had unavoidable dependencies on the `ChannelManager`, but with this patch we are able to perform the protocol without looking for the relevant `ChannelDetails` via the `ChannelManager`.

---

Currently untested beyond the automatic test! Maybe @bonomat wants to give this a go since he created the original issue and implemented the feature.

---

A large chunk of the diff is formatting and style fixes. I should have split it up, but too late now 😞

### DIFF
--- a/coordinator/src/collaborative_revert.rs
+++ b/coordinator/src/collaborative_revert.rs
@@ -8,13 +8,13 @@ use crate::storage::CoordinatorTenTenOneStorage;
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
-use axum::Json;
-use bdk::bitcoin::Transaction;
+use anyhow::Result;
 use bitcoin::hashes::hex::ToHex;
+use bitcoin::secp256k1::ecdsa::Signature;
 use bitcoin::secp256k1::Secp256k1;
 use bitcoin::Amount;
-use coordinator_commons::CollaborativeRevert;
-use coordinator_commons::CollaborativeRevertData;
+use bitcoin::OutPoint;
+use bitcoin::Transaction;
 use diesel::r2d2::ConnectionManager;
 use diesel::r2d2::Pool;
 use diesel::r2d2::PooledConnection;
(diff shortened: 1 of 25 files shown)


coordinator/src/collaborative_revert.rs
```diff
@@ -28,254 +29,242 @@ use dlc_manager::Storage;
...
+                    .expect("outbound liquidity to fit into u64");
+
+                (ln_inbound_liquidity_sat, ln_outbound_liquidity_sat)
+            }
+            None => {
+                let ln_liquidity = ln_liquidity
+                    .context("Unknown Lightning liquidity. Must provide that information")?;
+
+                (
+                    ln_liquidity.inbound_liquidity,
```
holzeis, line 90: ❓ So basically if we do not have the channel details we have to provide the channel capacity split ourselves? That could be very failure prune, I guess we'd have to consider the reserve here as well?

coordinator/src/collaborative_revert.rs
```diff
@@ -28,254 +29,242 @@ use dlc_manager::Storage;
...
-        .calculate_settlement_amount(revert_params.price)
-        .context("Could not calculate settlement amount")?;
+        // The minimum reserve is _at least_ 1_000 sats.
+        let reserve = reserve.max(Decimal::ONE_THOUSAND);
 
-    let trade_settled = sub_channel.fund_value_satoshis == channel_details.channel_value_satoshis;
+        reserve.to_u64().expect("Reserve to fit into u64")
+    };
+
+    let coordinator_amount = if subchannel.state == SubChannelState::OffChainClosed {
```
holzeis, line 122: 🔧 I don't think that this substitutes this `let trade_settled = sub_channel.fund_value_satoshis == channel_details.channel_value_satoshis;`

I added this `trade_settled` flag because it could happen that the lightning channel balance had already been updated even if the dlc protocol is still stuck (this is typically the reason why we want to collab revert).

Just checking if the subchannel state is already `OffChainClosed` is (as far as I remember) not enough.

coordinator/src/collaborative_revert.rs
```diff
@@ -28,254 +29,242 @@ use dlc_manager::Storage;
...
+    let trader_amount = subchannel.fund_value_satoshis - coordinator_amount as u64;
+
+    let fee = weight_to_fee(COLLABORATIVE_REVERT_TX_WEIGHT, fee_rate_sats_vb)
+        .expect("To be able to calculate constant fee rate");
 
     let coordinator_address = node.get_unused_address();
     let coordinator_amount = Amount::from_sat(coordinator_amount as u64 - fee / 2);
     let trader_amount = Amount::from_sat(trader_amount - fee / 2);
 
-    // TODO: check if trader still has more than dust
```
holzeis, line 129: ❓ did you check that? 😅

coordinator/src/collaborative_revert.rs
luckysori: Yes, if the `ChannelDetails` cannot be fetched from the `ChannelManager`, we will have to provide the liquidity split through the API. A "simple" way to get this information might be to look for it in a backed up version of the coordinator.

There is no perfect solution here. This whole collaborative revert protocol is inherently imperfect.

Moving forward we could:

- Periodically save the liquidity split in the database.
- Modify LDK to print the `ChannelDetails` before erasing the channel.

> I guess we'd have to consider the reserve here as well?

I'm not sure what you mean, but the reserve should not be included in these values.

coordinator/src/collaborative_revert.rs
luckysori: Okay, I'll look into this. The name `trade_settled` was misleading.

coordinator/src/collaborative_revert.rs
luckysori: I meant to move this TODO because I don't think we should check that here. The coordinator should forward their opinion on the values to the app and the app should check those values and build the transaction accordingly. If they find that either of the amounts is below dust, the _app_ should build a transaction with a single output.

luckysori: I've tackled the TODO the way I outlined above.

coordinator/src/collaborative_revert.rs
luckysori: I've decided to use the `ChannelDetails` here if available and let the caller pass in an `is_split` flag (together with the liquidity information) otherwise.

Given that this is an API that we must use with caution I think it's reasonable to expect us to provide these key values if everything else fails.

coordinator/src/collaborative_revert.rs
holzeis: I guess we could as well use the inbound and outbound liquidity stored on the shadow channel.

luckysori:
Relevant diff here, @holzeis.

coordinator/src/collaborative_revert.rs
```diff
@@ -25,257 +26,280 @@ use dlc_manager::subchannel::LnDlcChannelSigner;
...
+            let ln_outbound_liquidity_sat =
+                Decimal::from(details.outbound_capacity_msat) / Decimal::ONE_THOUSAND;
+            let ln_outbound_liquidity_sat = ln_outbound_liquidity_sat
+                .to_u64()
+                .expect("outbound liquidity to fit into u64");
+
+            (ln_inbound_liquidity_sat, ln_outbound_liquidity_sat)
+        }
+        None => {
+            let params = without_channel_details_params.context(
```
holzeis, line 88: 🤔 would it make sense to use the inbound and outbound capacity from the shadow channel? I feel like this would anyways the spot where I'd be looking for these values.

coordinator/src/collaborative_revert.rs
```diff
@@ -25,257 +26,280 @@ use dlc_manager::subchannel::LnDlcChannelSigner;
...
-        inbound_capacity_msat = channel_details.inbound_capacity_msat,
-        outbound_capacity_msat = channel_details.outbound_capacity_msat,
-        trader_margin = position.trader_margin,
-        coordinator_margin = position.coordinator_margin,
-        position_id = position.id,
-        "Collaborative revert temporary values"
-    );
+    tracing::debug!(%is_channel_split, "Calculating coordinator revert amount");
+
+    let coordinator_amount = if is_channel_split {
```
holzeis, line 160: 🤔 Shouldn't this be `!is_channel_split` ?

coordinator/src/collaborative_revert.rs
```diff
@@ -25,257 +26,280 @@ use dlc_manager::subchannel::LnDlcChannelSigner;
...
+
+            reserve.to_u64().expect("Reserve to fit into u64")
         }
-        true => {
-            sub_channel.fund_value_satoshis as i64
-                - (channel_details.inbound_capacity_msat / 1000) as i64
-                - channel_details.unspendable_punishment_reserve.unwrap_or(0) as i64
+    };
+
+    let is_channel_split = match channel_details_option {
```
holzeis, line 126: 🤔 Is this always correct? Comparing the balances might not be as elegant, but it was simply and worked.

coordinator/src/collaborative_revert.rs
```diff
@@ -25,257 +26,280 @@ use dlc_manager::subchannel::LnDlcChannelSigner;
...
+            // The channel is _not_ split if the `original_funding_txo` has not been set.
+            false
+        }
+        Some(ChannelDetails {
+            funding_txo: None, ..
+        }) => {
+            bail!("Cannot collaboratively revert channel without funding TXO");
+        }
+        None => {
+            let params = without_channel_details_params.context(
```
holzeis, line 149: nit: I guess if we do not have a split channel, we could simple assume that the inbound and outbound capacity have the correct values and we could return always true here.

The only reason I added this "flag" was that in some scenarios the inbound/outbound liquidity already included the position funds and sometimes not. The easiest way to check that was by comparing the ln balance and the sub channel capacity. If they match the amount was already settled into the lightning channel.

coordinator/src/collaborative_revert.rs
luckysori: It's a fair point. I sort of assumed that the shadow channel would be unreliable, but we seem to do our best to keep it synchronised.

coordinator/src/collaborative_revert.rs
luckysori: Good catch! I was meaning to swap the `if`-`else` blocks instead, but I forgot.

This does highlight the fact that our test for this feature is very bad D:

coordinator/src/collaborative_revert.rs
luckysori: Yeah, this should work the same way and is simpler. The `funding_outpoint` and the channel balances are updated at the same time when calling `set_funding_outpoint`.

coordinator/src/collaborative_revert.rs
luckysori: > nit: I guess if we do not have a split channel, we could simple assume that the inbound and outbound capacity have the correct values and we could return always true here.

I think you mean `false` provided that I action [this](https://github.com/get10101/10101/pull/1607#discussion_r1398853853) comment of yours. And I tend to agree. My first version was defaulting to `false` here. I am happy to revert to that.

With all your comments I think we can go back to a version without optional parameters, which is nice.

holzeis:
Manually tested the following scenarios

- Collab revert of healthy ln and healthy dlc channel (position open): works
- Collab revert of healthy ln and unhealthy dlc channel (position stuck at processing CloseConfirm): fails, see logs below.

```
2023-11-20 13:45:07  INFO collaborative_revert: coordinator::collaborative_revert: Proposing collaborative revert channel_id="9355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e0" coordinator_address=bcrt1q3g2z9f283zm7q9jt87qhq7jm9k5nmrk9553azf coordinator_amount=197664 trader_amount=1664
2023-11-20 13:45:07  INFO collaborative_revert: coordinator::admin: Proposed collaborative revert channel_id="9355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e0"
2023-11-20 13:45:07  INFO coordinator::message: Sending trader message: CollaborativeRevert { channel_id: [147, 85, 23, 60, 211, 32, 255, 13, 182, 131, 191, 211, 88, 104, 189, 67, 104, 199, 93, 123, 19, 192, 181, 107, 39, 49, 166, 207, 15, 89, 151, 224], coordinator_address: bcrt1q3g2z9f283zm7q9jt87qhq7jm9k5nmrk9553azf, coordinator_amount: Amount(0.00197664 BTC), trader_amount: Amount(0.00001664 BTC), execution_price: 35396, funding_txo: OutPoint { txid: e097590fcfa631276bb5c0137b5dc76843bd6858d3bf83b60dff20d33c175593, vout: 0 } } trader_id=03ea287dd9081b7945a94525410075afd1b976a6cbae75f5c2bb139e71131078b8
2023-11-20 13:45:07  INFO collaborative_revert_confirm: coordinator::routes: Confirming collaborative channel revert channel_id="9355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e0"
2023-11-20 13:45:07  INFO collaborative_revert_confirm: coordinator::collaborative_revert: Confirming collaborative revert channel_id="9355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e0" txid="6460e7779573c11c85db2e99951b16c0c475b1a19708f1543dec3782e7ad22ea"
2023-11-20 13:45:07  INFO collaborative_revert_confirm: coordinator::collaborative_revert: Broadcasting collaborative revert transaction txid="6460e7779573c11c85db2e99951b16c0c475b1a19708f1543dec3782e7ad22ea"
2023-11-20 13:45:07  INFO collaborative_revert_confirm: ln_dlc_node::ldk_node_wallet: Broadcasting transaction txid=6460e7779573c11c85db2e99951b16c0c475b1a19708f1543dec3782e7ad22ea raw_tx=020000000001019355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e00000000000ffffffff0220040300000000001600148a1422a54788b7e0164b3f81707a5b2da93d8ec5800600000000000016001435904b3292c46ad0a6dffd5f154a51b57b4712e6040047304402202dfa39ed4f600d3cc5e17bc37acb2a4b18ec010f43292e8f44dec205391408bb02201054a916e137d0fb401ec1222b8832e100b7e8413150629d7e05397e946c5ce801473044022036b3c7a3e3b7a10b2317dcfe436eb3ce17450a4a0dbebd13a5fbe78132ba601602206a3c2f61f38e9f677d161b86ca883ad3ef3ce0e6dac230d4c18a5304503d461201475221033e80a2dc5f99a4626e23bfd5bd42610e139eec8dd4ddf9f00d5504d2235296c92103a42c769bca373d1e782cbf288b8a1b546fbf162154aeb11237fc5246a185f12452ae00000000
2023-11-20 13:45:07 ERROR collaborative_revert_confirm: coordinator::routes: Could not confirm collaborative revert: Could not broadcast transaction: Failed to broadcast transaction 6460e7779573c11c85db2e99951b16c0c475b1a19708f1543dec3782e7ad22ea. Esplora client error: HttpResponse(400) channel_id="9355173cd320ff0db683bfd35868bd4368c75d7b13c0b56b2731a6cf0f5997e0"
2023-11-20 13:45:07 ERROR collaborative_revert_confirm: coordinator::routes: error=InternalServerError("Could not confirm collaborative revert")
```


It appears the amounts are calculated incorrectly

```
    "outputs": [
        {
            "addresses": [
                "bc1q3g2z9f283zm7q9jt87qhq7jm9k5nmrk9umnrwn"
            ],
            "script": "00148a1422a54788b7e0164b3f81707a5b2da93d8ec5",
            "script_type": "pay-to-witness-pubkey-hash",
            "value": 197664
        },
        {
            "addresses": [
                "bc1qxkgykv5jc34dpfkll4032jj3k4a5wyhxrz6mud"
            ],
            "script": "001435904b3292c46ad0a6dffd5f154a51b57b4712e6",
            "script_type": "pay-to-witness-pubkey-hash",
            "value": 1664
        }
    ],
```

luckysori:
> Manually tested the following scenarios
> 
>     * Collab revert of healthy ln and healthy dlc channel (position open): works
> 
>     * Collab revert of healthy ln and unhealthy dlc channel (position stuck at processing CloseConfirm): fails, see logs below.
> 

Thanks for testing, @holzeis! I'll look into it locally. It would be nice if we could write better tests for this feature, but it's hard in the context of the current e2e tests.

luckysori:
> Note, I am not sure how we can test a scenario, where the ln channel is deleted, but the transaction has not been broadcasted.

I think the easiest way would be to prevent the nodes from broadcasting a transaction. Still a bit fiddly as some transactions need to be broadcast so there needs to be some hacky logic involved.

luckysori:
@holzeis: Yet another version! Unfortunately I messed up the diff, so you'll have to review the whole thing again :(

holzeis (approved):

holzeis:
LGTM 🚀 

Thanks for iterating over it so many times 🙂

coordinator/src/routes.rs
```diff
@@ -153,6 +154,10 @@ pub fn router(
         .route("/api/admin/sign/:msg", get(sign_message))
         .route("/api/admin/connect", post(connect_to_peer))
         .route("/api/admin/channels/revert", post(collaborative_revert))
+        .route(
+            "/api/admin/channels/revert-expert",
```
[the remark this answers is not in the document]
luckysori: Sure, I'll rename it!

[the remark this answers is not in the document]
luckysori: Renamed!

PR #1609, opened 2023-11-20
fix: Fetch only limit orders

holzeis:
I am surprised this didn't had a bigger impact, but apparently we where fetching also market orders.

### DIFF
--- a/coordinator/src/orderbook/db/orders.rs
+++ b/coordinator/src/orderbook/db/orders.rs
@@ -181,24 +181,12 @@ impl From<OrderbookNewOrder> for NewOrder {
     }
 }
 
-pub fn all(
-    conn: &mut PgConnection,
-    show_expired: bool,
-    show_failed: bool,
-) -> QueryResult<Vec<OrderbookOrder>> {
-    let orders = match (show_expired, show_failed) {
-        (true, true) => orders::dsl::orders.load::<Order>(conn)?,
-        (false, false) => orders::table
-            .filter(orders::expiry.gt(OffsetDateTime::now_utc()))
-            .filter(orders::order_state.ne(OrderState::Failed))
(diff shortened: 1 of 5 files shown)


coordinator/src/orderbook/db/orders.rs
```diff
@@ -181,24 +181,12 @@ impl From<OrderbookNewOrder> for NewOrder {
     }
 }
 
-pub fn all(
-    conn: &mut PgConnection,
-    show_expired: bool,
-    show_failed: bool,
```
luckysori, line 187: ❓ Why did you remove this?

I have no opinion on it, but maybe this was useful to someone? Cc @bonomat.

holzeis: because it was unused.

luckysori: It was part of the HTTP API of the coordinator. In theory someone can query this endpoint to get information about the orderbook.

holzeis: Hmm.. Unless we were using it for some admin task (in which case that api should be in the admin file) it is not used by either the app nor the maker. (Unless I am missing something.

Happy to add it back if we think it's needed. 🙂

holzeis: Note, even if it is used for an admin task, I think it would be better to connect to the database directly for this query. (instead of exposing it publicly)

luckysori: > I think it would be better to connect to the database directly for this query.

That's a good point! Let's remove it then.

bonomat: No hard feelings on this :)

PR #1615, opened 2023-11-21
Update `Cargo.lock` after release `1.6`

luckysori:


luckysori:
Merging as there should be no risk here.

PR #1616, opened 2023-11-21
Compute trade cost correctly in app wallet history

luckysori:
Before, the trade cost (the main number associated with a trade in the wallet history screen) did not include the PNL or the order-matching fee. As such, it was only reflecting the margin change.

Now we actually do include these values in the calculation. These values were already available as they were being displayed separately.


holzeis (approved):
Concept ACK.

luckysori:
I've tested it locally and the numbers add up.

PR #1620, opened 2023-11-21
chore: Increase next wait time

holzeis:
The sub_channel_manager_periodic_check is only running every 30 seconds. Waiting for only 60 seconds has some chances that the force closure was not finalized in time, leaving the sub channel state still in `Closing` when the test fails.

Doubling the timeout should hopefully increase the chances of this test to succeed. Tested here https://github.com/get10101/10101/actions/runs/6944655389/job/18892309832?pr=1619


luckysori (approved):
Let's give it a go!

PR #1627, opened 2023-11-22
Fix position resize calculations: average execution price and closed DLC channel settlement amount

holzeis:
We must not change the average entry price if the position is reduced as the user is not buying/selling more of the contract, but selling/buying it.

This resulted into the incorrect pnl calculation in case the position was reduced.

### DIFF
--- a/coordinator/src/node/resize.rs
+++ b/coordinator/src/node/resize.rs
@@ -54,21 +54,16 @@ impl Node {
             "Resizing position",
         );
 
-        // We start the position resizing by proposing a collaborative settlement of the existing
-        // DLC channel _as if nothing had happened_. That is, the `accept_settlement_amount` is
-        // equal to the amount the accept party originally put into the contract.
-        //
         // TODO: Use subchannel resize protocol to ensure atomicity, simplify the implementation and
         // improve UX.
         {
-            // The position.trader_margin has already been subtracted the previous order-matching
-            // fee, so we don't have to touch this.
-            let unchanged_accept_settlement_amount = position.trader_margin as u64;
+            let accept_settlement_amount =
+                position.calculate_accept_settlement_amount_partial_close(trade_params)?;
 
             self.inner
                 .propose_dlc_channel_collaborative_settlement(
                     channel_id,
-                    unchanged_accept_settlement_amount,
+                    accept_settlement_amount.to_sat(),
                 )
                 .await?;
 
--- a/coordinator/src/node/resize.rs
+++ b/coordinator/src/node/resize.rs
@@ -1,7 +1,7 @@
+use crate::compute_relative_contracts;
 use crate::db;
-use crate::node::compute_relative_contracts;
-use crate::node::decimal_from_f32;
-use crate::node::f32_from_decimal;
+use crate::decimal_from_f32;
+use crate::f32_from_decimal;
 use crate::node::Node;
 use crate::payout_curve;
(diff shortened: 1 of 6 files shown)


coordinator/src/node/resize.rs
```diff
@@ -340,6 +340,12 @@ fn compute_average_execution_price(
 ) -> Decimal {
     let starting_average_execution_price = decimal_from_f32(starting_average_execution_price);
 
+    if starting_direction != trade_direction {
```
bonomat, line 343: If he is selling more than he had before the execution price should update, i.e. if he had 20 contracts long and sold 30 short he is now 10 negative at the new price.

bonomat (commented):
Can you add tests for this please?

coordinator/src/node/resize.rs
holzeis: https://github.com/get10101/10101/pull/1628#issue-2006431156

holzeis: > If he is selling more than he had before the execution price should update, i.e. if he had 20 contracts long and sold 30 short he is now 10 negative at the new price.

How is that average_execution_price calculated? In this context we only have on order matched by the bid / ask price, but we would actually need two orders don't we? one for the ask and one for the bid (and vice versa)

mobile/native/src/trade/position/handler.rs
```diff
@@ -39,34 +39,47 @@ pub async fn trade(filled: FilledWith) -> Result<()> {
...
-    {
-        let position_contracts_relative = compute_relative_contracts(*quantity, *direction);
-        let trade_contracts_relative =
-            compute_relative_contracts(trade_params.quantity, trade_params.direction);
-
-        if *contract_symbol == trade_params.contract_symbol
-            && position_contracts_relative + trade_contracts_relative != Decimal::ZERO
-        {
-            set_position_state(PositionState::Resizing)?;
+    let execution_price = match get_positions()?.first() {
```
luckysori, line 50: 🤔 I don't get why we need to compute an execution price. Surely, the execution price should be the one from the trade params, no matter the state of the current position. Examples:

1. If we are 5 long and we go 5 long at price X, we are now 10 long with an average execution price that takes into account price X.
2. If we are 5 long and we go 2 short, we are now 3 long and the PNL uses the trade execution price as a closing price. Importantly the average execution price for the position should not change in such a scenario. I need to check that.
3. If we are 5 long and we got 5 short, we no longer have a position and the PNL uses the trade execution price as the closing price.
4. If we are 5 long and we go 10 short, we are now 5 short with the trade execution price as the new average entry price. The PNL for the closed long position is calculated using the trade execution price as the closing price.

I don't think we want to mess with the trade execution price we record in the database in any of these scenarios.

luckysori: > I need to check that.

I found one UI bug already. We were wrongly updating the average execution price of the position when partially reducing the position.

luckysori:
@holzeis: I've updated the PR with my take on this. Testing the scenario where the position is first reduced and then closed, the PNL changes correctly.

holzeis:
> @holzeis: I've updated the PR with my take on this. Testing the scenario where the position is first reduced and then closed, the PNL changes correctly.

Awesome, looks great 🚀

luckysori (approved):
What a great PR.

Issue #1625, opened 2023-11-22
Pnl is not updated after resizing a position


holzeis:
## Reproduce

1. open 10 long
2. close 5 short

## Expected

The pnl should be reduced by the pnl extracted through the close 5 short

## Actual

The pnl remains the same on the resized position. Note, apparently not just for UI reasons, as closing the remaining 5 long will payout the full pnl of the 10 long.

PR #1628, opened 2023-11-22
feat: Prevent resizing more than the current position quantity

holzeis:
While working on https://github.com/get10101/10101/issues/1625 I noticed that the way we do a resizing with changing the direction completely can't work the way we have it designed at the moment.

The issue is that we have a bid and an ask price. If we are changing the general direction of our position during resizing we are not using the correct price for the new position.

As a quick fix I've added a simple UI validation rule, that prevents the user from reducing a position further than the current quantity of the position.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Prevent reducing quantity more than the current position
+
 ## [1.6.1] - 2023-11-21
 
 - Fix missing context when opening alert dialog
--- a/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
@@ -255,9 +255,13 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
     bool hasPosition = positionChangeNotifier.positions.containsKey(contractSymbol);
 
     double? positionLeverage;
+    int? positionQuantity;
+    Direction? positionDirection;
     if (hasPosition) {
       final position = context.read<PositionChangeNotifier>().positions[contractSymbol];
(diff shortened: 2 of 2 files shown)


holzeis:
Alternatively we could also automatically split up the order into two. 1. for closing the existing position and 2. for opening the new position. But the current design would not be able to show for what price the new position will be opened so I lean to preventing the user from reducing past the quantity of the existing position. 

Simply close the position and open a new one. @luckysori @bonomat wdyt?

bonomat (commented):
Ah, related to my comment here: https://github.com/get10101/10101/pull/1627#discussion_r1402182079

bonomat:
> Alternatively we could also automatically split up the order into two. 1. for closing the existing position and 2. for opening the new position. But the current design would not be able to show for what price the new position will be opened so I lean to preventing the user from reducing past the quantity of the existing position.
> 
> Simply close the position and open a new one. @luckysori @bonomat wdyt?

I would be in favor of fixing it correctly if it is possible. 

> But the current design would not be able to show for what price the new position will be opened

Why not?

holzeis:
> Why not?

Because at the moment we only show the bid or the ask price on the sell or buy tab.

Assume the following scenario: The user is 10 short and wants to go 20 long. 

Our trade screen tab would need to show the ask price for the first 10 contracts and the bid price for the second 10 contracts. Our UI design is not able to show that properly at the moment.

bonomat:
> > Why not?
> 
> Because at the moment we only show the bid or the ask price on the sell or buy tab.
> 
> Assume the following scenario: The user is 10 short and wants to go 20 long.
> 
> Our trade screen tab would need to show the ask price for the first 10 contracts and the bid price for the second 10 contracts. Our UI design is not able to show that properly at the moment.

Not sure I follow, if the user is 10 short, and he wants to be 20 long afterwards, he needs to go 30 long. For 30 long he will need to pay the ask price. He only sees the ask price in the `Long` screen tab.

holzeis:
closed in favor of https://github.com/get10101/10101/pull/1627

PR #1644, opened 2023-11-26
chore: Set log level to trace

holzeis:
This should help us debugging the payment routing issues in production.

### DIFF
--- a/mobile/lib/util/preferences.dart
+++ b/mobile/lib/util/preferences.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:get_10101/features/stable/stable_screen.dart';
 import 'package:get_10101/features/trade/trade_screen.dart';
 import 'package:shared_preferences/shared_preferences.dart';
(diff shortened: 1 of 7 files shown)


bonomat:
I guess we don't want to merge this 😅 


https://github.com/get10101/10101/assets/224613/70a1b12d-00ff-4c18-9310-b312c4f43689

luckysori:
> I guess we don't want to merge this 😅
> Screen.Recording.2023-11-26.at.12.51.23.pm.mov

AFAIK, the big channel update from the RGS server should only happen on startup for a fresh app. And they should be considerably smaller after that.

A long time ago, the LDK team did recommend using Trace logs given that we're still in a beta phase.

holzeis:
> > AFAIK, the big channel update from the RGS server should only happen on startup for a fresh app. And they should be considerably smaller after that.
> 
> 
> 
> It's still 21MB of log data. 

Should we add an opt in feature in the setting to enable tracing logs? (With a hint on the size implications)

That would help us when trying to debug some issues with users. Eg please try again sending a payment with trace logs enabled.

bonomat:
> Should we add an opt in feature in the setting to enable tracing logs? (With a hint on the size implications)

I like it.

holzeis:
@bonomat @luckysori I've added an opt-in (opt-out in test) to enable trace logs. This will allow us to instruct the user to enable the trace logs in production for a short period of time to reproduce an issue and set it back again afterwards.

Please have another look.

mobile/lib/util/preferences.dart
```diff
@@ -10,6 +11,17 @@ class Preferences {
   static const emailAddress = "emailAddress";
   static const openPosition = "openPosition";
   static const fullBackup = "fullBackup";
+  static const logLevelTrace = "logLevelTrace";
+
+  Future<bool> setLogLevelTrace(bool trace) async {
```
bonomat, line 16: Instead of setting it to a boolean you could set it to a `loglevel` or a `string` so that we can support different log levels, but I guess this is good enough for now :)

bonomat (approved):
LGTM.

mobile/lib/util/preferences.dart
holzeis: I thought about it, but it would have been a little bit more complex for something we might never add.

Happy to change it once we see the need to add other log levels as well :)

PR #1647, opened 2023-11-27
fix: Handle error on channel receive

### LINKED ISSUE #1646: tx_user_feed stops broadcasting new logins
It appears we are stopping to process new logins every now and then. Although we continue to log `New login` on our websocket, the corresponding `NewUser` event does not get broadcasted.

This ends up that we do not check if there are any pending tasks for the user on login, e.g. collab revert, expired positions or rollover.

The issue can be resolved after a restart.

holzeis:
Without that fixed we would have simply broke out of the loop and the task completed.

Note, the capacity of a 100 should have been enough, and it doesn't look like the channel got closed. Hopefully we learn more by logging errors when receiving a message from the channel.

fixes #1646

### DIFF
--- a/coordinator/src/message.rs
+++ b/coordinator/src/message.rs
@@ -14,6 +14,7 @@ use parking_lot::RwLock;
 use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::broadcast;
+use tokio::sync::broadcast::error::RecvError;
 use tokio::sync::mpsc;
 
 /// This value is arbitrarily set to 100 and defines the message accepted in the message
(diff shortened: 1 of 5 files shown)


coordinator/src/message.rs
```diff
@@ -48,10 +49,21 @@ pub fn spawn_delivering_messages_to_authenticated_users(
...
+                match user_feed.recv().await {
+                    Ok(new_user_msg) => {
+                        traders
+                            .write()
+                            .insert(new_user_msg.new_user, new_user_msg.sender);
+                    }
+                    Err(RecvError::Closed) => {
+                        tracing::error!("New user message sender died! Channel closed.");
+                        break;
+                    }
```
bonomat, line 62: Eventually we could try to re-subscribe to the channel here.

holzeis: Even if the sender died? I think resubscribing would only make sense if the sender is also able to restart.

bonomat: True. Makes sense.

Issue #1572, opened 2023-11-14
Function pieces are not continuous


luckysori:
These parameters resulted in an error when calling `payout_curve::build_contract_descriptor`:

```
initial_price=36404.5
coordinator_margin=18313
trader_margin=27469
leverage_coordinator=2.0
leverage_trader=3.0
coordinator_direction=Short
fee=0
rounding_intervals=RoundingIntervals { intervals: [RoundingInterval { begin_interval: 0, rounding_mod: 457 }] }
quantity=20.0
symbol=BtcUsd
```

luckysori:
cc @bonomat.

bonomat:
Reproducible with this test: 

```
#[test]
fn calculating_payout_curve_doesnt_crash_1() {
    let initial_price = Decimal::from_f32(36404.5).expect("to be able to parse price");
    let leverage_trader = 3.0;
    let leverage_coordinator = 2.0;
    let fee = 0;
    let coordinator_direction = Direction::Short;
    let quantity = 20.0;

    let coordinator_collateral = 18313;
    let trader_collateral = 27469;

    // act: we only test that this does not panic
    computed_payout_curve(
        quantity,
        coordinator_collateral,
        trader_collateral,
        initial_price,
        leverage_trader,
        leverage_coordinator,
        fee,
        coordinator_direction,
    )
    .unwrap();
}
```

bonomat:
> These parameters resulted in a `Function pieces are not continuous` error when calling `payout_curve::build_contract_descriptor`:
> 
> ```
> initial_price=36404.5
> coordinator_margin=18313
> trader_margin=27469
> leverage_coordinator=2.0
> leverage_trader=3.0
> coordinator_direction=Short
> fee=0
> rounding_intervals=RoundingIntervals { intervals: [RoundingInterval { begin_interval: 0, rounding_mod: 457 }] }
> quantity=20.0
> symbol=BtcUsd
> ```

Note, something is wrong in this example, the trader's margin is higher than the coordinator's margin even though the trader's leverage is `3`.

bonomat:
If you call it with the correct margins, it works. 

```
/// taken from a past crash
#[test]
fn calculating_payout_curve_doesnt_crash_1() {
    let initial_price = Decimal::from_f32(36404.5).expect("to be able to parse price");
    let leverage_trader = 3.0;
    let leverage_coordinator = 2.0;
    let fee = 0;
    let coordinator_direction = Direction::Short;
    let quantity = 20.0;

[...]

PR #1649, opened 2023-11-27
Fix: payout curve not continous

bonomat:
On the path to debug https://github.com/get10101/10101/issues/1563 I refactored the API a bit. 

I'm not sure why this refactoring fixed it, but I guess somewhere where the bounds incorrectly set 🙈.

Fixes [#1572](https://github.com/get10101/10101/issues/1572)

### DIFF
--- a/coordinator/src/payout_curve.rs
+++ b/coordinator/src/payout_curve.rs
@@ -174,4 +174,42 @@ mod tests {
             );
         }
     }
+
+    #[test]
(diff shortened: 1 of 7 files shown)


luckysori (approved):
LGTM.

I'm not sure if I agree with extracting the liquidation threshold from the `payout_curve` crate though. It seems like something that is part of computing an inverse payout curve.

PR #1655, opened 2023-11-28
chore: Add contract details to dlc channels api

holzeis:


bonomat (approved):
Does it serialize well? If so, let's get it in :)

PR #1657, opened 2023-11-28
Upgrade to `rust-lightning:0.0.117`

luckysori:
Fixes https://github.com/get10101/10101/issues/1651.

This includes an update to `rust-dlc` too.

TODOs:

- [x] Make sure that we haven't dropped any commits from our 10101 dedicated branches in the `rust-dlc` and `rust-lightning` repos. I think we may have dropped some from the `rust-dlc` branch.


crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -128,6 +132,7 @@ pub struct LiquidityRequest {
 }
 
 /// An LN-DLC node.
+#[derive()]
```
holzeis, line 135: ❓ Is this doing anything or can that be removed?

crates/ln-dlc-node/src/storage.rs
```diff
@@ -74,100 +40,82 @@ impl Default for TenTenOneInMemoryStorage {
...
-            if !entry.0.contains("monitors") {
-                continue;
+        primary_namespace: &str,
+        _secondary_namespace: &str,
+        key: &str,
+        _lazy: bool,
+    ) -> Result<(), Error> {
+        match key {
+            "manager" => *self.manager.write() = None,
+            "network_graph" => *self.network_graph.write() = None,
```
holzeis, line 97: 🔧 Not critical at all, but we should add a handling for the scorer here as well.

holzeis (approved):

crates/ln-dlc-node/src/ln/common_handlers.rs
```diff
@@ -403,21 +404,29 @@ pub async fn handle_funding_generation_ready<S: TenTenOneStorage, N: Storage>(
...
                 %err,
-                "Cannot open channel due to not being able to create funding tx"
+                "Cannot open channel due to not being able to create funding TX"
             );
+
             node.channel_manager
                 .close_channel(&temporary_channel_id, &counterparty_node_id)
                 .map_err(|e| anyhow!("{e:?}"))?;
 
+            tracing::info!("DID THIS DO ANYTHING?");
```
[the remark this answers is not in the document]
luckysori: Hahaha ooops!

crates/ln-dlc-node/src/node/mod.rs
luckysori: Nice catch!

PR #1663, opened 2023-11-30
fix: Return lsp config data in authenticated response

holzeis:
Before we have been bombarding our api everytime flutter decides to re-render a screen. I presume this is the reason, why we run into a timeout every once in a while, and why it took some time to load.

The lsp config data is now fetched once at authentication and then cached on the app side. This will significantly reduce the amount of external api calls and implicitely resolve our wait / timeout issues on the open order drawer as we are not fetching the lsp config anymore from an external api.

----
Not strictly necessary for this PR, but I took the liberty to clean up some things.

- c567a848d716d92a12a44dcff0923da5daeab1ea: Refactor the `coordinator-commons` and the `orderbook-commons` into a single commons crate, as they have anyways already been extremely close, and maintaining these two crates became an increasing pain.
- 8792eab273c357cc22b1d365675c09f42c859ff8: Remove check if the email is some. The email is optional we do not have to check if the email or nostr pubkey are set. I presume we did not run into the bad request because we are sending an empty string at the moment.
- 2a85eb62578209c8cef4cf7a66b8c2ef96add02e: As we have not been sending the authenticate message on a hot restart, my changes left the app in an intermediate / broken state. After trying different approaches I eventually decided to resend the `Authenticate` message on a hot-restart. This is only added for a nice development experience and does / should not have an impact on production. However, it also introduces some ground work to leverage the websocket connection more.


mobile/native/src/api.rs
```diff
@@ -246,16 +248,46 @@ pub fn subscribe(stream: StreamSink<event::api::Event>) {
...
+            let runtime = crate::state::get_or_create_tokio_runtime()?;
+            runtime.block_on(async {
+                tx_websocket.send(OrderbookRequest::Authenticate {
+                    fcm_token: Some(fcm_token),
+                    signature,
+                })
+            })?;
+
+            Ok(())
+        }
```
luckysori, line 289: 🤔 I don't like having to duplicate this logic just so that we can do a hot restart.

I would try to:

- Remove the `run`-`run_internal` separation and just have a single `run` function.
- Make it so that we can drop the orderbook WS task by returning a handle to it and keeping it in a static variable.
- On restart, always call `run` and, if the node is already initialised, clean up the old task and reconnect to the coordinator.

Or maybe we can refactor it a different way. But duplicating this logic is not very nice.

luckysori (approved):

mobile/native/src/state.rs
```diff
@@ -85,8 +86,6 @@ pub fn try_get_storage() -> Option<TenTenOneNodeStorage> {
 /// Lazily creates a multi threaded runtime with the the number of worker threads corresponding to
 /// the number of available cores.
 pub fn get_or_create_tokio_runtime() -> Result<&'static Runtime> {
-    static RUNTIME: Storage<Runtime> = Storage::new();
```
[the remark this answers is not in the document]
holzeis: It looked weird to me as it would create a new runtime everytime we calling this function? I am not sure about the rust syntax here. It might also be fine. I just aligned it with the other static storages that we keep in that file.

mobile/native/src/api.rs
holzeis: I improved it a little bit, but I guess it can still be done nicer. Leaving it as is for now, as this is not super critical.

Issue #1565, opened 2023-11-13
Failed to load DLC ChainMonitor even after overwriting old with empty one


holzeis:
I just found the below messages multiple times in a users logs, although they should appear only once.

```
[E] TIME: 2023-11-08T12:43:56.054569Z r: Failed to load DLC ChainMonitor from storage: Storage error Unknown required feature preventing decode 
[I] TIME: 2023-11-08T12:43:56.054805Z r: Overwriting old DLC ChainMonitor with empty one to be able to proceed 
but multiple times, shouldn't that only appear once, and then be recovered?
```

PR #1665, opened 2023-12-01
Read and write `ChainMonitor` using the same key

### LINKED ISSUE #1661: App is crashing at startup since upgrade to 1.6.0
The app is crashing for some of our users after upgrading to 1.6.0.

```
[E] TIME: 2023-11-24T18:42:12.320227Z Launching the app failed FfiException(RESULT_ERROR, Failed to start the backend

Caused by:
    0: Failed to initialise DlcManager
    1: Storage error Unknown required feature preventing decode, null)
```

luckysori:
Fixes #1661.
Fixes https://github.com/get10101/10101/issues/1565.

This fixes a kind of bug which could happen like this:

1. Create a `ChainMonitor` based on version 1.4.4 (for example). This would use the number "4" as the key.

2. Upgrade to version 1.6.5 (for example).

3. On startup, we would identify the original `ChainMonitor` as backwards-incompatible and attempt to overwrite it. Instead, we would insert it under the new key "chain_monitor".

4. Then we would build the `dlc_manager::Manager`, which would load the `ChainMonitor` internally. This would look for the _first_
`ChainMonitor` out of all the ones in the `CHAIN_MONITOR` tree, ignoring the key. Because the original one was inserted first, we
would load the original one which would once again fail to deserialise due to backwards-incompatibility.

Now we ensure that we always use the same key to read and write the `ChainMonitor`. This makes sense because there should only ever be one `ChainMonitor` per node.


crates/ln-dlc-storage/src/lib.rs
```diff
@@ -439,7 +443,7 @@ impl<K: DlcStoreProvider> dlc_manager::Storage for DlcStorageProvider<K> {
         self.store
             .write(
                 CHAIN_MONITOR,
-                "chain_monitor".to_string().into_bytes(),
+                CHAIN_MONITOR_KEY.to_string().into_bytes(),
```
holzeis, line 446: 🙈 ouch I've must have missed that, when reverting to the numbers for backwards compatibility. Nice catch! 💪

holzeis (approved):
Thanks for the fix 🚀

PR #1666, opened 2023-12-01
Reduce linking time by using Rust "unit" tests in `tests-e2e`

luckysori:
Rust "unit" tests are just tests that are located next to the code. By putting the e2e tests next to the auxiliary code in the `tests-e2e` crate, we avoid re-linking and building a separate binary for each e2e test.

Advice taken from this blog post[^1], which we already applied to the `ln-dlc-node` crate.

[^1]: https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html.

---

I was motivated to do this because linking the tests was failing for me locally. My theory is that my linker gave up having to do linking for so many test binaries.

### DIFF
--- a/crates/tests-e2e/src/lib.rs
+++ b/crates/tests-e2e/src/lib.rs
@@ -1,5 +1,8 @@
 #![allow(clippy::unwrap_used)]
 
+#[cfg(test)]
+mod tests;
+
 pub mod app;
 pub mod bitcoind;
 pub mod coordinator;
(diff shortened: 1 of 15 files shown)


holzeis (approved):
LGTM 👍

luckysori:
Converting it to draft because the e2e tests don't work since we all the `static` variables are shared when a single binary is built!

I already did some work trying to deal with this problem, but it's not ready.

luckysori:
Closing because it's harder than I thought and it's not high priority.

PR #1668, opened 2023-12-01
chore: Reduce websocket reconnect timeout to 200ms

holzeis:
This was originally increased to ensure that a rollover is not failing if the app is open and the coordinator restarts while moving into the rollover window, as the rollover will fail.

However, this is a very edge case, that shouldn't happen very often and even if it's happening. The rollover will simply timeout and finish on the next restart of the app without any errors.

Reducing this timeout to 200ms gives the app a more ready feeling as the app will not wait 5s before reconnect and thus fetching the latest limit orders to get a price.

### DIFF
--- a/mobile/native/src/orderbook.rs
+++ b/mobile/native/src/orderbook.rs
@@ -257,6 +257,9 @@ async fn handle_orderbook_mesage(
 
             *orders = initial_orders;
 
+            // if we receive a full set of new orders, we can clear the cached best price as it is
+            // outdated information.
+            cached_best_price.clear();
             update_prices_if_needed(cached_best_price, &orders);
         }
         Message::NewOrder(order) => {
--- a/mobile/native/src/orderbook.rs
+++ b/mobile/native/src/orderbook.rs
@@ -32,9 +32,12 @@ use tokio::sync::watch;
 use tokio_tungstenite::tungstenite;
 use uuid::Uuid;
 
-/// The reconnect timeout should be high enough for the coordinator to get ready after a restart. If
-/// we reconnect too early we may not be ready process messages which require DLC actions.
-const WS_RECONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+/// FIXME(holzeis): There is an edge case where the app is still open while we move into the
+/// rollover window. If the coordinator restarts while the app remains open in that scenario, the
+/// rollover will fail. However the rollover will succeed on the next restart.
+/// This could be fixed by only sending the rollover message once the channel is usable with the
+/// trader.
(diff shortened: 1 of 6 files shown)


mobile/native/src/orderbook.rs
```diff
@@ -32,9 +32,7 @@ use tokio::sync::watch;
 use tokio_tungstenite::tungstenite;
 use uuid::Uuid;
 
-/// The reconnect timeout should be high enough for the coordinator to get ready after a restart. If
-/// we reconnect too early we may not be ready process messages which require DLC actions.
-const WS_RECONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const WS_RECONNECT_TIMEOUT: u64 = 200;
```
luckysori, line 35: 🔧 I think it's better to keep the constant as a `Duration`. Then it can be used without having to know whether we wanted it to be seconds or milliseconds.

mobile/native/src/orderbook.rs
```diff
@@ -176,11 +178,13 @@ pub fn subscribe(
                 tracing::warn!("Cannot update orderbook status: {e:#}");
             };
 
-            tokio::time::sleep(WS_RECONNECT_TIMEOUT).await;
+            let retry_interval = Duration::from_millis(WS_RECONNECT_TIMEOUT) * round;
```
luckysori, line 181: 🙃 I suppose we are going for some sort of backoff here. We could use [this](https://docs.rs/backoff/latest/backoff/backoff/trait.Backoff.html) library that we've used in the past which is a bit smarter and can use [exponential backoff](https://docs.rs/backoff/latest/backoff/exponential/struct.ExponentialBackoff.html).

luckysori (approved):

mobile/native/src/orderbook.rs
holzeis: I had a quick look at the library and it feels like I would need to make far more changes to make it work with that library than I would like to invest at the moment.

mobile/native/src/state.rs
```diff
@@ -113,3 +116,20 @@ pub fn get_websocket() -> Sender<OrderbookRequest> {
...
+        Some(l) => *l.write() = sink,
+        None => {
+            LOG_STREAM_SINK.set(RwLock::new(sink));
+        }
+    }
+}
+
+pub fn get_log_stream_sink() -> Arc<StreamSink<LogEntry>> {
+    LOG_STREAM_SINK.get().read().clone()
+}
```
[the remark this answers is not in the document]
holzeis: it's not used at all at the moment. I just added it for consistency.. I'll remove it.

Issue #1678, opened 2023-12-04
Replace speed dial with buttons below the balance


holzeis:
![Screenshot 2023-12-04 at 14 11 08](https://github.com/get10101/10101/assets/382048/77dd89d6-8996-48f0-9890-0ca109eee168)

Issue #1676, opened 2023-12-04
Replace USDP tab on home screen with a swap drawer opened from the bottom


holzeis:
The screens do not 100% reflect the intended design, but is close enough to get an idea on how it should look like.

Scope:

 - Add swap button to home screen
 - Upon clicking the swap button open a drawer from the bottom showing the swap option
 - Both Lightning and USD-P should be editable
 - The selection should allow for selecting lighting and usdp, if one changes the other should be changed accordingly
 - If the amount in usdp is changed it should change the lightning amount and vice versa
 - The trading amount is the best ask price from the orderbook
 - The summary should not include an EST time
 - Upon confirming the swap a loading screen should be shown spinning until the swap was successful or failed.
 - success or failed screen should also be dispayed as drawer from the bottom.

![Screenshot 2023-12-04 at 14 01 53](https://github.com/get10101/10101/assets/382048/c64922e2-1e1d-476f-b8c9-0bf5a3f698e8)

SXBT69:
@Restioson check this out : 


![image](https://github.com/get10101/10101/assets/137571657/c32b5539-9b13-4e9d-81ad-891502940952)
![image](https://github.com/get10101/10101/assets/137571657/885b0cf7-de75-48c9-95b2-a5a543a99309)

the correct way to display it is in the "sending" flow

D

https://figma.com/file/5BgcmpxFA7AHN2bs8Cd8wC/10101-Mobile-app-design?type=design&mode=design

Restioson:
The thing that's confusing me is this line:

> Upon clicking the swap button open a drawer from the bottom showing the swap option

Does that mean that we want the entire swap screen in a drawer?

SXBT69:
It was not designed like this on figma, but I think that's not a bad idea since there is a lot of free space on the swap screen...

PR #1687, opened 2023-12-05
Replace speed dial with 3 buttons

### LINKED ISSUE #1515: Replace speed dial
With buttons underneath the wallet selector. 

image

luckysori:
Fixes #1515.
Fixes https://github.com/get10101/10101/issues/1678.

https://github.com/get10101/10101/assets/9418575/4820fb6e-8b7f-43c3-9aa3-19a9d3c80a9c

---

- Choose a speficic style for the new buttons that diverges from the `WalletTheme`. This is purely subjective, I'm just proposing something (temporary) here. I also went for no icons, since I didn't like any of the options readily available. _I'm happy to change everything related to the styles._

- Remove fund button since the now always-visible `Receive` button serves the same purpose.

### DIFF
--- a/mobile/lib/features/wallet/wallet_screen.dart
+++ b/mobile/lib/features/wallet/wallet_screen.dart
@@ -1,14 +1,12 @@
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_speed_dial/flutter_speed_dial.dart';
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
 import 'package:get_10101/common/channel_status_notifier.dart';
 import 'package:get_10101/features/wallet/balance.dart';
 import 'package:get_10101/features/wallet/receive_screen.dart';
 import 'package:get_10101/features/wallet/onboarding/onboarding_screen.dart';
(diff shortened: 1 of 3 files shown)


mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -40,16 +37,41 @@ class WalletScreen extends StatelessWidget {
...
+                        child: const Text('Receive'),
+                      ),
+                    ),
+                    const SizedBox(width: 10.0),
+                    Expanded(
+                      child: ElevatedButton(
+                        style: balanceButtonStyle,
+                        onPressed: () {
+                          // TODO: Take me to the Lightning-USDp swap screen.
+                        },
```
Restioson, line 62: Do we want this done before merge?

Restioson (approved):
Concept ACK; LGTM but not sure if we want to push the styling further in this PR or wait for follow-up

mobile/lib/features/wallet/wallet_screen.dart
luckysori: I think that will just route to the screen in your ticket: https://github.com/get10101/10101/issues/1676.

PR #1689, opened 2023-12-05
feat: Swap drawer

### LINKED ISSUE #1676: Replace USDP tab on home screen with a swap drawer opened from the bottom

Restioson:
Fixes #1676 

https://github.com/get10101/10101/assets/382048/9977f10e-467e-44c8-83e4-0b200adb0610

Screenshots:
image

image

The only thing that this doesn't address is the confirmation/fail screen, which are still the old dialogues.

Issues/to-do:
- [x] It still is registered as a screen in gorouter (will be fixed when updating to main)
- [x] Still using a dialogue for success (not a drawer as in mocks)
    - [x] Should it drop user back to wallet after success?
- [x] Balances not shown in the other currency (e.g a USD amount shown only in USD, not SAT also like in mocks)
- [x] Fee box is not up to spec either
- [x] Do we want to show expiry time?
- [x] Fine detail from figma
~~    - [ ] 'max' button~~
~~    - [ ] Show currency equivalent below input boxes~~
    - [x] Remove border & edit symbol on input boxes
    - [x] Fonts and so on (?)


Restioson:
The colours (at least for box background) I copied from figma, it is that hex code already.

The info box was removed after discussing with Richie

holzeis:
> The info box was removed after discussing with Richie 

I figured that information is better suited on the educational screen?

Restioson:
As far as the divider color, that is not based on figma, but idk if it matters.

For the ticker alignment I think it's probably more trouble than it's worth as I did try but it is not easy with the border and background

holzeis (approved):
It's functional!

Restioson:
> Looks good.
> 
> Likely unrelated: I can only swap once at the moment (i.e. acquire USDP). Then it looks like the app loses connection to the coordinator, reconnects but does not continue with the protocol

Interesting, was not happening to me. I could freely swap in either direction

mobile/lib/common/domain/model.dart
```diff
@@ -11,6 +11,11 @@ class Amount {
     _sats = Decimal.fromInt(sats);
   }
 
+  // TODO: this is bad for precision
+  Amount.fromBtc(double btc) {
```
Restioson, line 15: Note: this is really not good, we should not be using Amount to represent USD. I just did it this way because it's what's being used everywhere else in the code. We should aim to address the model issue after LoL/when time crunch is off

PR #1694, opened 2023-12-06
feat: Open scanner on send button

### LINKED ISSUE #1681: Show scanner when clicking hitting the send button
- Upper part of the screen should be the scanner
- Lower part should allow for manually entering a destination (Invoice, Bip21, on-chain address)
- Add button to copy directly from the clipboard
- Once the destination was successfully parsed, navigate to the corresponding send screen.

Screenshot 2023-12-04 at 15 47 52

holzeis:
fixes #1681 

- If the user provided permission the upper part will show the camera otherwise a hint is shown that no permission has been granted to show the camera
- destination is directly validated on the scanner screen and only forwarded to the send screen if valid.

https://github.com/get10101/10101/assets/382048/80c0177c-0de7-4264-ae0f-db8702c23b03


mobile/lib/features/wallet/scanner_screen.dart
```diff
@@ -37,38 +51,171 @@ class _ScannerScreenState extends State<ScannerScreen> {
...
+          ),
+          const SizedBox(height: 20),
+          Expanded(
+              child: !cameraPermission
+                  ? const Center(
+                      child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
+                        Icon(FontAwesomeIcons.ban, size: 55),
+                        SizedBox(height: 20),
+                        Text("Grant 10101 camera permission to use the scanner!")
+                      ]),
```
luckysori, line 91: 🔧 This message makes me feel a bit anxious haha. I like a more matter-of-fact sentence e.g. (based on Phoenix):

```suggestion
                      child: Column(mainAxisAlignment: MainAxisAlignment.center, children: [
                        Text("To scan a QR code you must give 10101 permission to access the camera.")
                      ]),
```

luckysori (approved):
LGTM beyond the message.

mobile/lib/features/wallet/scanner_screen.dart
Restioson: Can we make it do a popup to direct the user to the settings for that, like many apps do?

holzeis: We already show a popup to request access. This screen is shown if the user denied access.

Restioson: Ah I see, lgtm in that case but agree with Lucas

holzeis: fyi: this is what the user sees the first time the scanner is opened.

IMG_4069

PR #1695, opened 2023-12-07
Fix overflow in `rust-miniscript` for iOS and Android builds

luckysori:
By bumping our two miniscript versions to later patch releases which include the fix backported from `miniscript:10`.

We were only hitting this when [building for Android and iOS](https://github.com/get10101/10101/actions/runs/7086216992), since this merge commit: 3615e5bd.

---

Related issue: https://github.com/rust-bitcoin/rust-miniscript/issues/549.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,7 +321,7 @@ dependencies = [
  "getrandom",
  "js-sys",
  "log",
- "miniscript 9.0.1",
+ "miniscript 9.0.2",
  "rand",
  "serde",
  "serde_json",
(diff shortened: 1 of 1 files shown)


luckysori:
We're hitting a pretty random [error](https://github.com/get10101/10101/actions/runs/7122414203/job/19393406455?pr=1695#step:3:65) seemingly [related to Rust infrastructure](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/toolchain.20.60X.60.20is.20not.20installable/near/406421534), so I'm gonna ignore that since there is nothing to format in this PR anyway.

luckysori:
We'll see the results here: https://github.com/get10101/10101/actions/runs/7122732869.

Issue #1684, opened 2023-12-04
Rework send on-chain screen


holzeis:
Scope
- [ ] Align send on-chain screen with visual design (note, the error screen doesn't have to be a dedicated screen)
- [ ] Add on-chain network fee selection (#1527)

Not in scope
- Switch currency
- Max button

![Screenshot 2023-12-04 at 15 58 24](https://github.com/get10101/10101/assets/382048/680816ef-c359-43af-b9f5-f33643812332)

PR #1706, opened 2023-12-08
feat: Pay lightning invoice with USDP

### LINKED ISSUE #1683: Pay lightning invoice with usdp
Scope:
- Add toggle to choose between what balances a lightning invoice should be paid from
- If usdp is selected, the corresponding balance should be resized before paying the invoice
- Align screen visuals with the design proposal, note the failure screen doesn't have to be a dedicated screen.
- If the usdp balance is selected the amount should be entered as USD
- if the lightning balance is selected the amount should be entered in sats.

Not in scope
- Fee estimation is covered in https://github.com/get10101/10101/issues/1682
- Max button
- Change currency button

Screenshot 2023-12-04 at 15 50 15

holzeis:
Not shown on the video, but ..
- the error message is not jumping on the latest version
- enter amount includes a hint that entering 0 will send the maximum amount (as well as the confirmation)

send-usdp

fixes #1683 
starts some rework on #1684


holzeis:
image

luckysori (approved):
Concept ACK, nice work!

mobile/lib/features/wallet/send/confirm_payment_modal.dart
```diff
@@ -27,45 +33,110 @@ void showConfirmPaymentModal(BuildContext context, Destination destination, Amou
...
   Widget build(BuildContext context) {
     final walletService = context.read<WalletChangeNotifier>().service;
+    final submitOderChangeNotifier = context.read<SubmitOrderChangeNotifier>();
+    final formatter = NumberFormat("#,###,##0.00", "en");
 
-    final amt = destination.amount.sats > 0 ? destination.amount : amount!;
+    final tradeValuesChangeNotifier = context.watch<TradeValuesChangeNotifier>();
+
+    final tradeValues = tradeValuesChangeNotifier.fromDirection(Direction.long);
+    tradeValues.updateLeverage(Leverage(1));
```
bonomat, line 69: Why do you need to update the leverage?

holzeis: because the default leverage is 2 and if I wouldn't set it to 1 we would get an incorrect sats amount from the quantity.

bonomat (commented):
Can't say much about the flutter code :) 

Why did you go for composing it through flutter and not through rust?

holzeis:
> Why did you go for composing it through flutter and not through rust?

I think it would be cleaner having one function in rust to pay an invoice using the usdp balance.

I guess I did it through flutter because I was faster that way. Happy to refactor in a follow up PR.

bonomat:
> > Why did you go for composing it through flutter and not through rust?
> 
> I think it would be cleaner having one function in rust to pay an invoice using the usdp balance.
> 
> I guess I did it through flutter because I was faster that way. Happy to refactor in a follow up PR.

No worries, let's go!

mobile/lib/features/wallet/send/confirm_payment_modal.dart
```diff
@@ -27,45 +33,110 @@ void showConfirmPaymentModal(BuildContext context, Destination destination, Amou
       builder: (BuildContext context) {
         return SingleChildScrollView(
             child: SizedBox(
-                height: 320,
+                height: 420,
```
bonomat, line 36: of course 420, it's a multiple of 42!

mobile/lib/features/wallet/send/confirm_payment_modal.dart
```diff
@@ -27,45 +33,110 @@ void showConfirmPaymentModal(BuildContext context, Destination destination, Amou
...
-    final amt = destination.amount.sats > 0 ? destination.amount : amount!;
+    final tradeValuesChangeNotifier = context.watch<TradeValuesChangeNotifier>();
+
+    final tradeValues = tradeValuesChangeNotifier.fromDirection(Direction.long);
+    tradeValues.updateLeverage(Leverage(1));
+
+    Amount amt = destination.amount;
+    if (destination.amount.sats == 0) {
+      if (payWithUsdp) {
+        // if the destination does not specify an amount and we ar paying with the usdp balance we
```
bonomat, line 74: 🙃
```suggestion
        // if the destination does not specify an amount and we are paying with the usdp balance we
```

mobile/lib/features/wallet/wallet_screen.dart
```diff
@@ -105,7 +105,7 @@ class WalletScreen extends StatelessWidget {
                               size: 14,
                             ),
                             SizedBox(
-                              width: 10,
+                              width: 7,
```
bonomat, line 108: man, today you have the eastereggs hidden!

3  `7` changed in this file = `3 * 7` = `21` = `10101`

bonomat (approved):
Ack.

SXBT69:
goose bumps

Issue #1690, opened 2023-12-06
App outbound payments often fail


luckysori:
We replaced `P2pGossipSync` with `RapidGossipSync` and that solved some problems but not all. Some app users still encounter problems making outbound payments.

```[tasklist]
### Things that could be done
- [x] Print channel details when payment fails
- [ ] https://github.com/get10101/10101/pull/1707
- [x] Deploy a release with `rust-lightning:0.0.117` to get more verbose TRACE logs on error.
- [x] Allow app users to delete network graph and sync via RGS from scratch. https://github.com/get10101/10101/pull/1744
- [ ] https://github.com/get10101/10101/issues/1632
```

bonomat:
https://github.com/get10101/10101/pull/1744 was recently merged which might improve things.

luckysori:
No longer relevant.

PR #1707, opened 2023-12-08
Outbound payment features

### LINKED ISSUE #1682: Add fee estimation to sending a lightning invoice
Scope

- Add a probing api to lndlc crate for paying an invoice. Should return a route if possible.

Screenshot 2023-12-04 at 15 53 57

Either in this ticket or a follow up

- Should include an estimate on the lightning transaction fees
- If usdp is selected the fees should additionally include the order matching fees.

luckysori:
Fixes #1682.

I also included 3c45356f6cc5dd9929c591e959d74fe4727d9082 because I discovered the trick whilst investigating things related to #1690. We can put it in a separate PR if it's controversial, but I think it should help.


crates/ln-dlc-node/src/ln/probes.rs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(Some(new_status.clone()));
+
+                tracing::debug!(%payment_id, ?new_status, "Updating unregistered probe");
+            }
+        }
+    }
+
+    pub async fn wait_until_finished(
```
holzeis, line 41: 🔧 Instead of polling here couldn't we simply subscribe to the `Event::ProbeSuccessful` and `Event::ProbeFailed` events?

holzeis: I think that would also allow us to get rid of the `RwLock`

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -102,7 +103,7 @@ pub use wallet::PaymentDetails;
 /// The interval at which the [`lightning::ln::msgs::NodeAnnouncement`] is broadcast.
 ///
 /// According to the LDK team, a value of up to 1 hour should be fine.
-const BROADCAST_NODE_ANNOUNCEMENT_INTERVAL: Duration = Duration::from_secs(3600);
+const BROADCAST_NODE_ANNOUNCEMENT_INTERVAL: Duration = Duration::from_secs(3);
```
holzeis, line 106: 🔧 I don't think you intend to change that, did you?

crates/ln-dlc-node/src/tests/probe.rs
```diff
@@ -0,0 +1,119 @@
+use crate::node::Node;
+use crate::tests::bitcoind::mine;
+use crate::tests::init_tracing;
+use bitcoin::Amount;
+use lightning::ln::channelmanager::ChannelDetails;
+use std::time::Duration;
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore]
+async fn estimate_payment_fee() {
```
holzeis, line 10: nice 👍

crates/ln-dlc-node/src/ln/probes.rs
luckysori: > Instead of polling here couldn't we simply subscribe to the Event::ProbeSuccessful and Event::ProbeFailed events?

We could and I tried that first. But this was a lot easier to implement _for me_, so I went this way instead.

If you're really keen on that idea I'm happy to review a PR that goes in that direction.

crates/ln-dlc-node/src/node/mod.rs
luckysori: Good catch, forgot to remove this!

PR #1709, opened 2023-12-09
Ensure that coordinator's `post_sync` endpoint syncs both wallets

luckysori:
The Lightning sync is way cheaper anyway, so adding this just ensures that both wallets very likely agree on the state of the blockchain.

This could fix some intermittent failures in our e2e tests, as we were relying on manually triggering a coordinator sync in several spots. In particular, when creating Lightning funding transactions, we noticed that the two wallets could disagree on the latest block height. This could lead to an immediate force-closure of the pending channel due to:

```
Funding transaction absolute timelock is non-final
```

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -373,14 +373,19 @@ pub async fn post_broadcast_announcement(
     Ok(())
 }
 
-/// Internal API for syncing the wallet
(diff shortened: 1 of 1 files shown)


holzeis (approved):
Looks reasonable to me. 🙂

bonomat (approved):

PR #1710, opened 2023-12-09
feat: Receive USDP

### LINKED ISSUE #1685: Receive usdp funds
Scope
- Add toggle to select lightning or usdp balance to receive to
- Add toggle to select receiving via lightning or on-chain
- Show amount in usd if usdp is selected and sats if lightning is selected
- Invoice is initially generated without an amount
- Add support to add a memo to the invoice
- If usdp is selected, the received payment should be automatically transformed into a short position of the amount that has been received.

Screenshot 2023-12-04 at 16 01 50

bonomat:
Resolves #1685

Took me way too long to get here ...

TODO: 
- [x] get latest best price for new order quantity
    - [ ] ~(optional) or ideally add a new order type which lets us post an order by quantity and the orderbook decides the price~ https://github.com/get10101/10101/issues/1712
- [x] make that we go back to home screen on closing the trade dialog
- [x] squash the last few commits. I've kept them separately for better reviewing. The first one is just frontend, the second is the logic

https://github.com/get10101/10101/assets/224613/471a9c74-de4f-4cca-bf85-ef5c2a0a08bd


holzeis:
Very nice 👌 

A couple of UI/UX remarks from just looking at the video.

- I kind of get why you've added the dialog when receiving a payment, but it's kind of weird, that it pops out of nowhere. Do you think we can skip it? Or at least make it a bit more subtle?
- Are we settled on the USD-P icon? I am not a fan of it and the other places where we show usdp it does not have an icon. Either way I guess we should at least align it.
- Can you use the same styling for the buttons as used on the wallet screen?

bonomat:
> * I kind of get why you've added the dialog when receiving a payment, but it's kind of weird, that it pops out of nowhere. Do you think we can skip it? Or at least make it a bit more subtle?

I think we have to as otherwise we risk the app to be closed and get into error states.

> 
> * Are we settled on the USD-P icon? I am not a fan of it and the other places where we show usdp it does not have an icon. Either way I guess we should at least align it.

What USD-P icon? 🤔 Do you mean the dollar sign when entering the amount? 

> 
> * Can you use the same styling for the buttons as used on the wallet screen?

This is soo hard 🙈

bonomat:
> Can you use the same styling for the buttons as used on the wallet screen?

Done by extracting the style into a widget: https://github.com/get10101/10101/pull/1710/commits/f1ef53baa2691910724a549652532866c302d6c9

image

holzeis:
> I think we have to as otherwise we risk the app to be closed and get into error states.

I do agree, but I really don't like the dialog, also given that the content is pretty random 😅.

How about showing a success screen that a payment has been received which will then inform the user that his sats are swapped into usdp. Similar to the send flow.

Alternatively, (definitely not scope for this change) we should explore if it's feasible to intercept the payment on the coordinator and create the position before forwarding the payment? That way we might be able to even fail the payment if the short order failed.


> What USD-P icon? 🤔 Do you mean the dollar sign when entering the amount? 


Yes 🙂

holzeis:
Something like that maybe?

Screenshot 2023-12-10 at 10 54 52

bonomat:
> Alternatively, (definitely not scope for this change) we should explore if it's feasible to intercept the payment on the coordinator and create the position before forwarding the payment? That way we might be able to even fail the payment if the short order failed.

That's not possible because the preimage will only be released by the app.

bonomat:
> Something like that maybe?
> 
> Screenshot 2023-12-10 at 10 56 18
> 
> Correspondingly we could inform the user if the swap to usdp failed and what that means.

I made a custom dialog.

https://github.com/get10101/10101/assets/224613/22e8f746-e77d-4edb-b866-5f7d3c4dd3a3

After review I'll squash the last few commits. It's too hard to clean them up now :/

bonomat:
> > t USD-P icon? 🤔 Do you mean the dollar sign when entering the amount?
> 
> Yes 🙂

That's actually not USDP, it just allows the user to enter the invoice amount in USD.

holzeis:
> > > t USD-P icon? 🤔 Do you mean the dollar sign when entering the amount?
> > 
> > 
> > Yes 🙂
> 
> That's actually not USDP, it just allows the user to enter the invoice amount in USD.

Yes, but still can we get rid of that icon? or just us `$` instead?

holzeis:
> > Alternatively, (definitely not scope for this change) we should explore if it's feasible to intercept the payment on the coordinator and create the position before forwarding the payment? That way we might be able to even fail the payment if the short order failed.
> 
> That's not possible because the preimage will only be released by the app.

Yeah, its not possible to use the funds from the payment directly, but that's not necessarily what we have to do.

holzeis:
> I made a custom dialog.
> Simulator.Screen.Recording.-.iPhone.15.Pro.-.2023-12-10.at.11.27.21.mp4

It's weird that you received 11$ on a 10$ invoice 😅 

Could you use a drawer instead of the dialog? Also, why don't you simply reuse the `execute_swap_modal`? That way we would have it consistent.

bonomat:
> > > > t USD-P icon? 🤔 Do you mean the dollar sign when entering the amount?
> > > 
> > > 
> > > Yes 🙂
> > 
> > 
> > That's actually not USDP, it just allows the user to enter the invoice amount in USD.
> 
> Yes, but still can we get rid of that icon? or just us `$` instead?

Be my guest and play around with it :/ It's really hard to make it look nice because the input field formatting.

bonomat:
> > I made a custom dialog.
> > Simulator.Screen.Recording.-.iPhone.15.Pro.-.2023-12-10.at.11.27.21.mp4
> 
> It's weird that you received 11$ on a 10$ invoice 😅

I know, there is no way around it unfortunately. It could also be less...

> 
> Could you use a drawer instead of the dialog? Also, why don't you simply reuse the `execute_swap_modal`? That way we would have it consistent. 🙂

No particular reason. I think I have enough of flutter for a while.

holzeis:
I will have a look at the code later today 😉

coordinator/src/orderbook/routes.rs
```diff
@@ -51,6 +53,16 @@ pub async fn get_order(
     Ok(Json(order))
 }
 
+#[instrument(skip_all, err(Debug))]
+pub async fn get_orders(State(state): State<Arc<AppState>>) -> Result<Json<Vec<Order>>, AppError> {
```
holzeis, line 57: ❓ why is that api needed?

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -179,6 +180,9 @@ pub struct Node<S: TenTenOneStorage, N: Storage> {
     esplora_client: Arc<NodeEsploraClient>,
     pub pending_channel_opening_fee_rates: Arc<parking_lot::Mutex<HashMap<PublicKey, FeeRate>>>,
     pub probes: Probes,
+    // TODO: we should make this persistent as invoices might get paid later - but for now this is
+    // good enough
+    pub pending_usdp_invoices: Arc<parking_lot::Mutex<HashSet<bitcoin::hashes::sha256::Hash>>>,
```
holzeis, line 185: 🤔 Do we really have to keep that list on the shared crate? I figure we could keep that on the app alone.

mobile/lib/common/application/switch.dart
```diff
@@ -5,9 +5,14 @@ class TenTenOneSwitch extends StatefulWidget {
   final bool value;
   final ValueChanged<bool> onChanged;
   final bool showDisabled;
+  final bool isDisabled;
```
holzeis, line 8: 🤔 Is this used anywhere?

mobile/lib/common/custom_app_bar.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+class CustomAppBar extends StatelessWidget {
```
holzeis, line 4: Nice 👍 

Thanks for creating that component, we should use it everywhere 🙂

Can we rename it to `TenTenOneAppBar` instead of `CustomAppBar` ? 🙃

mobile/lib/common/domain/model.dart
```diff
@@ -64,3 +64,49 @@ class Amount {
     return formatSats(this);
   }
 }
+
+class Usd {
```
holzeis, line 68: 🔧 It looks like this component is not strictly useful for usd but rather for any fiat currency. How about renaming it to Fiat?

mobile/lib/common/secondary_action_button_style.dart
holzeis, line 1: 🔧 it's weird that the name has a style prefix when it contains a widget. 🙃

mobile/lib/features/wallet/payment_claimed_change_notifier.dart
```diff
@@ -1,20 +1,124 @@
...
+                price: price,
+                liquidationPrice: liquidationPrice,
+                fee: fee,
+                fundingRate: fundingRate,
+                expiry: expiry,
+                tradeValuesService: tradeValuesService),
+            PositionAction.open,
+            stable: true);
+
+        // TODO: figure this one out how to go back to home screen
```
holzeis, line 97: ❓ in what context do you want to go back to the Home Screen?

mobile/lib/features/wallet/payment_claimed_change_notifier.dart
```diff
@@ -1,20 +1,124 @@
...
+        // _claimed = true;
+        showDialog(
```
holzeis, line 99: 🙃 would be nice if you could use the `execute_swap_modal` here 🙂

mobile/lib/features/wallet/payment_claimed_change_notifier.dart
```diff
@@ -1,20 +1,124 @@
...
+                liquidationPrice: liquidationPrice,
+                fee: fee,
+                fundingRate: fundingRate,
+                expiry: expiry,
+                tradeValuesService: tradeValuesService),
+            PositionAction.open,
+            stable: true);
+
+        // TODO: figure this one out how to go back to home screen
+        // _claimed = true;
```
holzeis, line 98: 🔧 Shouldn't we set _claimed to true here?

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -269,7 +269,6 @@ pub fn run(seed_dir: String, runtime: &Runtime) -> Result<()> {
         let mut ephemeral_randomness = [0; 32];
         thread_rng().fill_bytes(&mut ephemeral_randomness);
 
-        event::subscribe(position::subscriber::Subscriber {});
```
holzeis, line 272: ❓ Not needed anymore?

mobile/native/src/ln_dlc/node.rs
```diff
@@ -364,6 +364,12 @@ impl Node {
             tokio::time::sleep(reconnect_interval).await;
         }
     }
+
+    pub fn register_usdp_invoice(&self, payment_hash: &bitcoin::hashes::sha256::Hash) {
+        let mut invoices = self.inner.pending_usdp_invoices.lock();
+        invoices.insert(*payment_hash);
```
holzeis, line 370: 🤔 Should we consider cleaning up that list on received payments?

mobile/lib/features/wallet/send/send_onchain_screen.dart
```diff
@@ -78,39 +78,7 @@ class _SendOnChainScreenState extends State<SendOnChainScreen> {
...
-                                    style: TextStyle(fontWeight: FontWeight.w500, fontSize: 20),
-                                  ),
-                                ],
-                              ),
-                            ],
-                          ),
-                        ),
-                      ],
-                    ),
+                    const CustomAppBar(title: "Send"),
```
holzeis, line 81: Nice thanks 🙏

mobile/lib/features/wallet/send/send_lightning_screen.dart
```diff
@@ -253,7 +222,8 @@ class _SendLightningScreenState extends State<SendLightningScreen> {
                     ),
                     TenTenOneSwitch(
                         value: _payWithUsdp,
-                        showDisabled: false,
+                        isDisabled: usdpBalance <= 0,
```
holzeis, line 225: 🤔 hmm.. isn't it nicer to let the user select the usdp option and show the user directly that they are missing funds. I feel like this could get annoying not knowing why the switch is disabled.

holzeis (approved):
Looks good 🚀

I left minor remarks, but we can also address / discuss them in follow ups. 🙂

coordinator/src/orderbook/routes.rs
bonomat: I used it to manually retrieve the current orders from the coordinator. I think this endpoint is useful even though it's not being used in the app

crates/ln-dlc-node/src/node/mod.rs
bonomat: You are right, can be stored some levels up . Ideally it was stored in the DB

mobile/lib/features/wallet/payment_claimed_change_notifier.dart
bonomat: Oh , that's already resolved

mobile/native/src/ln_dlc/mod.rs
bonomat: It didn't do anything

mobile/native/src/ln_dlc/node.rs
bonomat: I figured it's in memory only so not hurting, but I guess if it's in the db we should

crates/ln-dlc-node/src/node/mod.rs
bonomat: a9125f2a2812743a7adad436e0bfba09e824c5ff

mobile/lib/common/custom_app_bar.dart
bonomat: ca87c51a183b6d4e22b2e9814cc247c99c7dc149

crates/ln-dlc-node/src/node/mod.rs
bonomat: a9125f2a2812743a7adad436e0bfba09e824c5ff

mobile/lib/common/domain/model.dart
bonomat: Let's deal with this once we have other currencies than USD :)

mobile/lib/features/wallet/send/send_lightning_screen.dart
bonomat: https://github.com/get10101/10101/issues/1713

coordinator/src/orderbook/routes.rs
holzeis: Gotya, should we move it to admin then and add a comment on why it's there? 🙂

mobile/native/src/state.rs
```diff
@@ -51,6 +52,18 @@ pub fn get_node() -> Arc<Node> {
     NODE.get().read().clone()
 }
 
+pub fn register_pending_usdp_invoices(hash: bitcoin::hashes::sha256::Hash) {
```
holzeis, line 55: 🔧 shouldn't these functions rather be on the node than in the `state.rs`? 

This file should only contain boilerplate code for getting and setting states. (Storage)

bonomat: done

SXBT69:
Minor suggestions :

- use this as icon in the USDp side : 💲
- Can we switch the positions of Lightning and on-chain at the bottom?🙈

bonomat:
> * use this as icon in the USDp side : 💲

Feels redundant with the Dollar sign below? 

image


> * Can we switch the positions of Lightning and on-chain at the bottom?🙈

Mhmm... any particular reason why? :) 

I'll go ahead and merge this for now :)

SXBT69:
> Mhmm... any particular reason why? :)

Yes, accidental thumb clic on "on-chain" but minor remark

PR #1717, opened 2023-12-11
Handle resizing position on app restart

### LINKED ISSUE #1645: Closed position with signed dlc channel
Going over our logs, I've found that we have closed position with a signed dlc channel. That should not be possible.

luckysori:
Fix #1645.


holzeis (approved):
It might not change a lot but at least we now do not attempt to fix a broken state that we shouldn't. 👍

luckysori:
> It might not change a lot but at least we now do not attempt to fix a broken state that we shouldn't. 👍

Locally I reproduced a situation with a `Resizing` position and an `OffChainClosed` subchannel and these changes ensured that the resize protocol was completed after restarting the app.

PR #1722, opened 2023-12-11
feat: upload released binaries to server

bonomat:
This will upload the binaries directly to our testvm. A script on the test vm will check for the existance of the file and deploy it if necessary


bonomat:
It's a non-invasive change on the repo, so I'm just going ahead and merge this.

PR #1724, opened 2023-12-12
Prevent `coordinator::orderbook::trading` task from stopping if getting DB connection times out

### LINKED ISSUE #1672: Prevent `orderbook::trading` task from stopping
As it is currently unsupervised, if the channel connecting the HTTP API routes and the trading task ever breaks, the coordinator is unable to accept new orders until it is restarted.

#### Proposal

By supervising the task we should be able to heal from a closed channel without having to restart the whole binary.

luckysori:
Fix #1672.

---

Fixing this has made me realise that this is an antipattern that we have (very likely) been running into for a long time. That is,
using a `?` when setting up a spawned task and assuming that it will be handled somewhere.

Here it caused two problems:

1. It stopped a task that we did not want to stop.
2. It failed silently.

In some cases it might be acceptable to stop a task, but we don't want this to happen silently.

I will keep an eye out for this from now on. And I will audit the code base for instances where we use `?` when calling `pool.get()`.

### DIFF
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -328,12 +328,19 @@ async fn main() -> Result<()> {
     // Start the metrics exporter
     autometrics::prometheus_exporter::init();
 
-    tracing::debug!("listening on http://{}", http_address);
-    axum::Server::bind(&http_address)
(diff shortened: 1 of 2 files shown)


bonomat (approved):
LGTM

Issue #142, opened 2023-02-15
Investigate diesel


bonomat:
For our databases (e.g. #26 ) we want to consider [Diesel](https://diesel.rs/) instead sqlx this time. 

The goal of this ticket is to have a simple PoC setup with CRUD functionality

Things to look into: 
- can we have a swappable DB, e.g. can we switch postgres for sqlite? 
- how complex does the model get compared to sqlx?
- any other obvious drawbacks?

1010Tom:
I've started with this today. 
In case someone wants to have a look (cc @da-kami ) how to use diesel in a simple setup, please checkout this branch: 
https://github.com/get10101/10101/compare/feat/orderbook?expand=1

Regarding the outlined questions

> can we have a swappable DB, e.g. can we switch postgres for sqlite?

Short answer: it is not easily possible to write generic code which would allow us to use either sqlite or postgres. 

Longer answer: https://github.com/diesel-rs/diesel/issues/2384


> how complex does the model get compared to sqlx?

So far I can't say much about this due to using very simple types.

> any other obvious drawbacks?

It's sync but that shouldn't be a problem (imho)

da-kami:
Note: Have you considered using [`diesel-async`](https://github.com/weiznich/diesel_async)? 
At least for Posres it supports a full async connection. It's created by the creator of diesel.
(Very long discussion here: https://github.com/diesel-rs/diesel/issues/399)

Not saying that we will *need* it from the getgo, but with sync code we will limit the number of db connections that can execute to the number of threads (and a threadpool). I think for the app and sqlite that should be fine, I don't expect heavy load there, but for the Ordinator this might be a different story, as there will be more load.

We can roll with a sync implementation initially and can see :)

1010Tom:
> Note: Have you considered using [`diesel-async`](https://github.com/weiznich/diesel_async)?

I don't see the need for async yet. I would assume that our db queries are super short running which should hopeflly hardly have an effect on the remaining application.

bonomat:
We decided to go with diesel for mobile #26  and the coordinator #27

PR #1725, opened 2023-12-12
Keep long-running tasks alive

luckysori:
Inspired by #1724.


coordinator/src/message.rs
```diff
@@ -71,7 +70,13 @@ pub fn spawn_delivering_messages_to_authenticated_users(
...
         async move {
             while let Some(notification) = receiver.recv().await {
-                let mut conn = pool.get()?;
+                let mut conn = match pool.get() {
+                    Ok(conn) => conn,
+                    Err(e) => {
+                        tracing::error!("Failed to get DB connection: {e:#}");
+                        continue;
+                    },
+                };
```
Restioson, line 79: Maybe it would be cleaner here to wrap the inside of the `while` with a `try` block (idk if stable, we could use an async block/closure instead possibly)? that way, `?` would only bubble up to that point.

coordinator/src/node/rollover.rs
```diff
@@ -156,14 +156,18 @@ impl Rollover {
...
         &self,
-        conn: &mut PgConnection,
+        pool: Pool<ConnectionManager<PgConnection>>,
         notifier: mpsc::Sender<OrderbookMessage>,
         trader_id: PublicKey,
         network: Network,
     ) -> Result<()> {
+        let mut conn = spawn_blocking(move || pool.get())
+            .await
+            .expect("task to complete")?;
```
Restioson, line 166: Why do we spawn_blocking here but not elsewhere? Additionally, can we not use a native async connection pool if this is a big issue?

coordinator/src/message.rs
luckysori: `try` would be super cool, but it's not stable.

I wasn't gonna bother originally, but since you mention it I can just extract the inside of the `while` loop into a function, like in all the other places where I fixed this.

luckysori: See https://github.com/get10101/10101/compare/7b236da36f5196ad73e6dc0da9088c677556d3e1..e0a341c90840f14dbf17b9ae3b4e0b21c59b3aa2.

coordinator/src/node/rollover.rs
luckysori: > Why do we spawn_blocking here but not elsewhere?

What do you mean by "elsewhere"? In this PR I have tried to consistently use `spawn_blocking` when calling `pool.get()`. 
It's very possible that we aren't doing this consistently, but we should because `get` blocks until the connection is retrieved or until a timeout expires.

> Additionally, can we not use a native async connection pool if this is a big issue?

Yep, we can, although we didn't prioritise it [thus far](https://github.com/get10101/10101/issues/142#issuecomment-1436169522). It looks like we could use https://github.com/djc/bb8.

coordinator/src/message.rs
Restioson: Looks good, that works too!

coordinator/src/node/rollover.rs
Restioson: > What do you mean by "elsewhere"? In this PR I have tried to consistently use spawn_blocking when calling pool.get().

My mistake, maybe I was checking the deleted and not added part :sweat_smile:

bonomat (approved):
Looks good

PR #1734, opened 2023-12-12
Roll back DB transaction if resize protocol start fails 

bonomat:
If we do this we are less likely to have inconsistent state between the DB and the subchannel data.

We should do this in more places.

### DIFF
--- a/coordinator/src/node/resize.rs
+++ b/coordinator/src/node/resize.rs
@@ -14,6 +14,7 @@ use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
 use commons::order_matching_fee_taker;
 use commons::TradeParams;
+use diesel::Connection;
 use diesel::PgConnection;
 use dlc_manager::contract::contract_input::ContractInput;
 use dlc_manager::contract::contract_input::ContractInputInfo;
@@ -23,6 +24,7 @@ use rust_decimal::prelude::Signed;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 use rust_decimal::RoundingStrategy;
+use tokio::task::block_in_place;
 use trade::cfd::calculate_long_liquidation_price;
 use trade::cfd::calculate_margin;
 use trade::cfd::calculate_short_liquidation_price;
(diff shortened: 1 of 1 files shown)


luckysori (commented):
This looks like a fix so that we don't run into https://github.com/get10101/meta/issues/262, nice!

If you don't mind, I'm going to refactor it slightly and also test that it works. An automatic test seems out of the question, but I think I can manually reproduce the original bug.

holzeis:
I also propose that we get rid of the task that lead to that issue originally.

Let's set the position to closed when processing the `CloseFinalize` resp. `CloseRevoke` message. IMHO that's easier to understand and maintain.

luckysori:
> I also propose that we get rid of the task that lead to that issue originally.
> 
> Let's set the position to closed when processing the `CloseFinalize` resp. `CloseRevoke` message. IMHO that's easier to understand and maintain.

100% agree.

I was actually unable to reproduce the bug, although maybe @holzeis had better luck. It's a bit frustrating, but at this stage I think the `closed_positions.rs` task is at fault and removing it would be for the better.

bonomat:
> Let's set the position to closed when processing the `CloseFinalize` resp. `CloseRevoke` message. IMHO that's easier to understand and maintain.

Oh, I misunderstood this task. I thought it was for cleaning up old positions. Totally agree with you that we should just update the db when receiveing `CloseFinalize` .

bonomat:
> > I also propose that we get rid of the task that lead to that issue originally.
> > Let's set the position to closed when processing the `CloseFinalize` resp. `CloseRevoke` message. IMHO that's easier to understand and maintain.
> 
> 100% agree.
> 
> I was actually unable to reproduce the bug, although maybe @bonomat had better luck. It's a bit frustrating, but at this stage I think the `closed_positions.rs` task is at fault and removing it would be for the better.

I was able to reproduce it here: https://github.com/get10101/10101/pull/1740

bonomat:
@luckysori : did you say today you would like to see this change happening but refactor it a bit?

luckysori:
> @luckysori : did you say today you would like to see this change happening but refactor it a bit?

Yep, I can push a couple changes to this PR later.

luckysori:
I've refactored as discussed, @bonomat. You can't give approval to your own PR, but let me know if this is okay.

bonomat (commented):
If this works, let's get it in. Thanks @luckysori

luckysori (approved):
Approved on behalf of @bonomat!

PR #1740, opened 2023-12-13
wip: reproduce timing issue

bonomat:
if a position is getting resized we close the current contract and reopen a new one. It might happen that the task  will pick it up in the meantime and set the  to closed. This breaks the renew protocl because we expect the position to be in a certain state/

How to reproduce: 
- start the app
- open position
- resize/add contracts to the position

the task `closed_position` will pick it up and set the position to closed breaking the renew protocol.

### DIFF
--- a/coordinator/src/node/resize.rs
+++ b/coordinator/src/node/resize.rs
@@ -98,7 +98,17 @@ impl Node {
             },
         )?;
 
-        db::positions::Position::set_open_position_to_resizing(conn, position.trader.to_string())?;
+        let pool = self.pool.clone();
+        tokio::spawn({
+            let string = position.trader.to_string();
+            let pool = pool.clone();
+            async move {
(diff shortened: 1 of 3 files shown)


bonomat:
Closing this as it's not meant to be merged

coordinator/src/node/resize.rs
```diff
@@ -98,7 +98,17 @@ impl Node {
             },
         )?;
 
-        db::positions::Position::set_open_position_to_resizing(conn, position.trader.to_string())?;
+        let pool = self.pool.clone();
+        tokio::spawn({
```
luckysori, line 102: Interestingly, I wasn't able to reproduce it, although I was using `std::thread::spawn` instead.

luckysori (commented):
Thanks for the repro!

PR #1744, opened 2023-12-13
feat: Add delete network graph in settings

holzeis:
My mainnet app was not able to send any payments. After deleting and recreating the network graph it was working again. This is certainly not a fix for the payment issues, but it may help as a quick fix for others as well to unblock payments.

Screenshot 2023-12-13 at 17 49 24


bonomat (approved):
LGTM

PR #1745, opened 2023-12-13
feat(ci): auto assigns PRs to the author

bonomat:
So that @holzeis doesn't have to manually assign us all the time 😅

-----
image

-----

image

-----

image

-----

...

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,5 @@
+# Set to true to add reviewers to pull requests
+addReviewers: false
+
+# Set to true to add assignees to pull requests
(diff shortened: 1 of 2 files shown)


.github/auto_assign.yml
```diff
@@ -0,0 +1,5 @@
+# Set to true to add reviewers to pull requests
+addReviewers: false
+
+# Set to true to add assignees to pull requests
```
holzeis, line 4: The comment suggests that this is a Boolean attribute, but it is set to author. Is the comment wrong?

holzeis (approved):

PR #1749, opened 2023-12-14
Trigger on/off-chain syncs back to back in tests

luckysori:
Motivated by this [error](https://github.com/get10101/10101/actions/runs/7196185313/job/19600541210#step:12:2970) spotted on CI (thanks, @Restioson).

---

I think we actually need to do the same for production code. You want BDK and LDK to agree on the state of the blockchain at all times. Obviously this problem is more obvious in tests because we were explicitly mining a block and then triggering 1 out of 2 syncs. But we could get unlucky and things could blow up on mainnet too.


holzeis (approved):
LGTM 👍

PR #1751, opened 2023-12-14
Increase likelihood that payments below dust go through

luckysori:
With transaction fees going up, certain payments that would normally succeed can fail because they end up below the dust limit.

By increasing the `max_dust_htlc_exposure_msat` from a dynamic value of `5 * high_priority_feerate_per_kw` sats to a fixed value of 10_000_000 sats, we increase the likelihood that payments below dust go through.

For example, in a high transaction fee environment where the high priority fee rate is 200 sat/vByte (50_000 sat/KW), before this patch we would allow up to `50_000 * 5 = 250_000` sats in dust HTLCs, whereas now we would allow up to 10_000_000 sats.

This does come at the risk of losing more funds to transaction fees if dust HTLCs are part of the channel when it is force-closed.

### DIFF
--- a/crates/ln-dlc-node/src/config.rs
+++ b/crates/ln-dlc-node/src/config.rs
@@ -3,6 +3,7 @@ use lightning::ln::channelmanager::MIN_CLTV_EXPIRY_DELTA;
 use lightning::util::config::ChannelConfig;
 use lightning::util::config::ChannelHandshakeConfig;
 use lightning::util::config::ChannelHandshakeLimits;
+use lightning::util::config::MaxDustHTLCExposure;
 use lightning::util::config::UserConfig;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


holzeis:
Will that have an effect if we do not adapt the coordinator channels as well? Also 250k sats doesn't sound too bad.

holzeis (commented):
Will that have an effect if we do not adapt the coordinator channels as well? Also 250k sats doesn't sound too bad.

holzeis (approved):
Let's give this a go!

holzeis:
@luckysori In case the negotiated fees on that channel are very high it would mean we can send the max amount of 10k sats right?

holzeis:
@luckysori Is there anything left on this PR or can we get it merged?

luckysori:
> @luckysori In case the negotiated fees on that channel are very high it would mean we can send the max amount of 10k sats right?

I'm not sure I get you.

The example I left in the docs might be helpful:

```rust
/// Fee rate multiplier used to dynamically calculate how many msats we allow to have in incoming,
/// dust HTLCs for a channel.
///
/// The larger this value the less likely we are to reject a payment that would produce a dust HTLC.
///
/// ### Example
///
/// - High priority fee: 200 sat/vByte (50_000 sat/KW).
///
/// - Fee rate multiplier: 100_000.
///
/// Total dust HTLC sats: (50_000 * 100_000) / 1_000 = 5_000_000.
const MAX_DUST_HTLC_EXPOSURE_FEE_RATE_MULTIPLIER: MaxDustHTLCExposure =
    MaxDustHTLCExposure::FeeRateMultiplier(100_000);
```

We set the `max_dust_htlc_exposure` to `MaxDustHTLCExposure::FeeRateMultiplier(100_000)` for both parties, so the app-coordinator channels will have a variable max dust HTLC exposure. E.g. if the current fee rate is 200 sat/vByte (like in the example above) the parties will accept up to 5 million sats in dust HTLCs (including any other inbound, unclaimed dust HTLCs already in the channel).

Also I don't think the fee rate is negotiated. It simply depends on the fee rate source for each party. It allows either party to reject inbound dust HTLCs based on _their view_ of the blockchain transaction fees.

luckysori:
> @luckysori Is there anything left on this PR or can we get it merged?

I don't think so, but I wasn't sure if we were all happy with the tradeoff. What do you think, @bonomat?

bonomat:
> > @luckysori Is there anything left on this PR or can we get it merged?
> 
> I don't think there's anything left, but I wasn't sure if we were all happy with the tradeoff. What do you think, @bonomat?

If I understand it correctly than we will simply keep _dust_ HTLCs in the holding cells until they are resolved. If we were to force close in the mean time, then this amount will go to the miners. 
Meaning, if we have a fee of 200 sats/vbyte we might lose up to 5m sats to miners?

luckysori:
Not needed anymore.

PR #1752, opened 2023-12-14
Miscellaneous improvements

luckysori:
This is the more important one: 15963d97e2d18991f6b0f78fb9cdb71af3b9e0c5. I wasted some time debugging because of this yesterday.

### DIFF
--- a/coordinator/src/position/models.rs
+++ b/coordinator/src/position/models.rs
@@ -21,7 +21,7 @@ use trade::cfd::calculate_pnl;
 use trade::ContractSymbol;
 use trade::Direction;
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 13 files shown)


coordinator/src/position/models.rs
```diff
@@ -827,37 +827,24 @@ pub mod tests {
...
             self
         }
 
-        pub(crate) fn with_direction(mut self, direction: Direction) -> Self {
+        fn with_direction(mut self, direction: Direction) -> Self {
             self.direction = direction;
             self
         }
     }
 }
```
holzeis, line 850: 🙃 I guess that should have gone into the previous commit.

holzeis (approved):

PR #1754, opened 2023-12-14
Instruct maker to send random probes every minute

### LINKED ISSUE #1632: Add prober service
A service that is probing payments to major nodes regularly to produce an accurate scorer file. This scorer file can then be shared with the 10101 apps to improve the probability of a successful payment.

luckysori:
Fix #1632.

As the maker sends random probes through the coordinator and collects the results, the maker's scorer will get updated over time.

We can then share the scorer file with other peers connected to the coordinator (10101 apps), which should improve their payment experience.

This last part is not included in this PR because I want to see how this works in production before we actually support the app pulling the maker's scorer file.

---

Based on this [patch](https://git.bitcoin.ninja/index.cgi?p=ldk-sample;a=blobdiff;f=src/main.rs;h=1d6edb45006a70d53e8a3d466dd9968c3c8d1dfb;hp=6480805d92759eed4d90407ada06a3f6eb7f202f;hb=f47b6f50b85ef3770173cc8e4bcddd3662cb5d5e;hpb=37da86a8a1a3294f991657ecca49c19e7c2b043d) shared by Matt from LDK.

### DIFF
--- a/maker/src/probing.rs
+++ b/maker/src/probing.rs
@@ -1,0 +1,122 @@
+use anyhow::Context;
+use anyhow::Result;
+use bitcoin::secp256k1::PublicKey;
+use lightning::routing::gossip::ReadOnlyNetworkGraph;
+use lightning::routing::router::PaymentParameters;
+use lightning::routing::router::RouteParameters;
+use lightning::routing::router::ScorerAccountingForInFlightHtlcs;
+use lightning::routing::scoring::ProbabilisticScoringFeeParameters;
+use ln_dlc_node::node::Node;
+use ln_dlc_node::node::Storage;
+use ln_dlc_node::storage::TenTenOneStorage;
+use rand::thread_rng;
+use rand::Rng;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::task::spawn_blocking;
+
+const MAX_AMOUNT_MSAT: u64 = 500_000_000;
+
+const FINAL_CLTV_EXPIRY_DELTA: u32 = 144;
+
+const MAX_PATH_COUNT: u8 = 1;
+
+pub async fn send_payment_probes_regularly<
+    S: TenTenOneStorage + 'static,
+    N: Storage + Sync + Send + 'static,
+>(
+    node: Arc<Node<S, N>>,
+) {
+    let mut interval = tokio::time::interval(Duration::from_secs(60));
+    loop {
+        interval.tick().await;
+        if let Err(e) = spawn_blocking({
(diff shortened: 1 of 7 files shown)


maker/src/probing.rs
```diff
@@ -0,0 +1,122 @@
...
+        &channel_manager.get_our_node_id(),
+        &route_parameters,
+        &node.network_graph,
+        Some(&usable_channels),
+        node.logger.clone(),
+        &inflight_scorer,
+        &ProbabilisticScoringFeeParameters::default(),
+        &[32; 32],
+    );
+    if let Ok(route) = route_res {
```
holzeis, line 96: ❓ Shouldn't we at least log if the result of find route was an error?

maker/src/probing.rs
```diff
@@ -0,0 +1,122 @@
...
+        for path in route.paths {
+            let _ = channel_manager.send_probe(path);
```
holzeis, line 98: ❓ shouldn't we log something here as well?

maker/src/probing.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+
+fn random_node(network_graph: ReadOnlyNetworkGraph) -> Result<Option<PublicKey>> {
```
holzeis, line 103: 🙃 Might be very unlikely, but should we prevent sending to our own node?

holzeis (commented):
🤔 I guess it was the easiest to put that code to the maker, but IMHO it shouldn't be there.

Also would that work if the maker does not have a channel with the coordinator? or would we need to create a channel?

maker/src/probing.rs
luckysori: No need to log, events are already logged.

maker/src/probing.rs
luckysori: Same as above.

maker/src/probing.rs
luckysori: Doesn't matter, given that we probe every minute. It's not useful info, but it doesn't hurt.

luckysori:
> 🤔 I guess it was the easiest to put that code to the maker, but IMHO it shouldn't be there.

Yeah, it should not be there, but probing is not exactly a heavy or risky activity, so I don't see any harm in putting this there temporarily.

As soon as we want to use the maker in a more interesting way, moving the code out is trivial.

> Also would that work if the maker does not have a channel with the coordinator? or would we need to create a channel?

I need to test that. I would assume you do need a channel.

PR #1756, opened 2023-12-14
chore: remove trade values in sending on chain

Restioson:
This could cause issues and is unused in any case


holzeis (approved):
Thanks 👍

Issue #1761, opened 2023-12-15
Broken state: Storage error: No sk for provided pk


bonomat:
I received this error today after finally being able to reject an outaded offer

```
flutter: [E] TIME: 2023-12-15T14:26:04.576926Z r: Could not accept subchannel offer Failed to accept DLC channel offer for channel db0bcee4f9cc27da2235baa6f7990a9d70e834bbf7abfba8de4dbc709cd8774d: Invalid state: Storage error No sk for provided pk. Continuing with rejecting it
```

In this state I cannot create a new subchannel.

Note: it could be because I recovered from seed once :/

bonomat:
TODO: @bonomat check if logs are still present

PR #1763, opened 2023-12-15
fix: reject pending offers

bonomat:
I was able to reject an old unacceptable sub channel offer with this, however, I was then in a broken state. See here: #1761

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -107,6 +107,7 @@ const PROCESS_INCOMING_DLC_MESSAGES_INTERVAL: Duration = Duration::from_millis(2
 const UPDATE_WALLET_HISTORY_INTERVAL: Duration = Duration::from_secs(5);
 const CHECK_OPEN_ORDERS_INTERVAL: Duration = Duration::from_secs(60);
 const ON_CHAIN_SYNC_INTERVAL: Duration = Duration::from_secs(300);
+const WAIT_FOR_CONNECTING_TO_COORDINATOR: Duration = Duration::from_secs(2);
 
 /// The weight estimate of the funding transaction
 ///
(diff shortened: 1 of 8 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -390,7 +393,32 @@ pub fn run(seed_dir: String, runtime: &Runtime) -> Result<()> {
             }
         });
 
-        runtime.spawn(track_channel_status(node.clone()));
+        runtime.spawn({
+            let mut iteration_count = 0;
+
+            while !node
```
holzeis, line 399: 🤔 is this really needed. It think the channel status is derived from the shadow channel in the database.

holzeis (approved):
LGTM 👍

> I was able to reject an old unacceptable sub channel offer with this, however, I was then in a broken state. See here: #1761

I think your guess that a key was not successfully backed up is very likely for why you ended up in the broken state.

mobile/native/src/ln_dlc/mod.rs
bonomat: I failed once to accept the order because I could not send a message to the coordinator to accept the offer because I was not connected :/

holzeis: But then it's not related to the `track_channel_status` but to the `sync_position_to_subchannel_state`

bonomat: You are completely right, this might have happened in a rebase. I'll fix it.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -392,15 +395,44 @@ pub fn run(seed_dir: String, runtime: &Runtime) -> Result<()> {
 
         runtime.spawn(track_channel_status(node.clone()));
 
-        if let Err(e) = node.sync_position_with_subchannel_state().await {
-            tracing::error!("Failed to sync position with subchannel state. Error: {e:#}");
-        }
+        runtime.spawn({
```
holzeis, line 398: 🤔 Could that be potentially dangerous as it's not blocking?

What if the user is opening / closing a position in the meantime?

holzeis:
@bonomat is there anything left on this PR or can we get it merged?

bonomat:
> @bonomat is there anything left on this PR or can we get it merged?

I thought I had pushed the fix for your last comment 🙈

bonomat:
@holzeis : I'm going ahead and merge this with my last fix.

PR #1775, opened 2024-01-02
Move past multiple orders in `Filling` state

### LINKED ISSUE #1762: More than one order is being filled at the same time, this should not happen
I have multiple orders in the state `Filling` which should not happen. 
However, I cannot recover from this. 

```
More than one order is being filled at the same time, this should not happen

```
```
Caused by:
    0: Failed to load order being filled
    1: More than one order is being filled at the same time, this should not happen. [Order { id: "b7ef5d68-e1d3-434e-9a0c-432e51704319", leverage: 2.0, quantity: 10.0, contract_symbol: BtcUsd, direction: Long, order_type: Market, state: Filling, creation_timestamp: 1702649408, limit_price: None, execution_price: Some(42250.0), failure_reason: None, order_expiry_timestamp: 1702649468, reason: Manual, stable: false }, Order { id: "0dff1d34-d358-4772-961d-8dbab317dd53", leverage: 2.0, quantity: 10.0, contract_symbol: BtcUsd, direction: Long, order_type: Market, state: Filling, creation_timestamp: 1702650363, limit_price: None, execution_price: Some(42170.5), failure_reason: None, order_expiry_timestamp: 1702650423, reason: Manual, stable: false }]
flutter: [E] TIME: 2023-12-15T14:26:04.587982Z r:  error: Could not set order to failed
```

luckysori:
Fixes #1762.

Ultimately, it's still bad that we have no way of linking an order that is in `Filling` to a particular subchannel update. With such information we would be able to more confidently discard irrelevant orders and continue with the only relevant one.

I considered other approaches, but I think this is the simplest.

### DIFF
--- a/mobile/lib/features/trade/async_order_change_notifier.dart
+++ b/mobile/lib/features/trade/async_order_change_notifier.dart
@@ -42,8 +42,10 @@ class AsyncOrderChangeNotifier extends ChangeNotifier implements Subscriber {
           TaskStatus status = TaskStatus.pending;
           switch (asyncOrder?.state) {
             case OrderState.open:
+            case OrderState.filling:
               status = TaskStatus.pending;
             case OrderState.failed:
+            case OrderState.rejected:
               status = TaskStatus.failed;
             case OrderState.filled:
               status = TaskStatus.success;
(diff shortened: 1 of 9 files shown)


bonomat (approved):
I think this is a fair approach. Once we move to dlc-channels we can rethink this though.

PR #1776, opened 2024-01-02
No more `mod.rs`

### LINKED ISSUE #1329: Remove all instances of `mod.rs`
`mod.rs` files are a bit annoying because they all have the same name.

_Originally posted by @luckysori in https://github.com/get10101/10101/pull/1216#discussion_r1330074944_

luckysori:
Fix #1329.


holzeis (approved):
👍

luckysori:
Hmmm, I added it to the merge-queue but it just merged the PR even though certain checks did not pass.

In particular, `just gen` [no longer works](https://github.com/get10101/10101/actions/runs/7393083043/job/20112963603) with these changes! Probably due to some bug in our very old version of `flutter_rust_bridge`. I think we should prioritise #1777 soon.

For now I will just revert the patch that introduced the regression.

PR #1778, opened 2024-01-03
Revert "chore: Stop using mod.rs"

luckysori:
The original commit revealed a bug in our very old version of `flutter_rust_bridge`. I think we should just upgrade, but for the time being we can revert 16e916b1f64784ede52da3dac311578b7f0d722e :(


holzeis (approved):
What is the bug? 🙂

luckysori:
> What is the bug? 🙂

This one: https://github.com/get10101/10101/pull/1776#issuecomment-1874831547.

PR #1779, opened 2024-01-03
fix: Parse wallet settings from actual settings

holzeis:
This is needed to actually prevent opening new jit channels.


luckysori (approved):
LGTM.

Issue #1800, opened 2024-01-05
Show channel funding tx in tx history


bonomat:
I'm working on displaying the balance at the moment and things don't add up

For example: 

- user initial on-chain balance: 100,000,000
- order price: 40,000
- leverage: 2
-->
- margin: 12500
- fee: 75

We show this correctly in the tile for opening a position: 

<img width="318" alt="image" src="https://github.com/get10101/10101/assets/224613/15f37585-629e-4bfb-90d0-9e2386390f6f">

But the outgoing transaction shows a different value:

<img width="320" alt="image" src="https://github.com/get10101/10101/assets/224613/d41ae3cd-c737-4a4c-b4a5-09568fdff660">

- outgoing tx: 16,904
- on-chain fee: 4,698 (already included in 16,904)

So basically we fund the dlc-channel with `16,904-4,698`=`12,206`?

Related to https://github.com/get10101/10101/issues/1799

bonomat:
My conclusion is that the amounts are actually correct. We have funded the channel with 12500 and the remaining amount is for the transaction fees reserved for buffer/refund and cets. 

https://github.com/p2pderivatives/rust-dlc/blob/adc7cef5650cbba56f869a275b7b0a273c414ef1/dlc/src/lib.rs#L277-L328

I'll see to simply update the tile.

Issue #1799, opened 2024-01-04
Charge fees when closing position


bonomat:
It looks like we are not charging a fee when closing a position into the dlc-channel

old balance in dlc-channel | 11738
-- | --
new balance in dlc-channel | 23257
diff | 11519
  |  
coordinator collateral | 11266
profit | 185
fee opening | 68
fee closing | 68
  |  
missing | -68

bonomat:
Correction: this works as expected.

1. opened a position and contributed 12,575
   1. which consists out of `margin=12500` and `fee=75`
2. closed the position and ended up having `12424` which is `12500`-`75`(for closing)

https://github.com/get10101/10101/assets/224613/8db82a13-398f-4afc-afec-0e8436518f81

PR #1801, opened 2024-01-05
feat: show dlc-channel balance

bonomat:
resolves https://github.com/get10101/10101/issues/1780

Off-chain balance only shows the `available` balance, i.e. after opening a position, the off-chain balance will be `0` because all the amounts are locked in the position.

See video

https://github.com/get10101/10101/assets/224613/e2e0f8da-1009-468d-98e4-5ade94b54f4e

In this video the price has been fixed so that the user does not make any profit nor loss which showed this bug: https://github.com/get10101/10101/issues/1799 and this one https://github.com/get10101/10101/issues/1800

### DIFF
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -11,6 +11,7 @@ use anyhow::Context;
 use anyhow::Error;
 use anyhow::Result;
 use bitcoin::secp256k1::PublicKey;
+use bitcoin::Amount;
 use dlc_manager::channel::signed_channel::SignedChannel;
 use dlc_manager::channel::signed_channel::SignedChannelState;
 use dlc_manager::channel::Channel;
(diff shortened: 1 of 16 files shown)


mobile/native/src/api.rs
```diff
@@ -80,7 +80,7 @@ pub struct WalletInfo {
 #[derive(Clone, Debug, Default)]
 pub struct Balances {
     pub on_chain: u64,
-    pub lightning: u64,
+    pub off_chain: u64,
```
holzeis, line 83: nit: it's weird to me that on the coordinator you've changed the model from offchain to lightning and here you do it the other way around. It would be nice to have a common term for this kind of balance.

holzeis (approved):
LGTM 👍

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -305,6 +306,57 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
...
+
+                    SignedChannelState::SettledOffered { .. }
+                    | SignedChannelState::SettledReceived { .. }
+                    | SignedChannelState::SettledAccepted { .. }
+                    | SignedChannelState::SettledConfirmed { .. }
+                    | SignedChannelState::Established { .. } => {
+                        // if we are not yet settled or just established the channel, we have an
+                        // open position with the full amount being locked,
+                        // hence, the current balance is 0
+                        0
```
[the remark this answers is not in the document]
bonomat: > but I guess your intention was to show only a usable balance?

Yes, this was my intention: show only the usable balance.

mobile/native/src/api.rs
bonomat: I was thinking that we will release a new app to the user and force-close all channels to the users - hence, they don't need this anymore. For the coordinator in contrast we might want to see if there are still some channels open.

PR #1803, opened 2024-01-05
feat: show dlc-channel opening tx in tx history

### LINKED ISSUE #1800: Show channel funding tx in tx history

bonomat:
resolves #1800

https://github.com/get10101/10101/assets/224613/ecc94f66-1254-4bda-bba3-87b60227811e

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -113,6 +113,9 @@ const CHECK_OPEN_ORDERS_INTERVAL: Duration = Duration::from_secs(60);
 const ON_CHAIN_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 const WAIT_FOR_CONNECTING_TO_COORDINATOR: Duration = Duration::from_secs(2);
 
+/// Defines a constant from which we treat a transaction as confirmed
+const NUMBER_OF_CONFIRMATION_FOR_BEING_CONFIRMED: u64 = 1;
+
 /// The weight estimate of the funding transaction
 ///
 /// This weight estimate assumes two inputs.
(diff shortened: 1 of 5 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -113,6 +113,9 @@ const CHECK_OPEN_ORDERS_INTERVAL: Duration = Duration::from_secs(60);
 const ON_CHAIN_SYNC_INTERVAL: Duration = Duration::from_secs(300);
 const WAIT_FOR_CONNECTING_TO_COORDINATOR: Duration = Duration::from_secs(2);
 
+/// Defines a constant from which we treat a transaction as confirmed
+const NUMBER_OF_CONFIRMATION_FOR_BEING_CONFIRMED: u64 = 1;
```
luckysori, line 117: 🙃 Maybe `REQUIRED_CONFIRMATIONS` is clearer and simpler.

luckysori (approved):

luckysori:
It's a little annoying that the first trade appears before the channel opening event, but it's probably hard to fix and the impact is minor as the user keeps trading.

image

bonomat:
> It's a little annoying that the first trade appears before the channel opening event, but it's probably hard to fix and the impact is minor as the user keeps trading.
> 

Mhm.. that's hard to fix  with our current approach of sorting I believe

PR #1805, opened 2024-01-08
chore: refactor payout curve example

bonomat:
The example now produces a csv file which contains `<price>,<offerer>,<acceptor>`. This is useful to plot the payout for the coordinator(offerer) and trader (acceptor) in a single diagram.

I've also changed the gnuplot file which now produces an SVG so that one can zoom in and see more details.

Note: diagrams below where generated with a fee of 30%
payout_curve

### DIFF
--- a/crates/payout_curve/src/lib.rs
+++ b/crates/payout_curve/src/lib.rs
@@ -266,7 +266,7 @@ fn calculate_upper_range_payouts(
             },
             PayoutPoint {
                 event_outcome: BTCUSD_MAX_PRICE,
-                outcome_payout: 0,
+                outcome_payout: fee,
                 extra_precision: 0,
             },
         ),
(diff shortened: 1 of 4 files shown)


crates/payout_curve/src/lib.rs
```diff
@@ -266,7 +266,7 @@ fn calculate_upper_range_payouts(
             },
             PayoutPoint {
                 event_outcome: BTCUSD_MAX_PRICE,
-                outcome_payout: 0,
+                outcome_payout: fee,
```
luckysori, line 269: 👍

luckysori (approved):
Thank you for this!

I guess the second patch is not a refactor because you fixed some things that weren't working! `chore: Fix` would make sense to me since it's only auxiliary code.

luckysori:
@bonomat: I've pushed https://github.com/get10101/10101/pull/1805/commits/548bd88e3ef1bff22faea0d359152bdb3c5004bf to fix the tests after https://github.com/get10101/10101/pull/1805/commits/d0bdcff4e75f2e31c24f6cedfe6c1e771361ba0e.

I'll let you squash and merge if it looks good to you.

Issue #1791, opened 2024-01-04
Allow users to resize a position


bonomat:
aka implement channel resize on dlc-channels

bonomat:
Whoever picks this up, please ping Richie

PR #1806, opened 2024-01-08
feat: allow collab and force close DLC channel

### LINKED ISSUE #1785: Allow collab close dlc-channel

### LINKED ISSUE #1786: Allow force-close dlc-channel

bonomat:
resolves #1785 
resolves #1786

### DIFF
--- a/crates/fund/examples/fund.rs
+++ b/crates/fund/examples/fund.rs
@@ -53,7 +53,7 @@ async fn fund_everything(faucet: &str, coordinator: &str) -> Result<()> {
     let coordinator_balance = coordinator.get_balance().await?;
     tracing::info!(
         onchain = %Amount::from_sat(coordinator_balance.onchain),
-        offchain = %Amount::from_sat(coordinator_balance.offchain),
+        offchain = %Amount::from_sat(coordinator_balance.dlc_channel),
         "Coordinator balance",
     );
 
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -158,6 +158,73 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
         Ok(())
     }
 
+    pub async fn close_dlc_channel(&self, channel_id: DlcChannelId, force: bool) -> Result<()> {
+        let channel_id_hex = hex::encode(channel_id);
+        tracing::info!(channel_id = channel_id_hex, "Closing DLC channel");
+
+        let channel = self
+            .get_dlc_channel(|channel| channel.channel_id == channel_id)?
+            .context("DLC channel to close not found")?;
+
+        if force {
+            // TODO(bonomat): implement force closure
+        } else {
+            self.propose_dlc_channel_collaborative_close(channel)
+                .await?
+        }
+
+        Ok(())
+    }
+
+    /// Collaboratively close a DLC channel on-chain if there is no open position
+    async fn propose_dlc_channel_collaborative_close(&self, channel: SignedChannel) -> Result<()> {
+        let channel_id_hex = hex::encode(channel.channel_id);
+
+        tracing::info!(
+            channel_id = %channel_id_hex,
+            "Closing DLC channel collaboratively"
+        );
+
+        let counterparty = channel.counter_party;
+
+        match channel.state {
+            SignedChannelState::Settled { .. } | SignedChannelState::RenewFinalized { .. } => {
+                spawn_blocking({
+                    let dlc_manager = self.dlc_manager.clone();
+                    let dlc_message_handler = self.dlc_message_handler.clone();
+                    let peer_manager = self.peer_manager.clone();
+                    move || {
+                        let settle_offer = dlc_manager
+                            .offer_collaborative_close(
+                                &channel.channel_id,
+                                channel.counter_params.collateral,
+                            )
+                            .context(
+                                "Could not propose to collaboratively close the dlc channel.",
+                            )?;
+
+                        send_dlc_message(
+                            &dlc_message_handler,
+                            &peer_manager,
+                            counterparty,
+                            Message::Channel(ChannelMessage::CollaborativeCloseOffer(settle_offer)),
+                        );
+
+                        anyhow::Ok(())
+                    }
+                })
+                .await??;
+            }
+            _ => {
+                tracing::error!( state = %channel.state, "Can't collaboratively close a channel with an open position.");
+                bail!("Can't collaboratively close a channel with an open position");
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Collaboratively close a position within a DLC Channel
     pub async fn propose_dlc_channel_collaborative_settlement(
         &self,
         channel_id: DlcChannelId,
@@ -168,7 +235,7 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
         tracing::info!(
             channel_id = %channel_id_hex,
             %accept_settlement_amount,
-            "Settling DLC channel collaboratively"
+            "Settling DLC in channel collaboratively"
         );
 
         spawn_blocking({
@@ -190,6 +257,17 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
             }
         })
         .await?
+    }
+
+    pub fn accept_dlc_channel_collaborative_close(&self, channel_id: DlcChannelId) -> Result<()> {
+        let channel_id_hex = hex::encode(channel_id);
+
+        tracing::info!(channel_id = %channel_id_hex, "Accepting DLC channel collaborative close offer");
+
+        let dlc_manager = self.dlc_manager.clone();
+        dlc_manager.accept_collaborative_close(&channel_id)?;
+
+        Ok(())
     }
 
     pub fn accept_dlc_channel_collaborative_settlement(
--- a/crates/ln-dlc-node/src/node/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/node/dlc_channel.rs
@@ -167,12 +167,24 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
             .context("DLC channel to close not found")?;
 
         if force {
-            // TODO(bonomat): implement force closure
+            self.force_close_dlc_channel(&channel_id)?;
         } else {
             self.propose_dlc_channel_collaborative_close(channel)
                 .await?
         }
 
+        Ok(())
+    }
+
+    fn force_close_dlc_channel(&self, channel_id: &DlcChannelId) -> Result<()> {
+        let channel_id_hex = hex::encode(channel_id);
+
+        tracing::info!(
+            channel_id = %channel_id_hex,
+            "Force closing DLC channel"
+        );
+
+        self.dlc_manager.force_close_channel(channel_id)?;
         Ok(())
     }
 
--- a/crates/ln-dlc-node/src/tests/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/tests/dlc_channel.rs
@@ -245,4 +245,78 @@ async fn vanilla_dlc_channel() {
     })
     .await
     .unwrap();
+
+    let app_on_chain_balance_before_close = app.get_on_chain_balance().unwrap();
+    let coordinator_on_chain_balance_before_close = coordinator.get_on_chain_balance().unwrap();
+
+    tracing::debug!("Proposing to close dlc channel collaboratively");
+
+    coordinator
+        .close_dlc_channel(app_signed_channel.channel_id, false)
+        .await
+        .unwrap();
+
+    wait_until(Duration::from_secs(10), || async {
+        app.process_incoming_messages()?;
+
+        let dlc_channels = app
+            .dlc_manager
+            .get_store()
+            .get_signed_channels(Some(SignedChannelStateType::CollaborativeCloseOffered))?;
+
+        Ok(dlc_channels
+            .iter()
+            .find(|dlc_channel| dlc_channel.counter_party == coordinator.info.pubkey)
+            .cloned())
+    })
+    .await
+    .unwrap();
+
+    tracing::debug!("Accepting collaborative close offer");
+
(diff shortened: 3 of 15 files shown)


crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -158,6 +158,73 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
...
+            // TODO(bonomat): implement force closure
+        } else {
+            self.propose_dlc_channel_collaborative_close(channel)
+                .await?
+        }
+
+        Ok(())
+    }
+
+    /// Collaboratively close a DLC channel on-chain collaboratively if there is no open position
```
luckysori, line 191: 🔧 Too many "collaboratively".

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -158,6 +158,73 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
...
+                        );
+
+                        anyhow::Ok(())
+                    }
+                })
+                .await??;
+            }
+            _ => {
+                tracing::error!(  state = %channel.state, "Can't collaboratively close a channel with an open position.");
+                bail!("Can't collaboratively close a channel with an open position");
```
luckysori, line 232: 🔧 Why not?

Also, it's weird to log and return an error for the same error.

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -168,7 +235,7 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
         tracing::info!(
             channel_id = %channel_id_hex,
             %accept_settlement_amount,
-            "Settling DLC channel collaboratively"
+            "Settling position in dlc channel collaboratively"
```
luckysori, line 250: 🙃 "Position" is weird at this level. I propose DLC instead of position.

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -371,6 +449,36 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
                 .sum(),
         ))
     }
+
+    /// Check if it is safe to close the DLC channel collaboratively.
+    ///
+    /// In general, it is NOT safe to close an DLC channel if there still is a position attached
+    /// to it. This is because this can lead to loss of funds.
```
luckysori, line 468: 🤔 I thought this only applied to Lightning channels.

With a split Lightning channel, if we triggered a collaborative close via LDK it would fail and the channel would be force-closed (IIRC). But here we are only operating with `rust-dlc`.

I'm surprised collab-close doesn't work in more states (particularly `Established`), since the API asks for the `accept_settlement_amount` anyway.

crates/ln-dlc-node/src/tests/dlc_channel.rs
```diff
@@ -245,4 +245,78 @@ async fn vanilla_dlc_channel() {
...
+            .find(|dlc_channel| dlc_channel.counter_party == coordinator.info.pubkey)
+            .cloned())
+    })
+    .await
+    .unwrap();
+
+    tracing::debug!("Accepting collaborative close offer");
+
+    let app_on_chain_balance_before_close = app.get_on_chain_balance().unwrap();
+    let coordinator_on_chain_balance_before_close = coordinator.get_on_chain_balance().unwrap();
```
luckysori, line 275: 🙃 I would expect this to appear before we send the proposal.

crates/ln-dlc-node/src/tests/dlc_channel.rs
```diff
@@ -245,4 +245,78 @@ async fn vanilla_dlc_channel() {
...
+            > app_on_chain_balance_before_close.confirmed;
+        if app_balance_changed {
+            tracing::debug!(
+                old_balance = app_on_chain_balance_before_close.confirmed,
+                new_balance = app_on_chain_balances_after_close.confirmed,
+                "Balance updated"
+            )
+        }
+
+        Ok(app_balance_changed.then_some(true))
```
luckysori, line 318: 🔧 You can just use `then_some(())` if you don't care about the return type (the `bool` is not used AFAIK).

mobile/lib/common/channel_status_notifier.dart
```diff
@@ -17,23 +17,15 @@ class ChannelStatusNotifier extends ChangeNotifier implements Subscriber {
...
-  bool hasLnChannel() {
-    return getChannelStatus() == ChannelStatus.LnOpen;
-  }
-
   bool hasDlcChannel() {
-    return getChannelStatus() == ChannelStatus.LnDlcOpen;
+    return getChannelStatus() == ChannelStatus.Open;
   }
 
   /// Whether the current LN-DLC channel is closed or not.
```
luckysori, line 29: 🔧 Let's change the docs then.

mobile/lib/common/settings/collab_close_screen.dart
```diff
@@ -125,7 +125,7 @@ class _CollabCloseScreenState extends State<CollabCloseScreen> {
 }
 
 RichText getCloseChannelText(ChannelStatusNotifier channelStatusNotifier) {
-  if (channelStatusNotifier.hasDlcChannel()) {
+  if (channelStatusNotifier.hasOpenPosition()) {
     return RichText(
         text: const TextSpan(
       style: TextStyle(fontSize: 18, color: Colors.black, letterSpacing: 0.4),
```
luckysori, line 131: The text below is weird (you didn't change it here but still). It should be:

```
Please, close your open position before collaboratively closing your channel
```

The comma is in the wrong place. Or just remove the comma altogether.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -19,19 +18,14 @@ pub enum ChannelStatus {
     /// there was a previous channel nor does it imply that a previous channel was completely
     /// closed i.e. there might be pending transactions.
```
luckysori, line 19: 🔧 Let's rewrite the docs for this variant too.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -104,100 +93,63 @@ async fn channel_status(node: impl Borrow<Node>) -> Result<ConcreteChannelStatus
 #[derive(Debug, Clone, Copy)]
```
luckysori, line 93: The docs should now read: "The complete status of the app's DLC channel". It talked about channel**S** before because we had LN and DLC channels.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -104,100 +93,63 @@ async fn channel_status(node: impl Borrow<Node>) -> Result<ConcreteChannelStatus
...
+        | Some(SignedChannel {
+            state: SignedChannelState::CollaborativeCloseOffered { .. },
+            ..
+        }) => ConcreteChannelStatus::Closing,
+        Some(SignedChannel {
+            state: SignedChannelState::Established { .. },
+            ..
+        })
+        | Some(SignedChannel {
+            state: SignedChannelState::RenewFinalized { .. },
```
luckysori, line 145: 🤔 Is this really how it works? Doesn't the channel go back to `Established` after this one? I'd consider this state part of `Pending`.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -104,100 +93,63 @@ async fn channel_status(node: impl Borrow<Node>) -> Result<ConcreteChannelStatus
...
-    ln_channel: Channel,
-    // All the subchannels that we have ever recorded.
-    subchannels: Vec<SubChannel>,
-) -> ConcreteChannelStatus {
-    match ln_channel {
-        // If the LN channel is open (or practically open).
-        Channel {
-            channel_id: Some(channel_id),
-            channel_state: ChannelState::Pending | ChannelState::Open | ChannelState::OpenUnpaid,
+fn derive_dlc_channel_status(dlc_channel: Option<&SignedChannel>) -> ConcreteChannelStatus {
```
luckysori, line 102: Given that we take an `Option<&SignedChannel>` I think that means that we consider a new channel that is not yet `Established` `NotOpen`. That seems fine, even if it would be nice to track the state of the opening protocol too.

But that would require looking at `Channel`s beyond `SignedChannel`, which is a bit trickier.

mobile/native/src/ln_dlc/channel_status.rs
```diff
@@ -19,19 +18,14 @@ pub enum ChannelStatus {
...
-    /// The LN-DLC channel is in the process of being force-closed.
-    LnDlcForceClosing,
-    /// The LN-DLC channel is in an inconsistent state.
-    Inconsistent,
+    /// DLC Channel is open
+    Open,
+    /// DLC Channel is open and has an open position
+    WithPosition,
+    /// DLC Channel is currently in progress of being renewed
+    Pending,
```
luckysori, line 26: According to the logic below, I think the channel can be `Pending` if it's being renewed or if it is being settled.

"Pending" makes me think of a channel that is going from `NotOpen` to `Open`, so maybe we should rename this to `BeingUpdated` or something like that.

luckysori: Or we could even introduce `Renewing` and `Settling`.

I think we may even want to use the same model that `rust-dlc` uses now that we are not dealing with both LN channels and DLC channels.

crates/fund/examples/fund.rs
```diff
@@ -53,7 +53,7 @@ async fn fund_everything(faucet: &str, coordinator: &str) -> Result<()> {
     let coordinator_balance = coordinator.get_balance().await?;
     tracing::info!(
         onchain = %Amount::from_sat(coordinator_balance.onchain),
-        offchain = %Amount::from_sat(coordinator_balance.offchain),
+        offchain = %Amount::from_sat(coordinator_balance.dlc_channel),
```
luckysori, line 56: 👍 Thanks for fixing this! I ran into it locally.

crates/ln-dlc-node/src/node/dlc_channel.rs
```diff
@@ -167,7 +167,7 @@ impl<S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send + 'static> Nod
             .context("DLC channel to close not found")?;
 
         if force {
-            // TODO(bonomat): implement force closure
+            self.force_close_dlc_channel(&channel_id)?;
```
luckysori, line 170: ❓ Should we add a test?

It might time to split up the new `ln-dlc-node` test into multiple ones.

luckysori (commented):
Looks mostly good to me!

crates/ln-dlc-node/src/node/dlc_channel.rs
bonomat: > Also, it's weird to log and return an error for the same error.

I started doing this because I stopped trusting where errors are being logged (or even ignored) and with this it's easier to find them.

> 🔧 Why not?

Mhm, good question. We had this in the past, so I kept it. Happy to change it.

crates/ln-dlc-node/src/node/dlc_channel.rs
bonomat: I honestly haven't tried... 🙈

mobile/native/src/ln_dlc/channel_status.rs
bonomat: Afaik we end in `RenewFinalized` or `Settled` after any off-chain action and `Established` is only once the channel has been set up in the first place

mobile/native/src/ln_dlc/channel_status.rs
bonomat: > Or we could even introduce Renewing and Settling.

Yes, I thought we'd introduce these temporary state with https://github.com/get10101/10101/issues/1791

> I think we may even want to use the same model that rust-dlc uses now that we are not dealing with both LN channels and DLC channels.

That makes sense 🙈

crates/ln-dlc-node/src/node/dlc_channel.rs
bonomat: Thinking about it more. It's easier to tell the user to close the position first because otherwise we will need to calculate the PnL here and think about how to handle the otherside of the trade.

crates/ln-dlc-node/src/node/dlc_channel.rs
bonomat: See here: https://github.com/get10101/10101/pull/1806#discussion_r1445795987: if there is still an open DLC we would need to close the counter position as well or hedge otherwise. I've updated the doc and will leave it as is.

holzeis (approved):
LGTM 🚀

mobile/native/src/ln_dlc/channel_status.rs
bonomat: I've simplified this part and replicate the libraries states in our code.

crates/ln-dlc-node/src/node/dlc_channel.rs
bonomat: Done https://github.com/get10101/10101/pull/1806/commits/61e5ed46974a02facac7d0760a3364779741de8b

crates/ln-dlc-node/src/node/dlc_channel.rs
luckysori: > I started doing this because I stopped trusting where errors are being logged (or even ignored) and with this it's easier to find them.

Hmmm, I'm not happy with this because we will end up spamming the logs with redundant messages, which makes debugging harder. We just need to fix the spots where errors are being swallowed. I don't think there can be many of those.

luckysori: > Thinking about it more. It's easier to tell the user to close the position first because otherwise we will need to calculate the PnL here and think about how to handle the otherside of the trade.

Fair. It's not strictly necessary and it's nice to keep things simple.

mobile/native/src/ln_dlc/channel_status.rs
luckysori: Hmm, we should verify this.

luckysori (commented):
Thanks for addressing my review!

PR #1812, opened 2024-01-10
chore: input validation on trade screen

### LINKED ISSUE #1782: Reintroduce validations in the frontend when opening a position
- if you don't have a dlc-channel, check your on-chain balance for how much you can trade
- if you do have a dlc-channel, look at the channel balance.

bonomat:
I believe this is much simpler code wise and also to understand from a user perspective. 
Note: small UI update: the info icon is now part of the margin field and hence, the input fields don't jump as they used to.

Video shows two different validations 
1. if the user has enough margin available --> leads to `insuffucient balance` error
2. if the counterparty (the coordinator) has enough margin --> leads to `max amount` error

https://github.com/get10101/10101/assets/224613/61f76e30-46a7-4857-96d9-8af6d3ad1819

resolves #1782


mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -264,6 +190,25 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
...
+        channelTradeConstraints.coordinatorLeverage;
+
+    String text =
+        "The usable balance of ${formatSats(Amount(usableBalance))} are your on-chain funds. If you need more, you can always deposit more into you wallet. "
+        "\nHowever, the maximum you can trade is ${formatUsd(Usd(maxQuantity.toInt()))}";
+    if (isChannelBalance) {
+      text =
+          "The usable balance of ${formatSats(Amount(usableBalance))} are your off-chain funds. At the moment you can't add use more than this as we do not support splicing. \n"
+          "If you want to trade more than this, you will need to close the channel and open a bigger one. "
+          "\nHowever, the maximum you can trade is ${formatUsd(Usd(maxQuantity.toInt()))}";
```
luckysori, line 209: 🔧 "However" here and above reads a bit weird. Maybe just:

```
With your current balance, the maximum you can trade is X.
```

mobile/native/src/api.rs
```diff
@@ -416,6 +416,66 @@ pub fn channel_info() -> Result<Option<ChannelInfo>> {
...
+    /// has in the channel. In the future we can consider splice in and allow the user to use more
+    /// than just his channel balance.
+    pub is_channel_balance: bool,
+}
+
+pub fn channel_trade_constraints() -> Result<TradeConstraints> {
+    // TODO(bonomat): retrieve these values from the coordinator. This can come from the liquidity
+    // options.
+    let coordinator_leverage = 2.0;
+    let min_quantity = 1;
```
luckysori, line 447: 🤔 I guess this needs to be higher. Maybe 100?

mobile/native/src/api.rs
```diff
@@ -416,6 +416,66 @@ pub fn channel_info() -> Result<Option<ChannelInfo>> {
...
+    pub coordinator_leverage: f64,
+    /// Smallest allowed amount of contracts
+    pub min_quantity: u64,
+    /// If true it means that the user has a channel and hence the max amount is limited by what he
+    /// has in the channel. In the future we can consider splice in and allow the user to use more
+    /// than just his channel balance.
+    pub is_channel_balance: bool,
+}
+
+pub fn channel_trade_constraints() -> Result<TradeConstraints> {
```
luckysori, line 441: 🔧 I think the `api` module is supposed to be a dumb layer to interface between Flutter and Rust. I would put the logic below in a different module.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -722,6 +724,16 @@ pub async fn close_channel(is_force_close: bool) -> Result<()> {
         .await
 }
 
+pub fn get_dlc_channel() -> Result<Vec<SignedChannel>> {
+    let node = state::try_get_node().context("failed to get ln dlc node")?;
+    node.inner.list_dlc_channels()
+}
```
luckysori, line 730: 🔧 Name should be plural.

luckysori (approved):
I like where this is going! And the tooltip seems very helpful.

Issue #1621, opened 2023-11-21
Improve error message on why an order failed


holzeis:
At the moment we only say it failed, try again later.

PR #1814, opened 2024-01-11
feat: show error message details in dialog

bonomat:
We don't have good error messages yet but with this we have a mode to display them to the user. 
I think this is a good start and we can build up on this as we go and as we identify more errors.

Simulator Screenshot - iPhone 15 - 2024-01-11 at 20 36 10

Simulator Screenshot - iPhone 15 - 2024-01-11 at 20 38 42

Resolves https://github.com/get10101/10101/issues/1621


mobile/lib/features/trade/trade_dialog.dart
```diff
@@ -125,3 +164,47 @@ Future<void> shareTweet(PositionAction action) async {
...
+          TextSpan(
+            text: 'https://t.me/get10101',
+            style: const TextStyle(
+              color: Colors.blue,
+              decoration: TextDecoration.underline,
+            ),
+            recognizer: TapGestureRecognizer()
+              ..onTap = () async {
+                final httpsUri = Uri(scheme: 'https', host: 't.me', path: 'get10101');
+                if (await canLaunchUrl(httpsUri)) {
```
holzeis, line 187: ❓ i remember that this url launcher can go external or within the app. I think the later is the default. But can you clarify which one it is here?

FYI: I'd prefer to open an external app than opening a browser in app.

mobile/lib/features/trade/trade_dialog.dart
```diff
@@ -63,55 +68,89 @@ Widget createSubmitWidget(
...
-        padding: EdgeInsets.only(left: 10, right: 10, bottom: 5),
-        child: Text("Do not close the app!",
-            style: TextStyle(fontSize: 15, fontWeight: FontWeight.bold)),
+              Padding(
+                padding: const EdgeInsets.fromLTRB(0, 15.0, 0, 15.0),
+                child: Container(
+
+                  color: Colors.grey.shade300,
+                  child: Text(
+                    getPrettyJSONString(pendingOrder.failureReason?.details ?? ""),
```
holzeis, line 92: nit: it would be nice if the error text could be copied. Resp. If we'd add a share logs button.

mobile/native/src/db/custom_types.rs
```diff
@@ -146,7 +146,7 @@ impl ToSql<Text, Sqlite> for FailureReason {
             FailureReason::TradeResponse => "TradeResponse",
             FailureReason::NodeAccess => "NodeAccess",
             FailureReason::NoUsableChannel => "NoUsableChannel",
-            FailureReason::ProposeDlcChannel => "ProposeDlcChannel",
+            FailureReason::CollabRevert => "ProposeDlcChannel",
```
luckysori: 📣 You have to update the strings too.

mobile/native/src/trade/order/mod.rs
```diff
@@ -22,14 +22,20 @@ pub enum OrderType {
...
+    /// The order failed because we failed sending the trade request
     TradeRequest,
+    /// A failure happened during the initial phase of the protocol. I.e. after sending the trade
+    /// request
     TradeResponse,
     NodeAccess,
     NoUsableChannel,
-    ProposeDlcChannel,
+    /// The order failed due to collaboratively reverting the position
+    CollabRevert,
```
luckysori, line 33: 🤔 That's probably not the reason why the order failed though. We can say that the _channel_ was collaboratively reverted before the order was `Filled`, but it's very likely that the order didn't get filled for a different reason.

I would adapt the docs, the name is okay though.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -910,6 +910,8 @@ fn update_state_after_collab_revert(
                 order_expiry_timestamp: OffsetDateTime::now_utc(),
                 reason: OrderReason::Expired,
                 stable: position.stable,
+                // TODO(bonomat): why do we create an order here if we filling failed?
+                failure_reason: None,
```
luckysori: I guess so that we can continue with the rest of this function since we need an `Order`. Did you write this, @holzeis?

mobile/native/src/db/custom_types.rs
```diff
@@ -161,20 +150,27 @@ impl ToSql<Text, Sqlite> for FailureReason {
...
-            "SubchannelOfferOutdated" => Ok(FailureReason::SubchannelOfferOutdated),
-            "SubchannelOfferDateUndetermined" => Ok(FailureReason::SubchannelOfferDateUndetermined),
-            "SubchannelOfferUnacceptable" => Ok(FailureReason::SubchannelOfferUnacceptable),
-            "OrderRejected" => Ok(FailureReason::OrderRejected),
-            _ => Err("Unrecognized enum variant".into()),
-        };
+        match serde_json::from_str(string.as_str()) {
+            Ok(reason) => Ok(reason),
+            Err(_) => {
+                // backwards compatibility
```
luckysori, line 156: 🔧 Do we need backwards compatibility here? Maybe just say "unknown" for all of them.

mobile/lib/features/trade/trade_dialog.dart
```diff
@@ -125,3 +164,47 @@ Future<void> shareTweet(PositionAction action) async {
...
 }
+
+class ClickableHelpText extends StatelessWidget {
+  const ClickableHelpText({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return RichText(
+      text: TextSpan(
+        text: 'Please help us fix this issue and join our telegram group: ',
```
luckysori, line 175:
```suggestion
        text: 'If the issue persists, please report it in our Telegram group, on GitHub or send us a Twitter DM.',
```

Maybe we could use hyperlinks too.

mobile/native/src/trade/order/mod.rs
bonomat: This was my understanding though, i.e. we fail the order because we collaboratively reverted before it was filled

mobile/lib/features/trade/trade_dialog.dart
bonomat: It would have used the platform default.

```
///   - [LaunchMode.platformDefault] is supported on all platforms:
///     - On iOS and Android, this treats web URLs as
///       [LaunchMode.inAppWebView], and all other URLs as
///       [LaunchMode.externalApplication].
///     - On Windows, macOS, and Linux this behaves like
///       [LaunchMode.externalApplication].
///     - On web, this uses `webOnlyWindowName` for web URLs, and behaves like
///       [LaunchMode.externalApplication] for any other content.
```

I'll fix this.

mobile/native/src/trade/order/mod.rs
luckysori: Fair enough, I think you are right.

mobile/lib/features/trade/trade_dialog.dart
bonomat: I rather pipe them all into one channel.

bonomat:
Please don't forget to review/approve @luckysori / @holzeis :) Thank you

holzeis (approved):
LGTM 👍

PR #1821, opened 2024-01-16
fix: Accept collaborative close offer from coordinator

holzeis:
I noticed that both parties can trigger a collaborative dlc channel close, but only the coordinator had the logic implemented to also accept it. This will allow the app also to accept a collab close offer from the coordinator.


bonomat (commented):
Afaik @luckysori also added this in one of his branches. 
Let's wait for his approval before merging please.

luckysori:
> Afaik @luckysori also added this in one of his branches. Let's wait for his approval before merging please.

I don't think I added this logic :D

luckysori (approved):

PR #1822, opened 2024-01-16
feat: Recover from pending offer

holzeis:
Checks on `on_connect` if the current dlc channel status is in offered.

- `OfferedChannel`: Automatically tries to accept the offer, if that fails the offer is attempted to be rejected. I think we do not need to check the maturity date outside of accepting the offer as the offer will fail anyways if the maturity date is in the past. If so we are continuing with rejecting the dlc channel offer.
- `SignedChannel(state: SettledReceived)`: Automatically accept the offer to close the position (settle the dlc channel)
- `SignedChannel(state: CollaborativeCloseOffered)`: Automatically accept the offer to collab close the dlc channel.

### Future work
- `SignedChannel(state: RenewOffered)`: To be addressed with #1791 .
- Add plausibility checks before automatically accepting any offer.


bonomat (approved):
Very nice.

Issue #1798, opened 2024-01-04
Re-add end2end tests


bonomat:
https://github.com/get10101/10101/pull/1788 changes our trading logic to use dlc-channels instead of ln/sub-channels. This obviously breaks all e2e tests. We should re-add them and make them work with the new protocol.

PR #1823, opened 2024-01-16
feat: collaborative revert dlc channels

### LINKED ISSUE #1809: Implement collab-revert for dlc-channels
This will come useful when needing to bump the fees

bonomat:
resolves #1809 

Tested manually. 

https://drive.google.com/file/d/1798JHsB6gIOfUsmmcBNE8cpCVdoBIPrP/view?usp=sharing

@luckysori : how far are you with https://github.com/get10101/10101/issues/1798? I thought I'd wait for your changes to add an e2e test for this feature.

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -12,7 +12,6 @@ use bdk::FeeRate;
 use bdk::LocalUtxo;
 use bdk::TransactionDetails;
 use bitcoin::secp256k1::PublicKey;
-use commons::CollaborativeRevertCoordinatorExpertRequest;
 use commons::CollaborativeRevertCoordinatorRequest;
 use dlc_manager::channel::Channel;
 use dlc_manager::contract::Contract;
@@ -339,14 +338,6 @@ pub async fn collaborative_revert(
     tracing::info!(channel_id = channel_id_hex, "Proposed collaborative revert");
 
     Ok(())
-}
-
-#[instrument(skip_all, err(Debug))]
-pub async fn expert_collaborative_revert(
-    State(_state): State<Arc<AppState>>,
-    _revert_params: Json<CollaborativeRevertCoordinatorExpertRequest>,
-) -> Result<(), AppError> {
-    unimplemented!("This needs to be reimplemented on top of dlc-channels");
 }
 
 #[instrument(skip_all, err(Debug))]
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -1,6 +1,5 @@
 use crate::collaborative_revert;
 use crate::db;
-use crate::parse_channel_id;
 use crate::parse_dlc_channel_id;
 use crate::routes::AppState;
 use crate::AppError;
@@ -13,7 +12,6 @@ use bdk::FeeRate;
 use bdk::LocalUtxo;
 use bdk::TransactionDetails;
 use bitcoin::secp256k1::PublicKey;
-use bitcoin::OutPoint;
 use commons::CollaborativeRevertCoordinatorExpertRequest;
 use commons::CollaborativeRevertCoordinatorRequest;
 use dlc_manager::channel::Channel;
@@ -321,22 +319,17 @@ pub async fn collaborative_revert(
     revert_params: Json<CollaborativeRevertCoordinatorRequest>,
 ) -> Result<(), AppError> {
     let channel_id_hex = revert_params.channel_id.clone();
-    let channel_id = parse_channel_id(channel_id_hex.as_str())
+    let channel_id = parse_dlc_channel_id(channel_id_hex.as_str())
         .map_err(|e| AppError::BadRequest(format!("Invalid channel ID provided: {e:#}")))?;
-
-    let funding_txo = OutPoint {
-        txid: revert_params.txid,
-        vout: revert_params.vout,
-    };
 
     collaborative_revert::propose_collaborative_revert(
         state.node.inner.clone(),
         state.pool.clone(),
         state.auth_users_notifier.clone(),
         channel_id,
-        revert_params.price,
         revert_params.fee_rate_sats_vb,
-        funding_txo,
+        revert_params.counter_payout,
+        revert_params.price,
     )
     .await
     .map_err(|e| {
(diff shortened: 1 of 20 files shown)


coordinator/src/admin.rs
```diff
@@ -339,36 +331,10 @@ pub async fn collaborative_revert(
...
-    )
-    .await
-    .map_err(|e| {
-        AppError::InternalServerError(format!("Could not collaboratively revert channel: {e:#}"))
-    })?;
-
-    tracing::info!(channel_id = channel_id_hex, "Proposed collaborative revert");
-
-    Ok(())
+    unimplemented!("This needs to be reimplemented on top of dlc-channels");
```
holzeis, line 337: ❓ Do we still need an "expert" collab revert option? Now that we have moved away from ldk, we do not run into the issue that the lightning channel might have been gone, which required us to implement an alternative collab revert in the first place.

coordinator/src/collaborative_revert.rs
```diff
@@ -347,159 +164,50 @@ pub fn confirm_collaborative_revert(
         "Proposed collaborative revert transaction doesn't pay the coordinator"
     );
 
-    let funding_txo = &funding_tx.output[record.vout as usize];
-
     let subchannels = node
```
holzeis, line 167:
```suggestion
    let signed_dlc_channels = node
```

coordinator/src/collaborative_revert.rs
```diff
@@ -347,159 +164,50 @@ pub fn confirm_collaborative_revert(
...
-                &Secp256k1::new(),
-                &revert_transaction,
-                &funding_redeemscript,
-                fund_value_satoshis,
-            )
-            .context("Could not get own signature for collaborative revert transaction")?
-    };
+        .list_signed_dlc_channels()
+        .context("Failed to list dlc channels")?;
+    let signed_channel = subchannels
```
holzeis, line 170:
```suggestion
    let signed_channel = signed_dlc_channels
```

coordinator/src/collaborative_revert.rs
```diff
@@ -347,159 +164,50 @@ pub fn confirm_collaborative_revert(
...
-            .get_holder_split_tx_signature(
-                &Secp256k1::new(),
-                &revert_transaction,
-                &funding_redeemscript,
-                fund_value_satoshis,
-            )
-            .context("Could not get own signature for collaborative revert transaction")?
-    };
+        .list_signed_dlc_channels()
+        .context("Failed to list dlc channels")?;
```
holzeis, line 169:
```suggestion
        .context("Failed to list signed dlc channels")?;
```

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -930,7 +884,8 @@ fn update_state_after_collab_revert(
                 direction: position.direction.opposite(),
                 order_type: OrderType::Market,
                 state: OrderState::Filled {
-                    execution_price: execution_price.to_f32().expect("to fit into f32"),
+                    // the price doesn't matter here. It's just to clean up some state
+                    execution_price: 21.0,
```
holzeis, line 888: Even if its just a clean up, it would have been nice to keep the execution price for reference. This has typically been discussed with the user, if we had to run a collab revert. I don't see the improvement in now showing `21` as execution price for these collab reverts.

But yet again, why do we even create an order for that? 😅 I think it shouldn't be needed anymore now that we've moved away from our ominous reliance on a specific order of our orders to make our UI work 🙈

coordinator/src/collaborative_revert.rs
```diff
@@ -347,159 +164,50 @@ pub fn confirm_collaborative_revert(
...
         .context("Could not set position to closed")?;
 
-    if let Some(mut subchannel) = optional_subchannel.cloned() {
-        subchannel.state = SubChannelState::OnChainClosed;
-        node.dlc_manager
-            .get_store()
-            .upsert_sub_channel(&subchannel)?;
-    }
-
     db::collaborative_reverts::delete(conn, channel_id)?;
```
holzeis, line 211: 🔧 It would be nice to also cleanup the dlc channel state on the coordinator.

```suggestion
node.dlc_manager
        .get_store()
        .upsert_channel(
            dlc_manager::channel::Channel::CollaborativelyClosed(ClosedChannel {
                counter_party: signed_channel.counter_party,
                temporary_channel_id: signed_channel.temporary_channel_id,
                channel_id: signed_channel.channel_id,
            }),
            // The contract doesn't matter anymore
            None,
        )
        .map_err(|e| anyhow!("{e:#}"))
```

mobile/pubspec.lock
holzeis, line 1: 🤔 Any particular reason why you downgrade the dependencies here?

holzeis (commented):
Very nice! I like how everything get so much simpler :)

coordinator/src/admin.rs
bonomat: I can't say I fully understand all states of DLC channel yet, but I believe my solution only works if the DlcChannel is in `Signed`. Maybe all DlcChannels are always `Signed`?

mobile/native/src/ln_dlc/mod.rs
bonomat: Oh, I didn't know we show this information 🙈

Well, for that to work, we need to define the execution price in the collab_revert endpoint and send it to the user.

coordinator/src/collaborative_revert.rs
bonomat: I thought I did this? ... thanks for pointing it out.

mobile/pubspec.lock
bonomat: Nope, just me and my old flutter version ...

coordinator/src/admin.rs
holzeis: No, not all dlc channels are `Signed`. But channels which are not signed should either mean the channel is not on-chain yet or (about to go) on-chain already.

But fair enough, lets look at it some other time then.

bonomat: I've removed this function for now

bonomat:
@holzeis : please rereview

holzeis (approved):
LGTM 👍

PR #1824, opened 2024-01-16
Update just fund: add more utxos to coordinator wallet

bonomat:


crates/fund/examples/fund.rs
```diff
@@ -42,10 +42,13 @@ async fn fund_everything(faucet: &str, coordinator: &str) -> Result<()> {
 
     let bitcoind = bitcoind::Bitcoind::new(client, faucet.to_string() + "/bitcoin");
 
-    bitcoind
-        .fund(&coord_addr, Amount::ONE_BTC)
-        .await
-        .context("Could not fund the faucet's on-chain wallet")?;
+    for _ in 0..5 {
```
holzeis, line 45: 🤔 I think the `generatetoaddress` api that is called for funding accepts a count argument.

holzeis (approved):
Thanks

crates/fund/examples/fund.rs
bonomat: I tried it but it didn't have the expected result and then I gave up because this has the same effect

PR #1825, opened 2024-01-16
chore: Publish recover dlc background task

holzeis:
I've added a panic before accepting the `SettleOffer` to demonstrate how the recovery would look like.

https://github.com/get10101/10101/assets/382048/5a872140-f50a-4187-95b6-9c7583d32986

---
Note, there is still a bug left in which sometimes the dialog would not be shown, due to the shellNavigationKey.currentContext being empty. That was definitely already an issue before. Unfortunately, I couldn't find a fix for that in this PR.


bonomat (approved):
LGTM

PR #1826, opened 2024-01-17
fix: Set default quantity to 100

holzeis:


bonomat (approved):
Thanks, good enough for now.  Eventually we should set it to `min_quantity` as defined in `TradeConstraints` .

https://github.com/get10101/10101/blob/75738a6de5d6acb30c08664afdd3db4d9e3cf9c1/mobile/native/src/api.rs#L436

PR #1836, opened 2024-01-18
Set required confirmations on DLC channel to 1

### LINKED ISSUE #1820: Set required channel confirmations to 1
Remember to test that the error is displayed properly in the app i.e. based on https://github.com/get10101/10101/pull/1814.

luckysori:
Fixes #1820.

- [x] Confirm that the associated error is displayed nicely on the app.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1147,7 +1147,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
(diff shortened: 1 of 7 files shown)


holzeis (approved):
LGTM, although it would be nice to give the user some hint about the pending status of the first trade. E.g. we could show the trade record as pending.. although we are already doing that with the open channel record 🤔

PR #1842, opened 2024-01-18
fix: error if the channel is not confirmed

bonomat:
This looks like a mistake. 

I noticed because `can_open_close_open_close_position` was failing locally. With this it works.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -497,6 +497,10 @@ impl Node {
         closing_price: Decimal,
         channel_id: DlcChannelId,
     ) -> Result<()> {
+        if !self.inner.is_dlc_channel_confirmed(&channel_id)? {
+            bail!("Underlying DLC channel not yet confirmed");
+        }
+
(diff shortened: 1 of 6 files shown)


luckysori (approved):
Oh, nice catch! I'm surprised the e2e test doesn't catch it though. I'll look into that.

I'm gonna refactor this slightly post merge, because I think the logic I introduced here should be part of the `is_dlc_channel_confirmed` function to make this cleaner.

luckysori:
Actually, a related e2e test [failed](https://github.com/get10101/10101/actions/runs/7575945735/job/20633740058?pr=1842#step:12:884), so I'll just take over and try to fix it.

luckysori:
@bonomat: all fixed AFAICT. See https://github.com/get10101/10101/pull/1842/commits/918a378d711626b5572c9f2c433b8c6313cfbb3d and https://github.com/get10101/10101/pull/1842/commits/1904d68daf52532fdba53adaf14c40a5071eb466 (not the same problem, but related).

PR #1849, opened 2024-01-19
feat: Add version endpoint

holzeis:
Demonstrates a simple cut through from the web ui to the backend.

### DIFF
--- a/webapp/frontend/lib/common/version_service.dart
+++ b/webapp/frontend/lib/common/version_service.dart
@@ -1,0 +1,36 @@
+import 'package:http/http.dart' as http;
+import 'dart:convert';
+
+class Version {
+  final String version;
+
+  const Version({required this.version});
+
+  factory Version.fromJson(Map<String, dynamic> json) {
+    return switch (json) {
+      {
+        'version': String version,
+      } =>
+        Version(version: version),
+      _ => throw const FormatException('Failed to load version.'),
+    };
+  }
+}
+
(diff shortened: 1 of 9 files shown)


webapp/frontend/lib/common/version_service.dart
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+class VersionService {
+  const VersionService();
+
+  Future<String> fetchVersion() async {
+    // TODO(holzeis): this should come from the config
+    const port = "3001";
+    const host = "localhost";
```
bonomat, line 26: how do you see this evolve to support relative paths from where it's being hosted from?

webapp/src/main.rs
```diff
@@ -120,4 +124,4 @@ async fn serve(app: Router, port: u16) -> Result<()> {
     tracing::debug!("listening on {}", listener.local_addr()?);
     axum::serve(listener, app).await?;
     Ok(())
-}
+}
```
bonomat, line 127: 🙃
Missing new line

bonomat (approved):
Nice

webapp/frontend/lib/common/version_service.dart
holzeis: I fear we can't the URI path argument required by the http package does not support relative paths. At least I haven't found a way of doing that. I presume we will have to set that in a config. (see todo).

PR #1855, opened 2024-01-19
chore: Start backend in webapp

holzeis:
Start node backend with the web app.

Demonstrates how to use the native crate api to get a new (unused) address

---

Note, we see constantly the following log, coming from our health check.

```
2024-01-19 19:08:07 DEBUG reqwest::connect: starting new connection: http://127.0.0.1:8000/
```

It is trying every second to ping the coordinator. Do we really need that? Isn't our websocket and p2p connection indication enough if the coordinator is available?


bonomat:
> It is trying every second to ping the coordinator. Do we really need that? Isn't our websocket and p2p connection indication enough if the coordinator is available?

In the app it should be every 10 seconds though. We have it at 1 second for the tests. To answer your question. Not sure 🤷‍♂️🙈

https://github.com/get10101/10101/blob/043a588541393d8295c26fc11c4809f4aa5c8f87/mobile/lib/util/environment.dart#L32

bonomat (approved):
Very simple. Let's go 🚀

PR #1857, opened 2024-01-20
feat: Send and receive on-chain funds

holzeis:
send-and-receive-onchain

### DIFF
--- a/webapp/frontend/lib/wallet/send_screen.dart
+++ b/webapp/frontend/lib/wallet/send_screen.dart
@@ -17,7 +17,10 @@ class SendScreen extends StatefulWidget {
 
 class _SendScreenState extends State<SendScreen> {
   final GlobalKey<FormState> _formKey = GlobalKey<FormState>();
+
   final TextEditingController _addressController = TextEditingController();
+  final TextEditingController _amountController = TextEditingController();
+  final TextEditingController _feeController = TextEditingController();
 
   String? address;
   Amount? amount;
@@ -72,6 +75,7 @@ class _SendScreenState extends State<SendScreen> {
                 AmountInputField(
                   value: amount != null ? amount! : Amount.zero(),
                   label: "Amount in sats",
+                  controller: _amountController,
                   validator: (value) {
                     return null;
                   },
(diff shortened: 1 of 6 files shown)


webapp/src/api.rs
```diff
@@ -17,3 +26,45 @@ pub async fn version() -> Json<Version> {
...
+    let balance = subscribers
+        .wallet_info()
+        .map(|wallet_info| Balance {
+            on_chain: wallet_info.balances.on_chain,
+            off_chain: wallet_info.balances.off_chain,
+        })
+        .unwrap_or(Balance {
+            on_chain: 0,
+            off_chain: 0,
+        });
```
bonomat, line 49: We should probably return an error to the caller instead of setting it to 0 here. Could be quite scary if the balance shows suddenly 0.

bonomat: Oh, you did it already below. nice

webapp/src/main.rs
```diff
@@ -165,3 +177,24 @@ async fn serve(app: Router, port: u16) -> Result<()> {
...
+}
+
+impl<E> From<E> for AppError
+where
+    E: Into<anyhow::Error>,
+{
+    fn from(err: E) -> Self {
+        Self(err.into())
+    }
+}
```
bonomat, line 200: Nit: I'd probably put this in api.rs.

bonomat (approved):
Looks good. A few minor nits.

webapp/frontend/lib/wallet/send_screen.dart
```diff
@@ -72,6 +75,7 @@ class _SendScreenState extends State<SendScreen> {
                 AmountInputField(
                   value: amount != null ? amount! : Amount.zero(),
                   label: "Amount in sats",
+                  controller: _amountController,
                   validator: (value) {
                     return null;
```
bonomat, line 80: When merged I'll see to make this look the same as the other input fields

webapp/src/api.rs
holzeis: I suggest we do that when we consider implementing polling or push. Otherwise, the user would simply get an error only because his request came in earlier than the wallet info event has been pusblished.

PR #1858, opened 2024-01-20
feat(web): add minimalistic frontend logger

bonomat:
this will print in the console when run through `flutter run` - if hosted, it will only print in the browser console.


webapp/frontend/lib/main.dart
```diff
@@ -8,6 +9,11 @@ import 'common/color.dart';
 import 'common/theme.dart';
 
 void main() {
+  WidgetsFlutterBinding.ensureInitialized();
+  buildLogger(false);
+  logger.i("Logger initialized");
+
+
```
holzeis, line 16: nit: this new line is too much 🙃

holzeis (approved):
👍

PR #1859, opened 2024-01-21
chore: Add web target to justfile

holzeis:


justfile
```diff
@@ -78,6 +78,10 @@ android args="":
 android-release:
     cd mobile/native && cargo ndk -t armeabi-v7a -t arm64-v8a -o ../android/app/src/main/jniLibs build --release --verbose
 
+# Build flutter webapp for cargo run --bin webapp
```
luckysori, line 81: ❓ Shall we add another recipe for `cargo run --bin webapp` then? Or combine them i.e. `just run-web`.

luckysori (approved):
I thought I was the only `justfile` user!

PR #1860, opened 2024-01-21
feat: Add wallet history to webapp

holzeis:
Screenshot 2024-01-21 at 20 12 06


webapp/src/api.rs
```diff
@@ -8,6 +8,7 @@ use axum::Json;
 use native::api;
 use native::api::Fee;
 use native::api::SendPayment;
+use native::api::WalletHistoryItemType;
```
luckysori, line 11: 🤔 I thought the idea was that we would reuse everything in `native` _except_ for the stuff in the `api` modules.

webapp/src/api.rs
```diff
@@ -73,6 +74,47 @@ pub async fn get_balance(
...
+    confirmations: u64,
+    fee: Option<u64>,
+}
+
+pub async fn get_onchain_payment_history(
+    State(subscribers): State<Arc<AppSubscribers>>,
+) -> Result<Json<Vec<OnChainPayment>>, AppError> {
+    ln_dlc::refresh_wallet_info().await?;
+
+    let history = match subscribers.wallet_info() {
```
luckysori, line 92: 🔧 `map`ping the `Option` is more idiomatic than `match`ing here.

luckysori (approved):
LGTM.

webapp/src/api.rs
holzeis: I fear the models are only on the `api` layer.

luckysori: Right! I think it might pay off to fix that, because we probably don't want to change the web app models when we are working on the mobile app.

For now we could just duplicate a couple of times.

webapp/src/api.rs
luckysori: After you showed me the alternative I disagree with original self 😄

webapp/src/api.rs
holzeis: I am not so sure about that. In the past we've copied the models extensively for the different modules (common, database, api, etc) and it was very cumbersome to write all this boilerplate code to copy data from one model to the other (which was mostly exactly the same). 

For now I suggest we keep it as is and revisit if we run into any issues with it. Note, the api models used between webapp ui and webapp backend are decoupled from the native crate.

luckysori: It would at least be sensible to move the shared code to a place that is not explicitly designed to be consumed by `flutter_rust_bridge` though.

holzeis: ok, I am going to duplicate the code then 🙈

Issue #1840, opened 2024-01-18
Could not handle trade request: Failed to open new position: Coordinator cannot trade with more than their total collateral in the DLC channel


holzeis:
`
{"error":"Could not handle trade request: Failed to open new position: Coordinator cannot trade with more than their total collateral in the DLC channel: margin (1167041) > collateral (489594) + order_matching_fee (7002)"}
`

Traded a couple of times, and then went for the max trading amount. Lesser amount are then not working either anymore.

![IMG_A9B43FBBB4F8-1](https://github.com/get10101/10101/assets/382048/92067416-6e73-4768-bb46-169dbe2c7e33)

bonomat:
How to reproduce: 

1. open channel with $1000
2. close position
3. open new position with max amount (for me it was $970) with leverage 3
4. See error: 

`
{
   "error":"Could not handle trade request: Failed to open new position: Coordinator cannot trade with more than their total collateral in the DLC channel: margin (1165571) > collateral (1144725) + order_matching_fee (6993)"
}
`

PR #1861, opened 2024-01-22
Use correct direction when computing settlement

luckysori:
All the tests I ran locally where with a fixed price, so of course this only showed up in an environment where the price can actually change.

This also highlighted the fact that we were missing some tests, so I added them.

Fixes https://github.com/get10101/10101/issues/1840.


holzeis (approved):
Thanks 👍

PR #1864, opened 2024-01-22
fix: Run all of the db backup in a dedicated thread

### LINKED ISSUE #1847: Trade modal does not disappear after order is filled
A closing order was filled and the position disappeared correctly, but the modal stayed open. I was able to close it after the timeout and everything was fine, but that was really weird.

The code that sends the `PositionCloseNotification` was definitely executed, because we sent the next message in the protocol to the coordinator, which correctly moved the position to `Closed` too.

This is probably unlikely to happen, but it's still worth reporting.

holzeis:
Without that fix the open and close position protocol was affected by the db backup to finish. That could lead to the trade dialog not showing a success message.

fixes #1847

### DIFF
--- a/mobile/native/src/backup.rs
+++ b/mobile/native/src/backup.rs
@@ -20,7 +20,6 @@ use std::fs;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::task::spawn_blocking;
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Yay!

PR #1865, opened 2024-01-22
chore: configure cors for webapp

bonomat:
This allows us to send request from the webapp while in dev mode

### DIFF
--- a/webapp/src/main.rs
+++ b/webapp/src/main.rs
@@ -22,11 +22,13 @@ use axum::response::Response;
 use axum::routing::get;
 use axum::routing::post;
 use axum::Router;
+use bitcoin::Network;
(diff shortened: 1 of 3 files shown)


webapp/src/main.rs
```diff
@@ -116,6 +117,7 @@ fn using_serve_dir(subscribers: Arc<AppSubscribers>) -> Router {
                     },
                 ),
         )
+        .layer(CorsLayer::permissive())
```
holzeis, line 120: 🔧 We should probably only add this when network is set to "regtest"

holzeis (approved):

holzeis:
Thanks, that will be very useful! 👍

PR #1866, opened 2024-01-22
feat: Add app info to settings

holzeis:
Screenshot 2024-01-22 at 12 29 50

### DIFF
--- a/webapp/frontend/lib/settings/settings_screen.dart
+++ b/webapp/frontend/lib/settings/settings_screen.dart
@@ -1,4 +1,7 @@
 import 'package:flutter/material.dart';
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
+import 'package:get_10101/common/color.dart';
+import 'package:get_10101/settings/app_info_screen.dart';
 
 class SettingsScreen extends StatefulWidget {
   static const route = "/settings";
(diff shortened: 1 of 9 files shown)


webapp/frontend/lib/settings/settings_screen.dart
```diff
@@ -9,17 +12,62 @@ class SettingsScreen extends StatefulWidget {
...
-          'Settings Screen',
-        ),
-      ],
+    return SizedBox(
+      width: 500,
+      child: Column(
+        mainAxisAlignment: MainAxisAlignment.center,
+        crossAxisAlignment: CrossAxisAlignment.center,
+        children: [
+          TabBar(
```
bonomat, line 26: You love tabs ha? 😅

webapp/frontend/lib/settings/settings_service.dart
```diff
@@ -0,0 +1,23 @@
...
+class SettingsService {
+  const SettingsService();
+
+  Future<String> getNodeId() async {
+    // TODO(holzeis): this should come from the config
+    const port = "3001";
+    const host = "localhost";
+
+    try {
+      final response = await http.get(Uri.http('$host:$port', '/api/node'));
```
bonomat, line 12: I've recently introduced a new HttpClient  which unifies these calls

https://github.com/get10101/10101/blob/dd57e3e6ff645685267929440c50daf9a86b9f40/webapp/frontend/lib/common/http_client.dart#L12-L11

bonomat (approved):
LGTM

webapp/frontend/lib/settings/settings_screen.dart
holzeis: It's fast, we can redesign when we have more time.

PR #1870, opened 2024-01-22
chore: Run axum server on https

holzeis:
We need to enforce https if we want to support hosting the webapp in the cloud. This change will now host the app solely on https using self-signed certificates. The cert path can be configured through the cli to point to production certs.


bonomat (commented):
Is this the right approach? Wouldn't it make more sense to run it behind an nginx reverse proxy ?

holzeis:
> Is this the right approach? Wouldn't it make more sense to run it behind an nginx reverse proxy ?

Terminating ssl encryption directly on the service is definitely the more secure approach.

I opted against a reverse proxy setup as it would make the deployment more complex (requires an additional service to be run and configured), while adding ssl termination on the service is simpler to setup.

webapp/src/cli.rs
```diff
@@ -74,6 +78,15 @@ impl Opts {
         Ok(data_dir)
     }
 
+    pub fn cert_dir(&self) -> Result<PathBuf> {
+        let cert_dir = match self.cert_dir.clone() {
+            None => current_dir()?.join("webapp/certs"),
```
luckysori, line 83: So if we don't specify and this fallback directory doesn't exist, we would just fail, right?

I guess this fallback is convenient if we run the `webapp` from the root of this repo.

luckysori (approved):
LGTM.

webapp/README.md
```diff
@@ -16,9 +16,9 @@ flutter build web
 ## Run the rust app
 
 ```bash
-cargo run
+cargo run - --cert-dir certs --data-dir ../data
```
bonomat, line 19:
```suggestion
cargo run -- --cert-dir certs --data-dir ../data
```

webapp/src/cli.rs
holzeis: Exactly! 👍

webapp/frontend/lib/wallet/wallet_service.dart
```diff
@@ -15,7 +15,7 @@ class WalletService {
     const host = "localhost";
 
     try {
-      final response = await http.get(Uri.http('$host:$port', '/api/balance'));
+      final response = await http.get(Uri.https('$host:$port', '/api/balance'));
```
bonomat, line 18: After rebase this change should not be necessary

webapp/src/main.rs
```diff
@@ -82,12 +86,23 @@ async fn main() -> Result<()> {
...
+    // configure certificate and private key used by https
+    let config =
+        RustlsConfig::from_pem_file(cert_dir.join("cert.pem"), cert_dir.join("key.pem")).await?;
+
+    let app = router(Arc::new(rx), network);
+
+    // run https server
+    let addr = SocketAddr::from(([0, 0, 0, 0], 3001));
+    tracing::debug!("listening on {}", addr);
+    axum_server::bind_rustls(addr, config)
```
bonomat, line 98: I'm still divided whether we should simply run on http and add https using nginx

bonomat (approved):
LGTM

holzeis:
closed in favor of https://github.com/get10101/10101/pull/1876

PR #1876, opened 2024-01-23
chore: Add authentication to web app

holzeis:
Adds authentication to our web app api and frontend.

- The user is automatically logged out after 5 mins of inactivity (not navigating)
- The session is invalidated after 1 hour of inactivity.
- The session is kept in memory, that means if the web app server is restarted the user is also logged out.
- The password can be specified via the cli (default: "satoshi")
- To run on https you will have to specify the --secure flag on cli
- Use just run-web to make use of flutter development

authentication


webapp/src/main.rs
```diff
@@ -94,7 +93,30 @@ async fn main() -> Result<()> {
...
-    let app = router(Arc::new(rx), network);
+    let session_store = InMemorySessionStore::new();
+    let deletion_task = tokio::task::spawn(
+        session_store
+            .clone()
+            .continuously_delete_expired(Duration::from_secs(60)),
+    );
+
+    let session_layer = SessionManagerLayer::new(session_store.clone())
+        .with_secure(false)
```
bonomat, line 104: What is this and why shouldn't it be `true`? :D

webapp/src/main.rs
```diff
@@ -103,21 +125,14 @@ async fn main() -> Result<()> {
         .serve(app.into_make_service())
         .await?;
 
+    deletion_task.await??;
```
bonomat, line 128: I think this will never be called because the call above blocks the thread and if you hit `ctrl+c` in the console the app will terminate.

webapp/src/auth.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        Ok(user)
+    }
+
+    async fn get_user(&self, _user_id: &UserId<Self>) -> Result<Option<Self::User>, Self::Error> {
+        Ok(Some(User {
+            password: self.hashed_password.clone(),
+        }))
+    }
+}
```
bonomat, line 72: Can we still use curl to test the endpoints?

webapp/src/main.rs
holzeis: Thanks for pointing it out. This specifies whether the cookie should get the secure flag. We should definitely set this to true in production.

webapp/src/main.rs
holzeis: I think you are right. But it doesn't hurt either does it? :)

webapp/src/auth.rs
holzeis: yes.

webapp/src/main.rs
bonomat: no, but it's useless 😅

bonomat (approved):
lets go

PR #1885, opened 2024-01-24
chore: Only depend on the certs directory if secure flag is set

holzeis:
It's cumbersome to deal with providing a valid cert path if we don't even want to start securely.


bonomat (approved):
LGTM

PR #1886, opened 2024-01-24
fix: Create certs as 1000 user

holzeis:
Also generates the certs on the default path so that the certs directory doesn't have to be set if the default certs should be used.

tested locally with the following
```
mkdir data

docker run \
  --name webapp \
  --rm \
  -v ./data:/data \
  -p 3001:3001 -it \
ghcr.io/holzeis/webapp:0.1.2  \
  --coordinator-endpoint=021025118ede915d626f7da184d75af7281dca965dd91a06993cf9d21fe4ca54e5@34.89.166.156:9045 \
  --data-dir=/data \
  --esplora=http://api.10101.finance:3000 \
  mainnet
```

### DIFF
--- a/webapp/Dockerfile
+++ b/webapp/Dockerfile
@@ -10,11 +10,12 @@ RUN apt-get update && \
     apt-get install sqlite3 -y && \
     apt-get install ca-certificates openssl -y
 
+RUN mkdir -p webapp/certs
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Thank you!

PR #1891, opened 2024-01-24
fix: Use fee estimator for dlc transaction fees

holzeis:
The fee rate was taken from our setting, which was too low. That was most likely a left over from the sub channel implementation where the funding transaction was not dependent on the fee rate set on the offered contract.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -34,6 +34,7 @@ use dlc_manager::ContractId;
 use dlc_manager::DlcChannelId;
 use dlc_messages::ChannelMessage;
 use dlc_messages::Message;
+use lightning::chain::chaininterface::ConfirmationTarget;
 use lightning::ln::ChannelId;
 use lightning::util::config::UserConfig;
 use ln_dlc_node::dlc_message::DlcMessage;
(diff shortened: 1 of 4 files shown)


coordinator/src/routes.rs
```diff
@@ -126,6 +127,7 @@ pub fn router(
     Router::new()
         .route("/", get(index))
         .route("/api/version", get(version))
+        .route("/api/fee_rate_estimate", get(get_fee_rate_estimation))
```
bonomat, line 130: would be cool if we could pass in a target here or simply return all of them.

coordinator/src/node.rs
```diff
@@ -266,7 +267,16 @@ impl Node {
         let maturity_time = trade_params.filled_with.expiry_timestamp;
         let maturity_time = maturity_time.unix_timestamp();
 
-        let fee_rate = self.settings.read().await.contract_tx_fee_rate;
+        let sats_per_vbyte = self
+            .inner
+            .fee_rate_estimator
+            .get(ConfirmationTarget::Normal)
+            .as_sat_per_vb()
+            .floor();
```
bonomat, line 275: I guess it should be always a round number but maybe _rounding_ makes more sense here?

bonomat (approved):
Looks good.

PR #1901, opened 2024-01-26
fix: Compare on upper case direction name

holzeis:


bonomat (approved):

webapp/frontend/lib/trade/order_and_position_table.dart
```diff
@@ -117,7 +117,7 @@ class OpenPositionTable extends StatelessWidget {
         for (var position in positions)
           TableRow(
             children: [
-              buildTableCell(Text(position.direction == "short"
+              buildTableCell(Text(position.direction == "Short"
```
bonomat, line 120: nit, alternative you could do `position.direction.toLowerCase == "short"`

bonomat (approved):

PR #1906, opened 2024-01-26
feat: Add min margin to channel trade constraints

holzeis:
- Sets the min quantity to 1 dollar
- Sets the min margin to 1 sat or 250k sat if there is no dlc channel yet.

https://github.com/get10101/10101/assets/382048/1a0d8d33-98ec-424b-971f-009f24888156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Chore: Set minimum quantity to 1 if dlc channel is open
+- Chore: Enforce minimum margin of 250k instead of min quantity
+
 ## [1.8.2] - 2024-01-26
 
 - Feat: Add endpoint to force close ln-dlc channels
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Chore: Set minimum quantity to 1 if dlc channel is open
```
luckysori, line 10: Won't this result in tiny DLCs that cannot be claimed on-chain?

holzeis: > Won't this result in tiny DLCs that cannot be claimed on-chain?

Aren't we considering the full amount on the CET transaction  including any potential off chain amount.

So I think we only have to care about the initial margin to be big enough for a channel but afterwards there shouldn't be a limit. 

Either party may loose so much that their payout can't be claimed anymore. But I don't think that's a problem.

luckysori: I think you are right!

I do think eventually we should track whether either side is getting too close to the dust limit, to at least inform about this.

bonomat: > I do think eventually we should track whether either side is getting too close to the dust limit, to at least inform about this.

afaik rust-dlc is taking care of this.

PR #1908, opened 2024-01-27
fix: Log error for failed channels to deserialize

holzeis:
Otherwise only one broken dlc channel breaks the whole function.


bonomat (approved):
Let's go

crates/ln-dlc-storage/src/lib.rs
luckysori, line 1: I guess this popped up when we applied the update to the production coordinator database? Although I thought we had already introduced this layer a while ago 🤔

holzeis: We only had it on the sub channels but not on the dlc channels.

Issue #1911, opened 2024-01-27
App crashes after upgrade to 1.8.3


bonomat:
On my android the app crashes after upgrading to 1.8.3

bonomat:
The crash comes from LDK panicing because we seemingly fall behind in our state. We cannot force close the channel because we would risk getting punished. 
I've overwritten said path in LDK locally and was able to force close the channel from the coordinator side.

This is the commit I made. 

https://github.com/bonomat/rust-lightning-p2p-derivatives/commit/e49030e785408f0fd4da077f63f8101cc0b2436e

PR #1928, opened 2024-01-30
fix: remove one possible panic

bonomat:
We had an issue where the app would panic because we fell behind in the protocol. This panic has been removed in this commit https://github.com/bonomat/rust-lightning-p2p-derivatives/commit/e49030e785408f0fd4da077f63f8101cc0b2436e.
This commit was built on top of our latest commit https://github.com/p2pderivatives/rust-lightning/commit/121bc324fec54cfb8b0335ec013f596ffc9d6f9f

I didn't feel good about pushing this to our usual fork in https://github.com/p2pderivatives/rust-lightning because it sounds dangerous and hence pushed it onto my own fork. We can drop this again in the future. 

(hopefully) resolves most panics.

Resolves https://github.com/get10101/10101/issues/1911

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,13 +30,13 @@ dlc = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "69d63e1" }
 p2pd-oracle-client = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "69d63e1" }
 dlc-trie = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "69d63e1" }
 
-# We should usually track the `p2pderivatives/split-tx-experiment[-10101]` branch.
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -30,13 +30,13 @@ dlc = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "69d63e1" }
 p2pd-oracle-client = { git = "https://github.com/p2pderivatives/rust-dlc", rev = "69d63e1" }
```
holzeis, line 30: I think we also need to update our rust-dlc lightning dependencies.

bonomat: It's not needed, I've tested it locally.

holzeis (approved):

PR #1916, opened 2024-01-29
feat: replace fee rate estimator with mempool.space client

bonomat:
Replace our esplora fee rate estimator with a mempool.space client.

PR #1930, opened 2024-01-31
Smarter coin selection for DLC funding transaction

### LINKED ISSUE #1912: Funding amount input is not enough to cover collateral, fund fee cet or refund fee and extra fee
After upgrading to 1.8.3 I tried to open a dlc channel with a quantity of 260 and a leverage of 2. When trying to accept the offer I fail to do so.

In my scenario I had the following values:

required collateral:  `360,664`
required fund_fee: `15,086` sats 
required cet_or_refund_fee: `3,230` sats
required extra_fee: `10,260`

===
total required input funds: `389,240` sats

However the total funds reserved for the funding transaction have only been `367,470` which is not enough to create the funding transaction. Consequently accepting the offer fails.

-----

I suspect that the following section does not take all required fees into account.

https://github.com/p2pderivatives/rust-dlc/blob/49ace9b4bc5eea6360d06773c53348af9efb002c/dlc-manager/src/utils.rs#L62-L128

luckysori:
Fixes #1912.

There are more problems that can lead to #1912, but this is definitely one of them as was demonstrated with ccbd0c55cb5682b2c5e38de4f8e9046f5e55013f.

In a different PR I will address the problem of `rust-dlc` underestimating how much money each party needs to bring in to cover the fees for all the transactions that are involved in a DLC channel.

### DIFF
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -179,7 +179,11 @@ where
         Ok(utxos)
     }
 
-    pub fn get_utxos_for_amount(&self, amount: u64, lock_utxos: bool) -> Result<Vec<Utxo>> {
+    pub fn get_utxos_for_dlc_funding_transaction(
+        &self,
+        amount: u64,
+        lock_utxos: bool,
+    ) -> Result<Vec<Utxo>> {
         let utxos = self.get_utxos()?;
         // get temporarily reserved utxo from in-memory storage
         let mut reserved_outpoints = self.locked_outpoints.lock();
--- a/crates/ln-dlc-node/src/ldk_node_wallet.rs
+++ b/crates/ln-dlc-node/src/ldk_node_wallet.rs
@@ -15,6 +15,12 @@ use bdk::FeeRate;
 use bdk::SignOptions;
 use bdk::SyncOptions;
 use bdk::TransactionDetails;
+use bdk_coin_select::metrics::LowestFee;
+use bdk_coin_select::Candidate;
+use bdk_coin_select::ChangePolicy;
+use bdk_coin_select::CoinSelector;
+use bdk_coin_select::DrainWeights;
+use bdk_coin_select::Target;
 use bitcoin::consensus::encode::serialize_hex;
 use bitcoin::psbt::PartiallySignedTransaction;
 use bitcoin::Address;
@@ -23,13 +29,14 @@ use bitcoin::BlockHash;
 use bitcoin::OutPoint;
 use bitcoin::Script;
 use bitcoin::Transaction;
+use bitcoin::TxIn;
 use bitcoin::Txid;
+use bitcoin::VarInt;
 use dlc_manager::Utxo;
 use lightning::chain::chaininterface::BroadcasterInterface;
 use lightning::chain::chaininterface::ConfirmationTarget;
 use parking_lot::Mutex;
 use parking_lot::MutexGuard;
-use rust_bitcoin_coin_selection::select_coins;
 use std::sync::Arc;
 use std::time::Instant;
 use tokio::sync::RwLock;
@@ -49,6 +56,8 @@ where
     inner: Mutex<bdk::Wallet<D>>,
     settings: RwLock<WalletSettings>,
     fee_rate_estimator: Arc<F>,
+    // Only cleared upon restart. This means that if a locked outpoint ends up unspent, it will
+    // remain locked until the binary is restarted.
     locked_outpoints: Mutex<Vec<OutPoint>>,
     node_storage: Arc<N>,
 }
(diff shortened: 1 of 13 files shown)


crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -182,55 +191,118 @@ where
...
+                        let script_sig_size = tx_in.script_sig.len();
+                        (36 + VarInt(script_sig_size as u64).len() + script_sig_size + 4) * 4
+                    };
+
+                    legacy_weight + tx_in.witness.serialized_len()
+                };
+
+                // The 10101 wallet always generates SegWit addresses.
+                //
+                // TODO: Rework this once we used Taproot.
```
bonomat, line 239:
```suggestion
                // TODO: Rework this once we use Taproot.
```

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -182,55 +191,118 @@ where
...
-
-        // update our temporarily reserved utxos with the selected once.
-        // note: this storage is only cleared up on a restart, meaning, if the protocol
-        // fails later on, the utxos will remain reserved
-        if lock_utxos {
-            for utxo in selected_local_utxos.clone() {
-                reserved_outpoints.push(utxo.utxo.outpoint);
+        // This is a standard base weight (without inputs or change outputs) for on-chain DLCs. We
+        // assume that this value is still correct for DLC channels.
+        let funding_tx_base_weight = 212;
```
bonomat, line 248: The weight depends on how many inputs will be selected, so very likely this will be wrong if there are more than 1 input per party.

bonomat (approved):
Looks good. Thanks.

crates/ln-dlc-node/src/ldk_node_wallet.rs
luckysori: True, although this is the _base_ weight. According to `bdk_coin_select`:

```rust
/// The `base_weight` is the weight of the transaction without any inputs and without a change
/// output.
```

bonomat: oh, thanks for the explanation.

luckysori:
FYI: we're running into weird `bitcoind`-related problems on CI. I assume this is because of the many transactions and blocks that we are mining for the new test, so I'm gonna rework this test setup by generating a single transaction with many outputs instead.

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -50,6 +56,8 @@ where
     inner: Mutex<bdk::Wallet<D>>,
     settings: RwLock<WalletSettings>,
     fee_rate_estimator: Arc<F>,
+    // Only cleared upon restart. This means that if a locked outpoint ends up unspent, it will
+    // remain locked until the binary is restarted.
     locked_outpoints: Mutex<Vec<OutPoint>>,
```
holzeis, line 61: Unrelated to this PR, but maybe these `locked_outpoints` should be attached with a timestamp with which we can reuse locked outputs if they haven't been used e.g. more than 30 mins?

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -180,58 +188,121 @@ where
...
     ) -> Result<Vec<Utxo>> {
-        let utxos = self.get_utxos()?;
-        // get temporarily reserved utxo from in-memory storage
+        let network = {
+            let bdk = self.bdk_lock();
+            bdk.network()
+        };
+
+        let fee_rate = fee_rate.map(|fee_rate| fee_rate as f32).unwrap_or_else(|| {
+            self.get_fee_rate(ConfirmationTarget::Normal)
```
holzeis, line 203: 🤔 Our fee rate estimator is pretty pessimistic on the fees. `Normal` has been overpaid 2 times the other day. For the dlc channel offer we use a fee rate with a `ConfirmationTarget::Background`. Maybe we should use the same here?

crates/ln-dlc-node/src/ldk_node_wallet.rs
```diff
@@ -180,58 +188,121 @@ where
...
+                .as_sat_per_vb()
+        });
+
+        // Get temporarily reserved UTXOs from in-memory storage.
         let mut reserved_outpoints = self.locked_outpoints.lock();
 
-        // filter reserved utxos from all known utxos to not accidentally double spend and those who
-        // have actually been spent already
+        let utxos = self.get_utxos()?;
```
holzeis, line 210: 🤔 Could we run into race conditions here? e.g. if two dlc channels are setup at the same time and they pick the same utxos? i guess we need some locking around selecting and reserving utxos.

crates/ln-dlc-node/src/ldk_node_wallet.rs
bonomat: Let's switch to a better feerate estimator https://github.com/get10101/10101/pull/1916 .

crates/ln-dlc-node/src/ldk_node_wallet.rs
luckysori: Hmm, there is a mutex around locked_outpoints.

luckysori: Actually, I think you have a point, but for a different scenario.

Two independent calls to `get_utxos_for_dlc_funding_transaction` cannot run into a race condition, because the `locked_outpoints` mutex will be held by one of them first, and it is only dropped at the end of the function.

But a different method not protected by this specific mutex could pick the same UTXOs and try to use them too. I think Philipp already fixed this problem with d9178a45354f271fc29fad917cdabd35fc8048ec, but it wouldn't be hard to misuse this mutex.

Maybe with the new version of `bdk` UTXO management is a bit easier.

crates/tests-e2e/src/bitcoind.rs
```diff
@@ -75,6 +71,106 @@ impl Bitcoind {
         Ok(response)
     }
 
+    pub async fn send_multiple_utxos_to_address<F>(
```
luckysori, line 74: I had to add this bad boy to avoid weird CI bugs related to calling `sendrawtransaction` and `generatetoaddress` too many times in a row.

luckysori: Actually that was a gigantic waste of time, because an actual problem was that we cannot have more than one e2e test per file 🤦

crates/ln-dlc-node/src/ldk_node_wallet.rs
holzeis: Right, I missed the lock just a few lines above 🙈

PR #1931, opened 2024-01-31
feat: add 'Max' button to sending on-chain screen

### LINKED ISSUE #1925: Add `Send Max` button in sending on-chain screen

Restioson:
Proof of concept for #1925; major issue is how this should be handled in the amount field (see https://github.com/get10101/10101/issues/1925#issuecomment-1918602574). This is one possible solution but isn't the greatest as
1) If the fee increases it'd be over the amount again
2) It doesn't actually use the draining logic

That being said, it is still editable by the user instead of being an entirely new mode of interaction, which is a plus.

Resolves #1925

### DIFF
--- a/mobile/lib/features/wallet/send/send_onchain_screen.dart
+++ b/mobile/lib/features/wallet/send/send_onchain_screen.dart
@@ -34,7 +34,8 @@ class _SendOnChainScreenState extends State<SendOnChainScreen> {
 
   ChannelInfo? channelInfo;
 
-  Amount _amount = Amount.zero();
+  // null = max
+  Amount? _amount = Amount(1000);
   Fee _fee = PriorityFee(ConfirmationTarget.normal);
   Map<ConfirmationTarget, FeeEstimation>? _feeEstimates;
   late WalletService _walletService;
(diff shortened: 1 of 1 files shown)


Restioson:
> Mhm, I would just leave the logic with 0 and add a Max button. If the Max button is clicked, the input field is not editable and it says Max. Then we can keep the logic with 0 in the backend.

Maybe it can be some kind of toggle - when pressed, it becomes surrounded by blue and enabled. When tapped again, the background becomes white, unlocking the edit thing

Restioson:
image
image

I changed the default amount to 1000 and added an 'enter an amount' error when the amount is zero - this is to prevent unticking max showing an error immediately

bonomat:
> I changed the default amount to 1000 and added an 'enter an amount' error when the amount is zero - this is to prevent unticking max showing an error immediately

I like it.

mobile/lib/features/wallet/send/send_onchain_screen.dart
```diff
@@ -62,11 +64,21 @@ class _SendOnChainScreenState extends State<SendOnChainScreen> {
 
     setState(() {
       _feeEstimates = fees;
-      _amount = widget.destination.amount;
-      _controller.text = _amount.formatted();
+      Amount amt = widget.destination.amount;
+      amt = amt.sats == 0 ? Amount(1000) : amt;
+      logger.i(amt);
```
bonomat, line 69: I guess the logger is not needed anymore.

bonomat (approved):
I haven't run it yet, but from what I see this should be fine.

mobile/lib/features/wallet/send/send_onchain_screen.dart
```diff
@@ -163,7 +177,7 @@ class _SendOnChainScreenState extends State<SendOnChainScreen> {
                                   keyboardType: TextInputType.number,
                                   textAlign: TextAlign.center,
                                   decoration: const InputDecoration(
-                                      hintText: "0.00",
+                                      hintText: "1000",
```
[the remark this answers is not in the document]
Restioson: It could be, but we are setting it to 1000 as the default in any case

mobile/lib/features/wallet/send/send_onchain_screen.dart
Restioson: Good catch, thanks!

PR #1932, opened 2024-01-31
fix: share address without BIP-21 query component

### LINKED ISSUE #1929: Creating a payment request on-chain with an amount leads to a malformed address
Steps to reproduce:
- Navigate to balances screen
- Select on-chain
- Click 'receive'
- Click 'edit'
- Enter an amount
- Click 'continue'
- Click the copy icon next to 'Address'

In this case, what is copied is something like `bcrt1qh2dgxfxy9wmwsmspl49saf0tas54jaq8gns7ln?amount=0.00000123`. As far as I am aware this is not a valid address. We create this by removing the BIP21 `bitcoin:` prefix, but we don't remove the query parameters such as the amount. Thus this is neither a valid address nor a valid BIP21 URI

The copy icon under the QR code is still okay as it does not truncate the `bitcoin:` prefix

Restioson:
Fixes #1929

### DIFF
--- a/mobile/lib/features/wallet/application/wallet_service.dart
+++ b/mobile/lib/features/wallet/application/wallet_service.dart
@@ -51,7 +51,7 @@ class WalletService {
       rust.PaymentRequest req =
           await rust.api.createPaymentRequest(amountSats: amount?.sats, description: description);
       logger.i("Successfully created payment request.");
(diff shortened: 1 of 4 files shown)


bonomat (approved):
Much better :D

Issue #1894, opened 2024-01-25
Flaky `can_revert_channel` test


luckysori:
I left a TODO related to this, but actually the test recently failed on [CI](https://github.com/get10101/10101/actions/runs/7649698400/job/20844576121#step:12:1529) on a different assertion! I've added some logs to debug this intermittent problem (see https://github.com/get10101/10101/commit/a3fd47e80c74e80b0486e7c4b07c9cb8986134f7).

luckysori:
I'm glad you're looking into this, @bonomat, because one of my PRs keeps running into this issue.

bonomat:
Uff... it's really hard to figure our why this test is flaky. @luckysori , do you have an assumption? 

My guess was it's because we fail at broadcasting the transaction but the logs suggest that it was successful. 
Locally it always works for me unfortunately which makes it really hard to debug. 

If you have a CI failure, can you add the failure here below please?

bonomat:
I start 😅

PR: https://github.com/get10101/10101/pull/1942 - CI failure: https://github.com/get10101/10101/actions/runs/7747015212/job/21126592601?pr=1942

PR #1942, opened 2024-02-01
chore: print latest height after sync

bonomat:
This is useful to see what height we have synced up to.

On the path of debugging https://github.com/get10101/10101/issues/1894 I wandered down the path of questioning our wallet design and if we miss a block or don't sync up to the latest height. This additional log field might help.


holzeis (approved):

luckysori (approved):

bonomat:
Merging this without CI to get over flaky e2e tests.

PR #1944, opened 2024-02-02
chore: Reject all pending dlc channel offers

holzeis:
We need that as due to the bug https://github.com/get10101/10101/issues/1912 users have tried multiple times to submit their order. And since we don't check if the user already has an offered channel, we simply offer another one, which then again fails.

That will leave us with multiple positions in state `Proposed` and offered dlc channels. By simply rejected all offered on reconnect we clean up those positions and dlc channels.

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -662,6 +662,7 @@ impl Node {
         is_stable_order: bool,
     ) -> Result<()> {
         let trader_peer_id = trade_params.pubkey;
+
         match self
             .inner
             .get_signed_dlc_channel_by_counterparty(&trader_peer_id)?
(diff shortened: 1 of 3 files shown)


coordinator/src/node.rs
```diff
@@ -672,6 +673,16 @@ impl Node {
...
                 );
 
+                ensure!(
+                    !self
+                        .inner
+                        .list_dlc_channels()?
+                        .iter()
+                        .filter(|c| c.get_counter_party_id() == trader_peer_id)
+                        .any(|c| matches!(c, Channel::Offered(_) | Channel::Accepted(_))),
+                    "Previous DLC Channel offer still pending."
```
bonomat, line 683: How do we recover from this? 
Does an `Offer` time out?

bonomat (approved):
Looks good. Just a nit-pick and one question

coordinator/src/node.rs
holzeis: The app needs to reject the offer.

bonomat: Thanks for the explanation.

PR #1948, opened 2024-02-02
feat: Add emergency kit screen to set pending orders to failed

holzeis:
Introduces an emergency kit section to the settings to cleanup pending orders. I think this pattern might be useful in the future to selectively fix broken states on the app site.

https://github.com/get10101/10101/assets/382048/451155d8-b11a-4ebf-8a8c-1be8e3be7845


mobile/native/src/db/models.rs
```diff
@@ -83,6 +83,16 @@ impl Order {
         }
     }
 
+    /// Sets all filling orders to failed. Only be used for emergency recoveries!
+    pub fn set_all_filling_orders_to_failed(conn: &mut SqliteConnection) -> Result<()> {
+        diesel::update(orders::table)
+            .filter(schema::orders::state.eq(OrderState::Filling))
+            .set(orders::state.eq(OrderState::Failed))
+            .execute(conn)?;
```
bonomat, line 91: Do we get a `updated_rows` count here? If so, it might be worth logging this. This could help us debug things in the future.

bonomat (approved):
What's the side effect of setting `Filling` orders to failed`? If we get a match later on, we would simply fail right?

holzeis:
> What's the side effect of setting `Filling` orders to failed`? If we get a match later on, we would simply fail right?

Can't predict all side effects, but in the case if we would have a regular order which gets executed. We would get an error when trying to set the order to `Filled` as the `Filling` order was not found. That would also prevent the position from getting created on the app side, but the dlc channel would be set up.

PR #1953, opened 2024-02-03
chore: Set maturity time of dummy contract to the future

holzeis:
The maturity time was hard coded to Feb 3rd 2024 which has expired now and was breaking our ln-dlc-node tests.

This change dynamically calculates a maturity time 7 days in the future.


bonomat (approved):
Thanks for looking into this

PR #1954, opened 2024-02-03
Alwasy force-close LN-DLC channel via `rust-lightning` first

luckysori:
We already have better tested logic in `rust-dlc` to handle any channel closures coming from `rust-lightning`.

This change is important because the logic to choose between closing via `rust-dlc` or `rust-lightning` was flawed, as some vanilla LN channels still had old subchannel state associated with them. Now we can just attempt to close via `rust-lightning` _first_ and then fall back to `rust-dlc` if the channel is currently split.


bonomat (approved):
Permission granted

PR #1965, opened 2024-02-05
fix: use correct path

bonomat:
this action is executed from the workspace dir, hence, we need to provide the full path


bonomat:
Merging right away.

PR #1967, opened 2024-02-06
feat(coordinator): sort dlc channels by state

bonomat:
This will sort the channels retrieved via `/api/admin/dlc_channels` by their `state` first and then by their signed_channel_state` if the `state` is equal.


holzeis (approved):
cool thanks! would be great if the dlc channel would also include a timestamp 🙈

PR #1969, opened 2024-02-06
Feat(settings): Manage dlc channels in the app settings

holzeis:
Allows the user to view all dlc channels and to delete selected dlc channels in the emergency kit.

https://github.com/get10101/10101/assets/382048/aab6e29d-ce4e-4866-9fe2-b0f76d5228bf


mobile/lib/common/dlc_channel_change_notifier.dart
```diff
@@ -0,0 +1,154 @@
...
+      }
+      return false;
+    });
+  }
+
+  bool canForceClose() {
+    return hasDlcChannel();
+  }
+
+  /// Whether the current LN-DLC channel is closing or not.
```
bonomat, line 137: 🙃 It's not a **LN**-DLC channel anymore

mobile/lib/common/settings/channel_screen.dart
```diff
@@ -71,60 +89,244 @@ class _ChannelScreenState extends State<ChannelScreen> {
...
+                              ),
+                            ],
+                          )),
+                      const Divider(height: 1, thickness: 1, indent: 5, endIndent: 5),
+                      Column(
+                        children: <Widget>[
+                          const SizedBox(height: 10.0),
+                          Theme(
+                            data: Theme.of(context).copyWith(dividerColor: Colors.transparent),
+                            child: ExpansionTile(
```
bonomat, line 118: Most of those `ExpansionTile` look the same with only being the title and the list of channels being different. 
Would you mind extracting this into a function or a widget to remove duplication please?

mobile/lib/common/settings/emergency_kit_screen.dart
```diff
@@ -124,6 +128,64 @@ class _EmergencyKitScreenState extends State<EmergencyKitScreen> {
                           SizedBox(width: 10),
                           Text("Cleanup filling orders", style: TextStyle(fontSize: 16))
                         ])),
+                    const SizedBox(height: 30),
+                    Row(
```
bonomat, line 132: Why not having the delete button next to the channel?

bonomat (approved):
In general this is good. I'd say the design needs some more love but that's subjective :)

mobile/lib/common/settings/emergency_kit_screen.dart
holzeis: because I don't want somebody to accidentally delete their channel. IMO, it's safer to make that a bit more cumbersome :)

PR #1975, opened 2024-02-07
Feat/ci e2e matrix

bonomat:
This patch runs our e2e tests isolated in CI which was inspired by rust-dlc. 

**Note: an e2e test file should be called `e2e_*` otherwise it won't be executed**

See here: https://github.com/bonomat/10101/actions/runs/7805792927/job/21291279292?pr=7


luckysori (approved):
Very happy to try this, thank you.

bonomat:
I'm merging this so that we get separate e2e tests

PR #1978, opened 2024-02-07
feat(coordinator): leadership board api

bonomat:
The new api supports 3 query params:
- `top` - how many entries you want in the board
- `reverse` - top or flop
- `category` - whether you want by PnL or Volume

example: 

```bash
curl localhost:8000/api/leaderboard\?top\=3\&reverse=false\&category\='Pnl' -v | jq .

{
  "entries": [
    {
      "trader": "02d5aa8fce495f6301b466594af056a46104dcdc6d735ec4793aa43108854cbd4a",
      "pnl": "181905",
      "volume": "3156",
      "rank": 1
    },
    {
      "trader": "03b6fbc0de09815e2eb508feb8288ba6ac7f24aa27bd63435f6247d010334eaff2",
      "pnl": "35406",
      "volume": "1815",
      "rank": 2
    },
    {
      "trader": "02d8166a761dfe894575f848f419b6985500fdcb07a47e68f172b8c46ccd6a120a",
      "pnl": "0",
      "volume": "1000",
      "rank": 3
    }
  ]
}
```

Resolves https://github.com/get10101/meta/issues/322

The second commit (0fd674969065ca1c02d3ee3a1f5c39d5e8fd2f8b) fixes a bug where the PnL was calculated wrongly. However, this will only work for open and closing a position. Once we have position resize, this will be wrong. Imho, we should calculate PnL on a trade and not store it on the position. 

The last commit (889c00f9c998d73a1a2dcfa76b03b76b1668557f) is a bit of renaming. This buggs me everytime I look at the database 🙈


holzeis (approved):
LGTM, alternatively to the api we could have added a query to our metabase :)

bonomat:
> LGTM, alternatively to the api we could have added a query to our metabase :)

I wanted it as an API so that we can also display it in the app :)

PR #1987, opened 2024-02-08
fix: Refresh dlc channels when opening collab close screen

holzeis:
Otherwise the channel state is not correctly considered.


luckysori (approved):
Nice. I just ran into something similar that was caught by a UI test. Maybe we should add one for this flow.

PR #1988, opened 2024-02-08
chore: Cleanup docker compose

holzeis:
Removes lnd and chopsticks from our dev setup.

The nigiri chopsticks container was only useful to create a wallet and generate the first 200 blocks. Other than that its just a technical burden.

This patch removes the chopstick service from our dev setup by manually creating a wallet and mining 300 blocks when running just docker.

‼️ docker-compose up -d on an wiped environment will not be sufficient anymore. You will have to use `just docker` to start the environment, which ensures that wallet is created and 101 blocks are mined.
‼️ Chopsticks weirdly always mined a block when broadcasting a transaction. This is changed now that we got rid of chopsticks.

### DIFF
--- a/justfile
+++ b/justfile
@@ -106,7 +106,7 @@ run args="":
       --dart-define="COMMIT=$(git rev-parse HEAD)" \
       --dart-define="BRANCH=$(git rev-parse --abbrev-ref HEAD)" \
       --dart-define="REGTEST_FAUCET=http://localhost:8080" \
-      --dart-define="REGTEST_MAKER_FAUCET=http://localhost:18000/api/pay-invoice" \
+      --dart-define="ESPLORA_ENDPOINT=http://localhost:3000" \
       --dart-define="HEALTH_CHECK_INTERVAL_SECONDS=2"
 
 # Run against our public regtest server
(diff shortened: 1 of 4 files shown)


holzeis:
Note, while working on this is noticed that we have a confic ESPLORA_ENDPOINT, which is actually the ELECTRS_ENDPOINT. We should clean that up.

luckysori (approved):
LGTM if it works.

justfile
```diff
@@ -314,8 +312,18 @@ ln-dlc-node-test args="": docker
...
+    set -euxo pipefail
     docker-compose up -d
 
+    # wait a few seconds to ensure that Docker containers started
+    sleep 2
+
+    docker exec -it bitcoin bitcoin-cli --regtest -rpcuser=admin1 -rpcpassword=123 createwallet "test"
+    address=$(docker exec -it bitcoin bitcoin-cli --regtest -rpcuser=admin1 -rpcpassword=123 getnewaddress)
+    address="${address%%[[:cntrl:]]}"
+    docker exec bitcoin bitcoin-cli --regtest -rpcuser=admin1 -rpcpassword=123 generatetoaddress 300 $address
```
bonomat, line 325: NIT: I wouldn't generate 300 blocks but only the necessary amount (101). Generating 300 blocks will take (a bit) longer for no value.

justfile
```diff
@@ -314,8 +312,18 @@ ln-dlc-node-test args="": docker
 
 # Runs background Docker services
 docker:
+    #!/usr/bin/env bash
+    set -euxo pipefail
     docker-compose up -d
 
+    # wait a few seconds to ensure that Docker containers started
+    sleep 2
```
bonomat, line 320: This asks for flakyness :/ We might need some busy wait here.

bonomat (approved):
Thank you for addressing this.

justfile
holzeis: we actually don't have to wait. bitcoin is up instantly.

PR #1994, opened 2024-02-09
feat(mobile): allow users to change contact details

### LINKED ISSUE #1727: Add feature to add/update/detele an email

bonomat:
resolves #1727 

https://github.com/get10101/10101/assets/224613/78c2d9fa-0da6-46b1-8e76-33d0a2d94c4e

### DIFF
--- a/coordinator/src/routes.rs
+++ b/coordinator/src/routes.rs
@@ -21,6 +21,7 @@ use crate::collaborative_revert::confirm_legacy_collaborative_revert;
 use crate::db;
 use crate::db::liquidity::LiquidityRequestLog;
 use crate::db::user;
+use crate::db::user::User;
 use crate::is_liquidity_sufficient;
 use crate::leaderboard::generate_leader_board;
 use crate::leaderboard::LeaderBoard;
@@ -160,7 +161,11 @@ pub fn router(
         .route("/api/orderbook/websocket", get(websocket_handler))
         .route("/api/trade", post(post_trade))
         .route("/api/rollover/:dlc_channel_id", post(rollover))
+        // Deprecated: we just keep it for backwards compatbility as otherwise old apps won't
+        // pass registration
         .route("/api/register", post(post_register))
+        .route("/api/users", post(post_register))
+        .route("/api/users/:trader_pubkey", get(get_user))
         .route("/api/admin/wallet/balance", get(get_balance))
         .route("/api/admin/wallet/utxos", get(get_utxos))
         .route("/api/admin/channels", get(list_channels).post(open_channel))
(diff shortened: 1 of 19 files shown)


coordinator/src/routes.rs
```diff
@@ -160,7 +161,11 @@ pub fn router(
         .route("/api/orderbook/websocket", get(websocket_handler))
         .route("/api/trade", post(post_trade))
         .route("/api/rollover/:dlc_channel_id", post(rollover))
+        // Deprecated: we just keep it for backwards compatbility as otherwise old apps won't get
```
luckysori, line 164:
```suggestion
        // Deprecated: we just keep it for backwards compatbility as otherwise old apps won't
```

mobile/lib/features/welcome/welcome_screen.dart
```diff
@@ -136,21 +129,17 @@ class _WelcomeScreenState extends State<WelcomeScreen> {
...
                                   labelStyle: const TextStyle(color: Colors.black87, fontSize: 14),
                                   hintText: 'Let us know how to reach you'),
                               validator: (value) {
                                 if (value == null || value.isEmpty) {
                                   return null;
                                 }
-
-                                if (!isEmailValid(value)) {
-                                  return 'Please enter a valid email address';
-                                }
```
luckysori, line 149: Should we at least validate that it is not longer than a certain number of characters? I'm not too keen on letting users insert arbitrary stuff in the database.

mobile/native/src/trade/users/mod.rs
```diff
@@ -5,37 +5,61 @@ use anyhow::anyhow;
...
 }
+
+/// Retrieve latest user details
+pub async fn get_user_details() -> Result<User> {
+    let key = ln_dlc::get_node_pubkey();
+
+    let client = reqwest_client();
+    let response = client
+        .get(format!(
+            "http://{}/api/users/{}",
```
luckysori, line 54: It would be nice if we could cache this rather than sending a HTTP request every time the user opens the user settings.

luckysori (approved):
Generally LGTM!

I think eventually we will want a bit more control over all the IDs the user might want to provide, but I like a simple approach for now.

mobile/native/src/trade/users/mod.rs
bonomat: I postpone this to once it is an issue :)

coordinator/src/routes.rs
```diff
@@ -160,7 +161,11 @@ pub fn router(
         .route("/api/orderbook/websocket", get(websocket_handler))
         .route("/api/trade", post(post_trade))
         .route("/api/rollover/:dlc_channel_id", post(rollover))
+        // Deprecated: we just keep it for backwards compatbility as otherwise old apps won't
+        // pass registration
         .route("/api/register", post(post_register))
+        .route("/api/users", post(post_register))
+        .route("/api/users/:trader_pubkey", get(get_user))
```
holzeis, line 168: 🔧 That api is concerning for me. If somebody learns about a users node id they can use this api to query user data.

I propose we either validate a users signature on that api or rather put that information into the authenticated message received by the app after a successful login. (there we already verify the users signature)

coordinator/src/routes.rs
```diff
@@ -160,7 +161,11 @@ pub fn router(
         .route("/api/orderbook/websocket", get(websocket_handler))
         .route("/api/trade", post(post_trade))
         .route("/api/rollover/:dlc_channel_id", post(rollover))
+        // Deprecated: we just keep it for backwards compatbility as otherwise old apps won't
+        // pass registration
         .route("/api/register", post(post_register))
+        .route("/api/users", post(post_register))
```
holzeis, line 167: same here, anybody could use this api to update the contact details of a user.

Issue #1993, opened 2024-02-09
Can't click buttons in UI


bonomat:
User feedback on Telegram

![telegram-cloud-photo-size-4-5922500184967855841-y](https://github.com/get10101/10101/assets/224613/8cd2b598-68e1-4b54-9aa3-134e24f8f535)

PR #1996, opened 2024-02-09
fix: don't use fix width but expand row

bonomat:
this will allow us to display nicely on a smaller device as well.

Resolves https://github.com/get10101/10101/issues/1993


holzeis (approved):
LGTM 👍

PR #1999, opened 2024-02-09
fix(dlc_handler): Only accept collab close offer if not our own

holzeis:
The `Channel::CollaborativelyCloseOffered` state did not include who offered, so the `on_reconnect` logic incorrectly always tried to accept a collab close offer even if it has been proposed by themselves.

`rust-dlc` does not store that information, so I fix that in https://github.com/p2pderivatives/rust-dlc/commit/ba68a7bc5db4bdafb24f27fd485507fbf4322f81 by adding a `is_offer` flag to the `Channel::CollaborativelyCloseOffered` state. (analog to how it is done on the `Channel::RenewOffered` state)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1196,7 +1196,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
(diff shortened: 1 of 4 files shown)


luckysori (approved):
LGTM!

PR #2000, opened 2024-02-09
feat: allow to increase the number of addresses to sync for

bonomat:
I've decided against putting it into the recovery screen as the recommendation should be to increase this number incrementally. The reason is, the higher they gap, the slower the sync. 
If this setting would have been only available on the recover screen, a user might need to reset the app again to sync again, hence, it is kinda required to be used from a fully setup app. 
I've created a new category in the settings screen, which can come handy in the future, e.g. edit the esplora endpoint.

This fixes a problem where we don't see money in the wallet after recovering the wallet.

https://github.com/get10101/10101/assets/224613/10dc53a7-45cc-4537-a707-5ca83077986c

### DIFF
--- a/mobile/lib/common/settings/wallet_settings.dart
+++ b/mobile/lib/common/settings/wallet_settings.dart
@@ -1,0 +1,139 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:get_10101/common/settings/settings_screen.dart';
+import 'package:get_10101/common/snack_bar.dart';
+import 'package:get_10101/logger/logger.dart';
+import 'package:go_router/go_router.dart';
+import 'package:get_10101/ffi.dart' as rust;
+
+class WalletSettings extends StatefulWidget {
+  static const route = "${SettingsScreen.route}/$subRouteName";
+  static const subRouteName = "walletSettings";
+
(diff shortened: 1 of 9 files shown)


mobile/lib/common/settings/wallet_settings.dart
```diff
@@ -0,0 +1,139 @@
...
+            const SizedBox(
+              height: 20,
+            ),
+            Expanded(
+              child: Column(
+                mainAxisAlignment: MainAxisAlignment.start,
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  const Text(
+                    "The amount of addresses to sync for (at least). Once you confirm, a full wallet sync will be performed. The higher the gap is, the longer the sync will take. Hence, we recommend syncing incrementally.",
```
holzeis, line 74: 🤔 Shouldn't we also recommend to set the gap back to a smaller number after the wallet has been successfully synced?

holzeis (approved):
LGTM 👍

mobile/lib/common/settings/wallet_settings.dart
luckysori: I wonder if these syncs get less painful with the newest versions of `bdk`, since the wallet can be used during the sync.

PR #2005, opened 2024-02-10
fix: Propose the settled amount when collab closing a channel

holzeis:


bonomat (approved):
Looks good to me. Thanks @holzeis .

I've added a test which asserts the balance after closing and pushed this test before the fix to show that it first fails and then works. 

Note: [this](https://github.com/get10101/10101/issues/2006) is now an issue, but that's not critical

PR #2009, opened 2024-02-11
fix: e2e tests

### LINKED ISSUE #1894: Flaky `can_revert_channel` test

bonomat:
I might have caught one or the other flaky test with this. 

Might have fixed #1894

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,9 +220,6 @@ jobs:
         run: cargo test -p tests-e2e --no-run
       - name: build coordinator and maker
         run: cargo build --bin maker --bin coordinator
-      - name: set-matrix1
-        id: set-matrix1
-        run: echo "matrix=$(scripts/get_test_list.sh)}"
       - name: set-matrix
         id: set-matrix
         run: echo "matrix=$(scripts/get_test_list.sh)}" >> $GITHUB_OUTPUT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
       - uses: extractions/setup-just@v1
       - name: Setup rust toolchain
         run: rustup show
-      - uses: actions/cache@v2
+      - uses: actions/cache@v4
         env:
           cache-name: test-cache
         with:
@@ -236,7 +236,7 @@ jobs:
         tests: ${{ fromJson(needs.e2e_tests_prepare.outputs.matrix) }}
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/cache@v2
+      - uses: actions/cache@v4
         env:
           cache-name: test-cache
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,10 @@ jobs:
         env:
           cache-name: test-cache
         with:
-          path: target/debug/deps
+          path: |
+            target/debug/deps
+            target/debug/coordinator
+            target/debug/maker
           key: test-cache-${{ github.run_id }}-${{ github.run_number }}
       - name: Download RUST FFI bindings
         uses: actions/download-artifact@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -249,8 +255,34 @@ jobs:
       - name: Test containers are up
         run: |
           curl http://localhost:8080/bitcoin -d '{"jsonrpc": "1.0", "method": "sendtoaddress", "params": ["bcrt1qylgu6ffkp3p0m8tw8kp4tt2dmdh755f4r5dq7s", 0.1]}'
-      - name: Run coordinator, maker, etc
```
luckysori, line 253: ❓ Why do `just coordinator`  and `just maker` not work?

crates/tests-e2e/tests/e2e_collaborative_close_channel.rs
```diff
@@ -43,6 +43,7 @@ async fn can_open_and_collab_close_channel() {
 
     // wait until there is no balance off-chain anymore
     wait_until!({
+        test.bitcoind.mine(1).await.unwrap();
```
luckysori, line 46: 🔧 Not wrong, but shouldn't it be enough to mine 1 block i.e. just before the `wait_until`?

crates/tests-e2e/src/setup.rs
```diff
@@ -57,14 +57,20 @@ impl TestSetup {
...
         let address = self.coordinator.get_new_address().await.unwrap();
 
-        self.bitcoind
-            .send_to_address(&address, amount)
-            .await
-            .unwrap();
+        let sats_per_fund = amount.to_sat() / in_utxos;
+        for _ in 0..in_utxos {
+            self.bitcoind
+                .send_to_address(&address, Amount::from_sat(sats_per_fund))
```
luckysori, line 70: 🙃 Almost certainly unnecessary, but you could use my recently introduced `send_multiple_utxos_to_address`.

With your approach, I think in testing I ran into a problem where I was generating too many transactions, so they wouldn't all fit in a single block!

luckysori: 🙃 Can you document in the commit message why this change is needed? (A comment here is okay too.)

luckysori (approved):
Thank you for iterating on this new setup!

.github/workflows/ci.yml
bonomat: It would rebuild the binaries even though they are in `target/debug`

crates/tests-e2e/tests/e2e_collaborative_close_channel.rs
bonomat: Potentially, 1 block should be enough.

crates/tests-e2e/src/setup.rs
bonomat: the `reject_offer` test failed because we didn't have enough UTXOs on the second order. I first generated an other block in between the test (before opening the second position) but figured this might still be flaky - or we might need an other busy wait there. Hence, I changed the setup so that one can allow for multiple UTXOs at the getgo. I'll update the commit message

.github/workflows/ci.yml
luckysori: We could split it into `build-coordinator` and `run-coordinator` and then `coordinator` would call both. Then you could use `run-coordinator` here.

Not a huge deal, but it might be worth it to avoid duplication.

bonomat: I'd do this when we have to touch it again.

PR #2021, opened 2024-02-12
feat: finetune leader board

bonomat:
Leader board has now a nickname which can be configured in the app (or is random generated on registration). 
There is no way to opt-out of the leader board at the moment. 

Also, the leaderboard api allows for defining a start and end date. 

e.g. use with 

`curl 'localhost:8000/api/leaderboard?top=3&category=Pnl&start=2024-02-11&end=2024-02-13' -v | jq .`

resolves [#341](https://github.com/get10101/meta/issues/341)

### DIFF
--- a/coordinator/migrations/2024-02-09-152833_add_nickname_to_user_table/up.sql
+++ b/coordinator/migrations/2024-02-09-152833_add_nickname_to_user_table/up.sql
@@ -1,2 +1,2 @@
 ALTER TABLE
-    users ADD COLUMN nickname TEXT NOT NULL DEFAULT '';
+    users ADD COLUMN nickname TEXT;
--- a/coordinator/migrations/2024-02-09-152833_add_nickname_to_user_table/up.sql
+++ b/coordinator/migrations/2024-02-09-152833_add_nickname_to_user_table/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE
+    users ADD COLUMN nickname TEXT NOT NULL DEFAULT '';
--- a/coordinator/src/db/user.rs
+++ b/coordinator/src/db/user.rs
@@ -19,7 +19,7 @@ pub struct User {
     pub timestamp: OffsetDateTime,
     pub fcm_token: String,
     pub last_login: OffsetDateTime,
-    pub nickname: String,
+    pub nickname: Option<String>,
 }
 
 impl From<RegisterParams> for User {
@@ -28,7 +28,7 @@ impl From<RegisterParams> for User {
             id: None,
             pubkey: value.pubkey.to_string(),
             contact: value.contact.unwrap_or("".to_owned()),
-            nickname: value.nickname.unwrap_or("".to_owned()),
+            nickname: value.nickname,
             timestamp: OffsetDateTime::now_utc(),
             fcm_token: "".to_owned(),
             last_login: OffsetDateTime::now_utc(),
@@ -56,7 +56,6 @@ pub fn upsert_user(
 ) -> QueryResult<User> {
     // If no name or contact has been provided we default to empty string
     let contact = contact.unwrap_or_default();
-    let nickname = nickname.unwrap_or_default();
 
     let timestamp = OffsetDateTime::now_utc();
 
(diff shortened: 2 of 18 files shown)


coordinator/src/orderbook/trading.rs
```diff
@@ -127,7 +127,7 @@ pub async fn process_new_order(
     network: Network,
     oracle_pk: XOnlyPublicKey,
 ) -> Result<Order> {
-    tracing::info!(
+    tracing::trace!(
```
luckysori, line 130: 👍 Ah, right. Because of our constant limit orders.

coordinator/migrations/2024-02-09-152833_add_nickname_to_user_table/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE
+    users ADD COLUMN nickname TEXT NOT NULL DEFAULT '';
```
luckysori, line 2: 🔧 Missing newline.

coordinator/src/db/user.rs
```diff
@@ -49,26 +51,59 @@ pub fn by_id(conn: &mut PgConnection, id: String) -> QueryResult<Option<User>> {
 pub fn upsert_user(
     conn: &mut PgConnection,
     trader_id: PublicKey,
-    contact: String,
+    contact: Option<String>,
+    nickname: Option<String>,
 ) -> QueryResult<User> {
+    // If no name or contact has been provided we default to empty string
+    let contact = contact.unwrap_or_default();
+    let nickname = nickname.unwrap_or_default();
```
luckysori, line 59: 🤔 What is the advantage of this?

I think it would be better to model this as a nullable type, so that it's more explicit when the user has decided to not provide one of these.

mobile/native/src/api.rs
```diff
@@ -848,3 +851,12 @@ pub fn delete_dlc_channel(dlc_channel_id: String) -> Result<()> {
 pub fn sync_for_addresses(gap: u32) -> Result<bool> {
     ln_dlc::sync_for_addresses(gap)
 }
+
+pub fn get_new_random_name() -> SyncReturn<String> {
+    SyncReturn(crate::names::get_new_name())
+}
+
+#[tokio::main(flavor = "current_thread")]
+pub async fn update_nickname(nickname: String) -> Result<()> {
```
luckysori, line 860: ❓ On the topic of the nullable nickname, I think FRB should be able to accept an `Option<String>` here, right?

mobile/native/Cargo.toml
```diff
@@ -31,6 +31,7 @@ ln-dlc-storage = { path = "../../crates/ln-dlc-storage" }
 openssl = { version = "0.10.60", features = ["vendored"] }
 orderbook-client = { path = "../../crates/orderbook-client" }
 parking_lot = { version = "0.12.1" }
+petname = "1.1.3"
```
luckysori, line 34: 👍 Haha, great.

mobile/native/src/names.rs
```diff
@@ -0,0 +1,26 @@
+pub fn get_new_name() -> String {
+    let names = petname::Petnames::default();
+    let input = names.generate_one(2, " ");
```
luckysori, line 3: 🔧 I would expect it to be separated with a hyphen!

mobile/native/src/names.rs
```diff
@@ -0,0 +1,26 @@
...
+    if let (Some(first), Some(second)) = (parts.next(), parts.next()) {
+        return format!(
+            "{} {}",
+            uppercase_first_character(first),
+            uppercase_first_character(second)
+        );
+    }
+
+    input.to_string()
+}
```
luckysori, line 18: 🙃 I prefer lowercase!

luckysori (approved):
LGTM, although I would opt to model the contact details and nickname as `Option<String>`.

Is the nickname mandatory? I would maybe make it mandatory and _not_ customisable, to avoid having to moderate any names. The user is just able to roll their nickname as many times as they want.

Also, if we find duplicates in the leader board, we could just append the start of their pubkey, just like Discord does: `fried-otter#a1b7`.

mobile/native/src/names.rs
bonomat: No offense, but my name is 

image

which reads much better with capital letters and a space in between :D

PR #2026, opened 2024-02-13
chore: the maker is dead, long live the maker

bonomat:
If you have been running the maker using `just` then nothing should change for you. 

- `just maker`: starts the maker attached
- `just run-maker-detached`: starts the maker detached

Note the maker image is public at the moment, hence, there is no login required.


holzeis (approved):
Nice 👍

PR #2030, opened 2024-02-13
fix(trade): Spawn trade execution in dedicated task

holzeis:
If the trade execution takes longer than 30 seconds the http request times out and the thread gets killed.

However, the spawned blocking thread to propose the dlc channel is finishing and sending the offer message. That leads to the coordinator sending the dlc message without updating the 10101 meta data (e.g. creating the position, updating order, etc.)

This also applies to the renew and close position flows.

Note, this is not fixing the issue of why proposing a dlc channel would take such a long time, but will fix the issue that the user may end up in a broken state.

https://github.com/get10101/10101/assets/382048/0e15657b-a98c-4dfb-aacd-258b0b1c96e6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fix(validation): Show correct max counterparty collateral in validation message.
 - Fix(mobile): Calculate counterparty balance correctly when checking the validity of trade parameters.
+- Fix(trade): Spawn dedicated tokio task when executing trade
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,12 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fix(validation): Show correct max counterparty collateral in validation message.
+- Fix(trade): Spawn dedicated tokio task when executing trade
```
bonomat, line 11: Nit: I would have mentioned what problem was fixed, not how.

bonomat (approved):
LGTM

PR #2041, opened 2024-02-14
fix(validation): Calculate counterparty balance from channel state

### LINKED ISSUE #2002: Max quantity validation is incorrect
I just tried a trade with the max quantity allowed by the validation (see image below)

image

But run into an error that the coordinator does not have enough funds to collateralise the trade.

image

holzeis:
The balance of the counterparty is the difference from the users balance and the total collateral.

Before we were calculating the max amount based on the initial collateral of the counterparty which can obviously change.

fixes #2002

### DIFF
--- a/mobile/native/src/channel_trade_constraints.rs
+++ b/mobile/native/src/channel_trade_constraints.rs
@@ -45,14 +45,19 @@ pub fn channel_trade_constraints() -> Result<TradeConstraints> {
                 min_margin,
             }
         }
-        Some(channel) => TradeConstraints {
-            max_local_margin_sats: ln_dlc::get_usable_dlc_channel_balance()?.to_sat(),
-            max_counterparty_margin_sats: channel.counter_params.collateral,
-            coordinator_leverage,
-            min_quantity,
-            is_channel_balance: true,
-            min_margin,
-        },
+        Some(_) => {
+            let local_balance = ln_dlc::get_usable_dlc_channel_balance()?.to_sat();
+            let counterparty_balance =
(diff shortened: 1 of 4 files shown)


mobile/native/src/channel_trade_constraints.rs
```diff
@@ -45,14 +45,20 @@ pub fn channel_trade_constraints() -> Result<TradeConstraints> {
...
-            coordinator_leverage,
-            min_quantity,
-            is_channel_balance: true,
-            min_margin,
-        },
+        Some(channel) => {
+            let total_collateral =
+                channel.own_params.collateral + channel.counter_params.collateral;
+            let local_balance = ln_dlc::get_usable_dlc_channel_balance()?.to_sat();
+            let counterparty_balance = total_collateral - local_balance;
```
luckysori, line 52: I like where this is going, but `local_balance` and `counterparty_balance` now mean different things:

- `local_balance` is the usable balance of the app, without the app trade margin (maybe including the fee reserve, I don't remember).
- `counterparty_balance` is the `counter_params.collateral`, plus the app trade margin and the coordinator trade margin (maybe including the fee coordinator reserve too).

luckysori:
@holzeis: I've updated it as discussed.

I documented some of the limitations in the commit message. They also apply to the local balance, so I don't see them as a blocker.

luckysori (commented):
I approve, but I won't merge until @holzeis approves.

luckysori (approved):

holzeis:
> I approve, but I won't merge until @holzeis approves.

Thanks for pushing it over the finish line :)

PR #2047, opened 2024-02-15
feat(wallet): Unreserve locked outpoints if channel gets rejected

holzeis:
Unreserves locked outpoints if a channel gets rejected. This would otherwise only happen if the coordinator gets restarted.

Adds the following commits to rust-dlc

- Set contract to rejected if channel gets cancelled. https://github.com/get10101/rust-dlc/pull/4/commits/12ee7b717a70e2b9da8d15cc919aa3c95543e10e
- Adds the unreserve utxos api to the wallet trait: https://github.com/get10101/rust-dlc/pull/4/commits/38c91beefa2e11938fae74e19593a1e707230be1


bonomat (approved):
LGTM

PR #2048, opened 2024-02-16
Use a fork of `rust-dlc` for now

luckysori:
We will be using our own fork of `rust-dlc` until:
   - We can drop all the LN-DLC features. 
   - `p2pderivatives/rust-dlc#master` includes certain patches currently only found in the LN-DLC branch.

Perhaps we will always want to maintain our own fork, although ideally we don't diverge too much from the original.


holzeis (approved):
Thanks 👍

PR #2054, opened 2024-02-16
fix(channel): use higher fee rate when opening a channel

bonomat:
In the past we set it to  because our feerate estimation was too far off. Now,  would be too low, hence, we switch to  which translates to  from mempool, i.e. it should get confirmed within 1h


holzeis (approved):
LGTM 👍

PR #2056, opened 2024-02-17
fix: Insert channel params before posting new order

### LINKED ISSUE #2055: Failed to post new order: operation timed out
Just happened on regtest. The post order http request timed out. 

```
[DEBUG] TIME: 2024-02-17T06:37:51.714612Z r: Pruning expired orders 
[ERROR] TIME: 2024-02-17T06:37:51.714855Z r: Failed to post new order: error sending request for url (http://34.32.0.52/api/orderbook/orders): operation timed out: operation timed out order_id: 58c11fda-d6ea-4129-a962-875098d5ab65
[DEBUG] TIME: 2024-02-17T06:37:51.715370Z r: lightning::ln::channelmanager: Marking channels with 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802 disconnected and generating channel_updates. 
[DEBUG] TIME: 2024-02-17T06:37:51.715581Z r: Connection lost; reconnecting peer: 03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@34.32.0.52:9045,reconnect_interval: 1s
[INFO] TIME: 2024-02-17T06:37:51.715634Z r: Updated order state new_state: Failed,order_id: 58c11fda-d6ea-4129-a962-875098d5ab65
[ERROR] TIME: 2024-02-17T06:37:51.716063Z Failed to submit order: FfiException(RESULT_ERROR, Failed to post order to orderbook: error sending request for url (http://34.32.0.52/api/orderbook/orders): operation timed out, null)
```

Not sure if this ERROR is part of this bug or something else. However it happened in the same milli second as the post order request timed out.

```
[ERROR] TIME: 2024-02-17T06:37:51.713745Z r: Failed to handle event: Trade request sent to coordinator for order 58c11fda-d6ea-4129-a962-875098d5ab65 failed: Could not load order from db: cannot acquire database connection: timed out waiting for connection 
```

The good news is that this bug doesn't seem to block the user a lot. After the order times out (1 minute). The user is able to trade again. I was able to retry successfully.

holzeis:
I am not exactly sure why this is fixing the issue, but my assumption is the following.

If we receive the order match (async via websockets) before the `post_new_order` http request has finished. We'd still hold an open connection and the write lock on the order.

I presume this leads to the order match processing to fail because of `Could not load order from db: cannot acquire database connection: timed out waiting for connection`. But I am not sure why that would result into the processing of the order match running into a timeout when trying to get a connection.

When the `post_new_order` finally finishes the lock is freed up. I guess afterwards the pool would return connections again but we are already stuck because we do not send the trade request anymore (since processing the match already failed)

Additionally I had to remove the order_id check on `the submit_order_change_notifier`, since it could happen that we don't have the id yet if the order match is already processed before the http post new order request finished. IMHO that should be fine because we only ever have one active order anyways and we don't have to separate them.

fixes #2055

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix(mobile): Calculate counterparty balance correctly when checking the validity of trade parameters.
 - Fix(trade): Spawn dedicated tokio task when executing trade.
 - Chore(wallet): Unreserve locked utxos if offer get rejected.
+- Fix(order): Insert channel-opening parameters before posting order to the coordinator. This prevents the app from running into a race condition between the completion of the HTTP request and the app being notified about the order being matched.
 
 ## [1.8.8] - 2024-02-14
 
--- a/mobile/native/src/trade/order/handler.rs
+++ b/mobile/native/src/trade/order/handler.rs
@@ -18,7 +18,6 @@ use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
-use diesel::Connection;
 use reqwest::Url;
 use time::Duration;
 use time::OffsetDateTime;
@@ -84,40 +83,32 @@ pub async fn submit_order(
         });
     }
 
+    db::insert_order(order.clone()).map_err(SubmitOrderError::Storage)?;
+
+    if let Some(channel_opening_params) = channel_opening_params {
+        let runtime = get_or_create_tokio_runtime().expect("to be able to get runtime");
+        runtime
+            .spawn_blocking({
+                move || {
+                    let mut conn = db::connection()?;
+                    tracing::debug!(
+                        ?channel_opening_params,
+                        "Recording channel-opening parameters"
+                    );
+
+                    db::insert_channel_opening_params(&mut conn, channel_opening_params)
+                }
+            })
+            .await
+            .expect("task to complete")
+            .map_err(SubmitOrderError::Storage)?;
+    }
+
     let url = format!("http://{}", config::get_http_endpoint());
     let url = Url::parse(&url).expect("correct URL");
     let orderbook_client = OrderbookClient::new(url);
 
-    db::insert_order(order.clone()).map_err(SubmitOrderError::Storage)?;
-
-    let runtime = get_or_create_tokio_runtime().expect("to be ablet to get runtime");
-    if let Err(err) = runtime
-        .spawn_blocking({
-            let order = order.clone();
-            move || {
-                let mut db = db::connection()?;
-                // We need to ensure that the channel-opening parameters are inserted after the
-                // order has been posted, but before order has been match and we need them to open
-                // the DLC channel. Without this database transaction we can easily end up with a
-                // match before we insert them!
-                db.transaction(|conn| {
-                    if let Some(channel_opening_params) = channel_opening_params {
-                        tracing::debug!(
-                            ?channel_opening_params,
-                            "Recording channel-opening parameters"
-                        );
-
-                        db::insert_channel_opening_params(conn, channel_opening_params)
-                            .map_err(SubmitOrderError::Storage)?;
-                    }
-
-                    runtime.block_on(orderbook_client.post_new_order(order.into()))
-                })
-            }
-        })
-        .await
-        .expect("task to complete")
-    {
+    if let Err(err) = orderbook_client.post_new_order(order.clone().into()).await {
         let order_id = order.id.clone().to_string();
 
         tracing::error!(order_id, "Failed to post new order: {err:#}");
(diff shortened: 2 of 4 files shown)


mobile/native/src/trade/order/handler.rs
```diff
@@ -90,34 +89,28 @@ pub async fn submit_order(
 
     db::insert_order(order.clone()).map_err(SubmitOrderError::Storage)?;
 
-    let runtime = get_or_create_tokio_runtime().expect("to be ablet to get runtime");
-    if let Err(err) = runtime
-        .spawn_blocking({
```
luckysori, line 95: 📣 We should not block and async task. Removing the `spawn_blocking` is not a good idea.

mobile/native/src/trade/order/handler.rs
```diff
@@ -90,34 +89,28 @@ pub async fn submit_order(
...
-                    }
-
-                    runtime.block_on(orderbook_client.post_new_order(order.into()))
-                })
+    {
+        let mut db = db::connection().map_err(SubmitOrderError::Storage)?;
+        // We need to ensure that the channel-opening parameters are inserted after the
+        // order has been posted, but before order has been match and we need them to open
+        // the DLC channel. Without this database transaction we can easily end up with a
+        // match before we insert them!
```
luckysori, line 97: 🤔 You have changed the code in a way that breaks what this comment is promisinh.

Maybe the original implementation was buggy, but by changing it I think we can now run into the problem outlined here.

In any case, you should delete the comment as part of the same patch (whether it is a good solution or not 😛).

mobile/native/src/trade/order/handler.rs
holzeis: > 📣 We should not block and async task. Removing the `spawn_blocking` is not a good idea.

Can you explain that? It looks weird to me because at the end we are still awaiting the task.

I've removed it because we are not doing it when calling the trade api and I haven't seen any regressions during testing.

The fix should be possible without removing the spawn_blocking. The important part is to not call the post order api before the order has been committed.

luckysori: > Can you explain that? It looks weird to me because at the end we are still awaiting the task.

You can wait for the task to complete, but the important thing is that with `spawn_blocking` the task will be run in a _blocking_ thread pool where blocking is safe.

luckysori: Yep, we have an issue with this kind of thing in several places.

mobile/native/src/trade/order/handler.rs
holzeis: What would happen if we have channel opening params inserted before posting the order?

holzeis:
@luckysori I've addressed your remarks! Please have another look cd1cb10b2fd8b1cfd5fad60aceb278890c2d4b58

mobile/native/src/trade/order/handler.rs
luckysori: I think you are right that it's okay to insert the data _before_ posting the order.

I'm not 100% sure if it could have any effect on attempting to open the channel a second time if the first one fails, so it might be worth checking. But I think it should be okay if this data is truly ephemeral.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix(mobile): Calculate counterparty balance correctly when checking the validity of trade parameters.
 - Fix(trade): Spawn dedicated tokio task when executing trade.
 - Chore(wallet): Unreserve locked utxos if offer get rejected.
+- Fix(order): Commit order before posting order to the coordinator. Preventing the app to run in a race condition where it could happen that the app receives the match before the http request finishes.
```
luckysori, line 14:
```suggestion
- Fix(order): Insert channel-opening parameters before posting order to the coordinator. This prevents the app from running into a race condition between the completion of the HTTP request and the app being notified about the order being matched.
```

mobile/native/src/trade/order/handler.rs
```diff
@@ -84,40 +83,33 @@ pub async fn submit_order(
...
-                            "Recording channel-opening parameters"
-                        );
-
-                        db::insert_channel_opening_params(conn, channel_opening_params)
-                            .map_err(SubmitOrderError::Storage)?;
-                    }
-
-                    runtime.block_on(orderbook_client.post_new_order(order.into()))
-                })
+                let mut conn = db::connection()?;
```
luckysori, line 92: 🔧 Should go within the `if let Some` block.

luckysori (approved):
LGTM. A couple of minor things. Thanks for fixing this bug!

mobile/native/src/trade/order/handler.rs
holzeis: Wouldn't we swallow the error with a `if let Some` ?

PR #2059, opened 2024-02-19
Replace TCP with WebSockets (when feature gate enabled)

Restioson:
This PR replaces all instances of TCP with WebSockets when the `ln_net_native_ws` or `ln_net_wasm_ws` features are enabled. This is step 1 for 10101 WASM :tada: 

The new, default, `ln_net_tcp` carries out the old behaviour. When it is not enabled, the node will not listen for incoming TCP connections. The coordinator supports both TCP and WS connections. WS connections simply connect to its HTTP endpoint at `/` with a Websocket upgrade header.

There is a _potential_ breaking protocol change in how we handle pings/pongs with the orderbook. I have swapped uses of `tungstenite` in transitive dependencies of `native` to the `tokio_tungstenite_wasm` crate, which is a facade over `tungstenite` and WASM websockets (depending on target). Since WASM websockets can't directly intercept `Ping`/`Pong` frames, instead we use `Text` messages with the content `"ping"` and `"pong"`. Additionally, we do the same in the BitMex protocol (which supports this natively as well).

The coordinator's axum websocket implementation is not perfect as it kind of 'lies' to rust-lightning about when the data gets sent exactly. See the SocketDescriptor `send_data` implementation comment for more info. FWIW, this is how Mutiny does it too so hopefully it's okay. This also applies to the `ln_net_native_ws` implementation, but this is only intended for testing anyway and should be removed later down the road.

Because of the large potential impact, I've requested a review from everyone.

To do:
- [x] ~~Support WSS (web socket secure)~~ Probably not needed as we don't use this currently (rust-lightning encrypts everything anyway)
- [x] Fix wasm implementation

### DIFF
--- a/crates/ln-dlc-node/Cargo.toml
+++ b/crates/ln-dlc-node/Cargo.toml
@@ -13,6 +13,7 @@ bdk = { version = "0.28.0", default-features = false, features = ["key-value-db"
 bdk_coin_select = "0.2.0"
 bip39 = { version = "2", features = ["rand_core"] }
 bitcoin = "0.29.2"
+cfg-if = "1.0.0"
 dlc = { version = "0.4.0" }
 dlc-manager = { version = "0.4.0", features = ["use-serde"] }
 dlc-messages = { version = "0.4.0" }
@@ -24,7 +25,6 @@ hkdf = "0.12"
 lightning = { version = "0.0.117", features = ["max_level_trace", "std"] }
 lightning-background-processor = { version = "0.0.117", features = ["futures"] }
 lightning-invoice = { version = "0.25" }
-lightning-net-tokio = { version = "0.0.117" }
 lightning-persister = { version = "0.0.117" }
 lightning-rapid-gossip-sync = { version = "0.0.117" }
 lightning-transaction-sync = { version = "0.0.117", features = ["esplora-blocking"] }
@@ -42,11 +42,19 @@ serde = "1.0.147"
 serde_with = "3.1.0"
 sha2 = "0.10"
 time = "0.3"
-tokio = { version = "1", default-features = false, features = ["io-util", "macros", "rt", "rt-multi-thread", "sync", "net", "time", "tracing"] }
+tokio = { version = "1", default-features = false, features = ["io-util", "macros", "rt", "rt-multi-thread", "sync", "time", "tracing"] }
 tracing = "0.1.37"
 tracing-log = "0.1.3"
 ureq = "2.5.0"
 uuid = { version = "1.3.0", features = ["v4", "serde"] }
+
+axum = { version = "0.6", features = ["ws"], optional = true }
+tokio-tungstenite-wasm = { version = "0.3.0", features = ["native-tls"], optional = true }
+
+# To enable JS support when compiling under wasm
+[dependencies.getrandom]
+version = "*"
+features = ["js"] # Has no effect on other targets
 
 [dev-dependencies]
 time = { version = "0.3", features = ["serde"] }
@@ -54,4 +62,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 uuid = { version = "1.3.0", features = ["v4", "serde"] }
 
 [features]
+default = ["ln_net_tcp"]
 load_tests = []
+ln_net_axum_ws = ["dep:axum"]
+ln_net_ws = ["dep:tokio-tungstenite-wasm"]
+ln_net_tcp = ["tokio/net"]
--- a/crates/ln-dlc-node/src/networking.rs
+++ b/crates/ln-dlc-node/src/networking.rs
@@ -1,0 +1,87 @@
+use crate::node::NodeInfo;
+use futures::future;
+use lightning::ln::peer_handler::APeerManager;
+use lightning::ln::peer_handler::SocketDescriptor;
+use std::future::Future;
+use std::ops::Deref;
+use tracing::debug;
+
+#[cfg(feature = "ln_net_axum_ws")]
+pub mod axum;
+#[cfg(feature = "ln_net_tcp")]
+pub mod tcp;
+#[cfg(feature = "ln_net_ws")]
+mod tungstenite;
+
+#[allow(clippy::diverging_sub_expression, unused_variables, unreachable_code)] // From the panic!() below
+pub async fn connect_outbound<PM: Deref + 'static + Send + Sync + Clone>(
+    peer_manager: PM,
+    peer: NodeInfo,
+) -> Option<impl Future<Output = ()>>
+where
+    PM::Target: APeerManager<Descriptor = DynamicSocketDescriptor>,
+{
+    if peer.is_ws {
+        debug!("Connecting over WS");
+
+        #[cfg(not(feature = "ln_net_ws"))]
+        let ws: Option<future::Either<future::Ready<()>, _>> =
+            panic!("Cannot connect outbound over WS when ln_net_ws is not enabled");
+
+        #[cfg(feature = "ln_net_ws")]
+        let ws = tungstenite::connect_outbound(peer_manager, peer)
+            .await
+            .map(future::Either::Left);
+
+        ws
+    } else {
+        debug!("Connecting over TCP");
+
+        #[cfg(not(feature = "ln_net_tcp"))]
+        let tcp: Option<future::Either<_, future::Ready<()>>> =
+            panic!("Cannot connect outbound over TCP when ln_net_tcp is not enabled");
+
+        #[cfg(feature = "ln_net_tcp")]
+        let tcp = tcp::connect_outbound(peer_manager, peer.pubkey, peer.address)
+            .await
+            .map(future::Either::Right);
+
+        tcp
+    }
+}
+
+/// A dynamic socket descriptor that could either be over WASM (JS) websockets, TCP sockets
+/// (lightning_net_tokio), or Axum websockets.
+#[derive(Hash, Clone, Eq, PartialEq)]
+pub enum DynamicSocketDescriptor {
+    #[cfg(feature = "ln_net_tcp")]
+    Tcp(tcp::SocketDescriptor),
+    #[cfg(feature = "ln_net_axum_ws")]
+    Axum(axum::SocketDescriptor),
+    #[cfg(feature = "ln_net_ws")]
+    Tungstenite(tungstenite::SocketDescriptor),
+}
+
+impl SocketDescriptor for DynamicSocketDescriptor {
+    fn send_data(&mut self, data: &[u8], resume_read: bool) -> usize {
+        match self {
+            #[cfg(feature = "ln_net_tcp")]
+            DynamicSocketDescriptor::Tcp(sock) => sock.send_data(data, resume_read),
+            #[cfg(feature = "ln_net_axum_ws")]
+            DynamicSocketDescriptor::Axum(sock) => sock.send_data(data, resume_read),
+            #[cfg(feature = "ln_net_ws")]
+            DynamicSocketDescriptor::Tungstenite(sock) => sock.send_data(data, resume_read),
+        }
+    }
+
+    fn disconnect_socket(&mut self) {
+        match self {
+            #[cfg(feature = "ln_net_tcp")]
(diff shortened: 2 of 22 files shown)


crates/ln-dlc-node/src/networking/tcp.rs
Restioson, line 1: This is largely a copy of `lightning-net-tokio`, but that is specific to a `PeerManager` with a `lighting_net_tokio::SocketDescriptor` in its type args. This one is for a `PeerManager` with a `DynamicSocketDescriptor`, thus supporting websockets and TCP in the same build (needed for the coordinator).

It is a shame that we have to fork it into our own repo though.

crates/orderbook-client/src/lib.rs
```diff
@@ -95,7 +93,7 @@ async fn subscribe_impl(
                     };
 
                     match msg {
-                        tungstenite::Message::Pong(_) => {
+                        tungstenite::Message::Text(s) if s.eq_ignore_ascii_case("pong") => {
```
Restioson, line 96: This is the aforementioned Pong -> "pong" change which _might_ be breaking (?)

crates/ln-dlc-node/src/networking/wasm.rs
Restioson, line 1: I've tested that this compiles but not that this works yet (we need the rest of `native` to compile first)

crates/ln-dlc-node/src/networking/tungstenite.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state);
+    }
+}
+
+impl peer_handler::SocketDescriptor for SocketDescriptor {
+    fn send_data(&mut self, data: &[u8], resume_read: bool) -> usize {
+        // This isn't so great as we should be waiting for this to be sent before returning the
+        // amount of data sent (which implies that the send operation is done). This is so that the
+        // backpressure stuff works properly
```
Restioson, line 160: Just calling this out as I did in the description so it's easier to find

Restioson:
Ah, it seems that the WASM websocket approach won't work due to the Send bound in BDK.

holzeis (approved):
Very nice 👍

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -178,7 +178,8 @@ pub struct Node<S: TenTenOneStorage, N: Storage> {
     pub dlc_storage: Arc<DlcStorageProvider<S>>,
 
     // fields below are needed only to start the node
-    listen_address: SocketAddr,
+    #[allow(dead_code)]
+    listen_address: SocketAddr, // Irrelevant when using websockets
```
luckysori, line 182: ❓Can we feature-flag this instead or is it too much of a pain?

crates/ln-dlc-node/src/node/mod.rs
```diff
@@ -1070,6 +1078,8 @@ async fn update_rgs_snapshot(
 
 impl Display for NodeInfo {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        format!("{}@{}", self.pubkey, self.address).fmt(f)
+        let scheme = if self.is_ws { "wss" } else { "tcp" };
+
+        format!("{scheme}://{}@{}", self.pubkey, self.address).fmt(f)
```
luckysori, line 1083: ♥️

crates/ln-dlc-node/Cargo.toml
```diff
@@ -42,16 +42,28 @@ serde = "1.0.147"
...
 tracing = "0.1.37"
 tracing-log = "0.1.3"
 ureq = "2.5.0"
 uuid = { version = "1.3.0", features = ["v4", "serde"] }
 
+axum = { version = "0.6", features = ["ws"], optional = true }
+tokio-tungstenite-wasm = { version = "0.3.0", features = ["native-tls"], optional = true }
+
+# To enable JS support when compiling under wasm
+[dependencies.getrandom]
```
luckysori, line 55: For some reason GitHub marks this in red 🤷

crates/ln-dlc-node/src/networking.rs
```diff
@@ -0,0 +1,89 @@
...
+use tracing::debug;
+
+#[cfg(feature = "ln_net_axum_ws")]
+pub mod axum;
+#[cfg(feature = "ln_net_tcp")]
+pub mod tcp;
+#[cfg(feature = "ln_net_ws")]
+mod tungstenite;
+
+// TODO we can let the connect functions be async as well
```
luckysori, line 16: Can you elaborate on this? I don't get what you mean.

crates/ln-dlc-node/src/networking.rs
```diff
@@ -0,0 +1,89 @@
...
+            panic!("Cannot connect outbound over TCP when ln_net_tcp is not enabled");
+
+        #[cfg(feature = "ln_net_tcp")]
+        let tcp = tcp::connect_outbound(peer_manager, peer.pubkey, peer.address)
+            .await
+            .map(future::Either::Right);
+
+        tcp
+    }
+}
```
luckysori, line 53: 🔧 I think you should be able to get rid of all of the `allow`s if you use the `cfg!` macro. You could handle all the valid combinations explicitly and panic for all the invalid cases in a single `else` branch.

crates/ln-dlc-node/src/networking/tungstenite.rs
```diff
@@ -0,0 +1,169 @@
...
+    PM::Target: APeerManager<Descriptor = DynamicSocketDescriptor>,
+{
+    let url = &format!(
+        "ws://{}:{}",
+        node_info.address.ip(),
+        node_info.address.port()
+    );
+    let mut ws = tokio_tungstenite_wasm::connect(url)
+        .await
+        .expect("Can't connect");
```
luckysori, line 40: ❓Is `expect`ing here dangerous?

crates/ln-dlc-node/src/networking/tungstenite.rs
```diff
@@ -0,0 +1,169 @@
...
+    PM::Target: APeerManager<Descriptor = DynamicSocketDescriptor>,
+{
+    let ws_next = if *emit_read_events {
+        Either::Left(ws.next())
+    } else {
+        Either::Right(future::pending())
+    };
+
+    tokio::select! {
+        task_msg = task_rx.recv() => match task_msg.context("rust-lighting SocketDescriptor dropped")? {
```
luckysori, line 104: Classic Lightning typo 😛 

```suggestion
        task_msg = task_rx.recv() => match task_msg.context("rust-lightning SocketDescriptor dropped")? {
```

crates/ln-dlc-node/src/networking/tungstenite.rs
luckysori: Do you have any idea how we could tackle this limitation? Can we just add another channel so that the other task can let us know when the data has been sent?

luckysori (approved):
LGTM, although I'm not the best reviewer of networking code.

It would be cool if we could get rid of some of those `allow`s.

crates/ln-dlc-node/src/networking/tungstenite.rs
Restioson: It's a little more complicated than that. If we don't send all the data when calling `send_data` then there is a function we need to call of the peer manager which results in `send_data` being called again. At first glance you'd think you can just make the 2nd a no-op, but there could be more data that's waiting to be sent by then. Therefore we need to keep track of how much we promised to send and how much extra must be sent.

Definitely possible to do this, just more work

crates/ln-dlc-node/src/node/mod.rs
Restioson: We could, but I thought that perhaps we may want to use it somewhere else (?). If not I can definitely get rid of it

crates/ln-dlc-node/src/node/mod.rs
Restioson: I've realized this is actually technically wrong and should be changed to just `ws` :sweat_smile:

crates/ln-dlc-node/Cargo.toml
Restioson: Not sure why... apparently google says that it could be because of a syntax error somewhere, but it does seem to be correct

crates/ln-dlc-node/src/networking.rs
Restioson: Oops, this is a fossil of a previous attempt. Good catch, thanks!

crates/ln-dlc-node/src/networking.rs
Restioson: I don't think we can use the cfg! macro here, as that resolves to a runtime bool. Hence, we'd end up with a call to `tcp::connect_outbound` still in the final code, which would not exist if the `ln_net_tcp` feature were not enabled.

I tried with the `cfg_if` crate but it doesn't support expression returns so regardless it's kind of the same.

However, one way we may be able to solve it is by explicit `return` statements sprinkled around, if you'd prefer?

crates/ln-dlc-node/src/networking/tungstenite.rs
Restioson: Ah thanks, this is one of the last two `expect`s that I missed

crates/ln-dlc-node/src/networking/tungstenite.rs
Restioson: Oops :sweat_smile:

crates/ln-dlc-node/src/networking.rs
Restioson: I tried this
```rs
    if peer.is_ws {
        debug!("Connecting over WS");

        #[cfg(not(feature = "ln_net_ws"))]
        return panic!("Cannot connect outbound over WS when ln_net_ws is not enabled");

        #[cfg(feature = "ln_net_ws")]
        tungstenite::connect_outbound(peer_manager, peer)
            .await
            .map(future::Either::Left)
    } else {
        debug!("Connecting over TCP");

        #[cfg(not(feature = "ln_net_tcp"))]
        return panic!("Cannot connect outbound over TCP when ln_net_tcp is not enabled");

        #[cfg(feature = "ln_net_tcp")]
        tcp::connect_outbound(peer_manager, peer.pubkey, peer.address)
            .await
            .map(future::Either::Right)
    }
    ```
    which is pleasing but doesn't give Rust enough info to infer the type of `Either` - which we can't know unless we know the features that are enabled, which is a bit of a pain.

crates/ln-dlc-node/src/node/mod.rs
Restioson: Fixed

crates/ln-dlc-node/Cargo.toml
Restioson: I would just ignore this personally. It works and also validates using an online toml linter (https://www.toml-lint.com/)

crates/orderbook-client/src/lib.rs
Restioson: I don't think it should be as it doesn't seem that we ever actually use Ping/Pong with the orderbook

crates/ln-dlc-node/src/networking.rs
luckysori: Okay, thanks for giving it a go! We can leave it. If I find time after we merge this I will play around with it and report back.

crates/ln-dlc-node/src/networking/tungstenite.rs
luckysori: Thanks for explaining! As discussed, we don't have to change this until we run into problems.

crates/ln-dlc-node/src/node/mod.rs
luckysori: Not a big deal, but I think this can be safely removed (from the version that doesn't use it).

luckysori:
> @luckysori just checking if we are ready for merge on this?

Sure thing! I will add it to the merge queue.

PR #2062, opened 2024-02-19
Display more fee information when opening a DLC channel

### LINKED ISSUE #2042: Include transaction fee info when opening a DLC channel
Opening a DLC channel involves a few transaction fees:

- Funding transaction fee (paid immediately when the channel is opened).
- Buffer transaction fee reserve (reserved until the channel is force-closed).
- CET or refund transaction fee reserve (reserved until the channel is force-closed).

These fees are evenly split between the two parties. At the moment, this information is not displayed in the app's UI when we are opening a new DLC channel.

The user needs to know the fees they are going to pay to decide if they want to open the channel and trade or not.

Before opening, the fees that we can display are only an estimate based on the current fee rate and the expected weights of the fund transaction, buffer transaction and CET (or refund transaction).

After the channel has been opened, we can display the actual transaction fees: how much we already paid for the funding transaction (half of the total displayed in a block explorer) plus the reserved fees for the buffer transaction and the CET or refund transaction.

luckysori:
Fixes #2042.

https://github.com/get10101/10101/assets/9418575/9976d4d2-155b-4c12-8be3-8ec65530b5e1

_The video does not correspond to the latest version, but it's close enough._

---

See https://github.com/get10101/rust-dlc/pull/5 for the `rust-dlc` side of things.

### DIFF
--- a/crates/ln-dlc-node/src/tests/dlc_channel.rs
+++ b/crates/ln-dlc-node/src/tests/dlc_channel.rs
@@ -260,39 +260,56 @@ async fn can_open_and_force_close_channel() {
 
 #[tokio::test(flavor = "multi_thread")]
 #[ignore]
-async fn can_open_channel_with_min_inputs() {
+async fn funding_transaction_pays_expected_fees() {
     init_tracing();
+
+    // Arrange
 
     let app_dlc_collateral = Amount::from_sat(10_000);
     let coordinator_dlc_collateral = Amount::from_sat(10_000);
 
-    // We must fix the fee rate so that we can predict how many sats `rust-dlc` will allocate
-    // for transaction fees.
-    let fee_rate_sats_per_vbyte = 2;
-    let expected_fund_tx_fee = 252 * fee_rate_sats_per_vbyte;
+    let fee_rate_sats_per_vb = 2;
 
-    // This also depends on the fee rate, but the formula is a bit more involved.
-    let fee_reserve = 880;
-
-    // Fee costs are evenly split.
-    let fee_cost_per_party = (expected_fund_tx_fee + fee_reserve) / 2;
-    let fee_cost_per_party = Amount::from_sat(fee_cost_per_party);
-
-    let (app, _running_app) = start_and_fund_app(app_dlc_collateral + fee_cost_per_party, 1).await;
+    // Give enough funds to app and coordinator so that each party can have their own change output.
+    // This is not currently enforced by `rust-dlc`, but it will be in the near future:
+    // https://github.com/p2pderivatives/rust-dlc/pull/152.
+    let (app, _running_app) = start_and_fund_app(app_dlc_collateral * 2, 1).await;
     let (coordinator, _running_coordinator) =
-        start_and_fund_coordinator(coordinator_dlc_collateral + fee_cost_per_party, 1).await;
+        start_and_fund_coordinator(app_dlc_collateral * 2, 1).await;
+
+    // Act
 
     let (app_signed_channel, _) = open_channel_and_position(
         app.clone(),
         coordinator.clone(),
         app_dlc_collateral,
         coordinator_dlc_collateral,
-        Some(fee_rate_sats_per_vbyte),
+        Some(fee_rate_sats_per_vb),
     )
     .await;
 
-    // No change output means that the inputs were spent in full by the fund output.
-    assert!(app_signed_channel.fund_tx.output.len() == 1);
+    // Assert
+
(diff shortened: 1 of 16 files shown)


crates/ln-dlc-node/src/tests/dlc_channel.rs
```diff
@@ -305,6 +305,75 @@ async fn funding_transaction_pays_expected_fees() {
...
+    .await;
+
+    let fund_output_vout = app_signed_channel.fund_output_index;
+    let fund_output_amount = &app_signed_channel.fund_tx.output[fund_output_vout].value;
+
+    // We cannot easily assert equality because both `rust-dlc` and us have to round in several
+    // spots.
+    let epsilon = *fund_output_amount as i64 - expected_fund_output_amount.to_sat() as i64;
+
+    assert!(
```
bonomat, line 346: Should this test really panic? If it panics, why do we need an assert, or is the assert panicking?

crates/ln-dlc-node/src/tests/dlc_channel.rs
```diff
@@ -307,7 +307,6 @@ async fn funding_transaction_pays_expected_fees() {
 
 #[tokio::test(flavor = "multi_thread")]
 #[ignore]
-#[should_panic]
```
bonomat: ah, got it, you made the test fail first, then fixed it, very nice :)

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -1193,6 +1194,24 @@ pub fn is_dlc_channel_confirmed() -> Result<bool> {
     node.inner.is_dlc_channel_confirmed(&dlc_channel.channel_id)
 }
 
+pub fn estimated_fee_reserve() -> Result<Amount> {
+    let node = state::get_node();
+
+    // TODO: Here we assume that the coordinator will use the same confirmation target AND that
```
bonomat, line 1202: I don't think this needs to be a TODO, it can just be a note given this function is just estimating.

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -512,9 +512,10 @@ fn keep_wallet_balance_and_history_up_to_date(node: &Node) -> Result<()> {
                     status,
                     wallet_type: WalletHistoryItemType::DlcChannelFunding {
                         funding_txid: details.txid.to_string(),
-                        // this is not 100% correct as fees are not exactly divided by 2. The fee a
-                        // user has to pay depends on his final address.
-                        reserved_fee_sats: details.fee.map(|fee| fee / 2),
+                        // this is not 100% correct as fees are not exactly divided by 2. The share
```
bonomat, line 515: Maybe we should just fix this in rust-dlc and say the fee is always 50:50

justfile
```diff
@@ -289,10 +289,19 @@ coordinator args="":
 
     cargo run --bin coordinator -- {{args}}
 
+run_maker_args := if os() == "linux" {
```
bonomat, line 292: Nice, alternatively you could use 

```
docker run -d --name maker ghcr.io/get10101/aristides/aristides:main regtest --orderbook http://172.17.0.1:8000
```

It seems that on Linux `172.17.0.1` is always `localhost`

That being said, can you update the CI file to use this api as well?

https://github.com/holzeis/10101/blob/d29baf6acdfa262aa0c7e1b84ecc581e036f9b8b/.github/workflows/ci.yml#L276

bonomat (approved):
Looks good.

mobile/native/src/ln_dlc/mod.rs
luckysori: Hmm, I don't think it should be fixed there, because it wouldn't make sense to me to charge the user more just because we as the coordinator bring 7 inputs or something.

I think it would be nicer to record all the fee contributions in a channel metadata table. I think this is something we should be able to achieve after the new model is introduced.

luckysori:
I will keep this open until #2094 merges.

mobile/lib/features/trade/channel_configuration.dart
```diff
@@ -48,7 +50,7 @@ channelConfiguration({
                 },
                 child: SingleChildScrollView(
                   child: SizedBox(
-                    height: 450,
+                    height: 500,
```
luckysori, line 53: I've had to bump this number even further because otherwise the unit tests fail 🙈

holzeis: Noo 🫣 I worked so long to get the tests pass with the correct size 😅

Now it looks weird again 🙈

PR #2063, opened 2024-02-19
fix: relaxed constraint on failed orders

### LINKED ISSUE #2031: On my android the app crashes after upgrading to 1.8.7

bonomat:
So far, a failed order needed a reason when converting between types. However, we failed to do so when setting orders to failed. I've relaxed this constraint now. If there is no reason in the db, we assume an `Unknown` reason in the frontend. At the same time, we ensure that we set a reason when setting all orders to failed

Resolves #2031 

We should roll out a release after this is merged as it blocks at least one user and myself.


holzeis (approved):
LGTM 👍

Issue #1989, opened 2024-02-08
Rename ESPLORA_ENDPOINT to ELECTRS_ENDPOINT


holzeis:
Note, while working on this is noticed that we have a confic ESPLORA_ENDPOINT, which is actually the ELECTRS_ENDPOINT. We should clean that up.

_Originally posted by @holzeis in https://github.com/get10101/10101/issues/1988#issuecomment-1934610327_

PR #2065, opened 2024-02-20
Use Electrs name instead of Esplora where relevant

luckysori:
Fixes https://github.com/get10101/10101/issues/1989.


bonomat (approved):
Note that we will need to update our test and production environment.

bonomat:
Sorry, I've removed this from the merge queue to get a quick release out. Will re-add afterwards.

PR #2066, opened 2024-02-20
fix: use platform dependent cache

bonomat:
I believe our CI issue is coming from the fact that we use two different macos versions but cache flutter for the same key


bonomat:
Merged right away to get this in

PR #2068, opened 2024-02-20
Simplify cache

bonomat:
Using default cache settings will stop us from caching cargo binaries. This might slow down our production build but hopefully gets rid of random errors where libraries were already installed.


luckysori (approved):
[`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall) might be interesting here.

PR #2074, opened 2024-02-20
Webapp: Allow user to close/force-close channel

bonomat:
resolves [#2061](https://github.com/get10101/10101/issues/2061)

Note: the error in the video is already fixed.

https://github.com/get10101/10101/assets/224613/c6fdb9c1-62c4-4942-ae90-defdfe714425


holzeis:
Minor findings:

- force closing should not be possible if the channel is not yet confirmed. (the associated contract needs to be confirmed (on-chain) before force-closing is possible)
- force closing button should be removed after the channel is already in closing

holzeis (approved):

PR #2077, opened 2024-02-21
Fix backwards-incompatible CLI change

luckysori:
Patch https://github.com/get10101/10101/commit/cd04992884b88d911c04b48c075e4c23883cfa0f changed the name of the `esplora` CLI argument to `electrs` for correctness. But the change was introduced in a backwards-incompatible manner.

This patch solves that issue by introducing an alias for the CLI argument with the original name.

All this applies to both the coordinator and the webapp.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4863,7 +4863,7 @@ dependencies = [
 
 [[package]]
 name = "webapp"
-version = "0.2.0"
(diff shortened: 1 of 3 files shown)


luckysori:
Tested locally and it works. Merging to be able to use this ASAP.

PR #2079, opened 2024-02-21
Let app force-close after channel renewal

luckysori:
See https://github.com/get10101/rust-dlc/commit/0191dc4.


bonomat (approved):
Thank master of rust-dlc :)

PR #2080, opened 2024-02-21
fix: Add a holistic check if the dlc channel is confirmed.

holzeis:


bonomat (approved):
Looks good. Thanks.

luckysori (commented):
I think the fix makes sense if `is_dlc_channel_confirmed` is actually checking if the DLC channel is confirmed (including when the DLC channel has been closed).

holzeis:
> I think the fix makes sense if `is_dlc_channel_confirmed` is actually checking if the DLC channel is confirmed (including when the DLC channel has been closed).

Yeah, I had it differently before, but i think @bonomat is right with his remark that any other logic would not match the function name.

Anyhow, I think this code will get obsolete once we introduce our dlc channel representation, which will hold that information directly on the channel.

IMO, having to go over the contract to check if the funding tx got confirmed is a bit weird.

PR #2087, opened 2024-02-22
feat: sign orders and whitelist makers

bonomat:
only whitelisted makers are allowed to post limit orders

Note: this is a breaking change: 
- apps which do not upgrade, cannot post orders --> I think this is a feature and not a bug :)
- our old maker won't work anymore --> we need to deploy our new maker in production

### DIFF
--- a/crates/commons/src/order.rs
+++ b/crates/commons/src/order.rs
@@ -1,4 +1,8 @@
+use anyhow::Result;
+use bitcoin::hashes::sha256;
 use rust_decimal::Decimal;
+use secp256k1::ecdsa::Signature;
+use secp256k1::Message;
 use secp256k1::PublicKey;
 use serde::Deserialize;
 use serde::Serialize;
@@ -6,6 +10,21 @@ use time::OffsetDateTime;
 use trade::ContractSymbol;
 use trade::Direction;
 use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct NewOrderRequest {
+    pub value: NewOrder,
+    /// A signature of the sha256 of [`value`]
+    pub signature: Signature,
+}
+
+impl NewOrderRequest {
+    pub fn verify(&self) -> Result<()> {
+        let message = self.value.message();
+        self.signature.verify(&message, &self.value.trader_id)?;
+        Ok(())
+    }
+}
 
 #[derive(Serialize, Deserialize, Clone)]
 pub struct NewOrder {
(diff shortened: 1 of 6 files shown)


holzeis:
> Note: this is a breaking change:

we should release 1.9.0 then :)

crates/commons/src/order.rs
```diff
@@ -7,6 +11,23 @@ use trade::ContractSymbol;
...
+    /// A signature of the sha256 of [`value`]
+    pub signature: Signature,
+    /// The public key of the sending party
+    pub public_key: PublicKey,
+}
+
+impl NewOrderRequest {
+    pub fn verify(&self) -> Result<()> {
+        let message = self.value.message();
+        self.signature.verify(&message, &self.public_key)?;
```
holzeis, line 26: 🤔 That doesn't make much sense does it?

Any body could send any message signed with the public key that they provide. We need to verify the signature on the node id of the user. This information can be found in the `trader_id` field.

mobile/native/src/trade/order/orderbook_client.rs
```diff
@@ -15,10 +18,19 @@ impl OrderbookClient {
     }
 
     pub(crate) async fn post_new_order(&self, order: NewOrder) -> Result<OrderResponse> {
+        let secret_key = get_node_key();
+        let message = order.message();
+        let signature = secret_key.sign_ecdsa(message);
+        let new_order_request = NewOrderRequest {
+            value: order,
+            signature,
+            public_key: secret_key.public_key(SECP256K1),
```
holzeis, line 27: 🔧 why not use `order.trader_id`?

mobile/native/src/trade/order/orderbook_client.rs
```diff
@@ -15,10 +18,19 @@ impl OrderbookClient {
     }
 
     pub(crate) async fn post_new_order(&self, order: NewOrder) -> Result<OrderResponse> {
+        let secret_key = get_node_key();
+        let message = order.message();
+        let signature = secret_key.sign_ecdsa(message);
```
holzeis, line 23: 🔧 It would probably reasonable to put the signing into a function and do not leak the secret_key just like that. Have a look at the `AesCipher`.

holzeis (changes requested):

crates/commons/src/order.rs
bonomat: wups,that's the reason why we have PR reviews I guess 🙈

mobile/native/src/trade/order/orderbook_client.rs
bonomat: Where is the advantage? The `AesCipher` needs to be initialized using the `secret_key`, meaning, we just move this to a different place.

holzeis (approved):
LGTM 👍

mobile/native/src/trade/order/orderbook_client.rs
bonomat: @luckysori : I'm keen on hearing your thoughts on this and happily will change it in a follow-up PR.

PR #2088, opened 2024-02-22
fix: Add more resilience if async order has already been seen.

holzeis:
Submitting a trade may fail for many reasons. Before that patch the users expired position would get stuck because the app wouldn't accept another match with the same order id.

This patch adds some more resilience around that and handles the case where the async order might have already been stored to the database.


mobile/native/src/db/mod.rs
```diff
@@ -167,11 +167,16 @@ pub fn update_order_state(
     Ok(order.try_into()?)
 }
 
-pub fn get_order(order_id: Uuid) -> Result<trade::order::Order> {
+pub fn get_order(order_id: Uuid) -> Result<Option<trade::order::Order>> {
     let mut db = connection()?;
     let order = Order::get(order_id.to_string(), &mut db)?;
 
-    Ok(order.try_into()?)
+    let order = match order {
```
bonomat, line 174: Nit. you could do `order.map(|..|)...`

bonomat (approved):
LGTM

mobile/native/src/db/mod.rs
holzeis: Doesn't work because it's a `try_into`

Issue #2075, opened 2024-02-20
App still is crashing now after upgrading to 1.8.11 #2031 and only happens with my original restore phrase 


wksantiago:
This is an update to closed issue #2031 

App still is shutting down now after upgrading to v1.8.11 #2031 and only happens with my original restore phrase.

I created a new phrase and the app start OK no problem.  I clear storage and cache and restore from original phrase and then goes form:

Welcome 10101 to
...
Starting 10101

Then it just shuts down.

holzeis:
Thanks for opening another issue. 

Does the state that you are trying to restore includes a channel?

wksantiago:
The original a few weeks ago yes I had a channel for the stable token.

and I am running the apk form github

bonomat:
Thanks for letting us know. 
In case you know your way around in a terminal, could you connect your phone and run`adb logcat` (https://developer.android.com/tools/adb) and send us the logs. This would be tremendously helpful for debugging your case.

bonomat:
Thank you for this. 
This is the critical part: 

```
02-22 16:05:39.380 29178 29221 I flutter : [ERROR] TIME: 2024-02-22T20:05:39.380657Z Launching the app failed FfiException(RESULT_ERROR, Failed to start the backend
02-22 16:05:39.380 29178 29221 I flutter :
02-22 16:05:39.380 29178 29221 I flutter : Caused by:
02-22 16:05:39.380 29178 29221 I flutter :     0: Failed to get wallet balances
02-22 16:05:39.380 29178 29221 I flutter :     1: Could not find contract associated with channel to compute usable balance, null)
02-22 16:05:39.381 29178 29221 I flutter : [DEBUG] TIME: 2024-02-22T20:05:39.381026Z r: Got all spendable outputs outputs: []
02-22 16:05:39.389 29178 29221 I flutter : [DEBUG] TIME: 2024-02-22T20:05:39.389418Z r: Got all transactions transactions: []
02-22 16:05:39.389 29178 29221 I flutter : [DEBUG] TIME: 2024-02-22T20:05:39.389750Z r: Syncing 0 shadow transactions
```

bonomat:
We will look into it, I've prioritized it for this week.

holzeis:
The issue resulted out of a missing contract backup which fails wallet sync at startup. This let the app not start properly and crashed the app.

[...]

PR #2090, opened 2024-02-23
fix: Handle wallet sync error during startup

holzeis:
Let the app start even if the wallet sync fails.

If the wallet sync fails the balance will be set to `None`, telling the UI that we were unable to determine the balance.

fixes: #2075

### DIFF
--- a/mobile/lib/features/wallet/balance_row.dart
+++ b/mobile/lib/features/wallet/balance_row.dart
@@ -18,28 +18,35 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
   Widget build(BuildContext context) {
     WalletChangeNotifier walletChangeNotifier = context.watch<WalletChangeNotifier>();
 
+    final offchainBalance = walletChangeNotifier.offChain()?.formatted() ?? "n/a";
+    final onchainBalance = walletChangeNotifier.onChain()?.formatted() ?? "n/a";
+
     final amountText = switch (widget.walletType) {
       WalletType.lightning => Row(
             mainAxisAlignment: MainAxisAlignment.center,
             crossAxisAlignment: CrossAxisAlignment.center,
             children: [
-              Text(walletChangeNotifier.offChain().formatted(),
+              Text(offchainBalance,
                   style: const TextStyle(
                       fontSize: 30, color: Colors.white, fontWeight: FontWeight.bold)),
-              const Text(" sats",
-                  style:
-                      TextStyle(fontSize: 14, color: Colors.white, fontWeight: FontWeight.normal))
+              walletChangeNotifier.offChain() != null
+                  ? const Text(" sats",
+                      style: TextStyle(
+                          fontSize: 14, color: Colors.white, fontWeight: FontWeight.normal))
+                  : Container()
             ]),
       WalletType.onChain => Row(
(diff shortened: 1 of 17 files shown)


mobile/native/src/ln_dlc/node.rs
```diff
@@ -94,15 +94,27 @@ impl Node {
         self.inner.get_blockchain_height()
     }
 
-    pub fn get_wallet_balances(&self) -> Result<Balances> {
-        let on_chain = self.inner.get_on_chain_balance()?.confirmed;
+    pub fn get_wallet_balances(&self) -> Balances {
+        let on_chain = match self.inner.get_on_chain_balance() {
+            Ok(on_chain) => on_chain.confirmed,
+            Err(e) => {
+                tracing::error!("Failed to get onchain balance. {e:#}");
```
bonomat, line 101: I'm not sure this is a good idea. The user will have no idea that the sync failed and will only see a balance of 0. Imho we should catch this error in the frontend and at least show the user an error message with which he can come to us

holzeis: Ok, I will think of something.

holzeis: 179a336a5aa59c3525ca384f73297cb10bd81a3a

mobile/lib/features/wallet/balance_row.dart
```diff
@@ -18,28 +18,35 @@ class _BalanceRowState extends State<BalanceRow> with SingleTickerProviderStateM
   Widget build(BuildContext context) {
     WalletChangeNotifier walletChangeNotifier = context.watch<WalletChangeNotifier>();
 
+    final offchainBalance = walletChangeNotifier.offChain()?.formatted() ?? "n/a";
```
luckysori, line 21: 👍 Better than a zero or nothing at all.

I wonder if we should eventually use skeletons during loading and just display an error pop-up if some key data cannot be loaded.

luckysori (approved):

mobile/native/src/ln_dlc/mod.rs
```diff
@@ -369,10 +369,14 @@ pub fn run(seed_dir: String, runtime: &Runtime) -> Result<()> {
         // TODO: This might not be necessary once we rewrite the on-chain wallet with bdk:1.0.0.
```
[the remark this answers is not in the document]
holzeis: We will probably not need it for the on-chain sync. But the off-chain balance will definitely fail in regards to the linked issue, since the contract was not backed up breaking the restored app.

mobile/lib/features/wallet/balance_row.dart
holzeis: > I wonder if we should eventually use skeletons during loading and just display an error pop-up if some key data cannot be loaded.

Can you elaborate? I am not sure I completely follow.

luckysori: I was just thinking about a nicer way to display information that is not yet available, and also information that we have failed to load.

In this case, showing a skeleton until the balance is available would be cool. But I'm not sure what would be the nicest way to display the fact that the balance could not be loaded. The label `n/a` is serviceable, but probably not the nicest.

PR #2091, opened 2024-02-23
chore(coordinator): set expired orders to expired

bonomat:
this will allow us to filter failed from expired orders in the db


holzeis (approved):
LGTM 👍

PR #2092, opened 2024-02-23
feat: pending balance is good balance

### LINKED ISSUE #2053: Count trusted-pending as on-chain balance
Trusted-pending are change outputs and can be spent immediately. We should count them as spendable balance.

~This is somewhat cumbersome and blocking me atm to bump fees of a stuck transaction.~

bonomat:
Resolves #2053


holzeis (approved):
Thanks

PR #2099, opened 2024-02-26
Feat: Add closing tx

holzeis:
Adds the closing txid to the dlc channels api and the channels screen in the settings (see screenshot below).

depends on https://github.com/get10101/rust-dlc/pull/13

Note this is a breaking change. Old closed channels can't be read anymore. But given that old closed channel state don't provide any valuable information, I think this is acceptable.

image

### DIFF
--- a/coordinator/src/collaborative_revert.rs
+++ b/coordinator/src/collaborative_revert.rs
@@ -240,6 +240,7 @@ pub fn confirm_collaborative_revert(
             temporary_channel_id: signed_channel.temporary_channel_id,
             channel_id: signed_channel.channel_id,
             reference_id: None,
+            closing_txid: revert_transaction.txid(),
         }),
         // The contract doesn't matter anymore.
         None,
--- a/crates/ln-dlc-node/src/ln/dlc_channel_details.rs
+++ b/crates/ln-dlc-node/src/ln/dlc_channel_details.rs
@@ -1,5 +1,6 @@
 use bitcoin::hashes::hex::ToHex;
 use bitcoin::secp256k1::PublicKey;
+use dlc_manager::channel::signed_channel::SignedChannel;
 use dlc_manager::channel::Channel;
 use dlc_manager::DlcChannelId;
 use serde::Serialize;
@@ -17,6 +18,7 @@ pub struct DlcChannelDetails {
     pub fee_rate_per_vb: Option<u64>,
     pub funding_txid: Option<String>,
     pub funding_tx_vout: Option<usize>,
+    pub closing_txid: Option<String>,
 }
 
 #[derive(Serialize, Debug, Ord, PartialOrd, Eq, PartialEq)]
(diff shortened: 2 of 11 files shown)


coordinator/src/collaborative_revert.rs
```diff
@@ -239,6 +239,8 @@ pub fn confirm_collaborative_revert(
             counter_party: signed_channel.counter_party,
             temporary_channel_id: signed_channel.temporary_channel_id,
             channel_id: signed_channel.channel_id,
+            reference_id: None,
+            closing_txid: revert_transaction.txid(),
```
luckysori, line 243: 👍 Nice!

crates/ln-dlc-node/src/ln/dlc_channel_details.rs
```diff
@@ -52,16 +54,44 @@ pub enum ChannelState {
...
             match channel.clone() {
+                Channel::Signed(SignedChannel {
+                    update_idx,
+                    fee_rate_per_vb,
+                    fund_tx,
+                    fund_output_index,
+                    state: dlc_manager::channel::signed_channel::SignedChannelState::CollaborativeCloseOffered{close_tx, ..},
+                    ..
+                }) => (
+                    Some(update_idx), Some(SignedChannelState::CollaborativeCloseOffered), Some(fee_rate_per_vb), Some(fund_tx.txid().to_hex()), Some(fund_output_index), Some(close_tx.txid().to_hex())),
```
luckysori, line 67: 🔧 Doesn't `to_string` work too? There is no `ToHex` in `bitcoin:0.30`, so I would avoid using it.

crates/ln-dlc-node/src/ln/dlc_channel_details.rs
```diff
@@ -52,16 +54,44 @@ pub enum ChannelState {
...
     fn from(channel: Channel) -> Self {
-        let (update_idx, state, fee_rate_per_vb, funding_txid, funding_tx_vout) =
+        let (update_idx, state, fee_rate_per_vb, funding_txid, funding_tx_vout, closing_txid) =
             match channel.clone() {
+                Channel::Signed(SignedChannel {
+                    update_idx,
+                    fee_rate_per_vb,
+                    fund_tx,
+                    fund_output_index,
+                    state: dlc_manager::channel::signed_channel::SignedChannelState::CollaborativeCloseOffered{close_tx, ..},
```
luckysori, line 64: 🤔 Formatting seems off here.

mobile/lib/common/settings/channel_screen.dart
```diff
@@ -215,7 +222,12 @@ class ChannelsTile extends StatelessWidget {
                     ? ListTile(
                         leading: const Text('Buffer TxId', style: TextStyle(fontSize: 17)),
                         title: TransactionIdText(channel.bufferTxid))
-                    : Container()
+                    : Container(),
+                channel is ClosedDlcChannel
+                    ? ListTile(
+                        leading: const Text('Closing TxId', style: TextStyle(fontSize: 17)),
```
luckysori, line 228: 🙃 I would prefer "TXID".

luckysori: 🔧 It would also be nice to fix the "Dlc" above. That one should be "DLC".

luckysori (approved):
Nice, thank you!

bonomat (approved):
Can you update the webapp as well please, I think this can come very handy there :)

Issue #2076, opened 2024-02-20
Share logs is not clickable if installed on macos from testflight


holzeis:

luckysori:
@luckysori to reproduce on Linux.

PR #2111, opened 2024-02-28
Show log file location on desktop

luckysori:
Fix https://github.com/get10101/10101/issues/2076.

share-logs-desktop

The share button wasn't working on desktop anyway. On Linux it is just not supported. On Mac we heard that it did not work.

Desktop users (not officially supported) can now copy the log file location, find the log file using a file explorer, and share it with us.


holzeis (approved):
LGTM 👍

PR #2113, opened 2024-02-28
chore(app): Reduce channel config modal height to match order modal height

holzeis:
Screenshot 2024-02-28 at 11 00 10


bonomat (approved):
👍

PR #2121, opened 2024-02-29
chore(ci): Create pre-releases with generated release notes

holzeis:
As discussed in our last demo. We can skip creating manual change log entries if we are creating proper pull request titles.


bonomat (approved):
Nice

PR #2122, opened 2024-02-29
chore(app): Rework some styling changes after the flutter upgrade

holzeis:
|                                       |                                       |                                       |
|:-------------------------------------:|:-------------------------------------:|:-------------------------------------:|
| image | image | image


holzeis:
Note, I didn't manage to set the background of the candlestick chart back to white, but we might want to consider replacing that dependency altogether, since it doesn't seem to be maintained anymore.

luckysori (approved):
Love it, thanks!

PR #2124, opened 2024-02-29
chore(app): Remind about minimum channel collateral when receiving on chain funds

holzeis:
image

fixes https://github.com/get10101/meta/issues/362

### DIFF
--- a/mobile/lib/features/wallet/receive_screen.dart
+++ b/mobile/lib/features/wallet/receive_screen.dart
@@ -238,7 +238,7 @@ class BitcoinAddress extends StatelessWidget {
             });
           },
           child: Container(
-            padding: const EdgeInsets.all(20),
+            padding: const EdgeInsets.only(left: 20, right: 20, bottom: 15),
             child: Opacity(
                 opacity: 1.0,
                 child: Column(crossAxisAlignment: CrossAxisAlignment.start, children: [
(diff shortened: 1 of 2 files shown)


mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -470,7 +470,8 @@ class SelectableButton extends StatelessWidget {
       onPressed: onPressed,
       style: ButtonStyle(
         iconSize: MaterialStateProperty.all<double>(20.0),
-        elevation: MaterialStateProperty.all<double>(0), // this reduces the shade
+        elevation: MaterialStateProperty.all<double>(0),
+        // this reduces the shade
```
luckysori, line 509: 🔧 I think the comment belongs next to or on top of the previous line. Same thing happened below.

mobile/lib/features/wallet/receive_screen.dart
```diff
@@ -215,6 +222,34 @@ class _ReceiveScreenState extends State<ReceiveScreen> {
...
+                children: [
+                  const Icon(
+                    Icons.info,
+                    color: tenTenOnePurple,
+                    size: 22,
+                  ),
+                  const SizedBox(width: 10),
+                  Expanded(
+                    child: Text(
+                      "Receive around ${formatSats(Amount(minBalance))} to be able to trade.",
```
luckysori, line 245: 🔧 This doesn't take into account how much they might already have in the on-chain wallet. I think that's okay, but I would reword the message to:

```
"You will need around $minBalance in your on-chain wallet to start trading."
```

luckysori (approved):
Nice. Just a suggestion.

PR #2135, opened 2024-03-01
feat: Add admin api to send targeted push notifications

holzeis:
Adds an admin api allowing to send targeted push notifications to a list of users using their node ids. Note, if the user opted to not receive notifications, he will not get that campaign message.

With the `dry_run` flag you can test to how many users the message will be sent to and validate if the message is as expected before actually sending it.

```
$ curl localhost:8000/api/admin/campaign/push -d '{"node_ids":["027775c06af68dd0fbd1591a3da81afd13f7680c015ec48a76ae290bf48225a486"], "title": "test", "message": "this is a test", "dry_run": false }' -H "Content-Type: application/json"
Sending push notification campaign (title: test, message: this is a test to 1 users%
```

fixes https://github.com/get10101/meta/issues/364

### DIFF
--- a/coordinator/src/campaign.rs
+++ b/coordinator/src/campaign.rs
@@ -1,0 +1,80 @@
+use crate::db;
+use crate::notifications::FcmToken;
+use crate::notifications::Notification;
+use crate::notifications::NotificationKind;
+use crate::routes::AppState;
+use crate::AppError;
+use axum::extract::State;
+use axum::Json;
+use bitcoin::secp256k1::PublicKey;
+use serde::Deserialize;
+use serde::Serialize;
+use std::sync::Arc;
+use tracing::instrument;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PushCampaignParams {
+    pub node_ids: Vec<PublicKey>,
+    pub title: String,
+    pub message: String,
+    pub dry_run: Option<bool>,
+}
+
+#[instrument(skip_all, err(Debug))]
+pub async fn post_push_campaign(
+    State(state): State<Arc<AppState>>,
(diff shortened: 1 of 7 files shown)


coordinator/src/campaign.rs
```diff
@@ -0,0 +1,79 @@
...
+        .map_err(|e| AppError::InternalServerError(format!("Could not get connection: {e:#}")))?;
+
+    let users = db::user::get_users(&mut conn, params.node_ids)
+        .map_err(|e| AppError::InternalServerError(format!("Failed to get users: {e:#}")))?;
+
+    let fcm_tokens = users
+        .iter()
+        .map(|user| user.fcm_token.clone())
+        .filter(|token| !token.is_empty() && token != "unavailable")
+        .map(|token| FcmToken::new(token).expect("token to not be empty"))
```
bonomat, line 43: This seams to be error prone. It might be better to just do `.filter_map(Result::ok)`

coordinator/src/campaign.rs
```diff
@@ -0,0 +1,79 @@
...
+        .collect::<Vec<_>>();
+
+    let notification_kind = NotificationKind::Campaign {
+        title: params.title.clone(),
+        message: params.message.clone(),
+    };
+
+    tracing::info!(
+        "Sending push notification campaign (title: {}, message: {} to {} users",
```
bonomat, line 52: Nit: I prefer having the fields before the message, i.e. 
```suggestion
	title = params.title,
	message = params.message,
	users = fcm_tokens.len(),
        "Sending push notification campaign,
```

bonomat (approved):
Looks good, just some nits.

PR #2143, opened 2024-03-04
feat: websocket for sending position/trade updates

bonomat:
This new websocket will send trade/position updates as seen from the coordinator's perspective.

No need to do an indepth review, but please let me know what you think about this approach and I'll continue

### DIFF
--- a/coordinator/migrations/2024-03-10-042249_add_delete_state_to_order/down.sql
+++ b/coordinator/migrations/2024-03-10-042249_add_delete_state_to_order/down.sql
@@ -1,0 +1,2 @@
+-- We need to do something, or the migration fails
+select 1;
--- a/coordinator/migrations/2024-03-10-042249_add_delete_state_to_order/up.sql
+++ b/coordinator/migrations/2024-03-10-042249_add_delete_state_to_order/up.sql
@@ -1,0 +1,1 @@
+ALTER TYPE "OrderState_Type" ADD VALUE IF NOT EXISTS 'Deleted';
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -24,6 +24,7 @@ use coordinator::run_migration;
 use coordinator::scheduler::NotificationScheduler;
 use coordinator::settings::Settings;
 use coordinator::storage::CoordinatorTenTenOneStorage;
+use coordinator::trade::websocket::InternalPositionUpdateMessage;
 use diesel::r2d2;
 use diesel::r2d2::ConnectionManager;
 use diesel::PgConnection;
@@ -148,7 +149,17 @@ async fn main() -> Result<()> {
 
     let event_handler = CoordinatorEventHandler::new(node.clone(), None);
     let running = node.start(event_handler, false)?;
-    let node = Node::new(node, running, pool.clone(), settings.to_node_settings());
+
+    // an internal channel to send updates about our position
+    let (tx_position_feed, _rx) = broadcast::channel::<InternalPositionUpdateMessage>(100);
+
+    let node = Node::new(
+        node,
+        running,
+        pool.clone(),
+        settings.to_node_settings(),
+        tx_position_feed.clone(),
+    );
 
     // TODO: Pass the tokio metrics into Prometheus
     if let Some(interval) = opts.tokio_metrics_interval_seconds {
@@ -284,6 +295,7 @@ async fn main() -> Result<()> {
         NODE_ALIAS,
         trading_sender,
         tx_price_feed,
+        tx_position_feed,
         tx_user_feed,
         auth_users_notifier.clone(),
         notification_service.get_sender(),
--- a/coordinator/src/dlc_protocol.rs
+++ b/coordinator/src/dlc_protocol.rs
@@ -1,6 +1,7 @@
 use crate::db;
 use crate::position::models::PositionState;
 use crate::trade::models::NewTrade;
+use crate::trade::websocket::InternalPositionUpdateMessage;
 use anyhow::Context;
 use anyhow::Result;
 use bitcoin::secp256k1::PublicKey;
@@ -20,6 +21,7 @@ use std::fmt::Display;
 use std::fmt::Formatter;
 use std::str::from_utf8;
 use time::OffsetDateTime;
+use tokio::sync::broadcast::Sender;
 use trade::cfd::calculate_margin;
 use trade::cfd::calculate_pnl;
 use trade::Direction;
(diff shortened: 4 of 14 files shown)


holzeis:
It's probably the safest and easiest, but it seems wrong to put the logic around the takers positions.

IMO we should also reflect a position with the matched maker upon every successful trade.

Meaning we would have a consolidated position with the maker for all matched orders.

This would also make the api a bit nicer as a maker could ask for the coordinators point of view of his position and sync up his state.

It would also allow us to scale to multiple makers.

Also the api wouldn't have to be an admin api if we would verify the users signature.

holzeis:
I think in terms of maintenance it would probably also be easier to have a position with the maker, which we can patch as needed without affecting a takers position.

bonomat:
> It's probably the safest and easiest, but it seems wrong to put the logic around the takers positions.
> 
> IMO we should also reflect a position with the matched maker upon every successful trade.
> 
> Meaning we would have a consolidated position with the maker for all matched orders.
>
> This would also make the api a bit nicer as a maker could ask for the coordinators point of view of his position and sync up his state.

I'm not sure I understand what you are saying, because I believe that's what I'm doing?

Imho our matching logic should change though in a way that it either 
1. takes _our_ orders, or
2. matches against another maker

For the purpose of this websocket, we then should only care about all positions against 1. 

> It would also allow us to scale to multiple makers.

Imho, there should always be just 1 _maker_ for 1. 

> 
> Also the api wouldn't have to be an admin api if we would verify the users signature.

holzeis:
> I'm not sure I understand what you are saying, because I believe that's what I'm doing?

I am referring to this bit [`1e2be0e` (#2143)](https://github.com/get10101/10101/pull/2143/commits/1e2be0e5d751562981d863397f3e4b95663251e2#diff-52561b0a3ac3222faca2f414bcef5a156b174e79d4ae38a859b24d78927e7ff0R210).

We are aggregating all open takers positions (on the maker side), instead I think we should have one position with that particular maker and provide updates on that one. e.g. if we'd have multiple makers there would be a position per maker.

It's pretty close to what you have done already, but my proposal would be to not depend the updates on the takers positions, but rather aggregate the position we have with one maker on the coordinator side. And the maker should get updated based on this position, instead of getting all open takers positions. (then this api could also be public as you'd only get updates to your position. btw. that may be also useful with takers to sync up the apps position with the coordinators)

Having the ability to reflect a position with a maker (like we do with a taker) would also make the api a bit nice IMO. As we do not have to share "InternalPositionUpdates", but can share position updates with a particular user / maker.

crates/commons/src/position.rs
```diff
@@ -0,0 +1,970 @@
...
+use trade::bitmex_client::Quote;
+use trade::cfd::calculate_margin;
+use trade::cfd::calculate_pnl;
+use trade::ContractSymbol;
+use trade::Direction;
+
+pub type ContractId = [u8; 32];
+
+#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+pub enum PositionState {
```
luckysori, line 24: 🤔 I think this is a bad idea, because we are just exposing the internal state of the coordinator to the maker. Inevitably, we will have to iterate on this model (we already know this work is coming) and this will break the maker. Breaking changes across repositories are particularly annoying to deal with.

I don't think the maker should need all this context to make decisions. I would definitely introduce a dedicated model for the new websocket API.

bonomat:
> > I'm not sure I understand what you are saying, because I believe that's what I'm doing?
> 
> I am referring to this bit [`1e2be0e` (#2143)](https://github.com/get10101/10101/pull/2143/commits/1e2be0e5d751562981d863397f3e4b95663251e2#diff-52561b0a3ac3222faca2f414bcef5a156b174e79d4ae38a859b24d78927e7ff0R210).
> 
> We are aggregating all open takers positions (on the maker side), instead I think we should have one position with that particular maker and provide updates on that one. e.g. if we'd have multiple makers there would be a position per maker.
> 

> It's pretty close to what you have done already, but my proposal would be to not depend the updates on the takers positions, but rather aggregate the position we have with one maker on the coordinator side. And the maker should get updated based on this position, instead of getting all open takers positions. (then this api could also be public as you'd only get updates to your position. btw. that may be also useful with takers to sync up the apps position with the coordinators)
> 

Ah, gotya, I'll update the API to only return for the subscribed _maker_. (although there shouldn't be more than one _maker_ imho). 

> Having the ability to reflect a position with a maker (like we do with a taker) would also make the api a bit nicer IMO. As we do not have to share "InternalPositionUpdates", but can share position updates with a particular user / maker.

I would like to avoid mixing it with the [orderbook-based websocket](https://github.com/get10101/10101/blob/80c315e2b0c0888c08cb5035936e6b43cb87f18e/coordinator/src/routes.rs#L162) which gives updates to the trader as these positions are bound to DLCs and the one I'm interested here may not.

bonomat:
> Ah, gotya, I'll update the API to only return for the subscribed _maker_. (although there shouldn't be more than one _maker_ imho).

Correction, I think this goes into the wrong direction: 

In this websocket I'm actually interested in the _coordinator's_ position update, not for a particular maker. That's why I've also put it behind the admin api. 

Our current data model as a position currently is defined as **having a DLC with a _trader_**. Trader can be a maker or a taker. If we have a maker who also has a DLC with our coordinator, the overall position of the coordinator should be 0. 

That's why I need to aggregate all _trader_'s positions as all trades go against the coordinator.

bonomat:
> > Ah, gotya, I'll update the API to only return for the subscribed _maker_. (although there shouldn't be more than one _maker_ imho).
> 
> Correction, I think this goes into the wrong direction:
> 
> In this websocket I'm actually interested in the _coordinator's_ position update, not for a particular maker. That's why I've also put it behind the admin api.
> 
> Our current data model as a position currently is defined as **having a DLC with a _trader_**. Trader can be a maker or a taker. If we have a maker who also has a DLC with our coordinator, the overall position of the coordinator should be 0.
> 
> That's why I need to aggregate all _trader_'s positions as all trades go against the coordinator.

Long story short, I think my here proposed API/approach is correct :)

crates/commons/src/position.rs
bonomat: What in particular is a bad idea, moving it into `commons` ? 
Eitherway, it was not necessary after all and I've dropped this commit.

holzeis:
> Our current data model as a position currently is defined as having a DLC with a trader. Trader can be a maker or a taker. If we have a maker who also has a DLC with our coordinator, the overall position of the coordinator should be 0.
> 
> That's why I need to aggregate all trader's positions as all trades go against the coordinator.

That makes sense to me. Thanks for explaining. 👍

holzeis (approved):
LGTM 👍

crates/commons/src/position.rs
luckysori: I thibk sharing an internal position model with the maker is a bad idea. Glad to hear it is not necessary though.

PR #2144, opened 2024-03-04
fix(rollover): Remove outdated rollover recovery logic

### LINKED ISSUE #2137: Rollover failed with a reconnect after accepting renew offer
Environment: `test`
node id: `03b48395a708073df3ece6696af2cde63bd5a39b87d3bc4f197c13dc217482d99e`

1. App gets request to rollover
2. App rolls over
3. Coordinator sends the `RenewOffer` message
4. App accepts the `RenewOffer` message and transitions to `RenewAccepted`
5. App disconnects and connects again
6. Upon reconnecting the coordinator identifies that the app is in an intermediate state and rolls back to `Established`
```
Found signed channel contract in an intermediate state. Trying to rollback channel to the last stable state!
```
7. Coordinator again sends a `RenewOffer`referring the `Established` contract.
8. App rejects rollover because it has already accepted the new contract in the `RenewOffer`.

Now the app is stuck and unable to move forward.

The logic for rolling back is outdated and should be removed as it has been replaced with the retry logic.

holzeis:
The recovery logic has been replaced generally with retrying to send the `last_outbound_dlc_message` this also applies to the rollover. Hence we should not rollback to the latest state if the app reconnects in an intermediate state.

fixes #2137

### DIFF
--- a/coordinator/src/node/rollover.rs
+++ b/coordinator/src/node/rollover.rs
@@ -5,7 +5,6 @@ use crate::message::NewUserMessage;
 use crate::message::OrderbookMessage;
 use crate::node::Node;
 use crate::position::models::PositionState;
-use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thanks!!

PR #2145, opened 2024-03-04
chore(app): Change background color back to grey

holzeis:
This was the background color before the flutter upgrade. Also fixing the issue with the background color of the candlestick chart.

|                                       |						|
|:-------------------------------------:|:-------------------------------------:|
| Simulator Screenshot - iPhone 15 Pro - 2024-03-04 at 10 29 29 | Simulator Screenshot - iPhone 15 Pro - 2024-03-04 at 10 29 20 |

### DIFF
--- a/mobile/lib/common/scaffold_with_nav_bar.dart
+++ b/mobile/lib/common/scaffold_with_nav_bar.dart
@@ -25,7 +25,7 @@ class ScaffoldWithNavBar extends StatelessWidget {
         appBar: const PreferredSize(
             preferredSize: Size.fromHeight(40), child: SafeArea(child: AppBarWrapper())),
         bottomNavigationBar: BottomNavigationBar(
-          backgroundColor: Colors.white,
+          backgroundColor: const Color(0xFFFAFAFA),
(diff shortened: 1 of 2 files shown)


mobile/lib/common/scaffold_with_nav_bar.dart
```diff
@@ -25,7 +25,7 @@ class ScaffoldWithNavBar extends StatelessWidget {
         appBar: const PreferredSize(
             preferredSize: Size.fromHeight(40), child: SafeArea(child: AppBarWrapper())),
         bottomNavigationBar: BottomNavigationBar(
-          backgroundColor: Colors.white,
+          backgroundColor: const Color(0xFFFAFAFA),
```
luckysori, line 28: That's not grey, that's FFFAFAFA!

luckysori (approved):

mobile/lib/common/scaffold_with_nav_bar.dart
holzeis: Its copied from the color palette "light", didn't figure it out how to access it from there. 🙈

PR #2146, opened 2024-03-04
chore(emergency kit): Add option to rollback channel to last stable state

holzeis:
This might become very handy in case the protocol got stuck in an intermediate state. Unfortunately this will not help if one side has already reached the final state, while the other hasn't.

e.g. helps to recover from that https://github.com/get10101/10101/issues/2137

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -447,6 +447,50 @@ pub async fn delete_dlc_channel(
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


luckysori (approved):
I think it's okay to add this simple emergency button during the beta phase.

After the beta phase, I would not be comfortable letting the user delete their latest channel data, because they would have to trust the coordinator to do the same to do so safely.

PR #2147, opened 2024-03-04
feat(emergency kit): Add delete position and resend settle finalize message

holzeis:
This will allow our users to recover from a very specific symptom, which resulted out of a bug.


luckysori (approved):
LGTM!

PR #2148, opened 2024-03-04
Upgrade Rust to version 1.76

luckysori:
This upgrade highlighted the fact that we were using `Node::connect` in two different ways:

- Caring about the returned value, a future that will be ready once the connection is dropped.
- Not caring about the returned value.

Clippy was now complaining about us not always using the returned future. To fix this, I decided to split the API into two, so that the caller can be more explicit about what they want to do.


holzeis (approved):
Nice, thanks 👍

PR #2172, opened 2024-03-06
Check for old wallet data in correct location

luckysori:
I believe some old apps do have a `wallet` file that indicates that the app should probably be migrated to the new wallet DB via a full sync.

But this is not the case for fresh 1.9.0 apps. For those, we need to consider the presence of the `on_chain` directory.


holzeis:
> But this is not the case for fresh 1.9.0 apps. For those, we need to consider the presence of the on_chain directory.

Why isn't that for fresh 1.9.0 apps? did we change something there recently?

luckysori:
> > But this is not the case for fresh 1.9.0 apps. For those, we need to consider the presence of the on_chain directory.
> 
> Why isn't that for fresh 1.9.0 apps? did we change something there recently?

No idea. I assume that the location of the wallet directory has been like that for a long time. I don't really know what that other `wallet` file was for, but it's not present with 1.9.0.

Shall we just merge this though?

holzeis:
> > > But this is not the case for fresh 1.9.0 apps. For those, we need to consider the presence of the on_chain directory.
> > 
> > 
> > Why isn't that for fresh 1.9.0 apps? did we change something there recently?
> 
> No idea. I assume that the location of the wallet directory has been like that for a long time. I don't really know what that other `wallet` file was for, but it's not present with 1.9.0.
> 
> Shall we just merge this though, @holzeis?

How about renaming the directory instead of deleting.. just to be on the safe side?

luckysori:
> How about renaming the directory instead of deleting.. just to be on the safe side?

BDK is only storing stuff that can be found on-chain. If you have the same seed, you will find the same outputs and transactions after an on-chain sync.

holzeis (approved):
Thank you !

PR #2173, opened 2024-03-06
feat(webapp): add currency toggle to navbar to display balances in Sats/Btc or USD

### LINKED ISSUE #2155: Ability to switch : total balance/off-chain/on-chain from BTC to SATS to EUR (switch button OR dropdown menu)

bonomat:
This enables the user to display position values either in Sats/Btc or USD.

Values are a bit jumpy, but I guess it's good enough

Resolves #2155

https://github.com/get10101/10101/assets/224613/db986131-01f2-4daf-929e-36330463e0f3


holzeis (approved):
Nice 👍

PR #2174, opened 2024-03-06
feat(trade): Block users from trading with an old app version 

### LINKED ISSUE #1748: Update `Cargo.lock` during `draft-new-release` CI workflow
Otherwise after merging a release PR `main` has an outdated `Cargo.lock`, because we bump the `coordinator`'s version with every new release.

### LINKED ISSUE #2141: Block old versions from trading
More than once we had a user coming to us with errors related to not upgrading the app. We should make it explicit and block them in the app with a big warning that they are running an outdated version.

holzeis:
I opted to use the Cargo.toml version instead of the pubspec.yaml version as it is easier to access from rust and should probably anyways be bumped with every release.

A minor downside of this is now that we have to different sources for the version. However, both versions are bumped by the same ci pipeline during `draft_new_release` - so eventually the source is the same.

https://github.com/get10101/10101/assets/382048/e0b4c9d2-5e85-4373-8894-11df107a8a9d

fixes #2141
fixes #1748

### DIFF
--- a/.github/workflows/draft_new_release.yml
+++ b/.github/workflows/draft_new_release.yml
@@ -63,12 +63,22 @@ jobs:
         run: |
           cargo set-version --package webapp ${{ github.event.inputs.version }}
 
-      - name: Commit changelog and manifest files
+      - name: Bump mobile/native Cargo.toml version
+        uses: stellar/binaries@v13
+        with:
+          name: cargo-edit
+          version: 0.11.6
+      - id: set-mobile-version
+        continue-on-error: true
+        run: |
+          cargo set-version --package mobile/native ${{ github.event.inputs.version }}
+
+      - name: Commit manifest files
         id: make-commit
         run: |
           /home/runner/.dprint/bin/dprint fmt
 
-          git add mobile/pubspec.yaml coordinator/Cargo.toml webapp/Cargo.toml
+          git add mobile/native/Cargo.toml mobile/pubspec.yaml coordinator/Cargo.toml webapp/Cargo.toml
           git commit --message "Prepare release ${{ github.event.inputs.version }}"
 
           echo "::set-output name=commit::$(git rev-parse HEAD)"
--- a/.github/workflows/draft_new_release.yml
+++ b/.github/workflows/draft_new_release.yml
@@ -78,7 +78,7 @@ jobs:
         run: |
           /home/runner/.dprint/bin/dprint fmt
 
-          git add mobile/native/Cargo.toml mobile/pubspec.yaml coordinator/Cargo.toml webapp/Cargo.toml
+          git add mobile/native/Cargo.toml mobile/pubspec.yaml coordinator/Cargo.toml webapp/Cargo.toml Cargo.lock
           git commit --message "Prepare release ${{ github.event.inputs.version }}"
 
           echo "::set-output name=commit::$(git rev-parse HEAD)"
--- a/coordinator/src/db/user.rs
+++ b/coordinator/src/db/user.rs
@@ -20,6 +20,8 @@ pub struct User {
     pub fcm_token: String,
     pub last_login: OffsetDateTime,
     pub nickname: Option<String>,
+    // TODO(holzeis): Version is only optional for the first upgrade. Afterwards we should make it
+    // mandatory.
     pub version: Option<String>,
 }
 
--- a/coordinator/src/db/user.rs
+++ b/coordinator/src/db/user.rs
@@ -20,6 +20,7 @@ pub struct User {
     pub fcm_token: String,
     pub last_login: OffsetDateTime,
     pub nickname: Option<String>,
+    pub version: Option<String>,
 }
 
 impl From<RegisterParams> for User {
@@ -32,6 +33,7 @@ impl From<RegisterParams> for User {
             timestamp: OffsetDateTime::now_utc(),
             fcm_token: "".to_owned(),
             last_login: OffsetDateTime::now_utc(),
+            version: value.version,
         }
     }
 }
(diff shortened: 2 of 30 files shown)


coordinator/src/db/user.rs
```diff
@@ -160,3 +161,22 @@ pub fn get_users(conn: &mut PgConnection, trader_ids: Vec<PublicKey>) -> Result<
...
 }
+
+pub fn check_version(conn: &mut PgConnection, trader_id: &PublicKey) -> Result<()> {
+    let user: User = users::table
+        .filter(users::pubkey.eq(trader_id.to_string()))
+        .first(conn)?;
+
+    let app_version = match user.version {
+        Some(version) => version,
+        None => bail!("No version found"),
```
luckysori, line 172: :+1: I think this makes sense. If the information is not there, we can assume that we are dealing with an old app.

luckysori: :wrench: We usually prefer to use `context()?` instead of `match` + `bail` when dealing with `Option`s.

coordinator/src/db/user.rs
```diff
@@ -160,3 +161,22 @@ pub fn get_users(conn: &mut PgConnection, trader_ids: Vec<PublicKey>) -> Result<
...
+    };
+
+    let coordinator_version = env!("CARGO_PKG_VERSION").to_string();
+    ensure!(
+        app_version == coordinator_version,
```
luckysori, line 177: :wrench: What if we use semver here? Then we can deploy patch fixes to the coordinator without breaking trading for our users.

mobile/native/src/trade/order/orderbook_client.rs
```diff
@@ -34,8 +34,9 @@ impl OrderbookClient {
             let response = response.json().await?;
             Ok(response)
         } else {
-            tracing::error!("Could not create new order");
-            bail!("Could not create new order: {response:?}")
+            let error = response.text().await?;
+            tracing::error!(error, "Could not create new order");
```
luckysori, line 38: :wrench: Can we get rid of this log? It should be redundant with the error below.

luckysori (commented):
Looks good overall, but I left some suggestions.

crates/orderbook-client/src/lib.rs
```diff
@@ -33,12 +33,13 @@ pub async fn subscribe_with_authentication(
     url: String,
     authenticate: impl Fn(Message) -> Signature,
     fcm_token: Option<String>,
+    version: Option<String>,
```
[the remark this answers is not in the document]
holzeis: For downwards compatibility reasons.

coordinator/src/db/user.rs
```diff
@@ -160,3 +161,22 @@ pub fn get_users(conn: &mut PgConnection, trader_ids: Vec<PublicKey>) -> Result<
 
     Ok(users)
 }
+
+pub fn check_version(conn: &mut PgConnection, trader_id: &PublicKey) -> Result<()> {
```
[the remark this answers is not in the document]
holzeis: Where? 😅 that was the best place I could find. How about adding a new `version_check.rs`?

mobile/native/src/trade/order/orderbook_client.rs
holzeis: I thought the same, I just kept it because it was already there.

crates/orderbook-client/src/lib.rs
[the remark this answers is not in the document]
luckysori: Aren't we looking to _not_ be downwards-compatible though?

coordinator/src/db/user.rs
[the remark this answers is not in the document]
luckysori: Yeah, we can add a dedicated module.

I just don't like putting the logic in the DB module. When it is a database query it obviously makes sense.

coordinator/src/db/user.rs
luckysori: Fair enough. I just wanted to mention it.

crates/orderbook-client/src/lib.rs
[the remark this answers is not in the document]
holzeis: Haha, yeah but otherwise the user would get a bad error message.

I wanted to be able to tell the user that they need to upgrade.

luckysori (approved):
Thanks for going through my review.

crates/orderbook-client/src/lib.rs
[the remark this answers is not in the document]
luckysori: Okay, that makes sense. I think it would be nice to remove this `Option` later. Can you add a TODO?

.github/workflows/draft_new_release.yml
```diff
@@ -63,12 +63,22 @@ jobs:
         run: |
           cargo set-version --package webapp ${{ github.event.inputs.version }}
 
-      - name: Commit changelog and manifest files
+      - name: Bump mobile/native Cargo.toml version
+        uses: stellar/binaries@v13
+        with:
+          name: cargo-edit
+          version: 0.11.6
```
bonomat, line 70: ❌ This step already exists a few lines above

.github/workflows/draft_new_release.yml
```diff
@@ -63,12 +63,22 @@ jobs:
...
+      - id: set-mobile-version
+        continue-on-error: true
+        run: |
+          cargo set-version --package mobile/native ${{ github.event.inputs.version }}
```
bonomat, line 74: ❌ At least locally this does not work.
Correct would be
```suggestion
           cargo set-version --package native ${{ github.event.inputs.version }}
```

holzeis: 🙈 I tested it locally with a different command. Thanks for the fix 🙂

PR #2175, opened 2024-03-06
Allow coordinator to run full sync via HTTP API

luckysori:
This is necessary for one of the steps in https://github.com/get10101/meta/issues/381.


holzeis (approved):
👍

PR #2176, opened 2024-03-07
fix(ci) cargo set-version needs the package not path

bonomat:
Also, cleans up some redundant steps.

follow-up PR from https://github.com/get10101/10101/pull/2174


holzeis (approved):
Thank you 👍

PR #2192, opened 2024-03-08
fix(wallet): Sync on chain wallet periodically

### LINKED ISSUE #2191: Apps is not running on-chain sync periodically
Environment: test

### How to reproduce

1. Open and fund the app 750k sats
2. Send out 200k sats and do not manually refresh the app
3. Mine 5 blocks

The app should mine every 5 minutes on its own, but it never does and the wallet balance and history does not get updated.

holzeis:
fixes #2191


luckysori (approved):
Ooops, thanks for finding this!

Issue #2190, opened 2024-03-08
Send on-chain funds produces multiple `OutputBelowDustLimit` Errors


holzeis:
Environment: test

It seems the transaction got broadcasted, but at the same time we see the following error messages in the logs.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FfiException(RESULT_ERROR, OutputBelowDustLimit(0), null)
#0      FlutterRustBridgeBase._transformRust2DartMessage (package:flutter_rust_bridge/src/basic.dart:129:9)
#1      FlutterRustBridgeBase.executeNormal.<anonymous closure> (package:flutter_rust_bridge/src/basic.dart:70:9)
<asynchronous suspension>
#2      WalletService.calculateFeesForOnChain (package:get_10101/features/wallet/application/wallet_service.dart:54:18)
<asynchronous suspension>
#3      _SendOnChainScreenState.build.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:get_10101/features/wallet/send/send_onchain_screen.dart:194:47)
<asynchronous suspension>
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FfiException(RESULT_ERROR, OutputBelowDustLimit(0), null)
#0      FlutterRustBridgeBase._transformRust2DartMessage (package:flutter_rust_bridge/src/basic.dart:129:9)
#1      FlutterRustBridgeBase.executeNormal.<anonymous closure> (package:flutter_rust_bridge/src/basic.dart:70:9)
<asynchronous suspension>
#2      WalletService.calculateFeesForOnChain (package:get_10101/features/wallet/application/wallet_service.dart:54:18)
<asynchronous suspension>
#3      _SendOnChainScreenState.build.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:get_10101/features/wallet/send/send_onchain_screen.dart:194:47)
<asynchronous suspension>
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FfiException(RESULT_ERROR, OutputBelowDustLimit(0), null)
#0      FlutterRustBridgeBase._transformRust2DartMessage (package:flutter_rust_bridge/src/basic.dart:129:9)
#1      FlutterRustBridgeBase.executeNormal.<anonymous closure> (package:flutter_rust_bridge/src/basic.dart:70:9)
<asynchronous suspension>
[...]

PR #2194, opened 2024-03-08
Do not error when user is typing send amount

luckysori:
Fixes https://github.com/get10101/10101/issues/2190.

When the user is about to send on-chain and is typing the send amount, we dynamically calculate a fee estimate so that we can update the screen as they update the value.

This is nice, but the user will inevitably start by typing in small amounts for which calculating a fee does not make economical sense.

We enhance the library code to handle this case explicitly by using a error enum variant `EstimateFeeError::SendAmountBelowDust`. We can then leverage this in the app backend to simply ignore this kind of error.

Recording 2024-03-08 at 13 04 36

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2378,6 +2378,7 @@ dependencies = [
  "serde",
  "serde_with 3.2.0",
  "sha2",
+ "thiserror",
  "time",
  "tokio",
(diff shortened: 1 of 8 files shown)


holzeis (approved):
Thank you 👍

PR #2211, opened 2024-03-11
Only sync UTXOs owned by the wallet

luckysori:
It turns out that we don't need to watch literaly all the UTXOs ever :facepalm:


holzeis (approved):
LGTM 👍

Issue #2205, opened 2024-03-11
Replace `rust-dlc`'s `periodic_chain_monitor`


luckysori:
We want to rely on our `bdk` local state to confirm `rust-dlc` transactions.

PR #2212, opened 2024-03-11
Remove `periodic_chain_monitor`

luckysori:
Fixes https://github.com/get10101/10101/issues/2205.

With the update to `rust-dlc`, we no longer need to run `periodic_chain_monitor` periodically, nor does `rust-dlc` have to process blocks independently of the our on-chain wallet.

Matching pull request in our `rust-dlc` fork: https://github.com/get10101/rust-dlc/pull/14.

---

If the tests pass, I think this is okay :D

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1366,7 +1366,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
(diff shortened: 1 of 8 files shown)


holzeis (approved):
LGTM 👍

PR #2218, opened 2024-03-12
Prevent deadlock in `webapp`

luckysori:
Fixes https://github.com/get10101/10101/issues/1938.

There was definitely a deadlock here, so I'm closing optimistically closing #1938, as this may be the only deadlock.

Probably didn't need to bother with the repro, but it was a fun exercise.

Thanks for spotting it, @Restioson!

### DIFF
--- a/webapp/src/session.rs
+++ b/webapp/src/session.rs
@@ -68,3 +68,41 @@ impl InMemorySessionStore {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::time::Duration;
+    use time::ext::NumericalDuration;
+
+    #[tokio::test]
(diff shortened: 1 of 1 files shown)


luckysori:
Hey, @maxcountryman! We found the source of one deadlock. This does not necessarily settle our argument on async vs blocking locks here, but I wanted to keep you up to date anyway :D

holzeis (approved):
Thanks 👍

maxcountryman:
One more thing I noticed here: I believe you're awaiting the join handle of the task: this can lead to zombie processes depending on how the app is run. You may want to pass that handle to a graceful shutdown handler and abort it in a signal. We've added that to the examples that use continuous deletion a while back.

PR #2219, opened 2024-03-12
Do not run CI on push to branch

luckysori:
- The `staging` branch is no longer used since we migrated from `bors` to GitHub merge queue.

- Building on push to `main` is redundant if we only ever merge to `main` via the merge queue[^1].

[^1]: https://github.com/orgs/community/discussions/14333#discussioncomment-3077104.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,6 @@ name: CI
 on:
   pull_request:
   merge_group:
-  push:
-    branches:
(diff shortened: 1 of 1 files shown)


holzeis (approved):
Thanks

luckysori:
I'll wait for @bonomat's review in case this is controversial.

bonomat (approved):
I think that make sense. Thank you.

PR #2220, opened 2024-03-12
chore(test): Add environment file for esplora

holzeis:
This will allow us to overwrite the esplora environment variables to point to the correct electrs if run on a remote machine.

⚠️ Note, the file_env argument used in docker-compose requires at least v2.24

If you haven't already make sure to upgrade your docker-compose version as otherwise you won't be able to run `docker-compose up`

I also had to allow cors on our test setup as electrs and esplora are running on different ports. We could have used our nginx reverse proxy to circumvent that issue, but I figured its easier to simply allow cors on our test setup.

fixes https://github.com/get10101/cloud-conf/issues/257

### DIFF
--- a/.esplora.env
+++ b/.esplora.env
@@ -1,0 +1,1 @@
+API_URL=http://localhost:3000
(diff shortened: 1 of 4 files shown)


luckysori (approved):
Makes sense!

bonomat (commented):
Very nice. Thanks.

PR #2221, opened 2024-03-12
feat(webapp/app): show pnl as percent

### LINKED ISSUE #2187: Show PnL in % (App and Webapp)
We should also display the PnL in %

bonomat:
Resolves #2187 

App: 

image

Webapp: 

image


mobile/lib/features/trade/position_list_item.dart
```diff
@@ -70,6 +71,10 @@ class _PositionListItemState extends State<PositionListItem> {
 
     TextStyle dataRowStyle = const TextStyle(fontSize: 14);
 
+    Amount? unrealizedPnl = notNullPosition.unrealizedPnl;
+    double pnlPercent =
+        ((unrealizedPnl?.sats ?? Amount.zero().sats) / notNullPosition.collateral.sats) * 100.0;
```
holzeis, line 76: 🔧 we should probably round here to 2 numbers after the comma.

holzeis (approved):
Nice

mobile/lib/features/trade/position_list_item.dart
bonomat: Thanks for pointing this out. 

It's done when the pnlPercent is being used with `toStringAsFixed(2)`.  (I initially used the wrong function, but now it's correct).

PR #2224, opened 2024-03-13
Fix: make NewOrderRequest serialization deterministic

bonomat:
Note: this is a breaking change as old formats will be serialized differently. 

Creating NewOrders with an old version will simply fail until the user upgrades.

### DIFF
--- a/crates/commons/Cargo.toml
+++ b/crates/commons/Cargo.toml
@@ -17,3 +17,6 @@ time = { version = "0.3", features = ["serde", "std"] }
 tokio-tungstenite-wasm = { version = "0.3.0" }
 trade = { path = "../trade" }
 uuid = { version = "1.3.0", features = ["v4", "serde"] }
+
+[dev-dependencies]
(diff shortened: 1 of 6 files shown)


bonomat:
CI is failling because this needs a fix from our maker

crates/commons/Cargo.toml
```diff
@@ -17,3 +17,6 @@ time = { version = "0.3", features = ["serde", "std"] }
 tokio-tungstenite-wasm = { version = "0.3.0" }
 trade = { path = "../trade" }
 uuid = { version = "1.3.0", features = ["v4", "serde"] }
+
+[dev-dependencies]
+secp256k1 = { version = "0.27.0", features = ["serde", "rand", "global-context"] }
```
holzeis, line 22: nit: this shouldn't be needed if we use the `verify_ecdsa` function to verify signatures.

holzeis (approved):
LGTM 👍

crates/commons/Cargo.toml
bonomat: this is needed for generating the secret keys and a public key from it.

PR #2225, opened 2024-03-13
chore: Merge trade with post order endpoint

holzeis:
This change merges the trade endpoint into the post order endpoint.

- The post order api won't block for the order processing to be finished anymore. Potential errors will be returned via the websocket. That might have an impact on the maker as it is not handling errors returned via the websocket, which might be fine for now.
- The `channel_opening_params` aren't stored on the phone anymore, but at the coordinator. This is needed as the app might disconnect before the trade actually gets executed. In order to stay analog to the behaviour before this change, the `channel_opening_params` have to be persisted to be used on a reconnect.
- The pending matches (e.g. expired orders) are now checked on the p2p connection instead of the websocket connection.

----
Note, working on this change made me realize we will need a similar change for the rollover protocol, which is at the moment also websocket driven. https://github.com/get10101/10101/issues/2226


coordinator/src/node.rs
```diff
@@ -87,23 +87,13 @@ impl Node {
...
 
-    /// Returns true or false, whether we can find an usable channel with the provided trader.
-    ///
-    /// Note, we use the usable channel to implicitely check if the user is connected, as it
-    /// wouldn't be usable otherwise.
-    pub fn is_connected(&self, trader: &PublicKey) -> bool {
-        let usable_channels = self.inner.channel_manager.list_usable_channels();
-        let usable_channels = usable_channels
+    /// Returns true or false, whether the given peer_id is connected with us.
+    pub fn is_connected(&self, peer_id: PublicKey) -> bool {
```
luckysori, line 87: :question: I guess we weren't using this for anything critical? Nice fix.

coordinator/src/bin/coordinator.rs
```diff
@@ -251,7 +251,7 @@ async fn main() -> Result<()> {
     );
     let _handle = async_match::monitor(
         pool.clone(),
-        tx_user_feed.clone(),
+        node_event_handler.subscribe(),
```
luckysori, line 254: :+1: Wow, how simple.

coordinator/src/orderbook/trading.rs
```diff
@@ -111,10 +107,19 @@ pub fn start(
                             )
                             .await
                         }
-                    };
-
-                    if let Err(e) = new_order_msg.sender.send(result).await {
-                        tracing::error!("Failed to respond to NewOrderMessage: {e:#}");
+                    } {
+                        // TODO(holzeis): the maker is currently not subscribed to the websocket
+                        // api, hence it wouldn't receive the error message.
```
luckysori, line 120: :+1: I think this is okay. Do we log this error on the coordinator though?

coordinator/src/orderbook/trading.rs
```diff
@@ -134,15 +139,16 @@ pub async fn process_new_limit_order(
...
     new_order: NewOrder,
     order_reason: OrderReason,
-) -> Result<Order> {
+) -> Result<Order, TradingError> {
     let mut conn = spawn_blocking(move || pool.get())
         .await
         .expect("task to complete")?;
 
     if new_order.price == Decimal::ZERO {
+        tracing::warn!(trader_id=%new_order.trader_id, "Limit orders with zero price are not allowed");
```
luckysori, line 156: :+1: I guess this answers my question above.

crates/commons/Cargo.toml
```diff
@@ -6,13 +6,16 @@ edition = "2021"
 [dependencies]
 anyhow = "1"
 bitcoin = { version = "0.30", features = ["serde"] }
+diesel = { version = "2.0.0" }
+r2d2 = { version = "0.8.1" }
```
luckysori, line 10: :thinking: Can we try to do this without introducing this dependency here? I guess it's convenient to use `From`, but this is a bit annoying.

crates/commons/src/message.rs
```diff
@@ -43,10 +44,38 @@ pub enum Message {
...
     },
 }
 
+#[derive(Serialize, Deserialize, Clone, Error, Debug, PartialEq)]
+pub enum TradingError {
+    #[error("Invalid order: {0}")]
+    InvalidOrder(String),
+    #[error("{0}")]
+    NoMatchFound(String),
+    #[error("{0}")]
```
luckysori, line 57: :wrench: We can use `#[error(transparent)]` here.

crates/commons/src/message.rs
```diff
@@ -43,10 +44,38 @@ pub enum Message {
...
-        error: String,
+        error: TradingError,
     },
 }
 
+#[derive(Serialize, Deserialize, Clone, Error, Debug, PartialEq)]
+pub enum TradingError {
+    #[error("Invalid order: {0}")]
+    InvalidOrder(String),
+    #[error("{0}")]
```
luckysori, line 55: :wrench: I think this should be

```suggestion
    #[error("No match found: {0}")]
```

mobile/native/src/orderbook.rs
```diff
@@ -344,7 +344,7 @@ async fn handle_orderbook_message(
         Message::TradeError { order_id, error } => {
             order::handler::order_failed(
                 Some(order_id),
-                FailureReason::TradeResponse(error.clone()),
+                FailureReason::TradeResponse(format!("{error}")),
```
luckysori, line 350: :wrench: Maybe `error.to_string()`?

coordinator/src/orderbook/async_match.rs
```diff
@@ -112,6 +121,26 @@ async fn process_pending_match(
         if let Err(e) = notifier.send(msg).await {
             tracing::error!("Failed to send notification. Error: {e:#}");
         }
+
+        tracing::info!(trader_id = %order.trader_id, order_id = %order.id, "Executing trade for match");
+        let trade_executor = TradeExecutor::new(node, notifier);
+
+        tokio::spawn(async move {
```
luckysori, line 132: :question: Do we need to spawn here?

coordinator/src/orderbook/async_match.rs
```diff
@@ -112,6 +121,26 @@ async fn process_pending_match(
         if let Err(e) = notifier.send(msg).await {
             tracing::error!("Failed to send notification. Error: {e:#}");
         }
+
+        tracing::info!(trader_id = %order.trader_id, order_id = %order.id, "Executing trade for match");
```
luckysori, line 129: :wrench: Maybe we should include the order reason as a field.

coordinator/src/orderbook/trading.rs
```diff
@@ -291,6 +302,30 @@ pub async fn process_new_market_order(
...
+                        direction: order.direction,
+                        filled_with: matched_orders.taker_match.filled_with,
+                    },
+                    trader_reserve: channel_opening_params.map(|p| p.trader_reserve),
+                    coordinator_reserve: channel_opening_params.map(|p| p.coordinator_reserve),
+                })
+                .await;
+        });
+    } else {
+        tracing::warn!(trader_id = %order.trader_id, order_id = %order.id, "Skipping trade execution as trader is not connected");
```
luckysori, line 333: :+1: I think this warning is okay because we would normally expect the trade to be executed right away.

luckysori (approved):
Nice job! I think this goes in the right direction.

coordinator/src/orderbook/trading.rs
holzeis: Well, now that you mention it. We would most likely print that warning every time we post an async order.

Should we change it to info?

coordinator/src/node.rs
holzeis: It wasn't used at all.

crates/commons/src/message.rs
holzeis: I don't think this is possible here, because the error must be serializable and `anyhow::Error` is not.

coordinator/src/orderbook/trading.rs
holzeis: I opted to print a warning on an manual order and an info on an expired order

coordinator/src/orderbook/async_match.rs
holzeis: I don't think so

crates/commons/Cargo.toml
holzeis: yes.. I've removed the dependencies and map every error into an anyhow error.

crates/commons/src/message.rs
luckysori: Ah, sorry. It used to be `String` earlier in the diff.

Issue #2234, opened 2024-03-14
Can't click confirm buttons in create position dialog


bonomat:
When trying to open a position, the dialog is not clickable. 

https://github.com/get10101/10101/assets/224613/4f176df3-4dc5-45a4-863e-1570d917b3a8

PR #2081, opened 2024-02-22
chore: upgrade flutter to 3.19.1

bonomat:
resolves #1924 

Note: 
1. The overall design changed slightly, I think whenever working on the UI we need to adjust minor things... or live with it that the color tone changed a bit.
2. this does not upgrade flutter rust bridge which is a beast in itself

PR #2236, opened 2024-03-14
chore: introduce fvm (flutter version manager)

bonomat:
This allows us to have different flutter versions for webapp and mobile. It is needed, because the webapp can be built best with the current `beta` version. Using latest `stable` will break the webapp in firefox and we would need to fall back to a different renderer which comes at the cost of performance.

This might be controversial but I was debugging issue #2234 and upgraded flutter to `beta` which solved an issue at least @holzeis is aware off: the latest flutter version does not work flawlessly in firefox - to make it work we had to switch to a different renderer (which was done in[`a8646e1` (#2081)](https://github.com/get10101/10101/pull/2081/commits/a8646e153e9768ad5da32ccd2a8d5bc64563e63d)). However, this actually comes at performance and responsiveness cost - which I thought was the problem for #2234 (but is not).

Long story short, this PR allows us to use canvaskit rendered on the latest version which provides better performance and responsiveness than html renderer. 

Let me know what you think. Alternatively we could simply use `beta` channel for all apps for the time being and upgrade once the next flutter version is released.

### DIFF
--- a/.github/actions/setup-fvm/action.yml
+++ b/.github/actions/setup-fvm/action.yml
@@ -1,0 +1,45 @@
+name: Setup Flutter FVM Environment
+description: "Setup FVM, Flutter for building Flutter apps."
+
+inputs:
+  fvm_config:
+    description: "Path to fvm config file."
+    required: false
+  working_dir:
+    description: "The directory to run these steps in"
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - uses: kuhnroyal/flutter-fvm-config-action@v2
+      id: fvm-config-action
+      with:
+        path: ${{ inputs.fvm_config }}
+
+    - uses: subosito/flutter-action@v2
+      with:
+        cache: true
+        flutter-version: ${{ steps.fvm-config-action.outputs.FLUTTER_VERSION }}
+        channel: ${{ steps.fvm-config-action.outputs.FLUTTER_CHANNEL }}
+
+    - name: install fvm (linux)
+      if: runner.os == 'Linux'
+      shell: bash
+      working-directory: ${{inputs.working_dir}}
+      run: |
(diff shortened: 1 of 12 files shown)


holzeis:
@bonomat Not sure if I completely follow.

- Would that mean we have to actively change the flutter version using fvm before running the mobile or web app?
- Do we need this beta version to fix #2234 ?

Feels like a lot of complexity if its just for performance improvement, where we currently don't have any big issues.

bonomat:
> Would that mean we have to actively change the flutter version using fvm before running the mobile or web app?

No, your normal flow will _just_ work.  `just run` and `just run-web` have been changed already.
Only if you want to run `flutter` manually for anything you will need to remember to use `fvm flutter` before. 

> Do we need this beta version to fix https://github.com/get10101/10101/issues/2234 ?

No, but it allows us to use canvaskit again as the `beta`-channel fixes the issue with the flickering interface in Firefox.

Quote from: https://docs.flutter.dev/platform-integration/web/renderers

> ##  HTML renderer
> This renderer, which has a smaller download size than the CanvasKit renderer, uses a combination of HTML elements, CSS, Canvas elements, and SVG elements.
> 
> ## CanvasKit renderer
> This renderer is fully consistent with Flutter mobile and desktop, has faster performance with higher widget density, but adds about 1.5MB in download size. CanvasKit uses WebGL to render Skia paint commands.

holzeis (approved):
This might become a burden once the flutter beta version is getting stable, but I guess at this point we can simply revert that commit again.

luckysori (approved):

bonomat:
@luckysori / @holzeis 
I'll go ahead and merge this. Make sure to rebase your branches against main and install [FVM](https://fvm.app/)

PR #2237, opened 2024-03-14
chore(rollover): Directly propose rollover on p2p connection

### LINKED ISSUE #2200: Don't send reminder to rollover when position is already expired
image

### LINKED ISSUE #2226: Get rid of the rollover api
A rollover over should be triggered by the coordinator at the p2p connection and shouldn't take a full round over the app before getting triggered.

Note, the rollover api might still be useful, but should be moved to the admin api.

holzeis:
fixes #2226 
fixes #2200 

- Rollover api has been moved to admin
- The coordinator now checks on a p2p connection if the users position needs to get rolled over
- Combines the logic for checking rollover from job an on connect, fixing the issue of notifying the user about the rollover window if there is already an expired position. (Note, the job did not check if the position was already expired)
- Adds a small fix to wait for the current context to become available before opening a dialog (async match and rollover)
- Connects to the coordinator on the websocket before starting the node. That prevents a possible race condition where the p2p connection could be already established but the websocket connection isn't.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,7 +1367,7 @@ dependencies = [
 [[package]]
 name = "dlc"
 version = "0.4.0"
-source = "git+https://github.com/get10101/rust-dlc?rev=1010f95#1010f95f930d98219f8c28c2854391fbbd081ff8"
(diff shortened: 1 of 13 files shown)


bonomat (approved):
Nice. Looks good to me.

PR #2238, opened 2024-03-14
Claim settle transaction funds after force-close

luckysori:
Matching `rust-dlc` PR: https://github.com/get10101/rust-dlc/pull/15.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,11 +20,11 @@ resolver = "2"
 # We are using our own fork of `rust-dlc` at least until we can drop all the LN-DLC features. Also,
 # `p2pderivatives/rust-dlc#master` is missing certain patches that can only be found in the LN-DLC
 # branch.
-dlc-manager = { git = "https://github.com/get10101/rust-dlc", rev = "c3f8850" }
-dlc-messages = { git = "https://github.com/get10101/rust-dlc", rev = "c3f8850" }
-dlc = { git = "https://github.com/get10101/rust-dlc", rev = "c3f8850" }
-p2pd-oracle-client = { git = "https://github.com/get10101/rust-dlc", rev = "c3f8850" }
-dlc-trie = { git = "https://github.com/get10101/rust-dlc", rev = "c3f8850" }
+dlc-manager = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
+dlc-messages = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
+dlc = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
+p2pd-oracle-client = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
+dlc-trie = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
 
 # We should usually track the `p2pderivatives/split-tx-experiment[-10101]` branch. For now we depend
 # on a special fork which removes a panic in `rust-lightning`.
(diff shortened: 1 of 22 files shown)


crates/tests-e2e/tests/e2e_force_close_settled_channel.rs
```diff
@@ -0,0 +1,47 @@
...
+    let setup = setup::TestSetup::new_with_open_position().await;
+
+    let closing_order = {
+        let mut order = dummy_order();
+        order.direction = Direction::Short;
+        order
+    };
+
+    submit_order(closing_order.clone());
+    wait_until!(setup.app.rx.position_close().is_some());
```
holzeis, line 26: Note, this is not really reliable to judge whether the channel did actually get settled successfully as this event is already thrown at the before the `SettleFinalize` message is sent back to the coordinator.

crates/ln-dlc-node/src/tests/dlc_channel.rs
```diff
@@ -244,19 +245,21 @@ async fn can_open_and_force_close_channel() {
...
+            .iter()
+            .find(|dlc_channel| {
+                dlc_channel.counter_party == to_secp_pk_29(app.info.pubkey)
+                    && matches!(dlc_channel.state, SignedChannelState::SettledClosing { .. })
+            })
+            .cloned())
     })
     .await
     .unwrap();
-
```
holzeis, line 257: 😂 you don't like that comment?

webapp/frontend/lib/settings/channel_screen.dart
```diff
@@ -115,26 +115,26 @@ class _ChannelDetailWidgetState extends State<ChannelDetailWidget> {
...
                 Visibility(
-                  visible: widget.channel.subchannelState?.name != null,
+                  visible: widget.channel.signedChannelState?.name != null,
                   child: Padding(
                     padding: const EdgeInsets.all(4.0),
                     child: Row(
                       mainAxisAlignment: MainAxisAlignment.spaceBetween,
                       children: [
-                        const Text("Subchannel state"),
+                        const Text("Signed channel state"),
```
holzeis, line 124: thank you for that 🙂

Cargo.toml
```diff
@@ -20,11 +20,11 @@ resolver = "2"
 # We are using our own fork of `rust-dlc` at least until we can drop all the LN-DLC features. Also,
 # `p2pderivatives/rust-dlc#master` is missing certain patches that can only be found in the LN-DLC
 # branch.
-dlc-manager = { git = "https://github.com/get10101/rust-dlc", rev = "1010f95" }
```
holzeis: Should this be in this PR? Its not referred in the PR description is it?

holzeis (approved):
LGTM 👍

crates/ln-dlc-node/src/tests/dlc_channel.rs
luckysori: I love it, but it doesn't actually make sense! This was actually closing a _settled_ channel, which does not involve CETs. So we don't need a Tardis after all :D

crates/tests-e2e/tests/e2e_force_close_settled_channel.rs
luckysori: So you would just remove it? I think I just followed the same approach that we use in the other force-close e2e test, and then extended it to prove something more meaningful.

Cargo.toml
luckysori: I can drop it since we rely on PR names for the change log.

crates/tests-e2e/tests/e2e_force_close_settled_channel.rs
holzeis: I wouldn't remove it, but probably fetch the dlc channels and check if the coordinator really thinks the channel is settled.

Cargo.toml
luckysori: Done.

crates/tests-e2e/tests/e2e_force_close_settled_channel.rs
luckysori: Maybe I can do that another time. I think it's a bit more involved to add this to the e2e tests.

PR #2243, opened 2024-03-15
fix: make webapp usable in 🥖🇫🇷

### LINKED ISSUE #2133: Webapp login not scrollable
Which makes it hard to login on mobile because the login field is behind the keyboard and the keyboard always pops up when trying to paste a password

### LINKED ISSUE #2242: Login screen not working when  `french` is system language

bonomat:
resolves #2242 
resolves #2133


holzeis (approved):
LGTM 👍

Is there now a fallback if a user accesses the app with a language not on the supported list?

bonomat:
> Is there now a fallback if a user accesses the app with a language not on the supported list?

Yes, but this has been there before. The problem was that the local was not forwarded to all widgets.

If the local language is not supported it will take the first one from the list of supported locals, i.e. `en`.

PR #2253, opened 2024-03-15
feat(polls): Add editable choice option

holzeis:
targeted_polls


coordinator/src/routes.rs
```diff
@@ -442,12 +443,18 @@ pub async fn version() -> Result<Json<Version>, AppError> {
     }))
 }
 
-pub async fn get_polls(State(state): State<Arc<AppState>>) -> Result<Json<Vec<Poll>>, AppError> {
+pub async fn get_polls(
+    Path(node_id): Path<String>,
```
bonomat, line 447: new name for `trader_pubkey`?
😅

coordinator/src/db/polls.rs
```diff
@@ -44,6 +49,7 @@ pub struct Poll {
     pub question: String,
     pub active: bool,
     pub creation_timestamp: OffsetDateTime,
+    pub whitelisted: bool,
```
bonomat, line 52: So, the field `whitelisted` means, that if it's true, the poll would only be shown to traders in `polls_whitelist`?

Issue #2240, opened 2024-03-14
Can overwrite channel margin 


bonomat:
### I'm opening a new trade :

As you can see, for 4000+ USD in quantity, and check the collateral needed:
![IMG_0528](https://github.com/get10101/meta/assets/137571657/3757db1a-814a-41c5-bf9e-e0348701cd0f)

**As expected, just after this screen we have the channel opening config screen.**

### BUT
Interestingly, it allows me to put **LESS** margin than needed by the trade I want to open. (I was expecting to have a minimum forced / limited by the necessary margin for the first trade, and not only by the absolute necessary margin to open a channel).

> I didn't tempt the devil by trying to open this channel with less collateral than the requested trade so as not to have to close the channel and reopen it, which entails costs.

I don't know if this behavior is expected, so I'm filing this issue for information/discussion.
![IMG_0529](https://github.com/get10101/meta/assets/137571657/6321d694-554f-4692-9e69-aa5c4eb327ff)

bonomat:
It's only a UX bug because the input value will be ignored and the min required margin will be taken.

Issue #2139, opened 2024-03-03
Add a max trade button


holzeis:
We have a lot of users that try to trade with the max balance they have in their wallet (balance).

Two phases: 

1. [x] use max on-chain balance to open a channel
2. [ ] use max off-chain balance to trade

bonomat:
Show this max button at least when the channel is open.

bonomat:
Mhm, this is actually not easily possible. 
The only way to define how much you want to trade is by entering a `. However, while the `quantity` is fixed, the actual margin is dynamic and depends on the rate. 
A user doesn't want to trade max quantity but max `margin`, i.e. say to the orderbook, I want to buy/sell for all the sats I own.

This idea has been proposed before here: https://github.com/get10101/10101/issues/1712

holzeis:
> This idea has been proposed before here: #1712

I like it, I guess the user wouldn't need to see the actual quantity if we tell them we will buy the max possible. I would do it similar to how its done on the send on-chain funds screen. In the background we can run a margin order.

bonomat:
I'm throwing this back into the pile because I burned my fingers with https://github.com/get10101/10101/pull/2282 and need to lick my wounds before giving it another go.

PR #2259, opened 2024-03-18
feat(mobile): Select max channel size

bonomat:
This PR introduces a new slider type which is now also being used for the leverage slider. 

Part I of #2139 next is to introduce `Margin Orders`
Resolves https://github.com/get10101/10101/issues/2240

https://github.com/get10101/10101/assets/224613/90744cff-f03f-44d0-93de-5ae65d0575b3

### DIFF
--- a/mobile/lib/common/amount_text_input_form_field.dart
+++ b/mobile/lib/common/amount_text_input_form_field.dart
@@ -6,7 +6,7 @@ import 'package:get_10101/common/modal_bottom_sheet_info.dart';
 
 class AmountInputField extends StatelessWidget {
   /// If `decoration` is passed, then `isLoading`, `hint`, `label`, `infoText`,
-  /// and `isLoading` are overriden.
+  /// and `isLoading` are overwritten.
   const AmountInputField(
       {super.key,
       this.enabled = true,
--- a/mobile/lib/common/amount_text_input_form_field.dart
+++ b/mobile/lib/common/amount_text_input_form_field.dart
@@ -2,11 +2,10 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:get_10101/common/application/numeric_text_formatter.dart';
 import 'package:get_10101/common/domain/model.dart';
-import 'package:get_10101/common/modal_bottom_sheet_info.dart';
 
 class AmountInputField extends StatelessWidget {
   /// If `decoration` is passed, then `isLoading`, `hint`, `label`, `infoText`,
-  /// and `isLoading` are overwritten.
+  /// and `isLoading` are overridden.
   const AmountInputField(
       {super.key,
       this.enabled = true,
@@ -15,7 +14,7 @@ class AmountInputField extends StatelessWidget {
       this.onChanged,
       required this.value,
       this.isLoading = false,
-      this.infoText,
+      this.suffixIcon,
       this.controller,
       this.validator,
       this.decoration,
@@ -28,7 +27,7 @@ class AmountInputField extends StatelessWidget {
   final bool enabled;
   final String label;
   final String hint;
-  final String? infoText;
+  final Widget? suffixIcon;
   final bool isLoading;
   final Function(String)? onChanged;
   final Function()? onTap;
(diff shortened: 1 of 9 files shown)


holzeis:
The channel config modal is getting quite huge 🙈 . Have you tested if this is still viewable on smaller screens?

mobile/lib/features/trade/channel_configuration.dart
```diff
@@ -38,27 +38,21 @@ channelConfiguration({
...
-                  child: SizedBox(
-                    height: 500,
-                    child: ChannelConfiguration(
-                      tradeValues: tradeValues,
-                      onConfirmation: onConfirmation,
-                    ),
-                  ),
-                )),
-          ),
+        return Wrap(
```
holzeis: 🤔 Are you sure this will work when toggling the keyboard?

mobile/lib/features/trade/channel_configuration.dart
```diff
@@ -148,6 +147,10 @@ class _ChannelConfiguration extends State<ChannelConfiguration> {
               Text(
                   "Please specify your preferred channel size, impacting how much you will be able to win up to.",
                   style: DefaultTextStyle.of(context).style),
+              const SizedBox(height: 10),
```
holzeis, line 152: I think we should incorporate that into the text above and reduce the wall of text in general. I guess we will need to do this anyways to save space for the toggled keyboard.

mobile/lib/features/trade/channel_configuration.dart
```diff
@@ -144,156 +132,142 @@ class _ChannelConfiguration extends State<ChannelConfiguration> {
         (maxCounterpartyCollateral.sats * counterpartyLeverage).toInt();
 
     return Form(
-        key: _formKey,
+      key: _formKey,
+      child: SingleChildScrollView(
```
holzeis, line 136: I guess the better solution would be to reduce the content on this modal. This could be a blocker for users already because it's too complicated, now the might even have to scroll 🙈 

Can't we reduce the text a bit?

mobile/lib/features/trade/channel_configuration.dart
bonomat: Good question, let me try.

mobile/lib/features/trade/channel_configuration.dart
bonomat: I reduced the text a bit.

bonomat:
I believe it is impossible to use max balance with the current rust-dlc library. I believe with this patch we get as close as possible though but it's not guaranteed that we do not run into errors due to needing more for fees. 

I did a few more manual test runs locally but I'm afraid the real experience we will only get when testing in production with a real fee environment.

holzeis:
> I did a few more manual test runs locally but I'm afraid the real experience we will only get when testing in production with a real fee environment.

Was there any other downside than that the order failed? Meaning, is the user able to retry with a regular order afterwards, or is he than blocked?

holzeis (approved):
Looks great! 👍 I guess the channel configurator is still a spot where we can improve a lot in terms of UI, but in terms of functionality I like it a lot.

I added one commit centering the DLC Channel config title.

image

bonomat:
> Was there any other downside than that the order failed? Meaning, is the user able to retry with a regular order afterwards, or is he than blocked?

He should be able to retry as the utxo are only locked if we find enough utxos, meaning, he will fail because of BNB not found.

mobile/lib/common/amount_text_input_form_field.dart
```diff
@@ -6,7 +6,7 @@ import 'package:get_10101/common/modal_bottom_sheet_info.dart';
 
 class AmountInputField extends StatelessWidget {
   /// If `decoration` is passed, then `isLoading`, `hint`, `label`, `infoText`,
-  /// and `isLoading` are overriden.
+  /// and `isLoading` are overwritten.
```
luckysori, line 8: 🔧 There was a typo, but the correct word is "overridden".

mobile/lib/features/trade/collateral_slider.dart
```diff
@@ -0,0 +1,94 @@
+import 'dart:math';
+
+import 'package:flutter/material.dart';
+import 'package:get_10101/common/color.dart';
+import 'package:syncfusion_flutter_sliders/sliders.dart';
+import 'package:syncfusion_flutter_core/theme.dart' as slider_theme;
+
+/// Slider that allows the user to select a value between minValeu and maxValue.
```
luckysori, line 8:
```suggestion
/// Slider that allows the user to select a value between minValue and maxValue.
```

luckysori (approved):

luckysori:
> Looks great! 👍 I guess the channel configurator is still a spot where we can improve a lot in terms of UI, but in terms of functionality I like it a lot.
> 
> I added one commit centering the DLC Channel config title.
> 
> image

@bonomat: Could we merge the two sections titled `Your collateral`. I think it's weird that the slider, the value and the max button are not all in the same section.

holzeis:
> > image
> 
> @bonomat: Could we merge the two sections titled `Your collateral`? I think it's weird that the slider, the value and the max button are not all in the same section.

@luckysori I tried, but the best I could come up with was this

image

At least it saves a little bit more real estate.

PR #2271, opened 2024-03-18
chore: Show settle transaction txid on dlc channels api

### LINKED ISSUE #2269: SettledClosingDlcChannel is not a subtype of type ClosingDlcChannel
image

holzeis:
fixes #2269

### DIFF
--- a/mobile/native/src/dlc_channel.rs
+++ b/mobile/native/src/dlc_channel.rs
@@ -32,6 +32,21 @@ impl From<&Channel> for DlcChannel {
                 closing_txid: Some(close_tx.txid().to_string()),
                 state: SignedChannelState::CollaborativeCloseOffered,
             },
+            s @ Channel::Signed(SignedChannel {
+                state: dlc_manager::channel::signed_channel::SignedChannelState::SettledClosing {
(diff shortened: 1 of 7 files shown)


mobile/native/src/dlc_channel.rs
```diff
@@ -32,6 +32,21 @@ impl From<&Channel> for DlcChannel {
...
+                },
+                fund_tx,
+                fund_output_index,
+                ..
+            }) => ChannelState::Signed {
+                contract_id: s.get_contract_id().map(hex::encode),
+                funding_txid: fund_tx.txid().to_string(),
+                funding_tx_vout: *fund_output_index,
+                closing_txid: Some(settle_transaction.txid().to_string()),
+                state: SignedChannelState::SettledClosing,
```
bonomat, line 48: Did you check if this works as expected in the webapp as well?

bonomat (approved):
LGTM

mobile/native/src/dlc_channel.rs
holzeis: no, I will have to fix it on the webapp as well. Thanks for pointing it out.

PR #2282, opened 2024-03-19
Feature/BREAKING(app):  introduce margin orders

bonomat:
This patch introduces margin orders, i.e. a trader can define the margin they want to use for opening a order instead of defining the quantity. 

Check commit messages for more details.

Overall I'm happy that it works but our code based really didn't make it easy to add this feature. 

Let me know what you think, if deemed to brittle, I'm happy to drop this PR.

https://github.com/get10101/10101/assets/224613/de832a8a-c0d4-43c1-83fc-9cfdeae68a59

### DIFF
--- a/coordinator/migrations/2024-03-18-054329_add_margin_orders/down.sql
+++ b/coordinator/migrations/2024-03-18-054329_add_margin_orders/down.sql
@@ -1,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+
+ALTER TABLE
+    orders DROP COLUMN margin_sats;
--- a/coordinator/migrations/2024-03-18-054329_add_margin_orders/up.sql
+++ b/coordinator/migrations/2024-03-18-054329_add_margin_orders/up.sql
@@ -1,0 +1,8 @@
+-- Note that the `IF NOT EXISTS` is essential because there is no `down` migration for removing this value because it is not really feasible to remove enum values!
+-- In order to allow re-running this migration we thus have to make sure to only add the value if it does not exist yet.
+ALTER TYPE "OrderType_Type"
+ADD
+    VALUE IF NOT EXISTS 'margin';
+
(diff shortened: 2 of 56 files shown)


bonomat:
> @bonomat is it necessary for the user to enter the margin? 

Not at all. We don't need to make it editable but the flow will be the same: instead of entering an amount into the margin field you click `max`. Same thing, just different color.

holzeis:
> Let's discuss the UI changes at last please.
> 
> If we want this change I'd like to get it in fast because of all the files I had to touch.

I am not referring solely to a UI change, I am questioning if we need a margin order. Can't we just max out the quantity in the same way we do the input validation. That might not be 100% correct (not the full possible margin) but will solve the users issue of going back and forth.

bonomat:
> I am not referring solely to a UI change, I am questioning if we need a margin order. Can't we just max out the quantity in the same way we do the input validation. That might not be 100% correct (not the full possible margin) but will solve the users issue of going back and forth.

Right. I think a margin trade is a nice feature to have. Sometimes I just want to trade in sats and not USD terms. 
If you max out the quantity you will always have some sats left in your channel.

bonomat:
@luckysori : what are your thoughts? 

I'm slightly leaning to letting this one go again (even though it did cost me a few hours to write).

luckysori (commented):
> @luckysori : what are your thoughts?
>
> I'm slightly leaning towards letting this one go again (even though it did cost me a few hours to write).

I would say I am against margin orders now, given that the implementation is not that simple and that they don't seem all that reliable judging from your commit description. Also, I don't think it's a _very_ important feature, because a user can relatively easily play around with the quantity to get reasonably close to their desired margin.

But I like the first 5 patches! Can we keep those?

bonomat:
closed in favor of https://github.com/get10101/10101/pull/2316

PR #2283, opened 2024-03-20
feat: introduce dev-maker

### LINKED ISSUE #2260: Implement dev-maker
We should implement a "dev" maker which is in the main repository.

Ideally, the dev-maker does not require an internet connection. We could either just do static prices or put a price.json file which contains historic data. This way we get some "randomness". 

Having this dev-maker would simplify breaking api changes as one does not have to work on two repos at the same time which depend on each other.

bonomat:
resolves #2260

### DIFF
--- a/crates/dev-maker/fetch_rates.sh
+++ b/crates/dev-maker/fetch_rates.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# A script to fetch historic rates from Bitmex between a specifc time window
+
+# BitMEX API base URL
+API_URL="https://www.bitmex.com/api/v1"
+
+SYMBOL="XBTUSD"
+
+# Set start and end timestamps for the date range (March 20, 2023, to March 20, 2024)
+START_TIMESTAMP="2023-03-20T00:00:00.000Z"
+END_TIMESTAMP="2024-03-20T00:00:00.000Z"
+
+# Construct the request URL for hourly data
+REQUEST_URL="${API_URL}/trade/bucketed?symbol=${SYMBOL}&binSize=1h&partial=false&count=1000&reverse=false&startTime=${START_TIMESTAMP}&endTime=${END_TIMESTAMP}"
+
+# Fetch the data from BitMEX API
+RESPONSE=$(curl -s "${REQUEST_URL}")
+
+# Check if the request was successful
+if [ -z "${RESPONSE}" ]; then
+    echo "Failed to fetch data from BitMEX API."
+    exit 1
+fi
+
+# Parse JSON data and extract timestamp, symbol, and open price
+JSON_DATA=$(echo "${RESPONSE}" | jq -r '[.[] | {timestamp: .timestamp, symbol: .symbol, open: .open}]')
+
+# Store JSON data to a file
+echo "${JSON_DATA}" > bitmex_hourly_rates.json
+
+echo "Hourly rates stored in bitmex_hourly_rates.json"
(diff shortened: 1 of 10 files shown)


crates/dev-maker/fetch_rates.sh
```diff
@@ -0,0 +1,31 @@
...
+API_URL="https://www.bitmex.com/api/v1"
+
+SYMBOL="XBTUSD"
+
+# Set start and end timestamps for the date range (March 20, 2023, to March 20, 2024)
+START_TIMESTAMP="2023-03-20T00:00:00.000Z"
+END_TIMESTAMP="2024-03-20T00:00:00.000Z"
+
+# Construct the request URL for hourly data
+REQUEST_URL="${API_URL}/trade/bucketed?symbol=${SYMBOL}&binSize=1h&partial=false&count=1000&reverse=false&startTime=${START_TIMESTAMP}&endTime=${END_TIMESTAMP}"
```
holzeis, line 14: nit: We only fetch the first 1000 prices here, which gets us to April 2023. We might want to have more recent prices?

crates/dev-maker/fetch_rates.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# A script to fetch historic rates from Bitmex between a specifc time window
+
+# BitMEX API base URL
+API_URL="https://www.bitmex.com/api/v1"
+
+SYMBOL="XBTUSD"
+
+# Set start and end timestamps for the date range (March 20, 2023, to March 20, 2024)
+START_TIMESTAMP="2023-03-20T00:00:00.000Z"
```
holzeis, line 10: nit: How about making the script more generic by automatically calculating the start and the end timestamp based on the current date?

justfile
```diff
@@ -302,9 +302,7 @@ run_maker_args := if os() == "linux" {
 }
 
 maker args="":
-    # we always delete the old container first as otherwise we might get an error if the container still exists
-    docker rm -f maker || true
-    docker run {{run_maker_args}}
+    cargo run --bin dev-maker
```
holzeis, line 305: 🤔 I guess we will have to add the dev makers pubkey to the `whitelisted_makers` in the `test-coordinator-settings.toml`

holzeis (commented):
Nice 👍 Thank you!

crates/dev-maker/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    historic_rates.sort_by(|a, b| a.timestamp.cmp(&b.timestamp));
+
+    loop {
+        for historic_rate in &historic_rates {
+            post_order(
+                client.clone(),
+                secret_key,
+                public_key,
+                Direction::Short,
+                historic_rate.open + Decimal::from(1),
```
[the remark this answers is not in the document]
bonomat: No, I think this is correct... the order is my indication to go _short_, so you would go _long_, or differently said, I'm selling this to you for this price. Hence, I charge `1` more.

crates/dev-maker/fetch_rates.sh
bonomat: feel free to update it, I think for local testing it doesn't matter much.

justfile
bonomat: For local dev setup it shouldn't be needed because whitelisting should be disabled.

bonomat:
> Nice 👍 Thank you!

Feel free to approve things ;)

crates/dev-maker/fetch_rates.sh
bonomat: I asked chat-gtp and the `date` library on linux works differently than on mac, hence, I hardcoded it 😬

PR #2293, opened 2024-03-21
chore: Add admin api to resend renew revoke message

holzeis:
- Adds an admin api to resent the renew revoke message
- Adds an emergency kit to recreate the position from the signed channel and last failed order. This might be needed if the app set the order to fail, which will block the app from recreating the app. I opted to not make the code smarter when processing the `RenewRevoke` message but rather add an emergency kit feature. That will hopefully be less failure prune.

fixes https://github.com/get10101/meta/issues/413

### DIFF
--- a/coordinator/src/admin.rs
+++ b/coordinator/src/admin.rs
@@ -543,6 +543,27 @@ pub async fn rollover(
     Ok(())
 }
 
+pub async fn resend_renew_revoke_message(
+    State(state): State<Arc<AppState>>,
+    Path(trader_pubkey): Path<String>,
+) -> Result<(), AppError> {
+    let trader = trader_pubkey.parse().map_err(|err| {
+        AppError::BadRequest(format!("Invalid public key {trader_pubkey}. Error: {err}"))
+    })?;
+
+    state
+        .node
+        .resend_renew_revoke_message_internal(trader)
+        .map_err(|e| {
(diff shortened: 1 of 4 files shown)


mobile/native/src/emergency_kit.rs
```diff
@@ -44,6 +52,86 @@ pub fn delete_position() -> Result<()> {
...
+                .oracle_announcements
+                .first()
+                .context("missing oracle info")?
+                .oracle_event;
+
+            let expiry_timestamp =
+                OffsetDateTime::from_unix_timestamp(oracle_event.event_maturity_epoch as i64)?;
+
+            (
+                contract.accepted_contract.accept_params.collateral,
```
luckysori, line 102: 🤔 This is the total DLC collateral, which does not take into account the collateral reserve that is not being traded. I think the position's `collateral` is actually meant to be the trading margin, which can be calculated like this:

```
contract.accepted_contract.accept_params.collateral - trader_collateral_reserve
```

The reserve is given by `get_usable_dlc_channel_balance`.

luckysori (approved):
LGTM overall.

holzeis:
> Since we use PR names to generate the changelog and this PR is adding two `feat`s, we should consider renaming the PR.

I've created a separate PR for the emergency kit https://github.com/get10101/10101/pull/2296

PR #2294, opened 2024-03-21
fix: Only send dlc messages if processing was successful

### LINKED ISSUE #2127: Only send dlc message after successful processing.
If a task fails after processing the dlc message, we do send the following dlc message even if the action failed. That results in weird symptoms after a failure.

We should only send the dlc message if everything was processed successfully. That would also prevent any premature response to a dlc message that we have sent before completing our task.

holzeis:
The message will firstly be stored into the last outbound dlc messages table. This will ensure as well that once we create a response we won't lose it anymore if something goes wrong afterwards.

Once the processing was successful the last dlc message is sent.

resolves #2127


luckysori (commented):
LGTM, although I think this is a `fix`, not a `chore`.

luckysori (approved):

PR #2306, opened 2024-03-22
fix: allow to overwrite flutter version

bonomat:
This is needed for CI where we can't interact with the terminal and answer the question with `yes`.

For local flutter version and channel use

```
just gen
```

For overwriting the channel use

```
just gen beta
```


bonomat:
Disclaimer: better approach would be to use `flavors` 

https://fvm.app/documentation/guides/project-flavors

I'll create a follow-up ticket for this but this PR should help

holzeis (approved):

Issue #2309, opened 2024-03-23
The max button always leads to a dlc protocol error


holzeis:
Environment: test

1. Open a new channel
2. Choose max on the channel configuration.

For me this always ended up in a dlc protocol error. I presume the max does not correctly consider on-chain and order matching fees.

![image](https://github.com/get10101/10101/assets/382048/f4b08dfd-5703-409c-b1fe-366a6a50f5e8)

luckysori:
I wasn't able to reproduce this. Can you provide more context, @holzeis?

bonomat:
I wasn't able to reproduce it neither. 

1. closed existing channel
2. send out all funds but 500k sats
3. opened a channel with max amount. 
success

luckysori:
@holzeis: Philipp and I were able to reproduce it and figure out why this is happening.

1. Create fresh app.
2. Receive _two_ UTXOs. Keep the sum of both UTXOs below the actual maximum determined by the coordinator i.e. 6M sats.
3. Attempt to open a channel with maximum collateral.

Run into the error:

```
2024-03-25T02:22:45.103055619Z ERROR process_dlc_channel_offer{channel_id="70feedddf1f857a2d909906c8251a46eee27e5163615d102de7d840dbdaec424"}: native::ln_dlc::node: Failed to accept dlc channel offer. Wallet error Wallet does not hold enough UTXOs to cover amount 1009559 with fee rate 3. No bnb solution found after 0 rounds (max rounds is 100000).
```

The problem arises because we use an estimate of the funding TX fee based on a transaction with a single input per party. This is okay because the estimate was meant to be used... well, as an estimate for informational purposes.

Now that we are using this value for something more important we need to be a bit more conservative. I propose to simply multiply this value by two to cover our bases. The user will not be missing out on that much collateral.

PR #2315, opened 2024-03-25
Overestimate funding TX fee when calculating max usable on-chain balance

luckysori:
Fixes https://github.com/get10101/10101/issues/2309.

With an overestimation we can be sure that the user won't run into problems during coin selection. Obviously they will not be able to use the true maximum, but the difference is small. We can never realistically hit the true maximum anyway.

### DIFF
--- a/mobile/lib/features/trade/channel_configuration.dart
+++ b/mobile/lib/features/trade/channel_configuration.dart
@@ -141,8 +141,12 @@ class _ChannelConfiguration extends State<ChannelConfiguration> {
 
   @override
   Widget build(BuildContext context) {
+    // We add a buffer because the `fundingTxFee` is just an estimate. This
+    // estimate will undershoot if we end up using more inputs or change
+    // outputs.
+    final Amount fundingTxFeeWithBuffer = Amount(fundingTxFee.sats * 2);
     final maxUsableOnChainBalance =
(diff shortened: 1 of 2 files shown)


mobile/lib/features/trade/channel_configuration.dart
```diff
@@ -141,8 +141,12 @@ class _ChannelConfiguration extends State<ChannelConfiguration> {
 
   @override
   Widget build(BuildContext context) {
+    // We add a buffer because the `fundingTxFee` is just an estimate. This
+    // estimate will undershoot if we end up using more inputs or change
+    // outputs.
+    final Amount fundingTxFeeWithBuffer = Amount(fundingTxFee.sats * 2);
     final maxUsableOnChainBalance =
```
bonomat, line 148: Could we end up with dust value because of this ?

Let's assume the estimated fee is 1000 sats. If we overestimate * 2, we get 2000 sats estimate. We end up needing 1200 sats and 800 will remain.

bonomat (approved):

mobile/lib/features/trade/channel_configuration.dart
luckysori: It's a good question. I think so, although it should be the responsibility of the coin selection algorithm to ensure that no dust change output is generated.

PR #2317, opened 2024-03-25
Compute confirmations correctly

### LINKED ISSUE #2222: App shows one confirmation less
|	|	|
|:---:|:----:|
| image | image |

luckysori:
Fixes #2222.

The original logic was flawed because it would correctly show 1 confirmation when the transaction was included in a block, but it would subsequently deduct 1 confirmation after the second confirmation.

We were already doing this correctly for the `Blockchain` struct, so there is nothing to fix there.


holzeis (approved):
Thanks 👍

PR #2319, opened 2024-03-25
fix: Return error message on failed post order

### LINKED ISSUE #2310: User gets a bad request error instead of the error message
Environment: test

If a user with an older version tries to trade with the coordinator the request is rejected with a bad request.

Instead of the http error the error message should be shown.

More specifically the user should see that a new version is available if the request gets rejected due to the version check.

image

holzeis:
fixes #2310 

image


bonomat (approved):
I just wanted to fix the same thing... 😅

PR #2323, opened 2024-03-25
fix: Rename trader_realized pnl to coordinator_realized pnl

holzeis:


bonomat (commented):
Note: this was changed on purpose in the past because it has been wrong. Let me investigate what's happening...

luckysori:
Superseded by https://github.com/get10101/10101/pull/2325. It wasn't just a naming problem.

PR #2325, opened 2024-03-26
Calculate coordinator position PNL correctly

luckysori:
We were using the closing trade trader direction instead of the position direction.


bonomat (approved):
Thank you for fixing this.

holzeis:
> We were using the closing trade trader direction instead of the position direction.

@bonomat that might have been the reason why our leaderboard was not correct.

bonomat:
> > We were using the closing trade trader direction instead of the position direction.
> 
> @bonomat that might have been the reason why our leaderboard was not correct.

Maybe, although, it was correct for some time until someone introduced the bug again ~2 weeks ago.

PR #2326, opened 2024-03-26
feat: introduce semantic versioning

bonomat:
With this patch we say `patch`-releases are compatible with older versions

### DIFF
--- a/coordinator/src/check_version.rs
+++ b/coordinator/src/check_version.rs
@@ -5,17 +5,104 @@ use anyhow::Context;
 use anyhow::Result;
 use bitcoin::secp256k1::PublicKey;
 use diesel::PgConnection;
+use semver::Comparator;
+use semver::Op;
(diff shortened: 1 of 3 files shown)


coordinator/src/check_version.rs
```diff
@@ -5,17 +5,104 @@ use anyhow::Context;
...
+    let req = VersionReq {
+        comparators: vec![
+            Comparator {
+                op: Op::Exact,
+                major: coordinator_version.major,
+                minor: Some(coordinator_version.minor),
+                patch: None,
+                pre: Default::default(),
+            },
+            // We introduced sematic versioning only after this version, hence, we say each app
```
luckysori, line 37:
```suggestion
            // We introduced semantic versioning only after this version, hence, we say each app
```

luckysori (approved):

PR #2332, opened 2024-03-26
feat: Add max quantity button

holzeis:
Clicking the max button next to the quantity input field will lock the quantity to stay on the max value.

https://github.com/get10101/10101/assets/382048/01131a53-175d-41c8-9240-30c75819e3bb

### DIFF
--- a/mobile/lib/features/trade/domain/trade_values.dart
+++ b/mobile/lib/features/trade/domain/trade_values.dart
@@ -13,6 +13,7 @@ class TradeValues {
   double? price;
   double? liquidationPrice;
   Amount? fee; // This fee is an estimate of the order-matching fee.
+  Usd? maxQuantity;
 
   double fundingRate;
   DateTime expiry;
@@ -20,25 +21,27 @@ class TradeValues {
   // no final so it can be mocked in tests
   TradeValuesService tradeValuesService;
 
-  TradeValues(
-      {required this.direction,
-      required this.margin,
-      required this.quantity,
-      required this.leverage,
-      required this.price,
-      required this.liquidationPrice,
-      required this.fee,
-      required this.fundingRate,
-      required this.expiry,
-      required this.tradeValuesService});
+  TradeValues({
+    required this.direction,
+    required this.margin,
+    required this.quantity,
+    required this.leverage,
+    required this.price,
+    required this.liquidationPrice,
+    required this.fee,
+    required this.fundingRate,
+    required this.expiry,
+    required this.tradeValuesService,
+  });
 
-  factory TradeValues.fromQuantity(
-      {required Usd quantity,
-      required Leverage leverage,
-      required double? price,
-      required double fundingRate,
-      required Direction direction,
-      required TradeValuesService tradeValuesService}) {
+  factory TradeValues.fromQuantity({
+    required Usd quantity,
+    required Leverage leverage,
+    required double? price,
+    required double fundingRate,
+    required Direction direction,
+    required TradeValuesService tradeValuesService,
+  }) {
     Amount? margin =
         tradeValuesService.calculateMargin(price: price, quantity: quantity, leverage: leverage);
     double? liquidationPrice = price != null
@@ -63,13 +66,14 @@ class TradeValues {
         tradeValuesService: tradeValuesService);
   }
 
-  factory TradeValues.fromMargin(
-      {required Amount? margin,
-      required Leverage leverage,
-      required double? price,
-      required double fundingRate,
-      required Direction direction,
-      required TradeValuesService tradeValuesService}) {
+  factory TradeValues.fromMargin({
+    required Amount? margin,
+    required Leverage leverage,
+    required double? price,
+    required double fundingRate,
+    required Direction direction,
+    required TradeValuesService tradeValuesService,
+  }) {
     Usd? quantity =
         tradeValuesService.calculateQuantity(price: price, margin: margin, leverage: leverage);
     double? liquidationPrice = price != null
@@ -104,6 +108,7 @@ class TradeValues {
     this.margin = margin;
     _recalculateQuantity();
     _recalculateFee();
+    _recalculateMaxQuantity();
   }
 
   updatePriceAndQuantity(double? price) {
@@ -111,6 +116,7 @@ class TradeValues {
     _recalculateQuantity();
     _recalculateLiquidationPrice();
     _recalculateFee();
+    _recalculateMaxQuantity();
   }
 
   updatePriceAndMargin(double? price) {
@@ -118,12 +124,14 @@ class TradeValues {
     _recalculateMargin();
     _recalculateLiquidationPrice();
     _recalculateFee();
+    _recalculateMaxQuantity();
   }
 
   updateLeverage(Leverage leverage) {
     this.leverage = leverage;
     _recalculateMargin();
     _recalculateLiquidationPrice();
+    _recalculateMaxQuantity();
   }
 
   // Can be used to calculate the counterparty's margin, based on their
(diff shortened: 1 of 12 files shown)


luckysori (approved):
Concept ACK.

I am not completely sold on the sticky nature of the `max` button though. I'm nevertheless happy to try it and see what others think.

bonomat (commented):
I'm not a big fan of having all those services in `TradeValues` (which looks to me like a model) :/ Can we move them one layer up into the `TradeValueChangeNotifier`?

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -227,7 +230,29 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
           children: [
             Flexible(
                 child: AmountInputField(
-              initialValue: Amount(tradeValues.quantity?.toInt ?? 0),
+              controller: quantityController,
+              suffixIcon: TextButton(
+                onPressed: () {
```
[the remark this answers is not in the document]
holzeis: 🤔 Why do you think so? That would be an additional click if you want to change the quantity.

holzeis:
> Concept ACK.
> 
> 
> 
> I am not completely sold on the sticky nature of the `max` button though. I'm nevertheless happy to try it and see what others think.

I've decided to add the stickiness as I felt the UX would suffer from the max button.

E.g. you hit the max button and the price changes. Now the max quantity isn't valid anymore. It happened to me several times during testing and it was annoying 🙈. Note, it might have happened to me more often because I artificially changed the price every 5 seconds with bigger jumps.

mobile/lib/features/trade/domain/trade_values.dart
```diff
@@ -153,4 +179,59 @@ class TradeValues {
...
+
+    // Substract required on-chain fees with buffer if the trade is opening a channel.
+    Amount maxCounterpartyMarginSats =
+        Amount(channelTradeConstraints.maxCounterpartyMarginSats).sub(onChainFees);
+    Amount maxLocalMarginSats = Amount(channelTradeConstraints.maxLocalMarginSats).sub(onChainFees);
+
+    Usd? quantity = Usd.zero();
+    if (maxLocalMarginSats.sats * leverage.leverage >
+        maxCounterpartyMarginSats.sats * channelTradeConstraints.coordinatorLeverage) {
+      // if the local balance is greater than the coordinators balance we don't have
```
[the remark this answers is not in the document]
holzeis: > What fee are you talking about and why would the coordinator get this fee from the trader if the trader has more money?

I am talking about the order matching fees (see else path). The coordinator doesn't have to care if he has enough funds for the order matching fees as they are paid by the trader. But there might be an issue of the trader has more than the coordinator but less than the order matching fees. I will move that part to rust and add some testing to it..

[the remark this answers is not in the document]
holzeis: > Also, why do you multiply his balance times the leverage?

To get the leverage perspective on to the balance. e.g. if the local balance is 400k and the coordinator balance is 500k, we must not use the local balance if the user has a leverage of 5. As the coordinator would need to put up 400k*2 which is >= 500k.

holzeis:
@luckysori @bonomat I did quite some changes, can you please have another look? Thanks :)

bonomat:
> Note, it might have happened to me more often because I artificially changed the price every 5 seconds with bigger jumps.

This might also happen in production because price might change quite fast now.

mobile/native/src/max_quantity.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    Ok(max_quantity)
+}
+
+fn calculate_max_quantity(
+    price: Decimal,
+    max_coordinator_margin: Amount,
+    max_trader_margin: Amount,
+    on_chain_fee_estimate: Option<Amount>,
+    coordinator_leverage: f32,
```
bonomat, line 45: eventually we should get rid of f32 and use Decimal in as many places as possible.

bonomat (approved):
Concept Ack.

I did not verify the calculations in rust. If you want me to double check that, please hold back with merging... particularly since you want this to be in the next release. 
I'll see to give it a nother round tomorrow. 

Otherwise, YOLO.

holzeis:
> Concept Ack.
> 
> 
> 
> I did not verify the calculations in rust. If you want me to double check that, please hold back with merging... particularly since you want this to be in the next release. 
> 
> I'll see to give it a nother round tomorrow. 
> 
> 
> 
> Otherwise, YOLO. 

😂 I'll wait for your thorough review. 😊

mobile/native/src/api.rs
```diff
@@ -327,6 +329,17 @@ pub fn order_matching_fee(quantity: f32, price: f32) -> SyncReturn<u64> {
     SyncReturn(order_matching_fee)
 }
 
+/// Calculates the max quantity the user is able to trade considering the trader and the coordinator
+/// balances and constraints. Note, this is not an exact maximum, but a very close approximation.
+pub fn calculate_max_quantity(price: f32, trader_leverage: f32) -> SyncReturn<u64> {
+    let price = Decimal::from_f32(price).expect("price to fit in Decimal");
+
+    let max_quantity = max_quantity(price, trader_leverage).unwrap_or(Decimal::ZERO);
```
bonomat, line 337: Should we print the error if we get one?

bonomat (approved):
Approved. 
Can we have some comments please showing how to manually calculate the numbers you use in the tests. This can be very helpful if it suddenly breaks. Right now the "formula" is only in the code.

mobile/native/src/api.rs
holzeis: I don't think its very useful. If this errors we would see that message already in other places.

luckysori: Why not error here?

luckysori (commented):
I'm not sure I get why we need a0dc14bea8a7cdd3e2728b5912e08aa2fe9a8abc.

It's weird to have to handle the possibility that the node is not available in the backend, when the node is pretty much the backend. I would be keener on ensuring that we don't move forward until the node is actually available.

holzeis:
> I'm not sure I get why we need a0dc14bea8a7cdd3e2728b5912e08aa2fe9a8abc.
> 
> 
> 
> It's weird to have to handle the possibility that the node is not available in the backend, when the node is pretty much the backend. I would be keener on ensuring that we don't move forward until the node is actually available.

We subscribe to the websocket before starting the node.. that means we can get price updates before the node is started. For some price related updates we are using the node, which can lead to panics due to race conditions where the node is not set in the state yet.

Also I think we should be more resilient. The app is still crashing too often for insignificant things. E.g. all those functions where I've added a handling, I think we can continue even if this is returning a default value.

luckysori:
> We subscribe to the websocket before starting the node

Perhaps a more reliable fix would be to not do that.

> Also I think we should be more resilient. The app is still crashing too often for insignificant things. E.g. all those functions where I've added a handling, I think we can continue even if this is returning a default value.

I suppose that's fair. I'm not a big fan of having to check that a global variable is available every time I use it, when its presence should be treated as an invariant. I think this is just covering for programming mistakes.

holzeis:
> > We subscribe to the websocket before starting the node
> 
> 
> 
> Perhaps a more reliable fix would be to not do that.

It's a chicken and egg issue, we have some actions that already respond on the websocket connection upon a p2p connection. I guess the clean fix would be not to use different communication channels.

> I suppose that's fair. I'm not a big fan of having to check that a global variable is available every time I use it, when its presence should be treated as an invariant. I think this is just covering for programming mistakes.

Agree, it's definitely masking a programming mistake.

PR #2333, opened 2024-03-27
feat(app): show latest price in trading overview

### LINKED ISSUE #2258: Show price on trading screen
User feedback: 
it would be good to see the price on the trading screen where you see your position. Not just when you click the buy/sell button.

bonomat:
Resolves #2258 

I've implemented this while debugging our latest issues. 

image

### DIFF
--- a/mobile/lib/features/trade/trade_screen.dart
+++ b/mobile/lib/features/trade/trade_screen.dart
@@ -1,11 +1,13 @@
 import 'package:candlesticks/candlesticks.dart';
 import 'package:flutter/material.dart';
+import 'package:get_10101/common/domain/model.dart';
 import 'package:get_10101/features/trade/candlestick_change_notifier.dart';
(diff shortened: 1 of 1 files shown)


mobile/lib/features/trade/trade_screen.dart
```diff
@@ -256,3 +275,26 @@ class TradeScreen extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    return RichText(
+      text: TextSpan(
+        text: label,
+        style: DefaultTextStyle.of(context).style,
+        children: [
+          TextSpan(
+            text: "\$${price.formatted()}",
```
holzeis, line 293: 🔧 i think the to string implement action is doing that already so you could simply 

```suggestion
            text: "$price",
```

holzeis (approved):

PR #2334, opened 2024-03-28
chore: Add ios compatibility to starting from 12.0

holzeis:
I was only able to test this with an ios version 15.0 as older version are not available to download for me anymore. I would still keep compatibility to 12.0 as it may still work.


luckysori (approved):
Nice!

PR #2336, opened 2024-03-28
feat: Add order matching fee to trades

holzeis:


coordinator/src/dlc_protocol.rs
```diff
@@ -392,6 +393,12 @@ impl DlcProtocolExecutor {
                 .map_err(|_| RollbackTransaction)?,
         );
 
+        // TODO(holzeis): its a bit ugly to calculate the order matching fee like that. The user
+        // might get a discount at some point. So ideally the order matching fee is decided on the
+        // trade params at a single point and then only stored here.
```
bonomat, line 398: You could already pull this up and provide it as a param because we have this information available in `open_position` and `open_dlc_channel`. Only in `start_closing_position` I couldn't find it but maybe I just missed it.

bonomat (approved):
One improvement proposal, otherwise looks good to me.

coordinator/src/dlc_protocol.rs
holzeis: I fear it's not that simple. We'd have to add the order matching fee to the `Matches` and the `TradeParams` table and then pass it through.

This seems to be quite some changes. I'll do this in a follow up PR.

holzeis: Also that order matching fee should then be taken when calculating the payout curve.

PR #2338, opened 2024-03-28
feat: Add trader realized pnl to trades

### LINKED ISSUE #2322: Add realized pnl to trade
At the moment we keep the realized pnl on the position. As we are planning to move away from having multiple positions, we should move the realized pnl to the trade.

holzeis:
fixes #2322


bonomat (approved):
LGTM

Issue #2290, opened 2024-03-21
Don't derive `smart` channel states


bonomat:
Right now in the webapp we are overly smart and derive some human readable states. We sholdn't do this because it makes debugging harder

<img width="336" alt="image" src="https://github.com/get10101/10101/assets/224613/994f7988-7500-4231-9441-18f3b7b070e0">

PR #2342, opened 2024-03-29
feat(webapp): Create DLC Channel Dialog

bonomat:
it took me way longer than I'd like to admit it :/ 

resolves https://github.com/get10101/10101/issues/2279
resolves https://github.com/get10101/10101/issues/2290

https://github.com/get10101/10101/assets/224613/b2df9c71-8330-4924-8634-e296d11075a0


webapp/frontend/lib/common/calculations.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:get_10101/common/model.dart';
+import 'package:get_10101/services/quote_service.dart';
+
+Amount calculateFee(Usd? quantity, BestQuote? quote, bool isLong) {
```
holzeis, line 4: hmm.. I guess it's reasonable, you don't want to make a backend call for that, but yet we didn't want this kind of logic in dart.

I guess there is no way around that. 🤔

holzeis (approved):

webapp/frontend/lib/common/calculations.dart
bonomat: Unfortunately not. I think this is better then doing a roundtrip to the backend for every re-calculation. This could get quite expensive.

webapp/frontend/lib/services/proportional_fee.g.dart
```diff
@@ -0,0 +1,17 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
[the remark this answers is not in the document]
bonomat: Yes. In case you ever need to re-generate it
```bash
fvm dart run build_runner build --delete-conflicting-outputs
```

PR #2344, opened 2024-03-30
chore: Move trading pair to app bar title

holzeis:
I think it looks better if the trading pair is shown in the app bar instead directly below it. It makes the components a little bit better aligned. Happy to drop that PR if you disagree.. (but I hope you don't :))

| before |	after |
|:-:|:-:|
| IMG_A3F1E78B6A77-1 | image |


bonomat (approved):
Very nice

PR #2345, opened 2024-03-30
chore: Add dlc channel event to flutter

holzeis:
This allows us to subscribe to dlc channel events on the flutter side, which also allows us to get rid of the polling refresh dlc channel function.

It is also a preparation step for shadowing the dlc channel on the app side.


mobile/native/src/api.rs
```diff
@@ -814,6 +815,7 @@ pub fn get_dlc_channel_id() -> Result<Option<String>> {
 pub fn list_dlc_channels() -> Result<Vec<DlcChannel>> {
     let channels = ln_dlc::list_dlc_channels()?
         .iter()
+        .map(dlc::DlcChannel::from)
         .map(DlcChannel::from)
```
bonomat, line 819: why double mapping?

bonomat (approved):
LGTM. Eventually we should add some form of websocket to DLC Connect and subscribe to all of those events

mobile/native/src/api.rs
holzeis: Because I didn't want to write another from implementation for `dlc_channel::DlcChannel` to `api::DlcChannel`.


We have `dlc_channel::DlcChannel` > `dlc::DlcChannel` > `api::DlcChannel`.

PR #2356, opened 2024-04-04
Fix: Close open position if dlc channel got force closed

### LINKED ISSUE #2106: Position is not set to closed if the underlying dlc channel got force closed.
As a result the coordinator tries to close the eventually expired position. Which is producing a significant amount of failed orders.

Screenshot 2024-02-27 at 14 10 35

see demo environment.

holzeis:
fixes #2106

### DIFF
--- a/coordinator/src/node/channel.rs
+++ b/coordinator/src/node/channel.rs
@@ -4,7 +4,9 @@ use crate::dlc_protocol::ProtocolId;
 use crate::node::Node;
 use anyhow::bail;
 use anyhow::Result;
+use bitcoin::secp256k1::PublicKey;
 use bitcoin::Amount;
+use bitcoin::Txid;
 use dlc_manager::channel::signed_channel::SignedChannel;
 use dlc_manager::channel::signed_channel::SignedChannelState;
 use dlc_manager::channel::Channel;
@@ -12,10 +14,12 @@ use dlc_manager::channel::ClosedChannel;
 use dlc_manager::channel::ClosedPunishedChannel;
 use dlc_manager::channel::ClosingChannel;
 use dlc_manager::channel::SettledClosingChannel;
+use dlc_manager::DlcChannelId;
 use ln_dlc_node::bitcoin_conversion::to_secp_pk_30;
 use ln_dlc_node::bitcoin_conversion::to_txid_30;
 use ln_dlc_node::node::event::NodeEvent;
 use ln_dlc_storage::DlcChannelEvent;
+use time::OffsetDateTime;
 use tokio::sync::broadcast::error::RecvError;
 
 pub enum DlcChannelState {
@@ -25,6 +29,24 @@ pub enum DlcChannelState {
     Closed,
     Failed,
     Cancelled,
+}
+
+pub struct DlcChannel {
+    pub channel_id: DlcChannelId,
+    pub trader: PublicKey,
+    pub channel_state: DlcChannelState,
+    pub trader_reserve_sats: Amount,
+    pub coordinator_reserve_sats: Amount,
+    pub coordinator_funding_sats: Amount,
+    pub trader_funding_sats: Amount,
+    pub funding_txid: Option<Txid>,
+    pub close_txid: Option<Txid>,
+    pub settle_txid: Option<Txid>,
+    pub buffer_txid: Option<Txid>,
+    pub claim_txid: Option<Txid>,
+    pub punish_txid: Option<Txid>,
+    pub created_at: OffsetDateTime,
+    pub updated_at: OffsetDateTime,
 }
 
 impl Node {
@@ -40,7 +62,7 @@ impl Node {
                             if let Err(e) = node.process_dlc_channel_event(dlc_channel_event) {
                                 tracing::error!(
                                     ?dlc_channel_event,
-                                    "Failed to process dlc channel event. Error: {e:#}"
+                                    "Failed to process DLC channel event. Error: {e:#}"
                                 );
                             }
                         }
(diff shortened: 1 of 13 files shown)


coordinator/src/node/channel.rs
```diff
@@ -38,7 +38,10 @@ impl Node {
                     match receiver.recv().await {
                         Ok(NodeEvent::DlcChannelEvent { dlc_channel_event }) => {
                             if let Err(e) = node.process_dlc_channel_event(dlc_channel_event) {
-                                tracing::error!("Failed to get connection. Error: {e:#}");
+                                tracing::error!(
+                                    ?dlc_channel_event,
+                                    "Failed to process dlc channel event. Error: {e:#}"
```
luckysori, line 43: 🙃

```suggestion
                                    "Failed to process DLC channel event. Error: {e:#}"
```

coordinator/src/position/mod.rs
```diff
@@ -1 +1,208 @@
...
+        dlc_channel_event: DlcChannelEvent,
+    ) -> Result<()> {
+        let mut conn = self.pool.get()?;
+
+        let protocol_id = match dlc_channel_event.get_reference_id() {
+            Some(reference_id) => reference_id,
+            None => {
+                bail!("Can't process dlc channel event without reference id. dlc_channel_event = {dlc_channel_event:?}");
+            }
+        };
```
luckysori, line 67: 🔧 It's nicer to use `with_context()?`.

coordinator/src/position/mod.rs
```diff
@@ -1 +1,208 @@
...
+                    to_secp_pk_30(trader_id),
+                    vec![PositionState::Open],
+                )? {
+                    Some(position) => position,
+                    None => return Ok(()),
+                };
+
+                let closing_price = match position.is_expired() {
+                    true => {
+                        let oracle = self.inner.oracles.first().context("Missing oracle")?;
```
luckysori, line 85: ❓ How do we know that it is the first oracle?

coordinator/src/position/mod.rs
```diff
@@ -1 +1,208 @@
...
+                    calculate_pnl(
+                        Decimal::from_f32(position.average_entry_price)
+                            .expect("to fit into decimal"),
+                        closing_price,
+                        position.quantity,
+                        trader_position_direction,
+                        initial_margin_long as u64,
+                        initial_margin_short as u64,
+                    )?
+                };
```
luckysori, line 155: 📣 Surely we should get the PNL from the CET or the settle transaction if we get force-closed. Using this function won't be accurate.

mobile/native/src/position.rs
```diff
@@ -0,0 +1,50 @@
...
+use crate::dlc::ChannelState;
+use crate::dlc::DlcChannel;
+use crate::dlc::SignedChannelState;
+use crate::event;
+use crate::event::subscriber::Subscriber;
+use crate::event::EventInternal;
+use crate::event::EventType;
+use trade::ContractSymbol;
+
+#[derive(Clone)]
```
luckysori, line 11: 🔧 I guess it can be `Copy` 😆 But is this even needed?

holzeis: Yes it's needed for the event hub.

coordinator/src/position/mod.rs
holzeis: We are only interested in the CET as for the settle transaction the position must have been closed.

How would you get the pnl of that position from the CET? Doesn't it include the reserve?

mobile/native/src/position.rs
luckysori: It's weird to have to `Clone` and empty type.

coordinator/src/position/mod.rs
luckysori: > We are only interested in the CET as for the settle transaction the position must have been closed.

That's not true. A settled channel (one without a DLC in it) can be force-closed too.

> How would you get the pnl of that position from the CET? Doesn't it include the reserve?

You would take the payout and subtract the reserve, which is now stored in the `dlc_channels` table.

What you are doing here is an approximation, which maybe is good enough, but I wasn't sure if that's what we wanted to do.

holzeis: > You would take the payout and subtract the reserve, which is now stored in the `dlc_channels` table.
> 
> 
> 
> What you are doing here is an approximation, which maybe is good enough, but I wasn't sure if that's what we wanted to do.

I think that would work as well, but then the pnl would include on-chain fees. That wouldn't be consistent with the pnl calculation of other positions.

I was thinking to calculate the fees per channel differently. But I may be missing something.

luckysori: > I think that would work as well, but then the pnl would include on-chain fees. That wouldn't be consistent with the pnl calculation of other positions.

The CET payouts are not reduced by on-chain fees. The on-chain fees come from the fee reserve. See https://github.com/p2pderivatives/rust-dlc/blob/883ad23f4c6467c90754ccce0b8820db10479f08/dlc/src/lib.rs#L572-L592.

mobile/native/src/position.rs
holzeis: Its because of this `subscribers: HashMap<EventType, Vec<Box<dyn Subscriber + 'static + Send + Sync>>>,` weirdness.

holzeis: And I think this is needed because we keep the event hub in the storage state.

holzeis:
@luckysori I changed the `trader_realized_pnl` calculation to use the cet payouts derived from the closed contract. It's much nicer now because we don't have to query the oracle again. Can you have another look?

PR #2357, opened 2024-04-05
Remove unneeded Flutter dependency override

luckysori:
Since we no longer use `Flogs`, we don't need this dependency override[^1]. It was creating problems in CI[^2] (and locally too I suppose).

[^1]: https://github.com/zubairehman/Flogs/issues/87.
[^2]: https://github.com/get10101/10101/actions/runs/8565195140/job/23472971318?pr=2352#step:9:357.


holzeis (approved):
Thanks 👍

PR #2359, opened 2024-04-05
Reintroduce position resizing

luckysori:
Fixes https://github.com/get10101/10101/issues/1791.

https://github.com/get10101/10101/assets/9418575/44f31bf2-a752-4a16-a07c-3e5abd5188ed

- I fixed calculation bugs in our original implementation of position resizing that we used for LN-DLC channels. This is why some assert values changed.

- I reintroduced the resizing e2e tests, which are now much more helpful because they assert on the specific values of position and trade for the app. This can be further improved by asserting on the coordinator side. The same pattern should eventually be applied to most e2e tests.

- I had to add an `is_complete` column to the coordinator's `trade` table to be able to record the `trader_realized_pnl_sat` ahead of time.

Limitations:

- The app must have an order in `Filling` to be able to update the position after we get a resizing channel renewal offer. This pattern has caused problems in the past. We will probably fix this my sending extra metadata with the `rust-dlc` offer message.


coordinator/migrations/2024-03-18-083352_add_order_matching_fees_to_position/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE positions ADD COLUMN "order_matching_fees" BIGINT;
```
holzeis, line 1: ❓ Can't we derive the order matching fee from the data on the position support resizing of already existing positions?

coordinator/src/position/models.rs
```diff
@@ -87,6 +88,14 @@ pub struct Position {
 
     pub position_state: PositionState,
 
+    /// Accumulated order matching fees for the lifetime of the position.
+    ///
+    /// We use an [`Option`] to maintain backwards-compatibility with older positions. This is
+    /// important because we cannot safely resize old positions, since we do not know the accrued
+    /// order matching fees. Without this information the coordinator could end up losing these
+    /// fees.
+    pub order_matching_fees: Option<Amount>,
```
holzeis, line 94: I don't think we need to use an option here if we write a simple migration script. The position contains all data necessary to calculate the `order_matching_fees`.

mobile/lib/features/welcome/welcome_screen.dart
```diff
@@ -218,6 +219,11 @@ class _WelcomeScreenState extends State<WelcomeScreen> {
   void initState() {
     super.initState();
 
+    if (Environment.parse().network == "regtest") {
```
holzeis, line 222: Nice :)

mobile/lib/features/trade/domain/trade_values.dart
```diff
@@ -163,6 +163,6 @@ class TradeValues {
   }
 
   _recalculateMaxQuantity() {
-    maxQuantity = tradeValuesService.calculateMaxQuantity(price: price, leverage: leverage);
```
holzeis, line 166: Should be fixed with https://github.com/get10101/10101/pull/2362

coordinator/src/trade/mod.rs
```diff
@@ -520,7 +563,179 @@ impl TradeExecutor {
...
+        };
+
+        let temporary_contract_id = self
+            .node
+            .inner
+            .propose_dlc_channel_update(&dlc_channel_id, contract_input, protocol_id.into())
+            .await
+            .context("Could not propose DLC channel update")?;
+
+        // TODO: Shouldn't we include actually starting the DLC protocol in the DB transaction in
```
holzeis, line 708: Since we use different transactional systems, postgres and sled I think it doesn't matter. But once we are also using postgres that would be great. As we could atomically execute both rust dlc and 10101 changes.

coordinator/src/trade/mod.rs
```diff
@@ -740,18 +959,50 @@ impl TradeExecutor {
...
+                            channel_id,
+                            position: Box::new(position),
+                            resize_action: ResizeAction::ChangeDirection {
+                                contracts_new_direction,
+                                average_execution_price,
+                            },
+                        }
                     }
-                } else {
-                    ensure!(
```
holzeis, line 745: 🔧 this got lost and should be checked for every resize action.

mobile/native/src/trade/position/handler.rs
```diff
@@ -43,34 +41,50 @@ pub fn get_position_matching_order(order: &Order) -> Result<Option<Position>> {
...
     if let Some(position) = db::get_positions()?.first() {
-        tracing::debug!("Setting position to rollover");
-        db::rollover_position(position.contract_symbol, expiry_timestamp)?;
-        let mut position = position.clone();
-        position.position_state = PositionState::Rollover;
-        position.expiry = expiry_timestamp;
-        event::publish(&EventInternal::PositionUpdateNotification(position));
+        // Assume that if there is an order in filling we are dealing with position resizing.
+        //
+        // TODO: This is garbage as we very well know. Any other ideas? We could generate
```
holzeis, line 63: @bonomat and I were talking before, that it would probably make sense to add a custom message to wrap the renew protocol for rollover, resizing and async matching.. and enrich the message with some 10101 meta data. that would also allow us to get rid of the ugly websocket communication to send over some additional data.

holzeis (approved):
LGTM 👍

coordinator/migrations/2024-03-18-083352_add_order_matching_fees_to_position/up.sql
luckysori: I think you are right, working on it!

mobile/lib/features/trade/domain/trade_values.dart
luckysori: I will drop the commit before merging!

coordinator/src/trade/mod.rs
luckysori: Good catch, thanks!

mobile/native/src/trade/position/handler.rs
luckysori: That sounds good, but I think that belongs in a different PR. Given your approval, I assume you are okay with not tackling this TODO right now.

coordinator/src/trade/mod.rs
```diff
@@ -513,7 +568,205 @@ impl TradeExecutor {
...
+                // TODO: To be removed.
+                coordinator_margin: 0,
+                trader_direction: trade_params.direction,
+                average_price: trade_params
+                    .average_execution_price()
+                    .to_f32()
+                    .expect("to fit"),
+                order_matching_fee,
+                trader_realized_pnl_sat: realized_pnl.map(SignedAmount::to_sat),
+                is_complete: false,
```
holzeis, line 761: 🤔 It looks to me the only reason why you'd have to create the trade already here is because of the `trader_realized_pnl_sat`. Why not simply add that to the `trade_params` table which is anyways just temporary and intended to remember the state from when the protocol starts to when it ends.

I dislike the `is_complete` flag because

1. It introduces a state on the trade.
2. It is not consistent with the rest of the implementation. e.g. if we want to keep the `is_complete` flag there is no reason for a technical `trade_params` table. We could remove it and create the trade immediately with the required data and set it to "incomplete"

I am not saying we have to stay with the `trade_params` but mixing the two concepts feels wrong to me.

luckysori: You might be right. 

It seemed wrong to use the `trade_params` table because the PNL is not a trade parameter. It's the result of applying the trade to a position.

I agree that it is inconsistent. I can try to fix it, but the priority for me is fairly low.

luckysori: Did it anyway: https://github.com/get10101/10101/pull/2394.

holzeis: Thanks

PR #2372, opened 2024-04-08
chore: Rename to maintenance margin rate

holzeis:
follow up to review remarks on https://github.com/get10101/10101/pull/2354


luckysori (approved):
Thanks!

PR #2373, opened 2024-04-08
chore: Fix liquidation migration script for coordinator leverage = 1

holzeis:
follow up to review remarks on https://github.com/get10101/10101/pull/2354

### DIFF
--- a/coordinator/migrations/2024-04-03-105900_add_coordinator_liquidation_price/up.sql
+++ b/coordinator/migrations/2024-04-03-105900_add_coordinator_liquidation_price/up.sql
@@ -2,4 +2,6 @@ ALTER TABLE "positions"
     ADD COLUMN coordinator_liquidation_price REAL NOT NULL DEFAULT 0;
 
 UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage + 1) where trader_direction='short';
-UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage - 1) where trader_direction='long';
+UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage - 1) where trader_direction='long' and coordinator_leverage > 1;
+UPDATE positions SET coordinator_liquidation_price = 1048575 where trader_direction='long' and coordinator_leverage = 1;
+


coordinator/migrations/2024-04-03-105900_add_coordinator_liquidation_price/up.sql
```diff
@@ -2,4 +2,5 @@ ALTER TABLE "positions"
     ADD COLUMN coordinator_liquidation_price REAL NOT NULL DEFAULT 0;
 
 UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage + 1) where trader_direction='short';
-UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage - 1) where trader_direction='long';
+UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage - 1) where trader_direction='long' and coordinator_leverage > 1;
+UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage where trader_direction='long' and coordinator_leverage = 1;
```
luckysori, line 6: 🤔 The formula doesn't look right to me. Elsewhere we've set the liquidation price of a party going short at 1x leverage to the maximum price the oracle can attest to. I'm not sure if that's correct though cc @bonomat.

luckysori (commented):
Not sure about the formula.

coordinator/migrations/2024-04-03-105900_add_coordinator_liquidation_price/up.sql
bonomat: Yea, if a party `goes` short with a leverage of 1 he basically never gets liquidated. So we can set this to a very high number. 

On the contrary, if a party goes `long` with a leverage of 1 his liquidation price is at about 50%, i.e. 
`(average_entry_price*leverage)/((leverage+1-(maintenance_margin*leverage)))` --> `average_entry_price/2`

holzeis: Can I get an approval then? 🙂

luckysori: It still looks wrong to me though. AFAICT, you're setting the coordinator's liquidation price to

```
average_entry_price * 1
```

if the coordinator goes short.

bonomat, line 6: Imho, based on this formula: 

```
(price_opening*leverage)/((leverage+1-(maintenance_margin*leverage)))
```

it should look like this: 

```suggestion
UPDATE positions SET coordinator_liquidation_price = average_entry_price * coordinator_leverage / (coordinator_leverage + 1) where trader_direction='long';

```

holzeis: whoops I've overlooked the direction.. we need the short direction as well 🙈

bonomat: ah, things are so confusing... don't mind me then 🙈

bonomat: Here we go, this should fix it: 

```suggestion
UPDATE positions SET coordinator_liquidation_price = 21_000_000 where trader_direction='long' and coordinator_leverage = 1;
```

holzeis: fair, I will adapt that.

holzeis: @bonomat I used the const from the code.

```
pub const BTCUSD_MAX_PRICE: u64 = 1_048_575;
```

Issue #2232, opened 2024-03-14
Potential redundant delete of expired orders


bonomat:
Our orderbook is deleting expired orders whenever there is a new order coming in. 

This has the negative side effect that sometimes it looks like an order is being deleted multiple times.

We then see the following message in the app: 

```
flutter: [WARN] TIME: 2024-03-14T02:00:36.193620Z r: Could not remove non-existing order order_id: 93d675da-5666-4d0f-9103-4f36f0c1e142
```

This is where it is being called: 

https://github.com/get10101/10101/blob/304032173e760f87f429b2d65864defb54ed7758/coordinator/src/orderbook/trading.rs#L170-L174

This is not a bug nor harmful, just annoying as it pollutes our logs

PR #2379, opened 2024-04-09
fix: flickering prices

### LINKED ISSUE #2370: Flickering price updates
Happens more frequently a little bit later in the video

https://github.com/get10101/10101/assets/382048/180fb149-201c-41e2-b80e-9ea2ececafe2

bonomat:
resolves https://github.com/get10101/10101/issues/2232
resolves #2370


holzeis (approved):
LGTM 👍 not sure though why this will fix the flickering price updates.

Did you verify that this is actually fixing it?

bonomat:
> LGTM 👍 not sure though why this will fix the flickering price updates.
> 
> Did you verify that this is actually fixing it?

The flickering price came from us sending updates like this: 

`Price { bid: Some(71204.88), ask: None }` which set one or the prices to 0. 

I could have done a simple null check in the frontend... but where is the fun of that?

PR #2385, opened 2024-04-09
fix: Create dedicated ios privacy info manifest

holzeis:
Since solution 1 did not work, let's try with solution 2 as described [here](https://github.com/FlutterFlow/flutterflow-issues/issues/2527#issuecomment-1998885560)

fixes https://github.com/get10101/meta/issues/400


bonomat (approved):
Let's give it a try :)

bonomat:
What do you think of backporting this to 2.1.1 and releasing it again from local with a different build number?

PR #2386, opened 2024-04-09
fix: ensure filling/filled orders have matching fee

### LINKED ISSUE #2382: Failed to start 10101! (test app)
### Contact

@SXBT69 

### The Problem

App won't start

### The ScreenShots

image

### The Logs
[2024-04-08_223208.log](https://github.com/get10101/meta/files/14912744/2024-04-08_223208.log)

bonomat:
resolves #2382 (hopefully). 

It ensures that this bug cannot happen again and sets the matching fee for existing orders in the db. 
My interpretation is that this error came from the fact that we introduced a new field `matching_fee_sats` which was obviously not set for existing orders.

### DIFF
--- a/mobile/native/migrations/2024-04-09-105312_ensure_old_orders_have_matching_fee/up.sql
+++ b/mobile/native/migrations/2024-04-09-105312_ensure_old_orders_have_matching_fee/up.sql
@@ -1,6 +1,6 @@
 -- Your SQL goes here
 update orders
-set matching_fee_sats = quantity / (execution_price * leverage) * 100000000
+set matching_fee_sats = quantity / (execution_price * leverage) * 100000000 * 0.003
 where matching_fee_sats IS NOT NULL
     and execution_price IS NOT NULL
     and matching_fee_sats IS NULL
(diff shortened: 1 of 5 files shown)


mobile/native/migrations/2024-04-09-105312_ensure_old_orders_have_matching_fee/up.sql
```diff
@@ -0,0 +1,17 @@
+-- Your SQL goes here
+update orders
+set matching_fee_sats = quantity / (execution_price * leverage) * 100000000
```
holzeis, line 3: 🔧 I think this is missing the fee_rate

```suggestion
set matching_fee_sats = quantity / (execution_price * leverage) * 100000000 * 0.003
```

holzeis (approved):
LGTM 👍 

fyi: I've deployed this branch on my regtest app and can confirm that this is fixing the issue ;)

mobile/native/migrations/2024-04-09-105312_ensure_old_orders_have_matching_fee/up.sql
bonomat: Makes sense 🙈

PR #2387, opened 2024-04-09
fix: Post correct order direction

holzeis:
It looks like we always posted an order with the opposite direction from the webapp.

I guess it was a copy and paste error from the close trade dialog.


bonomat (approved):

bonomat:
closed in favor of : https://github.com/get10101/10101/pull/2392

PR #2388, opened 2024-04-09
feat: Get rid of buy and sell tabs

### LINKED ISSUE #2351: Get rid of BUY/SELL tab and add a BUY and SELL button under the trade configurator

holzeis:
Since we do not know the direction, the liquidation price and margin has been moved to the confirmation dialog.

Therefore, the new order confirmation dialog has also been introduced when opening a new channel.

|	|	|
|:-:|:-:|
| Screenshot 2024-04-09 at 13 18 23 |  Screenshot 2024-04-09 at 16 05 38 |

fixes #2351

### DIFF
--- a/webapp/frontend/lib/common/calculations.dart
+++ b/webapp/frontend/lib/common/calculations.dart
@@ -27,13 +27,17 @@ Amount calculateMargin(Usd quantity, BestQuote quote, Leverage leverage, bool is
 }
 
 Amount calculateLiquidationPrice(
-    Usd quantity, BestQuote quote, Leverage leverage, double maintenanceMargin, bool isLong) {
-  if (isLong && quote.ask != null) {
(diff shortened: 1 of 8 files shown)


bonomat (commented):
I'm not a fan of this new design :/ 

I think it has lots of value to see the margin requirements and liqidation point right when entering things. 

But as you said, if we have two button we can't calculate this... Maybe this means we shouldn't do this change then?

holzeis:
> I'm not a fan of this new design :/
> 
> I think it has lots of value to see the margin requirements and liqidation point right when entering things.
> 
> But as you said, if we have two button we can't calculate this... Maybe this means we shouldn't do this change then?

I am impartial about that, I am fine either way. I think the web app can use some face lifts in general. This doesn't make a huge difference to me 😅

holzeis:
@bonomat Please let me know if we want this in or not.

bonomat:
we could ask UAP, but overall I'm against it.

PR #2389, opened 2024-04-09
feat: Add ability to whitelist withdrawal addresses

### LINKED ISSUE #2350: Ability to whitelist addresses to withdraw

holzeis:
fixes #2350

### DIFF
--- a/mobile/native/src/ln_dlc/mod.rs
+++ b/mobile/native/src/ln_dlc/mod.rs
@@ -1008,6 +1008,19 @@ pub fn estimated_funding_tx_fee() -> Result<Amount> {
     Ok(fee)
 }
 
+/// Returns true if the provided address belongs to our wallet and false otherwise.
+/// Returns and error if the address is invalid.
+///
+/// Note, this may return false if the wallet doesn't know about the address. A full sync is
+/// required then.
+pub fn is_address_mine(address: &str) -> Result<bool> {
+    let address: Address<NetworkUnchecked> = address.parse().context("Failed to parse address")?;
+    let is_mine = state::get_node()
+        .inner
+        .is_mine(&address.payload.script_pubkey());
+    Ok(is_mine)
+}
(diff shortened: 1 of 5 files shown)


mobile/native/src/ln_dlc/mod.rs
```diff
@@ -1008,6 +1008,16 @@ pub fn estimated_funding_tx_fee() -> Result<Amount> {
     Ok(fee)
 }
 
+/// Returns true if the provided address belongs to our wallet and false otherwise.
+/// Returns and error if the address is invalid.
+pub fn is_address_mine(address: &str) -> Result<bool> {
```
bonomat, line 1013: Note: this might return a false-negative if the wallet has not generated enough addresses, i.e. if you recover your wallet and did not sync up to the questioned address this will return false.

webapp/src/api.rs
```diff
@@ -161,11 +164,23 @@ pub struct Payment {
     fee: u64,
 }
 
-pub async fn send_payment(params: Json<Payment>) -> Result<(), AppError> {
+pub async fn send_payment(
+    State(state): State<Arc<AppState>>,
+    Json(params): Json<Payment>,
+) -> Result<(), AppError> {
+    if !state
```
bonomat, line 171: With this we force users to define whitelisted addresses. Should we allow for having an empty list of whitelisted addresses to bypass this check?

bonomat (approved):
Looks good. I've added one question we might want to consider.

mobile/native/src/ln_dlc/mod.rs
holzeis: I will leave a comment.

webapp/src/api.rs
holzeis: I will make it more explicit and add a cli flag to enable or disable that feature.

PR #2395, opened 2024-04-10
feat: Neg user to share referral code

bonomat:
This dialog will be shown on first app start and then every 7 days

Resolves #a-ticket-which-has-not-been-created-but-I-m-sure-we-want-this-feature 

join-the-future

https://github.com/get10101/10101/assets/224613/6f10836a-f55c-4e1e-a62d-2955a33e0650


bonomat:
> I like it but maybe instead of showing it every 7 days we could show it after the user made a trade. I don't think that a user is going to refer the app the first time they open it. Or its the first thing they want to see when opening the app.
> 
> @SXBT69 what do you think?

Done in [`6c89996` (#2395)](https://github.com/get10101/10101/pull/2395/commits/6c89996b30e4007c3cfa021a37e8f179be50bf31). If the user has a single trade and comes back onto the wallet screen we show him the dialog.

holzeis (approved):
Nice 👍

PR #2413, opened 2024-04-11
feat: Show fee rebate on order matching fee

### LINKED ISSUE #2376: Show the `fee_rebate` on the order view, confirmation dialog and trade dialog
nit: It would be great if we could show the `fee_rabate` on the order view, confirmation dialog and trade dialog.

_Originally posted by @holzeis in https://github.com/get10101/10101/issues/2355#issuecomment-2043437700_

holzeis:
fixes #2376 

|		|		|		|
|:---:|:---:|:---:|
|	Simulator Screenshot - iPhone 15 Pro - 2024-04-11 at 16 12 20	|	Simulator Screenshot - iPhone 15 Pro - 2024-04-11 at 16 12 29	|	Simulator Screenshot - iPhone 15 Pro - 2024-04-11 at 16 12 44	|	
| Simulator Screenshot - iPhone 15 Pro - 2024-04-11 at 16 12 50	|	Simulator Screenshot - iPhone 15 Pro - 2024-04-11 at 16 00 33	|		|

### DIFF
--- a/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
@@ -1,3 +1,4 @@
+import 'package:get_10101/common/application/tentenone_config_change_notifier.dart';
 import 'package:get_10101/common/dlc_channel_change_notifier.dart';
 import 'package:get_10101/common/dlc_channel_service.dart';
 import 'package:flutter/material.dart';
@@ -149,6 +150,8 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
     TradeValues tradeValues =
         Provider.of<TradeValuesChangeNotifier>(context).fromDirection(direction);
 
+    final referralStatus = context.read<TenTenOneConfigChangeNotifier>().referralStatus;
+
     bool isClose = tradeAction == TradeAction.closePosition;
     bool isChannelOpen = tradeAction == TradeAction.openChannel;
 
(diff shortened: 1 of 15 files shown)


mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
```diff
@@ -163,6 +163,8 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
     // Fallback to 0 if we can't get the fee or the margin
     Amount total =
         tradeValues.margin != null ? Amount(tradeValues.margin!.sats).add(reserve) : Amount(0);
+    total = total.add(tradeValues.fee ?? Amount.zero());
```
luckysori, line 169: 🤔 It reads weird when we include this fee in the message:

```
X sats will be locked up in a DLC channel.
```

I think before this change it only included the margin + reserve, which I think made sense because everything else is fees. If we include the order-matching fee, I think we should include all the fees.

In general I don't like the way we present all this information, before or after this change, but this change makes it a bit worse for me.

luckysori (approved):
An improvement overall, although the summary screens are not getting any simpler.

mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
holzeis: It's just weird that we show a total which does not represent a total 😅.

holzeis:
> An improvement overall, although the summary screens are not getting any simpler.

Agreed, I've created a ticket for that https://github.com/get10101/10101/issues/2417

mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
luckysori: Before this change, it was the total money you own in the channel.

After this change, it is the total money you own in the channel + a fee that belongs to the coordinator. It seems like less of a total to me 🤷

PR #2416, opened 2024-04-12
fix: don't truncate sats

bonomat:
Fixes truncation of sats here: 

image

And overall wherever we are using Fiat values we remove the decimal place, e.g. here: 

image


luckysori (approved):
Maybe unrelated, but I think it would be nice to show the bid and ask with a single decimal point (like other platforms do 😉).

holzeis (approved):

PR #2418, opened 2024-04-13
Add insert and delete APIs to orderbook WS

luckysori:
Related to https://github.com/get10101/aristides/issues/39.

### DIFF
--- a/coordinator/src/orderbook/websocket.rs
+++ b/coordinator/src/orderbook/websocket.rs
@@ -2,12 +2,18 @@ use crate::db;
 use crate::db::user;
 use crate::message::NewUserMessage;
 use crate::orderbook::db::orders;
+use crate::orderbook::trading::NewOrderMessage;
 use crate::referrals;
 use crate::routes::AppState;
+use anyhow::bail;
+use anyhow::Result;
 use axum::extract::ws::Message as WebsocketMessage;
 use axum::extract::ws::WebSocket;
+use bitcoin::secp256k1::PublicKey;
 use commons::create_sign_message;
 use commons::Message;
+use commons::NewLimitOrder;
+use commons::OrderReason;
 use commons::OrderbookRequest;
 use commons::ReferralStatus;
 use commons::TenTenOneConfig;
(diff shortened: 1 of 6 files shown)


coordinator/src/orderbook/websocket.rs
```diff
@@ -18,9 +24,66 @@ use std::sync::Arc;
...
+        let mut conn = state.pool.clone().get()?;
+        move || {
+            orders::delete_trader_order(&mut conn, order_id, trader_id)?;
+
+            anyhow::Ok(())
+        }
+    })
+    .await??;
+
+    let _ = state.tx_orderbook_feed.send(Message::DeleteOrder(order_id));
```
bonomat, line 82: `order_id` is the `trader_order_id` right? If so, the traders receiving the `DeleteOrder` msg might not know what todo with this.

The user connected to the websocket shouldn't know about the `trader_order_id`, so did this ever work? 🙈

bonomat (approved):
I love the feature, but I'm wondering if we have a mxiup with `trader_order_id` and `order_id` 🙈

coordinator/src/orderbook/websocket.rs
luckysori: Maybe I made a mistake. I'll test it again locally.

bonomat: No, you didn't it's been like that before.

luckysori: Hmm, I've been thinking about it and I don't see the problem with those warnings we get on the app.

As far as I can tell, what happens is:

1. The maker generates an order.
2. The maker deletes the order.
3. The app starts and connects to the coordinator's WS, getting an `AllOrders` message that does not include the deleted order.
4. The WS still sends the app the corresponding `DeleteOrder` message from step 2.

I don't understand why the "outdated" message is delivered to the app, but I think we should demote the WARN to DEBUG because I think this is harmless.

bonomat: I looked into it and we are actually sending the `trader_order_id` to all users 🙈 It's probably not an issue, but I think it wasn't meant to be used like that: 

Here are we mapping the `trader_order_id` from the db to `order.id` for the orders we are returning to the users.

https://github.com/get10101/10101/blob/7c6f8ee0a4fc1574f7d4c8000da4b8bfa1b6012a/coordinator/src/orderbook/db/orders.rs#L114

bonomat: > I don't understand why the "outdated" message is delivered to the app, but I think we should demote the WARN to DEBUG because I think this is harmless.

Imho this happens because we are here also clinsing outdated orders [here](https://github.com/get10101/10101/blob/a862e02415deeba27f194539f495dba4c02c2224/coordinator/src/orderbook/trading.rs#L160-L166) and we might have a race condition where the maker deletes an order and this is run

luckysori: > I looked into it and we are actually sending the `trader_order_id` to all users 🙈 It's probably not an issue, but I think it wasn't meant to be used like that:
> 
> Here are we mapping the `trader_order_id` from the db to `order.id` for the orders we are returning to the users.
> 
> https://github.com/get10101/10101/blob/7c6f8ee0a4fc1574f7d4c8000da4b8bfa1b6012a/coordinator/src/orderbook/db/orders.rs#L114

Yeah, that makes sense.

Did we ever have an _order_ ID that wasn't the trader order ID though? I can't find another one.

bonomat: Yes, on DB level we have a different one.

luckysori: Okay, but that one has only ever been used internally:

```rust
    // this id is only internally but needs to be here or diesel complains
    #[allow(dead_code)]
    pub id: i32,
```

Issue #2406, opened 2024-04-11
Referrals which do not trade are not shown in total referrals


holzeis:
Refer another user but do not trade. The referral should see that the referred user is added to the total referrals, but this value remains the same until the user trades. Thus being the same as active referrals.

![image](https://github.com/get10101/10101/assets/382048/0dbdb123-a671-4f9d-a5d5-8d2de42c3342)

PR #2420, opened 2024-04-15
fix: return all referred users

bonomat:
we are filtering the users by their volume in code later

fixes [#2406](https://github.com/get10101/10101/issues/2406)


holzeis (approved):
LGTM 👍

Issue #2407, opened 2024-04-11
Referent does not get Referral bonus after 3 active referrals


holzeis:
1. Refer another wallet so this will get the Referent bonus
2. From the referent wallet refer additional 3 wallets and trade.

### Expected behaviour

Referent wallet should get the Referral bonus of 20%

### Actual behaviour

Referent remains with the Referent bonus of 10%

![image](https://github.com/get10101/10101/assets/382048/d40a25e8-9a08-4d68-b847-5553237ab3d5)

bonomat:
I'm happy to change this but it was by design (removed bug): 

https://github.com/get10101/10101/blob/8c46fb0fbe00bcc6eb58071f13d5bd11b41dab6d/coordinator/src/referrals.rs#L90-L103

holzeis:
> I'm happy to change this but it was by design (removed bug-label): 
> 
> https://github.com/get10101/10101/blob/8c46fb0fbe00bcc6eb58071f13d5bd11b41dab6d/coordinator/src/referrals.rs#L90-L103
> 

So it was a design bug then 😉

PR #2421, opened 2024-04-15
fix: return highest bonus

bonomat:
when calculating the trader's bonus we check if he has an existing bonus and if he has, we return the best bonus at the end.

resolves https://github.com/get10101/10101/issues/2407

### DIFF
--- a/coordinator/src/referrals.rs
+++ b/coordinator/src/referrals.rs
@@ -4,7 +4,6 @@ use crate::db::bonus_tiers::BonusTier;
 use anyhow::Context;
 use anyhow::Result;
 use bitcoin::secp256k1::PublicKey;
-use commons::BonusStatusType;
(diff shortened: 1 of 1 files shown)


coordinator/src/referrals.rs
```diff
@@ -87,20 +86,9 @@ pub fn update_referral_status_for_user(
...
-                "User has active bonus status"
-            );
-            return Ok(status);
-        }
-    }
+    // first we check his existing status. If we need to compare it later against the newly computed
+    // one
+    let existing_status = get_referral_status(trader_pubkey, connection)?;
 
     // next we need to calculate if he qualifier for a referral program
```
holzeis, line 93:
```suggestion
    // next we need to calculate if he qualifies for a referral program
```

holzeis (approved):
LGTM 👍

PR #2429, opened 2024-04-16
Forward `SubmitOrderErrors` from app to coordinator

luckysori:
Fixes https://github.com/get10101/meta/issues/269.

It's just a start, but we should be able to build on this.

### Example

User runs into this error twice:

image

Database looks like this:

```
orderbook=# select * from reported_errors;
 id |                           trader_pubkey                            |                            error                            
----+--------------------------------------------------------------------+-------------------------------------------------------------
  1 | 0315bb4b3561fc219c2ace72018600049a556ad788db430318cdeedd76a4281634 | DLC channel not yet confirmed: has 0 confirmations, needs 1
  2 | 0315bb4b3561fc219c2ace72018600049a556ad788db430318cdeedd76a4281634 | DLC channel not yet confirmed: has 0 confirmations, needs 1
(2 rows)
```

### DIFF
--- a/coordinator/migrations/2024-04-16-035239_add_reported_errors/down.sql
+++ b/coordinator/migrations/2024-04-16-035239_add_reported_errors/down.sql
@@ -1,0 +1,1 @@
+DROP TABLE reported_errors;
--- a/coordinator/migrations/2024-04-16-035239_add_reported_errors/up.sql
+++ b/coordinator/migrations/2024-04-16-035239_add_reported_errors/up.sql
@@ -1,0 +1,5 @@
+CREATE TABLE reported_errors (
+    id SERIAL PRIMARY KEY NOT NULL,
+    trader_pubkey TEXT NOT NULL,
+    error TEXT NOT NULL
+);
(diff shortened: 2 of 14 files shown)


holzeis (commented):
🤔 That's pretty cool, but do we really need this for errors that are returned by the coordinator? Couldn't we just report the error already on the coordinator side, before having a round trip over the app? Note, most errors are anyways returned via the `OrderbookMessage::TraderMessage > TradeError`. We could simply subscribe to that message on the coordinator as well and store it to the database when thrown.

More interesting would be the errors that are produced on the app that are not known to the coordinator.

luckysori:
> 🤔 That's pretty cool, but do we really need this for errors that are returned by the coordinator? Couldn't we just report the error already on the coordinator side, before having a round trip over the app? Note, most errors are anyways returned via the `OrderbookMessage::TraderMessage > TradeError`. We could simply subscribe to that message on the coordinator as well and store it to the database when thrown.
> 
> 
> 
> More interesting would be the errors that are produced on the app that are not known to the coordinator.

Hmm, but only the `Orderbook` error variant comes from the coordinator. The rest are local to the app.

We can filter out those variants if you prefer.

In any case, this PR is just a start. I agree that there are more interesting errors that are not yet captured by this PR.

holzeis (approved):

PR #2439, opened 2024-04-17
fix: Calculate max quantity for resizing

### LINKED ISSUE #2391: Max quantity calculation is not correct for resizing
Environment: `test`

It seems the max quantity calculation does not take reducing a position into account.

holzeis:
Calculates the max quantity considering

- open quantity: Quantity of an existing position of the opposite direction. (needed for fee calculation)
- accumulated order matching fee: As the coordinator is not using already collected order matching fees as collateral for new trades, this has to be deducted from the max coordinator margin.

fixes #2391

### DIFF
--- a/mobile/lib/features/trade/domain/trade_values.dart
+++ b/mobile/lib/features/trade/domain/trade_values.dart
@@ -15,7 +15,6 @@ class TradeValues {
   Amount? fee; // This fee is an estimate of the order-matching fee.
   Usd? maxQuantity;
 
-  double fundingRate;
   DateTime expiry;
 
   // no final so it can be mocked in tests
@@ -29,7 +28,6 @@ class TradeValues {
     required this.price,
     required this.liquidationPrice,
     required this.fee,
-    required this.fundingRate,
     required this.expiry,
     required this.tradeValuesService,
   });
@@ -38,7 +36,6 @@ class TradeValues {
     required Usd quantity,
     required Leverage leverage,
     required double? price,
-    required double fundingRate,
     required Direction direction,
     required TradeValuesService tradeValuesService,
   }) {
@@ -59,39 +56,6 @@ class TradeValues {
         quantity: quantity,
         leverage: leverage,
         price: price,
-        fundingRate: fundingRate,
-        liquidationPrice: liquidationPrice,
-        fee: fee,
-        expiry: expiry,
-        tradeValuesService: tradeValuesService);
-  }
-
-  factory TradeValues.fromMargin({
-    required Amount? margin,
-    required Leverage leverage,
-    required double? price,
-    required double fundingRate,
-    required Direction direction,
-    required TradeValuesService tradeValuesService,
-  }) {
-    Usd? quantity =
-        tradeValuesService.calculateQuantity(price: price, margin: margin, leverage: leverage);
-    double? liquidationPrice = price != null
-        ? tradeValuesService.calculateLiquidationPrice(
-            price: price, leverage: leverage, direction: direction)
-        : null;
-
-    Amount? fee = tradeValuesService.orderMatchingFee(quantity: quantity, price: price);
-
-    DateTime expiry = tradeValuesService.getExpiryTimestamp();
-
-    return TradeValues(
-        direction: direction,
-        margin: margin,
-        quantity: quantity,
-        leverage: leverage,
-        price: price,
-        fundingRate: fundingRate,
         liquidationPrice: liquidationPrice,
         fee: fee,
         expiry: expiry,
--- a/mobile/lib/features/trade/domain/trade_values.dart
+++ b/mobile/lib/features/trade/domain/trade_values.dart
@@ -163,9 +163,10 @@ class TradeValues {
   }
 
   recalculateMaxQuantity() {
-    final quantity = tradeValuesService.calculateMaxQuantity(price: price, leverage: leverage);
+    final quantity = tradeValuesService.calculateMaxQuantity(
+        price: price, leverage: leverage, direction: direction);
     if (quantity != null) {
-      maxQuantity = quantity;
+      _maxQuantity = quantity;
     }
   }
 }
--- a/mobile/lib/features/trade/domain/trade_values.dart
+++ b/mobile/lib/features/trade/domain/trade_values.dart
@@ -4,16 +4,32 @@ import 'package:get_10101/features/trade/domain/direction.dart';
 import 'package:get_10101/features/trade/domain/leverage.dart';
 
 class TradeValues {
+  /// Potential quantity already in an open position
+  Usd _openQuantity = Usd.zero();
+
+  get openQuantity => _openQuantity;
+
+  set openQuantity(quantity) => _openQuantity = quantity;
+
+  /// The difference between open quantity and quantity if contracts is bigger than the open quantity.
+  /// This value is used to calculate the required margin.
+  Usd quantity = Usd.zero();
+
+  /// The actual contracts entered. Any value between 0 and maxQuantity.
+  Usd contracts = Usd.zero();
+
   Amount? margin;
   Leverage leverage;
   Direction direction;
 
-  // These values  can be null if coordinator is down
-  Usd? quantity;
   double? price;
   double? liquidationPrice;
   Amount? fee; // This fee is an estimate of the order-matching fee.
-  Usd? maxQuantity;
+
+  Usd? _maxQuantity;
+
+  /// The max quantity of the contracts not part of the open quantity.
+  Usd get maxQuantity => _maxQuantity ?? Usd.zero();
 
   DateTime expiry;
 
@@ -41,6 +57,7 @@ class TradeValues {
   }) {
     Amount? margin =
         tradeValuesService.calculateMargin(price: price, quantity: quantity, leverage: leverage);
+
     double? liquidationPrice = price != null
         ? tradeValuesService.calculateLiquidationPrice(
             price: price, leverage: leverage, direction: direction)
@@ -64,6 +81,11 @@ class TradeValues {
 
   updateQuantity(Usd quantity) {
     this.quantity = quantity;
+    _recalculateMargin();
+  }
+
+  updateContracts(Usd contracts) {
+    this.contracts = contracts;
     _recalculateMargin();
     _recalculateFee();
   }
@@ -113,7 +135,7 @@ class TradeValues {
   _recalculateQuantity() {
     Usd? quantity =
         tradeValuesService.calculateQuantity(price: price, margin: margin, leverage: leverage);
-    this.quantity = quantity;
+    this.quantity = quantity ?? Usd.zero();
   }
 
   _recalculateLiquidationPrice() {
@@ -123,7 +145,7 @@ class TradeValues {
   }
 
   _recalculateFee() {
-    fee = tradeValuesService.orderMatchingFee(quantity: quantity, price: price);
+    fee = tradeValuesService.orderMatchingFee(quantity: contracts, price: price);
   }
 
   recalculateMaxQuantity() {
--- a/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
@@ -166,7 +166,6 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
     // Fallback to 0 if we can't get the fee or the margin
     Amount total =
         tradeValues.margin != null ? Amount(tradeValues.margin!.sats).add(reserve) : Amount(0);
-    total = total.add(tradeValues.fee ?? Amount.zero());
 
     Amount pnl = Amount(0);
     if (context.read<PositionChangeNotifier>().positions.containsKey(ContractSymbol.btcusd)) {
--- a/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
@@ -37,7 +37,8 @@ class TradeBottomSheetTab extends StatefulWidget {
   State<TradeBottomSheetTab> createState() => _TradeBottomSheetTabState();
 }
 
-class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
+class _TradeBottomSheetTabState extends State<TradeBottomSheetTab>
+    with AutomaticKeepAliveClientMixin<TradeBottomSheetTab> {
   late final TradeValuesChangeNotifier provider;
   late final TenTenOneConfigChangeNotifier tentenoneConfigChangeNotifier;
   late final PositionChangeNotifier positionChangeNotifier;
@@ -106,6 +107,8 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
 
   @override
   Widget build(BuildContext context) {
+    super.build(context);
+
     TradeTheme tradeTheme = Theme.of(context).extension<TradeTheme>()!;
     DlcChannelChangeNotifier dlcChannelChangeNotifier = context.watch<DlcChannelChangeNotifier>();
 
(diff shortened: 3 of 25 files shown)


mobile/native/src/max_quantity.rs
```diff
@@ -33,16 +35,75 @@ pub fn max_quantity(
...
-        Some(position) => position.order_matching_fees,
-        _ => Amount::ZERO,
+    let positions = position::handler::get_positions()?;
+    let position = positions.first();
+    let accumulated_order_matching_fees = position
+        .map(|p| p.order_matching_fees)
+        .unwrap_or(Amount::ZERO);
+
+    let (max_coordinator_margin, max_trader_margin) = match position {
+        Some(position) if trader_direction != position.direction => {
```
luckysori, line 51: 🙃 `trader_direction` is slightly ambiguous. I think it should be `trade_direction`.

mobile/native/src/max_quantity.rs
```diff
@@ -33,16 +35,75 @@ pub fn max_quantity(
...
+            let total_collateral = channel_trade_constraints
+                .total_collateral
+                .unwrap_or(Amount::ZERO);
+
+            let total_balance = channel_trade_constraints.max_counterparty_balance_sats
+                + channel_trade_constraints.max_local_balance_sats;
+            let total_balance = Amount::from_sat(total_balance);
+
+            let trader_margin = Amount::from_sat(position.collateral);
+            let counterparty_margin = total_collateral - (total_balance + trader_margin);
```
luckysori, line 61: 🔧 Probably needs to be `checked_sub`bed, given that `total_collateral` can be zero.

luckysori: Maybe audit the PR diff for places where we should use `checked_sub` in general.

mobile/native/src/max_quantity.rs
```diff
@@ -33,16 +35,75 @@ pub fn max_quantity(
...
+                ),
+                false => (
+                    max_coordinator_margin
+                        .checked_add(unsigned_trader_pnl)
+                        .unwrap_or(Amount::ZERO),
+                    max_trader_margin
+                        .checked_sub(unsigned_trader_pnl)
+                        .unwrap_or(Amount::ZERO),
+                ),
+            }
```
luckysori, line 96: 🙃 It might a bit simpler/clearer to use `SignedAmount`s when applying the PNL, then convert to `Amount` afterwards and `unwrap_or_default`. We could get rid of the branching logic.

luckysori (commented):
LGTM, particularly given the demo, but I don't want to approve because my review wasn't very thorough.

luckysori (approved):
Actually, get the approval now and wait for @bonomat's review or another look from me tomorrow, if you think it's necessary.

mobile/native/migrations/2024-04-16-144000_add_accumulated_order_matching_fee_to_position/up.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE positions
+    ADD COLUMN order_matching_fees BIGINT NOT NULL DEFAULT 0;
+
+-- To avoid division by zero.
+UPDATE positions
+SET order_matching_fees = 0 where average_entry_price=0;
+
+-- Before this migration, the taker fee was hard-coded to 0.30%.
+UPDATE positions
+SET order_matching_fees = quantity * (1 / average_entry_price) * 0.0030 where average_entry_price!=0;
```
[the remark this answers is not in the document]
holzeis: Yeah, when I've written it 2.2.0 wasn't released yet. 🙈

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -59,6 +59,39 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
...
+    shortTradeValues.updateContracts(shortTradeValues.maxQuantity);
+
+    // init the long trade values
+    final longTradeValues = provider.fromDirection(Direction.long);
+    longTradeValues.updateQuantity(longTradeValues.maxQuantity);
+    // overwrite any potential pre-existing state
+    longTradeValues.openQuantity = Usd.zero();
+
+    // by default we set the amount to the max quantity.
+    longTradeValues.updateContracts(longTradeValues.maxQuantity);
```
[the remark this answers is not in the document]
holzeis: Yes.. sorry for not mentioning it in the PR but 500 is just a weird default. I figured the max quantity makes more sense.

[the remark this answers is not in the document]
luckysori: Yeah, I'm okay with the change. I just was surprised to find it in this patch.

mobile/native/src/max_quantity.rs
holzeis: But it's the direction that the trader has chosen. I think `trade_direction` wouldn't be clear enough.

luckysori: `trader_direction` is ambiguous because the trader has both a trade/order direction and a position direction. If you want to be very specific I think it should be `trader_trade_direction` or `trader_order_direction`.

mobile/lib/features/trade/domain/trade_values.dart
```diff
@@ -4,16 +4,28 @@ import 'package:get_10101/features/trade/domain/direction.dart';
 import 'package:get_10101/features/trade/domain/leverage.dart';
 
 class TradeValues {
+  /// Potential quantity already in an open position
+  Usd openQuantity = Usd.zero();
+
+  /// The difference between open quantity and quantity if quantity is bigger than the open quantity.
+  /// This value is used to calculate the required margin.
+  Usd quantity = Usd.zero();
```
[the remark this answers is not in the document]
holzeis: it should have been, if the contracts is bigger than the open quantity.

mobile/native/src/max_quantity.rs
```diff
@@ -33,16 +35,75 @@ pub fn max_quantity(
...
+    let order_matching_fee_rate = Decimal::try_from(order_matching_fee_rate).expect("to fit");
 
-    let accumulated_order_matching_fees = match position::handler::get_positions()?.first() {
-        Some(position) => position.order_matching_fees,
-        _ => Amount::ZERO,
+    let positions = position::handler::get_positions()?;
+    let position = positions.first();
+    let accumulated_order_matching_fees = position
+        .map(|p| p.order_matching_fees)
+        .unwrap_or(Amount::ZERO);
```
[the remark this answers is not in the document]
holzeis: I prefer this over having to return a triple from the match below.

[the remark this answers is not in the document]
luckysori: Okay, that makes sense. I thought the value was used inside the `Some` branch. Never mind.

mobile/native/migrations/2024-04-16-144000_add_accumulated_order_matching_fee_to_position/up.sql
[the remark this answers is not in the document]
bonomat: Also, any trade between Friday, 12.04 and now has no order matching fee

mobile/lib/features/trade/domain/trade_values.dart
```diff
@@ -113,7 +132,13 @@ class TradeValues {
...
     Usd? quantity =
         tradeValuesService.calculateQuantity(price: price, margin: margin, leverage: leverage);
-    this.quantity = quantity;
+    this.quantity = quantity ?? Usd.zero();
+  }
+
+  // Note, this only works because it is only called from [`updatePriceAndQuantity`], which is only
+  // called if the max quantity lock is set.
+  _recalculateContracts() {
+    contracts = quantity + openQuantity;
```
bonomat, line 141: Wouldn't it be safer to do this from inside `_recalculateQuantity()`, or whenever `quantity` is being updated?

bonomat, line 141: `openQuantity` seems to be always zero?

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -59,6 +59,39 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
...
+
+    if (positionChangeNotifier.positions.containsKey(contractSymbol)) {
+      // in case there is an open position we have to set the open quantity for the trade values of
+      // the opposite direction
+      final position = positionChangeNotifier.positions[contractSymbol]!;
+      final tradeValues = provider.fromDirection(position.direction.opposite());
+
+      tradeValues.openQuantity = position.quantity;
```
bonomat, line 87: 🔧
Not you who introduced this pattern: it might be better to start using `setters` on `TradeValues` it's getting a bit hard to see where things are being updated.

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
```diff
@@ -189,14 +222,16 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab> {
     bool hasPosition = positionChangeNotifier.positions.containsKey(contractSymbol);
 
     double? positionLeverage;
+    int usableBalance = channelTradeConstraints.maxLocalBalanceSats;
     if (hasPosition) {
       final position = context.read<PositionChangeNotifier>().positions[contractSymbol];
       positionLeverage = position!.leverage.leverage;
+      if (direction != position.direction) {
```
bonomat, line 232: 🫣
```suggestion
      if (direction == position.direction.opposite())
```

mobile/lib/features/trade/trade_value_change_notifier.dart
```diff
@@ -71,7 +72,16 @@ class TradeValuesChangeNotifier extends ChangeNotifier implements Subscriber {
   }
 
   void updateQuantity(Direction direction, Usd quantity) {
-    fromDirection(direction).updateQuantity(quantity);
+    if (fromDirection(direction).openQuantity < quantity) {
+      // the user is changing direction of his position
```
bonomat, line 76: How do you know in this case that the user is changing direction?

mobile/native/src/max_quantity.rs
```diff
@@ -123,14 +189,50 @@ fn calculate_max_quantity(
...
+            max_trader_margin,
+            None,
+            coordinator_leverage,
+            trader_leverage,
+            order_matching_fee_rate,
+            Amount::from_sat(4459),
+            open_quantity,
+        );
+
+        assert_eq!(dec!(648), max_quantity);
```
bonomat, line 224: I would like to understand how you have come to this value. For me it's hard to derive this myself.
Can you add the formula, or explain it in prosa why this should be 648?

mobile/native/src/max_quantity.rs
```diff
@@ -123,14 +189,50 @@ fn calculate_max_quantity(
...
+
+        // Ensure that the coordinator has enough funds for the trade
+        let coordinator_margin = calculations::calculate_margin(
+            price.to_f32().unwrap(),
+            (max_quantity - open_quantity).to_f32().unwrap(),
+            coordinator_leverage,
+        );
+
+        assert!(Amount::from_sat(coordinator_margin) < max_coordinator_margin);
```
bonomat, line 233: 🙃
no need to convert it to amount which will internally take the u64 anyways 

```suggestion
assert!(coordinator_margin < max_coordinator_margin.to_sat());
```

mobile/native/src/max_quantity.rs
```diff
@@ -33,16 +35,75 @@ pub fn max_quantity(
...
+    let (max_coordinator_margin, max_trader_margin) = match position {
+        Some(position) if trader_direction != position.direction => {
+            let total_collateral = channel_trade_constraints.total_collateral.unwrap_or(0);
+
+            let total_balance = channel_trade_constraints.max_counterparty_balance_sats
+                + channel_trade_constraints.max_local_balance_sats;
+            let total_balance = Amount::from_sat(total_balance);
+
+            let trader_margin = Amount::from_sat(position.collateral);
+            let counterparty_margin = Amount::from_sat(total_collateral)
```
bonomat, line 59: we are mixing terminology of `counterparty_` and `coordinator_` in this function.

webapp/frontend/lib/services/trade_constraints_service.g.dart
```diff
@@ -7,8 +7,8 @@ part of 'trade_constraints_service.dart';
 // **************************************************************************
 
 TradeConstraints _$TradeConstraintsFromJson(Map<String, dynamic> json) => TradeConstraints(
-      maxLocalMarginSats: json['max_local_margin_sats'] as int,
-      maxCounterpartyMarginSats: json['max_counterparty_margin_sats'] as int,
+      maxLocalBalanceSats: json['max_local_balance_sats'] as int,
```
bonomat, line 10: Did it work to generate these files or did you edit it manually?

mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
```diff
@@ -166,7 +166,6 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
     // Fallback to 0 if we can't get the fee or the margin
     Amount total =
         tradeValues.margin != null ? Amount(tradeValues.margin!.sats).add(reserve) : Amount(0);
-    total = total.add(tradeValues.fee ?? Amount.zero());
```
bonomat, line 169: Why?

bonomat (commented):
I trust that the calculations is correct as I can't say that I follow the `max_quantity` function anymore :(

mobile/native/migrations/2024-04-16-144000_add_accumulated_order_matching_fee_to_position/up.sql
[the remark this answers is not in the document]
holzeis: I don't think it's too critical as it will only move the max quantity lower. Do the user might not be able to trade everything, but he won't run into errors.

mobile/lib/features/trade/domain/trade_values.dart
holzeis: It is set when the trade bottom sheet tab is initialised.

mobile/lib/features/trade/trade_bottom_sheet_tab.dart
holzeis: I don't think this is correct. If the user is long the open quantity can be additionally shorted.

webapp/frontend/lib/services/trade_constraints_service.g.dart
holzeis: Edited manually!

mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
holzeis: https://github.com/get10101/10101/pull/2413#discussion_r1562152519

webapp/frontend/lib/services/trade_constraints_service.g.dart
holzeis: I've also run it with `fvm dart run build_runner build --delete-conflicting-outputs` and the output didn't change.

mobile/lib/features/trade/domain/trade_values.dart
holzeis: I think it wouldn't be safer, as we want to do this only if the max quantity lock is set. otherwise that would be wrong. As mentioned in the comment, this only works because its only called from the `updatePriceAndQuantity` which is only called if the `maxQuantityLock` is set.

holzeis: Actually after some testing, it turns out it wasn't working as expected. We achieve the original intention by simply setting the contracts to the maxQuantity after every price update.

mobile/native/src/max_quantity.rs
holzeis: I changed it to `trader_trade_direction`

mobile/lib/features/trade/trade_value_change_notifier.dart
bonomat: I don't understand this 😅 What if he was 50 long and then goes 51 long?

PR #2441, opened 2024-04-17
fix(ui): Calculate fee rebate from full order matching fee

### LINKED ISSUE #2437: Fix `fee_rebate` calculation on order and confirmation view
The fee rebate is incorrectly calculated using the already rebated fee to calculate the rebate from.

holzeis:
fixes #2437

### DIFF
--- a/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
@@ -178,9 +178,8 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
 
     final orderMatchingFee = tradeValues.fee ?? Amount.zero();
     final feeRebate = referralStatus != null
-        ? Amount((referralStatus.referralFeeBonus * orderMatchingFee.sats).ceil())
+        ? Amount((referralStatus.referralFeeBonus * orderMatchingFee.sats).floor())
         : Amount.zero();
-    final feeBeforeRebate = orderMatchingFee + feeRebate;
 
     return Container(
         padding: EdgeInsets.only(left: 20, right: 20, top: (isClose ? 20 : 10), bottom: 10),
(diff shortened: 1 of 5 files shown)


mobile/lib/features/trade/trade_bottom_sheet_confirmation.dart
```diff
@@ -177,10 +177,10 @@ class TradeBottomSheetConfirmation extends StatelessWidget {
     TextStyle dataRowStyle = const TextStyle(fontSize: 14);
 
     final orderMatchingFee = tradeValues.fee ?? Amount.zero();
-    final feeRebate = referralStatus != null
-        ? Amount((referralStatus.referralFeeBonus * orderMatchingFee.sats).ceil())
+    final feeBeforeRebate = referralStatus != null && orderMatchingFee.sats > 0
+        ? Amount((orderMatchingFee.sats / (1.0 - referralStatus.referralFeeBonus)).floor())
```
luckysori, line 181: 🤔 I think this can lead to a division-by-zero error.

holzeis: If we have a fee rebate of 100%. I'll add a handling

luckysori: Yeah. I only mention it because we talked about this scenario earlier today.

holzeis: Hmm. But if we have a fee rebate of 100% the order matching fee will be 0.

holzeis:
@luckysori I fixed the issue if there is a fee rebate of 100%. I had to parameterize the `order_matching_fee` function to not deduct the fee_rebate for the front end so that I can show the rebate on the UI.

This will now also work for a 100% fee deduction.

luckysori (approved):

PR #2444, opened 2024-04-17
fix: Derive referral code from node id for referral widget

### LINKED ISSUE #2442: App doesn't work since 2.2.0 update
### App doesn't work since 2.2.0 update

image

I tried closing and opening the app again and again nothing works

impossible to extract the logs

iOS 17.4.1

- > Maybe it comes from https://github.com/get10101/10101/pull/2395#event-12436072634 since there is a white layer impeaching me to see correctly the screen and clicking anything

holzeis:
This is not ideal, but since the referral code can be derived deterministically from the node id its safe to do and will ensure that the app will not fail to start.

This fixes an issue where the referral status is not yet set because the app hasn't processed the authenticated message from the coordinator yet. (happened because the referral widget will be shown immediately on the wallet screen after init, which can lead to that race condition)

fixes #2442

### DIFF
--- a/mobile/lib/features/wallet/referral_widget.dart
+++ b/mobile/lib/features/wallet/referral_widget.dart
@@ -3,16 +3,14 @@ import 'dart:io';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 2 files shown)


bonomat (commented):
This works. I guess an alternative solution would have been to simply not show the dialog when the `referralStatus` is not yet set.

PR #2463, opened 2024-04-23
Tradingview chart

### LINKED ISSUE #2123: Replace candlestick chart
Note, I didn't manage to set the background of the candlestick chart back to white, but we might want to consider replacing that dependency altogether, since it doesn't seem to be maintained anymore.

_Originally posted by @holzeis in https://github.com/get10101/10101/issues/2122#issuecomment-1970884220_

Restioson:
Fixes #2123

I'm not sure that this will work well at all DPIs - I add a 3x scale, else it is too small

image

TODO:
- [x] Align background colours
- [ ] Try remove grey border
- [ ] Line instead of candlesticks (if we can)
- [x] Hide link back to tradingview
- [x] Change to BitMex
- [ ] Try remove reset control
- [ ] Timezone?


bonomat:
Nice. Doesn't look too sexy though 😬 I tried it on an iOS simulator and it looks even worse :/

bonomat:
Beautiful ads :D

image

holzeis:
@Restioson can you configure the widget to open in the browser if the tradeview logo is tapped?

justfile
```diff
@@ -204,7 +204,7 @@ wipe: wipe-docker wipe-coordinator wipe-maker wipe-app wipe-webapp
 wipe-docker:
     #!/usr/bin/env bash
     set -euxo pipefail
-    docker-compose down -v
+    docker compose down -v
```
holzeis, line 207: I fear this will break it for me since I did not install compose as plugin but as separate binary.

mobile/lib/features/trade/tradingview_candlestick.dart
```diff
@@ -0,0 +1,66 @@
...
+    </style>
+    <div id="container">
+    <!-- TradingView Widget BEGIN -->
+    <div class="tradingview-widget-container" style="height:100%;width:100%">
+      <div class="tradingview-widget-container__widget" style="height:calc(100% - 32px);width:100%"></div>
+      <div class="tradingview-widget-copyright"><a href="https://www.tradingview.com/" rel="noopener nofollow" target="_blank"><span class="blue-text">Track all markets on TradingView</span></a></div>
+      <script type="text/javascript" src="https://s3.tradingview.com/external-embedding/embed-widget-advanced-chart.js" async>
+      {
+      "autosize": true,
+      "symbol": "BITSTAMP:BTCUSD",
```
holzeis, line 27: 🔧 can we change that to BitMEX?

mobile/lib/features/trade/tradingview_candlestick.dart
```diff
@@ -0,0 +1,66 @@
...
+      "interval": "D",
+      "timezone": "Etc/UTC",
+      "theme": "dark",
```
holzeis, line 30: 🔧 I guess it would better fit with light mode.

mobile/lib/features/trade/tradingview_candlestick.dart
```diff
@@ -0,0 +1,66 @@
...
+      
+      body {
+        overflow: hidden;
+      }
+    </style>
+    <div id="container">
+    <!-- TradingView Widget BEGIN -->
+    <div class="tradingview-widget-container" style="height:100%;width:100%">
+      <div class="tradingview-widget-container__widget" style="height:calc(100% - 32px);width:100%"></div>
+      <div class="tradingview-widget-copyright"><a href="https://www.tradingview.com/" rel="noopener nofollow" target="_blank"><span class="blue-text">Track all markets on TradingView</span></a></div>
```
holzeis, line 23: 🔧 I guess we can remove that link.

mobile/lib/features/trade/tradingview_candlestick.dart
```diff
@@ -0,0 +1,66 @@
...
+      "theme": "dark",
+      "style": "1",
+      "locale": "en",
+      "enable_publishing": false,
+      "hide_top_toolbar": true,
+      "hide_legend": true,
+      "allow_symbol_change": false,
+      "save_image": false,
+      "calendar": false,
+      "support_host": "https://www.tradingview.com"
```
holzeis, line 39: ❓ what is the support host?

holzeis (commented):
If we could get rid of the in widget browser "feature" and instead open the browser on the phone it would be good to go.

Note, I haven't run it on my machine. We should check if it's properly rendered for iOS and Android.

Additionally it would be cool to recognise if the user is tilting their phone. If so we could show more details on the trading view. But that doesn't have to come with this PR.

Restioson:
> Nice. Doesn't look too sexy though 😬 I tried it on an iOS simulator and it looks even worse :/

I believe it's a scaling issue. The photo that I sent looks poor because it's a JPEG. It's perfectly crisp on my phone screen but the artifacts look really bad when cropped & zoomed like that. As we've discussed with iOS it seems like the scaling algorithm isn't great for the simulator and possibly real phone too. I'm not exactly sure how we should be scaling it. Maybe there's some viewport meta tags that can be provided to specify DPI? I will have to try these out.

> @Restioson can you configure the widget to open in the browser if the TradingView logo is tapped?

I can give it a shot for sure!

> If we could get rid of the in widget browser "feature" and instead open the browser on the phone it would be good to go.

How do you mean - get rid of it entirely and just show a link? If we will do this, we could also use the mini ticker view and have a tap for details feature. That way there is still something to show, especially just the price and recent movements, but it won't be scrollable.

> Beautiful ads :D

Do we still want to go with TradingView if we will have ads like this? :confused:

mobile/lib/features/trade/tradingview_candlestick.dart
Restioson: This was generated by the widget creator on their site

mobile/lib/features/trade/tradingview_candlestick.dart
Restioson: We could, but I'm not sure if we would be violating some usage policy by doing that

mobile/lib/features/trade/tradingview_candlestick.dart
Restioson: Agreed, this was mostly just a quick prototype

holzeis:
>> If we could get rid of the in widget browser "feature" and instead open the browser on the phone it would be good to go.
> How do you mean - get rid of it entirely and just show a link? If we will do this, we could also use the mini ticker view and have a tap for details feature. That way there is still something to show, especially just the price and recent movements, but it won't be scrollable.

I meant If we can get rid of the widget opening another website internally, when clicking the TradingView logo. I think this will also fix the ad issue.

Restioson:
> I meant If we can get rid of the widget opening another website internally, when clicking the TradingView logo. I think this will also fix the ad issue.

Do you know anything about how the ads are showing up? I haven't had it happen yet and was assuming that it just replaced the old view (i.e without having to navigate to a new site or anything)

Restioson:
I've added the functionality to only open links in a new tab in the latest commit

holzeis:
> Do you know anything about how the ads are showing up? I haven't had it happen yet and was assuming that it just replaced the old view (i.e without having to navigate to a new site or anything)

My theory is that the widget simply shows the tradingview.com website and that the ad is simply coming from there.

bonomat:
> > Do you know anything about how the ads are showing up? I haven't had it happen yet and was assuming that it just replaced the old view (i.e without having to navigate to a new site or anything)
> 
> My theory is that the widget simply shows the tradingview.com website and that the ad is simply coming from there.

yea, I think the ad only came because I clicked on the trading view tab which opened trading view.

justfile
Restioson: Hmm, on the other hand, I do not have compose and only have the plugin. According to the site the separate binary is not recommended:
image

Restioson: I can drop this commit if we want but it'd be good to find a solution that works for both of us

Restioson:
I'm not sure we can remove the grey border, as it's inside the iframe. Same goes for the reset control. We should be able to change it to a line graph and also could edit the timezone, though. Are these features desired?

justfile
holzeis: Happy to change my setup if it's fine for everybody else @bonomat @luckysori

luckysori:
> I'm not sure we can remove the grey border, as it's inside the iframe. Same goes for the reset control. We should be able to change it to a line graph and also could edit the timezone, though. Are these features desired?

With regards to the grey border, it's not a big deal.

justfile
bonomat: no hard feelings, on my machine both seems to work :)

bonomat:
I've tested it on an iPhone simulator... unfortunately it doesn't work.
I get the error `Failed to open link to TradingView`. 
When I remove this part it works again.

```dart
..setNavigationDelegate(NavigationDelegate(
          onNavigationRequest: (req) async {
           
            final uri = Uri.parse(req.url);
            if (await canLaunchUrl(uri)) {
              await launchUrl(uri, mode: LaunchMode.externalApplication);
            } else {
              showSnackBar(ScaffoldMessenger.of(rootNavigatorKey.currentContext!),
                  "Failed to open link to TradingView");
            }

            return NavigationDecision.prevent;
          },
        ))
```

it looks like the requested URL is `"about:blank"`.

justfile
luckysori: I'm okay with this change.

bonomat:
> > I've tested it on an iPhone simulator... unfortunately it doesn't work. I get the error `Failed to open link to TradingView`. When I remove this part it works again.
> > -- snip --
> > it looks like the requested URL is `"about:blank"`.
> 
> When you say 'works', do you mean that it launches into a new browser app, or that it redirects the embedded page in the app?
> 
> Also, can you see the page fine? Is the external launching the only issue?

Sorry, what I meant is that I don't see a chart at all. Only if I remove said code part I see the chart.

Restioson:
Hmm, I see. I'll maybe add an exception for about scheme urls and we can see if that works? Are you happy to test this?

bonomat:
Unfortunately it doesn't work on my simulator. When navigating to the trade screen it automatically opens the external browser and opens the tradingview. It only works if I remove the `NavigationDelegate` completely

bonomat:
Made it work for you in [`60f89ff` (#2463)](https://github.com/get10101/10101/pull/2463/commits/60f89ff6cf880705679cebbebb3ad535523e8af5) using the same plugin we use in DLC-Connect :) Tested on Android Simulator and iOS Simulator.

bonomat (approved):
Tested locally on Android and iOS simulator. I failed deploying it on the phone due to some local setting issue though, but I think we should give it a try.

bonomat:
@Restioson : I've rebased for you. Let's get this in.

PR #2471, opened 2024-04-25
feat: automatically add app for review in testflight

bonomat:
first step for https://github.com/get10101/meta/issues/443


bonomat:
Note that we do have a problem: 
reviewing the app takes some time, if in the meantime the deployed coordinator changes with some breaking changes, the released reviewed app won't work. 

Given we don't guarantee anything on regtest, I think we can live with this.

.github/workflows/testflight-for-review.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: cargo install just --force
+
+      - name: Install fastlane
+        run: |
+          cd mobile/ios/fastlane
+          gem install bundler
+          bundle install
+          bundle info fastlane
+
+      - name: Put for beta testing app fro review
```
holzeis, line 45:
```suggestion
      - name: Put for beta testing app for review
```

holzeis (approved):
Approved, although the App Store approval is a bit cumbersome to coordinate with a coordinator upgrade.

What do you think of only supporting users on our test vm with the apk?

bonomat:
> What do you think of only supporting users on our test vm with the apk?

Given this action shouldn't cost us anything I would just run it as well and just don't guarantee anything. We don't have breaking changes too often 🤞.

PR #2472, opened 2024-04-25
feat: create release and add APK to CI release

bonomat:
Whenever we build APKs for main, we create a new release with the _date_ format and attach the APKs

Untested... YOLO.

### DIFF
--- a/.github/workflows/android-fastlane.yml
+++ b/.github/workflows/android-fastlane.yml
@@ -197,6 +197,27 @@ jobs:
             mobile/build/app/outputs/flutter-apk/*.apk
             mobile/build/app/outputs/flutter-apk/*.apk.sha1
 
+      - name: Compute the release tag for continues ci build
+        if: ${{ inputs.tag  == 'main' }}
+        run: |
+          echo "release_tag=v`date '+%Y-%m-%d %H:%M:%S'`" >> $GITHUB_ENV
+
+      - name: Attach android apks to release
+        uses: softprops/action-gh-release@v2
+        if: ${{ inputs.tag  == 'main' }}
+        with:
+          prerelease: true
+          name: ${{ env.release_tag }}
+          tag_name: ${{ env.release_tag }}
+          target_commitish: main
+          make_latest: false
+          repository: get10101/10101-test-apks
+          token: ${{ secrets.GITHUB_TOKEN }}
+          files: |
+            mobile/build/app/outputs/flutter-apk/*.apk
+            mobile/build/app/outputs/flutter-apk/*.apk.sha1
+
+
       - name: Release to Google Play Store
         env:
           # secrets
--- a/.github/workflows/android-fastlane.yml
+++ b/.github/workflows/android-fastlane.yml
@@ -189,7 +189,7 @@ jobs:
           MEME_ENDPOINT: ${{ inputs.meme_endpoint }}
 
       - name: Attach android apks to release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@v2
         if: ${{ inputs.tag  != 'main' }}
         with:
           tag_name: ${{ inputs.tag }}


.github/workflows/android-fastlane.yml
```diff
@@ -189,14 +189,34 @@ jobs:
...
         with:
           tag_name: ${{ inputs.tag }}
           files: |
             mobile/build/app/outputs/flutter-apk/*.apk
             mobile/build/app/outputs/flutter-apk/*.apk.sha1
 
+      - name: Compute the release tag for continues ci build
+        if: ${{ inputs.tag  == 'main' }}
+        run: |
+          echo "release_tag=v`date '+%Y-%m-%d %H:%M:%S'`" >> $GITHUB_ENV
```
holzeis, line 203: ❓ why not simply call it latest?

.github/workflows/android-fastlane.yml
```diff
@@ -189,14 +189,34 @@ jobs:
...
+
+      - name: Attach android apks to release
+        uses: softprops/action-gh-release@v2
+        if: ${{ inputs.tag  == 'main' }}
+        with:
+          prerelease: true
+          name: ${{ env.release_tag }}
+          tag_name: ${{ env.release_tag }}
+          target_commitish: main
+          generate_release_notes: true
```
holzeis, line 213: 🔧 I don't think we need release notes for that.

.github/workflows/android-fastlane.yml
bonomat: imho `latest` is meant to be the "real" latest release, i.e. our proper release

.github/workflows/android-fastlane.yml
bonomat: Imho they can come handy to know what was changed.

.github/workflows/android-fastlane.yml
holzeis: How about preview? I don't know, I am a bit worried that we will get hundreds of tags which will pollute our release page.

.github/workflows/android-fastlane.yml
holzeis: Wouldn't this always be just one line?

bonomat: Right... it will always be the PR which was merged...

.github/workflows/android-fastlane.yml
bonomat: Tags are meant to be unique and immutable, i.e. we shouldn't have a _moving_ tag which is moved for every new release. 

> I am a bit worried that we will get hundreds of tags which will pollute our release page.

This will indeed happen, but why is it a problem?

holzeis: > Tags are meant to be unique and immutable, i.e. we shouldn't have a _moving_ tag which is moved for every new release. 

Are you sure about that? Some tags have special meaning and change frequently. E.g. latest.


> This will indeed happen, but why is it a problem? 

I don't know. It doesn't "feel" right to have a full release for every merged PR. But we can give it a go and see if it annoys us.

holzeis (approved):

.github/workflows/android-fastlane.yml
bonomat: > Are you sure about that? Some tags have special meaning and change frequently. E.g. latest.

Very certain about this. Imho there is no best practice around having a `latest` tag neither. 

Unlike branches, tags are meant to be immutable, you can delete them but this might result in problems down the lines, e.g. we would need to 
- delete the released files or overwrite it
- you don't know which commit _latest_ is based on
- you can use tags to checkout and build your work on, if you do this and the tag gets deleted and recreated you will have problems.

> I don't know. It doesn't "feel" right to have a full release for every merged PR. But we can give it a go and see if it annoys us.

If this is the case, maybe we shouldn't do this after all 🤷. 
An alternative would be to simply do nightly releases.

bonomat:
Had a quick chat with @luckysori : we created a separate repository for this purpose: https://github.com/get10101/10101-test-apks . We will create the releases there to not pollute this repository.

Issue #2013, opened 2024-02-11
Stop periodically syncing sub channels


holzeis:
This is an unnecessary overhead and should be removed from the interval tasks.

Note, make sure to add the `dlc_channel_manager.periodic_check` back to the `manage_dlc_manager` task. 

```
if let Err(e) = self.dlc_channel_manager.periodic_check() {
            error!(
                "Error performing periodic check of the channel manager: {}",
                e
            );
        }
```

bonomat:
If we remove this, we might not be able to detect closed channels

PR #2473, opened 2024-04-25
feat: Wrap dlc messages with 10101 messages

### LINKED ISSUE #2380: Wrap rust-dlc messages with 10101 messages
This will allow us to wrap `ChannelRenewOffer` or `ChannelSettleOffer` messages with 10101 meta data. This is needed for

- Rollover
- Async Match (Expiry, Liquidation, Stop Loss, Take Profit)
- Resize

and will allow us to get rid of a parallel websocket communication to send protocol relevant data.

Our current design is very failure prune, especially if the user has an unstable connection. e.g. Websocket messages are not retried.

Related bugs:
- https://github.com/get10101/10101/issues/2136
- https://github.com/get10101/10101/issues/2361

holzeis:
This is primarily a refactoring, moving the dlc message handler from the `rust-dlc` dependency to 10101, allowing us in following steps to enrich and adapt protocol messages with 10101 meta data.

I figured I already open a PR for this change, before adapting the messages with 10101 meta data. My aim is to create custom messages for 10101 custom flows

- Expiry
- Liquidation
- Rollover
- Resizing

fixes #2380
fixes https://github.com/get10101/10101/issues/2013


mobile/native/src/trade/order/mod.rs
```diff
@@ -222,12 +222,7 @@ impl Order {
                 execution_price: Some(execution_price),
                 ..
             } => Some(execution_price),
-            _ => {
-                // TODO: The caller should decide how to handle this. Always logging an error is
-                // weird.
-                tracing::error!("Executed price not known in state {:?}", self.state);
-                None
-            }
+            _ => None,
```
luckysori, line 225: 🔧 Let's adapt the docs too.

crates/ln-dlc-node/src/message_handler.rs
```diff
@@ -0,0 +1,627 @@
...
+            }
+            Message::OnChain(_) | Message::SubChannel(_) => bail!("Unexpected dlc message"),
+        };
+
+        Ok(msg)
+    }
+}
+
+impl TenTenOneMessage {
+    pub fn get_reference_id(&self) -> Option<ReferenceId> {
```
luckysori, line 403: 🤡 Eventually we can actually move the reference ID onto the `TenTenOneMessage`.

luckysori (approved):
Large but straightforward diff. LGTM!

I know you wanted to talk about something, so maybe I've missed something in my review.

The only question I have is if this PR is backwards-compatible. This definitely looks like a breaking change to me, but I was wondering if we have to migrate anything at the database level, since we have a `dlc_messages` table.

PR #2478, opened 2024-04-26
fix: missing input for workflow

bonomat:
This needed now and was missing


holzeis:
Weird, didn't i fix that already with? 

https://github.com/get10101/10101/pull/2475

PR #2487, opened 2024-04-26
fix: Store matching fee and trader pnl into trade params

### LINKED ISSUE #2484: Trader realized pnl is not updated on the trade when resizing a position

holzeis:
fixes #2484


coordinator/src/db/trade_params.rs
```diff
@@ -42,6 +42,8 @@ pub(crate) fn insert(
             trade_params::trader_pubkey.eq(params.trader.to_string()),
             trade_params::direction.eq(Direction::from(params.direction)),
             trade_params::average_price.eq(params.average_price),
+            trade_params::matching_fee.eq(params.matching_fee.to_sat() as i64),
```
luckysori, line 45: Oops, I think this was my fault.

We should destructure the `TradeParams` so that the compiler can complain when a new field is missing or not used

PR #2489, opened 2024-04-26
refactor: Combine common crates into xxi-node

holzeis:
Combines the crates

`ln-dlc-storage`, `ln-dlc-node`, `commons` and `trade` into a single crate `xxi-node`


luckysori (commented):
I think it's a good idea to consider a refactor like this.

It would be good to benchmark the effect of combining the crates on compilation time though.

Also, I would propose a different name to replace `ln-dlc-node`: just `node`. We primarily use a `Node` type from that crate. And there are no other nodes in the domain of 10101. Alternatively, if we end up merging several crates, we could use something like `core`.

holzeis:
@luckysori happy to rename the crate to `node` or `core`. Waiting for @bonomat's feedback to change it though.

holzeis:
@luckysori quick heads up. The names `core` and `node` are ambiguous and are not a simple rename.. 

I think `core` is used by the `stdlib` and node is also used on the `coordinator` and `app` side to wrap the `xxi-node`

bonomat:
To be honest, I'm not a big fan of renamings like these without a proper reason (maybe I don't see the reason?). It's just extra work for not much gain and it's always more work than "just" renaming and moving things which we will all have to pay the price for.
E.g. your CI is failing, or not running: 

image

luckysori:
> To be honest, I'm not a big fan of renamings like these without a proper reason (maybe I don't see the reason?).

There are at least a couple of things happening in this PR:

1. Renaming `ln-dlc-node` to `xxi-node`.
2. Moving crates into `xxi-node`.

I think dropping the `ln-dlc-node` name does make sense, because we no longer support LN, and we don't intend to go back to the original protocol AFAIK. When we support Lightning again, it will be a very different approach.

I don't love the chosen name, but I think I can live with it since `node` is problematic.

Moving code into `xxi-node` may or may not be a good idea. This is why I asked for measuring the effect of this change, @holzeis.

holzeis:
> Moving code into xxi-node may or may not be a good idea.

The change was actually all about moving things together. The naming can be changed fairly easily.

The reason why I want commons, trade and ln-dlc-node combined is because we want to enrich the ten ten one messages with meta data which we'd either have to replicate (which I really dislike since we already have about 7 different order definitions) or depend on the commons / trade crate from the ln-dlc-node.

These added dependency shows that these are not independent, that's why I think it should be combined.

I might have overdone it with the ln-dlc-storage crate, but this I can revert.

luckysori (approved):
> The reason why I want commons, trade and ln-dlc-node combined is because we want to enrich the ten ten one messages with meta data

Fair enough. This context was useful to make a decision. I forgot about it and I don't think it was mentioned in the PR.

PR #2493, opened 2024-04-29
feat: store os in db

bonomat:
resolves https://github.com/get10101/meta/issues/444


holzeis:
@bonomat how would that look for iOS? Does it include the version?

bonomat:
> @bonomat how would that look for iOS? Does it include the version?

it does not include a version, only the os, i.e. one of these values: https://doc.rust-lang.org/std/env/consts/constant.OS.html

holzeis:
> > @bonomat how would that look for iOS? Does it include the version?
> 
> it does not include a version, only the os, i.e. one of these values: [doc.rust-lang.org/std/env/consts/constant.OS.html](https://doc.rust-lang.org/std/env/consts/constant.OS.html)

With the [device_info_plus](https://pub.dev/packages/device_info_plus) (which I think we are already using) we can get the os version as well.

coordinator/migrations/2024-04-29-051613_add_os_to_user_table/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+    ADD COLUMN os TEXT;
```
holzeis, line 2: 🔧 Missing newline

holzeis (approved):

PR #2494, opened 2024-04-30
chore: Remove left-over lightning dependencies

holzeis:
@bonomat @luckysori do we still need to point to our lightning fork?


luckysori (approved):
Nice!

PR #2495, opened 2024-04-30
feat: Move match, async match and rollover websocket events to corresponding offer message

### LINKED ISSUE #2136: Failed to process incoming DLC message: No order to mark as Filled
At the moment, if we can't find an order in `Filling` we stop everything. In case of https://github.com/get10101/meta/issues/355 that got the protocol stuck again, because the position was not deleted and the last message was also not sent back to the user.

Now the state is broke and has to be cleaned up.

I propose we do not stop the process in case we can't find an order in `Filling` but rather print a warning as this is not a super critical thing.

### LINKED ISSUE #2361: Missed async match leads to expired position to be not removed
When a position expires the user gets a web socket notification containing the async order that will be created in the order history. As the protocol moves on the async order will eventually be set to filled when the protocol finishes and the position gets removed.

It appears the user managed to open the app and connect to the p2p endpoint, which triggered the closing of the expired position. However, before the app was able to receive the async match notification, the app was closed again. Upon opening the app again, the dlc protocol continued and succeeded. But the position remained.

```
Failed to process incoming DLC message: No order to mark as Filled
```

|	|	|	|
|:-:|:-:|:-:|
|IMG_0455|IMG_0454|IMG_0453|

### LINKED ISSUE #2476: Add order to offer messages
Following up on https://github.com/get10101/10101/pull/2473 we can add 10101 meta data to our custom messages. Adding the order to the offer messages (`Offer`, `RenewOffer`, `SettleOffer`) allows us to not having to rely on that the order is in state `Filling` to get processed.

With the order part of the offer message we can simply fallback to the order sent by the coordinator. This is especially useful for async orders like expiry and liquidation, which are not initiated on the app side and for which we are using the websocket connection at the moment.

Note, the app should still though verify that the received order / offer matches with the local data.

related to https://github.com/get10101/10101/issues/2136

holzeis:
- Introduces a new protocol message to differentiate between a Rollover and a regular Renewal
- Removes the async match websocket event and adds the filled_with and order to the `TenTenOneSettleOffer` message
- Removes the match websocket event and adds the filled_with tto the `TenTenOneChannelOffer` and `TenTenOneRenewOffer` messages.

fixes #2361 
fixes #2476
fixes #2136


crates/xxi-node/src/message_handler.rs
```diff
@@ -155,8 +161,13 @@ pub struct TenTenOneSignChannel {
     pub sign_channel: SignChannel,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct TenTenOneSettleOffer {
+    pub order: Order,
+    // TODO(holzeis): We might want to consider adding the order to the filled with struct. Having
+    // this separated doesn't make much sense anymore since, the filled with is not separately
+    // processed by the app anymore.
```
luckysori, line 169: 🙃 I think this TODO belongs on the `FilledWith` or the `Order` type.

crates/xxi-node/src/message_handler.rs
```diff
@@ -658,10 +706,102 @@ fn read_tentenone_message<R: ::std::io::Read>(
...
+
+        let handler = TenTenOneMessageHandler::new(Arc::new(NodeEventHandler::new()));
+        let message = handler
+            .read(message_type, &mut reader)
+            .expect("to be able to read the message")
+            .expect("to have a message");
+
+        assert!(matches!(
+            message,
+            WireMessage::Message(TenTenOneMessage::SettleOffer(_))
```
luckysori, line 804: 🔧 I think we should check against the whole body. It would be a good opportunity to use `insta` too.

crates/xxi-node/src/message_handler.rs
```diff
@@ -658,10 +706,102 @@ fn read_tentenone_message<R: ::std::io::Read>(
...
+
+        let mut reader = Cursor::new(&mut buf);
+        let message_type =
+            <u16 as Readable>::read(&mut reader).expect("to be able to read the type prefix.");
+
+        let handler = TenTenOneMessageHandler::new(Arc::new(NodeEventHandler::new()));
+        let message = handler
+            .read(message_type, &mut reader)
+            .expect("to be able to read the message")
+            .expect("to have a message");
```
luckysori, line 800: 🙃 We can `unwrap` in test code.

crates/xxi-node/src/message_handler.rs
```diff
@@ -658,10 +706,102 @@ fn read_tentenone_message<R: ::std::io::Read>(
...
+    use lightning::util::ser::Readable;
+    use lightning::util::ser::Writeable;
+    use secp256k1::PublicKey;
+    use std::io::Cursor;
+    use std::str::FromStr;
+    use std::sync::Arc;
+    use time::OffsetDateTime;
+
+    #[test]
+    fn test_impl_serde_writeable() {
```
luckysori, line 745: 🙃 Rename to `test_settle_offer_impl_serde_writeable`?

Thank you for adding a test, btw. We should do the same for all the messages with the help of `insta`. Then we can detect breaking changes here.

crates/xxi-node/src/node/dlc_channel.rs
```diff
@@ -302,8 +315,18 @@ impl<D: BdkStorage, S: TenTenOneStorage + 'static, N: LnDlcStorage + Sync + Send
                     Some(protocol_id),
                 )?;
 
+                // TODO(holzeis): Separate into different functions for renew and rollover.
```
luckysori, line 318: 🔧 I think we should do this straight away :D

Also, if we want to use 10101 terminology consistently, I think we should not use the word `renew`. The DLC channel renew protocol corresponds to open-position-after-settle, resize-position or rollover. I think we should stick with those at this level. Since the first two are interchangeable we could say `propose_reopen_or_resize`?

crates/xxi-node/src/message_handler.rs
```diff
@@ -432,18 +443,23 @@ impl TryFrom<Message> for TenTenOneMessage {
...
                 TenTenOneMessage::Reject(TenTenOneReject { reject })
             }
-            Message::OnChain(_) | Message::SubChannel(_) => bail!("Unexpected dlc message"),
+            Message::OnChain(_)
+            | Message::SubChannel(_)
+            | Message::Channel(ChannelMessage::RenewOffer(_))
+            | Message::Channel(ChannelMessage::SettleOffer(_))
+            | Message::Channel(ChannelMessage::Offer(_)) => {
+                unreachable!()
+            }
```
luckysori, line 452: 👍 I think I don't mind the `unreachable!`, but we should change the function signature to be infallible then.

crates/xxi-node/src/node/dlc_manager.rs
```diff
@@ -77,7 +77,7 @@ impl<D: BdkStorage, S: TenTenOneStorage + 'static, N: Storage + Sync + Send + 's
             .on_dlc_message(&message.into(), to_secp_pk_29(node_id))?;
 
         let response = match response {
-            Some(response) => Some(TenTenOneMessage::try_from(response)?),
+            Some(response) => Some(TenTenOneMessage::build_from_response(response)?),
```
luckysori, line 80: 🔧 It could be a `From` since it's infallible now 🙂

mobile/native/src/dlc/node.rs
```diff
@@ -517,21 +531,68 @@ impl Node {
...
+                event::publish(&EventInternal::BackgroundNotification(
+                    BackgroundTask::AsyncTrade(order_reason.into()),
+                ));
+
+                order::handler::async_order_filling(&offer.order, &offer.filled_with)
+                    .with_context(|| {
+                        format!(
+                        "Failed to process async match update from orderbook. order_id {order_id}"
+                    )
+                    })?;
```
luckysori, line 558: 🔧 `rustfmt` gave up here. Can you format it manually?

mobile/native/src/dlc/node.rs
```diff
@@ -517,21 +531,68 @@ impl Node {
...
+            .average_execution_price()
+            .to_f32()
+            .expect("to fit into f32");
+
+        let matching_fee = filled_with.order_matching_fee();
+
+        // TODO(holzeis): Add handling to be more resilient if the order could not be found.
+        order::handler::order_filling(order_id, execution_price, matching_fee).with_context(|| {
+            format!("Failed to process match update from orderbook. order_id = {order_id}")
+        })
```
luckysori, line 593: 🔧 I disagree with the TODO.

If the app doesn't know about the order match I don't think it should trust the coordinator. At the moment we don't even verify that the created order and the matched order are the same beyond the ID, which is already problematic.

luckysori (approved):
I love it. I think this is a great fix for old and future problems.

mobile/native/src/dlc/node.rs
holzeis: The todo is probably not clear, but this is actually for the scenario where the order exists but not in the expected state.

crates/xxi-node/src/message_handler.rs
holzeis: `insta` is really cool. thanks for the tip.

mobile/native/src/dlc/node.rs
luckysori: Ah, fair enough. That makes sense then.

crates/xxi-node/src/node/dlc_manager.rs
holzeis: I don't think its a good idea, because a `From` implementation would suggest it can transform any `Message` into a `TenTenOneMessage` which is not true. This only works because we know that we are dealing with a response which can never be an offer. I feel this context would get lost in a `From` implementation.

Also, if we were to enhance further messages, besides the `Offer` messages, this `From` implementation may need more arguments to transform a `Message` into a `TenTenOneMessage`.

crates/xxi-node/src/message_handler.rs
holzeis: I've added one additional test, but then stopped. Creating all this additional dummy data is just taking more effort than I'd like to invest right now.

mobile/native/src/dlc/node.rs
holzeis: Actually you've been right.. we use the order id here already so that should be fine. Nothing to do.

crates/xxi-node/src/node/dlc_manager.rs
luckysori: You are right; I am wrong.

😛

mobile/native/src/dlc/node.rs
luckysori: Ah, we don't actually check the original state when we set to `Filling`! Nice then.

PR #2518, opened 2024-05-09
chore: set android compile version to 34

bonomat:
This seems to be needed for `flutter_inappwebview`


bonomat:
Shouldn't be controversial.

PR #2536, opened 2024-05-14
Fix Flutter versions

luckysori:
To avoid random breaking changes showing up in CI.

AFAIK these were the versions we were using before `stable` was updated.

### DIFF
--- a/mobile/.fvmrc
+++ b/mobile/.fvmrc
@@ -1,3 +1,3 @@
 {
-  "flutter": "stable"
+  "flutter": "3.19.4"
 }
--- a/webapp/frontend/.fvmrc
+++ b/webapp/frontend/.fvmrc
@@ -1,3 +1,3 @@
 {
-  "flutter": "beta"
+  "flutter": "3.22.0"
 }


bonomat (changes requested):
Let me check if the `"3.22.0"` has the needed fix for us.

bonomat (approved):
Nice, works. Meaning we do not need `beta`-channel anymore. I'm happy to remove `fvm` again if you guys want? 

@luckysori / @holzeis ?

luckysori:
> Nice, works. Meaning we do not need `beta`-channel anymore. I'm happy to remove `fvm` again if you guys want?
> 
> @luckysori / @holzeis ?

Hmm, I don't mind FVM. I like the fact that I was able to use the `.fvmrc` to set a specific version.

holzeis:
> Nice, works. Meaning we do not need `beta`-channel anymore. I'm happy to remove `fvm` again if you guys want? 
> 
> 
> 
> @luckysori / @holzeis ?

I would like to get rid of it if we don't need it. One less tooling on our stack.

@bonomat I guess it also means we can use the same flutter version for web app and app?

luckysori:
> > Nice, works. Meaning we do not need `beta`-channel anymore. I'm happy to remove `fvm` again if you guys want? 
> > 
> > 
> > 
> > @luckysori / @holzeis ?
> 
> I would like to get rid of it if we don't need it. One less tooling on our stack.
> 
> @bonomat I guess it also means we can use the same flutter version for web app and app? 

We need to fix the versions for CI though.

This PR did not change the versions btw. I just changed the target from a moving target to a fixed target. Because a PR was failing after `stable` was updated.

PR #2538, opened 2024-05-15
fix: Don't refresh TradingView when switching from wallet to trading tab

### LINKED ISSUE #2525: Trading view is reloaded when switching between tabs
https://github.com/get10101/10101/assets/382048/8f0f9fad-19a7-4484-ac24-294ab7ca1585

Restioson:
Fixes #2525 

I can say for certain that going into settings still triggers a refresh - I'm not sure if we care about this, though. Otherwise, it seems to work and it does not refresh when swapping from wallet to trading tab. If this general approach is acceptable (which maybe has wider-reaching impacts?), then I can re-add the animation to swap from wallet to trade tabs and we can merge this.

### DIFF
--- a/mobile/lib/common/routes.dart
+++ b/mobile/lib/common/routes.dart
@@ -174,55 +174,58 @@ GoRouter createRoutes() {
                       },
                     )
                   ]),
-              ShellRoute(
-                navigatorKey: shellNavigatorKey,
+              StatefulShellRoute.indexedStack(
                 builder: (BuildContext context, GoRouterState state, Widget child) {
                   return ScaffoldWithNavBar(
                     child: child,
                   );
                 },
-                routes: <RouteBase>[
-                  GoRoute(
-                    path: WalletScreen.route,
-                    builder: (BuildContext context, GoRouterState state) {
-                      return const WalletScreen();
-                    },
-                    routes: <RouteBase>[
-                      GoRoute(
-                        path: SendOnChainScreen.subRouteName,
-                        // Use root navigator so the screen overlays the application shell
-                        parentNavigatorKey: rootNavigatorKey,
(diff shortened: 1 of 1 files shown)


holzeis:
I think the approach makes sense.

bonomat (approved):
Nice solution. What other implication could this have?

Restioson:
> Nice solution. What other implication could this have?

It keeps the state for all of the pages under the wallet & trade screen

bonomat:
> > Nice solution. What other implication could this have?
> 
> It keeps the state for all of the pages under the wallet & trade screen. This may or may not be an issue but I'm not quite sure how to evaluate it.

I guess this is ok given our application is quite small.

Restioson:
Going to update the animation now and then make this ready for merge

Restioson:
Here is how it looks on main:
video on main


And on this branch:
video on this branch

We could try implement a slide animation or leave it as is

holzeis:
> Actually, this branch might be fine.
> 
> Here is how it looks on main: [youtube.com/shorts/2LxbveDbIGc](https://www.youtube.com/shorts/2LxbveDbIGc)
> 
> And on this branch [youtube.com/shorts/BO8j9-eOpKo?feature=share](https://youtube.com/shorts/BO8j9-eOpKo?feature=share)
> 
> We could try implement a slide animation or leave it as is

The webview plugin we are using is not supported on desktop. Try running it in a simulator.

Restioson:
> The webview plugin we are using is not supported on desktop. Try running it in a simulator.

I'm not testing the webview plugin but rather the animation of switching between the two views themselves. It works fine with the webview plugin in the simulator

Restioson:
I'm going to merge this for now and we can change the animation in a followup if we like

PR #2539, opened 2024-05-15
feat: Rewrite order matching strategy

holzeis:
Refactors the current trading component into a clearly separated orderbook component and a trade execution component. The linking part is the `ExecutableMatch` which can be derived from the matches stored into the database.

At the moment we assume optimistically that the trade execution will succeed. However, we should consider that a pending match may never get filled or it fails at execution in such a scenario we would need to rollback the matched orders.

An overview of the changes:

- Adds the order id to the trade entry so we can link which order created the trade.
- Allows for orders to get partially matched and update the `matches` table accordingly
- Matched limit orders are created as `Filled` from the get go since we do not execute them with our own maker.
- (Partially) Matched limit orders are not deleted anymore, but the quantity is updated to how much has already been mapped. (e.g. if an open limit order of $1000 gets deleted with $300 matched it will be updated to `Taken` with a quantity of $300)

Example of a matches table after opening and closing a position of $3576. Note the maker generates $1000 orders.

```
orderbook=# select execution_price, quantity, order_id, match_order_id from matches where trader_id='032752136072951607b66dd85f236363efab91b315b1c4fc8d9530ee873875108a' and match_state='Filled' order by created_at;
 execution_price | quantity |               order_id               |            match_order_id            
-----------------+----------+--------------------------------------+--------------------------------------
           60001 |     1000 | 42dd8a30-a20d-4640-bddd-f5467a638d28 | 9acc7260-0f24-405e-ad84-550a797b4464
           60001 |     1000 | 42dd8a30-a20d-4640-bddd-f5467a638d28 | c486a21c-312e-4ae0-ab35-38a8e93d07d5
           60001 |     1000 | 42dd8a30-a20d-4640-bddd-f5467a638d28 | e30ff0dc-4f8a-4888-a3ea-cddf63c94179
           60001 |      576 | 42dd8a30-a20d-4640-bddd-f5467a638d28 | adae6065-d78a-4ed6-bab3-9852f395ef67
           59999 |     1000 | a6be5b42-8817-40f7-8f86-0615df33c57a | 19811f8f-7fe7-4dd1-9b3a-72f24a0fcee5
           59999 |     1000 | a6be5b42-8817-40f7-8f86-0615df33c57a | 6cf3b7bd-2652-4590-afed-f15ece0a1ded
           59999 |     1000 | a6be5b42-8817-40f7-8f86-0615df33c57a | 0c8b6575-2ba2-4e37-8aa2-9c0e8beb055d
           59999 |      576 | a6be5b42-8817-40f7-8f86-0615df33c57a | 17fcb395-4434-4ca8-9cb2-4ca75ec73063
(8 rows)
```

Example of updated limit orders when deleted.

```
orderbook=# select order_id, price, direction, quantity, order_type from orders where order_id in ('19811f8f-7fe7-4dd1-9b3a-72f24a0fcee5', '6cf3b7bd-2652-4590-afed-f15ece0a1ded', '0c8b6575-2ba2-4e37-8aa2
               order_id               | price | direction | quantity | order_type 
--------------------------------------+-------+-----------+----------+------------
 19811f8f-7fe7-4dd1-9b3a-72f24a0fcee5 | 59999 | long      |     1000 | limit
 6cf3b7bd-2652-4590-afed-f15ece0a1ded | 59999 | long      |     1000 | limit
 0c8b6575-2ba2-4e37-8aa2-9c0e8beb055d | 59999 | long      |     1000 | limit
 17fcb395-4434-4ca8-9cb2-4ca75ec73063 | 59999 | long      |      576 | limit
(4 rows)
```

TODO:

- [x] Do not await on database transactions when interacting with the orderbook.
- [x] Process any order update through the orderbook component.
- [x] Rework matching logic to not immediately mutate on every single order.


coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
-            }
-            // Descending order.
-            b.price.cmp(&a.price)
-        })
+    /// updates an order in the orderbook. Ignores the update if order does not exist.
+    pub fn update_order(&mut self, updated_order: Order) {
+        if let Some(order) = self.orders.iter_mut().find(|o| o.id == updated_order.id) {
+            *order = updated_order;
+        }
+        self.sort(updated_order.direction);
```
luckysori, line 70: ❓Do we need `update`? How about removing and adding a new one?

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+                commons::Direction::Long => self.longs.add_order(order),
             }
         }
 
         Ok(())
     }
 
-    async fn add_order(
+    /// Adds a limit order to the orderbook.
+    async fn add_limit_order(&mut self, new_order: NewLimitOrder) -> Result<Message> {
```
luckysori, line 192: ❓So we cannot match limit orders with limit orders yet?

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+            move || {
+                let mut conn = pool.get()?;
+                orders::set_order_state(&mut conn, order_id, OrderState::Failed)?;
+                anyhow::Ok(())
             }
         };
 
-        Ok(message)
+        let matched_quantity: Decimal = matched_orders.iter().map(|o| o.quantity).sum();
+        if matched_quantity != order.quantity {
```
luckysori, line 252: 🔧 It's simpler if a failed `match_order` returns an error and does not modify the `OrderbookSide`. It's weird to have to roll back here.

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+            true => None,
+            false => Some(self.orders.remove(0)),
+        }
+    }
+    /// matches orders from the orderbook for the given quantity.
+    pub fn match_order(&mut self, quantity: Decimal) -> Vec<Order> {
+        let mut matched_orders = vec![];
+
+        let mut quantity = quantity;
+        while let Some(mut order) = self.take_order() {
```
luckysori, line 88: 🤔 I'm not a big fan of mutating the `OrderbookSide` before you know if you are gonna match the orders or not. This leads to having to roll back your work afterwards.

I think we should instead iterate through the top of the orderbook side and only take the orders after we have confirmed that we should.

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+        match order.direction {
+            commons::Direction::Short => self.shorts.add_order(order),
+            commons::Direction::Long => self.longs.add_order(order),
+        }
+        Ok(Message::NewOrder(order))
+    }
+
+    /// Matches a market order against the orderbook. Will fail if the market order can't be fully
+    /// matched.
+    async fn match_market_order(
```
luckysori, line 214: 🤔 I'm not sure any of these methods need to be `async`.

coordinator/src/orderbook/mod.rs
```diff
@@ -171,27 +376,38 @@ impl Orderbook {
...
             let mut conn = self.pool.clone().get()?;
             move || {
-                let order = orders::delete_trader_order(&mut conn, order_id, trader_pubkey)?;
+                let matches = matches::get_matches_by_order_id(&mut conn, order_id)?;
+
+                let order = if !matches.is_empty() {
+                    // order has been at least partially matched.
+                    let matched_quantity = matches.iter().map(|m| m.quantity).sum();
+
+                    orders::set_order_state_partially_taken(&mut conn, order_id, matched_quantity)
```
luckysori, line 394: ❓I don't understand what's going on here. Why do we need to do this when a limit order is being removed?

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+                commons::Direction::Short => self.shorts.rollback_matched_orders(matched_orders),
+                commons::Direction::Long => self.longs.rollback_matched_orders(matched_orders),
+            }
+
+            spawn_blocking(fail_order).await??;
+            bail!("No match found.");
+        }
+
+        // apply match to database.
+        let matches = spawn_blocking({
```
luckysori, line 271: 🤔 This whole block scares me. It looks like we have already established a match using our in-memory orderbook, but now have to go to the database a bunch of times to get this done. I think we want to defer most of this work, and avoid having to load the `matching_fee` here.

crates/xxi-node/src/commons/message.rs
```diff
@@ -80,7 +80,11 @@ pub enum OrderbookRequest {
         os: Option<String>,
         signature: Signature,
     },
+    // TODO(holzeis): This orderbook request seems to be unused and a duplicate to our http
+    // endpoint. I guess we can remove it.
     InsertOrder(NewLimitOrder),
+    // TODO(holzeis): This orderbook request seems to be unused and a duplicate to our http
+    // endpoint. I guess we can remove it.
```
luckysori, line 87: Nope, this is how the maker communicates with the orderbook. We want to get rid of the HTTP requests instead and use a WS connection between app and coordinator.

luckysori (commented):
I like the idea of defining an orderbook component which keeps the orders in memory, can be optimised for speed (eventually) and has exclusive access to certain DB tables. I think this is the vision suggested by this PR.

coordinator/src/orderbook/db/matches.rs
```diff
@@ -37,16 +36,33 @@ struct Matches {
...
-    }
+pub fn insert(
+    conn: &mut PgConnection,
+    order: &commons::Order,
+    matched_order: &commons::Order,
+    matching_fee: Amount,
+    match_state: commons::MatchState,
+    quantity: Decimal,
+) -> QueryResult<commons::Matches> {
+    let match_ = Matches {
```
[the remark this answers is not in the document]
holzeis: It's because match is a key word and I was lazy to think of a better name 🙈

coordinator/src/orderbook/mod.rs
holzeis: Yes not yet. I thought about addressing this when implementing limit orders for real 🙂

coordinator/src/orderbook/mod.rs
holzeis: Because it might have been partially matched already.

Today we are setting an order to deleted regardless if it has been matched or not. IMHO we should only set open orders to deleted, but not taken ones.

However how to deal with orders that have been partially matched? I opted to set the state to taken and reduce the quantity to the amount that has already been matched.

crates/xxi-node/src/commons/message.rs
holzeis: Makes sense, the dev-maker is using the http endpoint though. I guess that confused me 🙂

coordinator/src/orderbook/mod.rs
```diff
@@ -6,3 +29,253 @@ pub mod websocket;
...
+impl Orderbook {
+    pub fn new(pool: Pool<ConnectionManager<PgConnection>>) -> Self {
+        Self {
+            pool,
+            shorts: Shorts(vec![]),
+            longs: Longs(vec![]),
+        }
+    }
+
+    async fn initialize(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
holzeis: Hmm.. definitely possible, but I usually refrain from adding too much logic into a constructor.

coordinator/src/orderbook/mod.rs
```diff
@@ -6,3 +29,253 @@ pub mod websocket;
...
+                    }
+                    NewOrder::Limit(o) => orders::insert_limit_order(&mut conn, o, order_reason),
+                }
+                .map_err(|e| anyhow!(e))
+                .context("Failed to insert new order into DB")?;
+
+                anyhow::Ok(order)
+            }
+        })
+        .await??;
```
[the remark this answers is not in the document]
holzeis: I am afraid if we simply do not await we might miss errors or even inserts e.g. when the coordinator restarts before the insert is done.

I was thinking of a task running e.g. every few seconds could sync the in memory orderbook with the database. On a graceful shutdown we can then sync any pending updates to ensure we do not miss any orders.

coordinator/src/orderbook/mod.rs
```diff
@@ -6,3 +29,253 @@ pub mod websocket;
...
+                    commons::Direction::Long => self.longs.add(order.clone()),
+                }
+                Some(Message::NewOrder(order))
+            }
+        };
+
+        Ok(message)
+    }
+
+    async fn update_order(&mut self, order: Order) -> Result<Option<Message>> {
```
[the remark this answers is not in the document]
holzeis: I think it only makes sense for limit orders.

coordinator/src/orderbook/mod.rs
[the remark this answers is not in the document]
luckysori: I think we need to prioritise speed over everything else. Restarts are very rare and we could handle them gracefully.

In case of a panic, I am not even sure we can easily run into trouble.

coordinator/src/orderbook/mod.rs
luckysori: Shouldn't we do that earlier and not on deletion though. Also, why do we go directly against the database instead of updating the state in the ordebook cache? It seems like the state is only in the DB? Why is that?

crates/xxi-node/src/commons/message.rs
luckysori: Fair enough. I was too lazy to update the dev maker back when I changed this 😔

coordinator/src/orderbook/db/orders.rs
```diff
@@ -410,6 +410,22 @@ pub fn delete(conn: &mut PgConnection, id: Uuid) -> QueryResult<OrderbookOrder>
...
+pub fn set_order_state_partially_taken(
+    conn: &mut PgConnection,
+    id: Uuid,
+    quantity: Decimal,
+) -> QueryResult<OrderbookOrder> {
+    let order: Order = diesel::update(orders::table)
+        .filter(orders::order_id.eq(id))
+        .set((
+            orders::order_state.eq(OrderState::Taken),
+            orders::quantity.eq(quantity.to_f32().expect("to fit")),
```
bonomat, line 422: 🔔 
I think we shouldn't change the quantity. This will make it hard to do any type of analysis on orders in the future. Instead, we should add an additional field which indicates how much has been used already, e.g. `used_quantity`.
This can all happen on the db layer if we don't care about this at the moment, i.e. when loading the order from the orderbook we set the in-memory `quantity`=`quantity`-`used_quantity`

coordinator/src/orderbook/mod.rs
bonomat, line 70: Imho we can assume that the rate never changes and hence we don't need to resort the list after an update.

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+            order_matching_fee_rate,
         }
     }
 
+    /// Initializes the orderbook with non expired open limit orders.
     async fn initialize(&mut self) -> Result<()> {
         let all_orders = spawn_blocking({
             let mut conn = self.pool.clone().get()?;
             move || {
+                // TODO(holzeis): join with trades to get partially matched orders.
```
bonomat, line 171: Is this needed? You are updating the order::quantity so that you have already the partially matched  orders.

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
                 let all_orders =
                     orders::get_all_orders(&mut conn, OrderType::Limit, OrderState::Open, true)?;
                 anyhow::Ok(all_orders)
             }
         })
         .await??;
 
+        // TODO(holzeis): maybe consider batch adding all orders and sort afterwards instead of
```
bonomat, line 179: Imho you should do this in `Orderbook::new` and then simply overwrite all existing orders. This way you can simply split `all_orders` by direction and sort the resulting lists.

bonomat (commented):
What's the status of this PR? Do we want to continue with it?

coordinator/src/orderbook/mod.rs
holzeis: The matched order quantities are only updated in memory but not in the database. When initialising we need to get load how much has already been matched.

holzeis:
> What's the status of this PR? Do we want to continue with it?

I will continue to work on this after the lightning onboarding flow.

coordinator/src/orderbook/mod.rs
holzeis: you mean if we aren't waiting for the blocking database transactions? if you you might be right.

coordinator/src/orderbook/mod.rs
holzeis: can you elaborate on that? Why shouldn't I update the price of a limit order?

coordinator/src/orderbook/mod.rs
holzeis: done

coordinator/src/orderbook/mod.rs
```diff
@@ -6,3 +29,253 @@ pub mod websocket;
...
+        })
+    }
+}
+
+pub struct Longs(Vec<Order>);
+
+impl Longs {
+    pub fn add(&mut self, order: Order) {
+        self.0.push(order);
+        self.sort();
```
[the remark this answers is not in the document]
holzeis: I've changed it to a BTreeMap implementation.

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
-}
+    /// removes the order by the given id from the orderbook.
+    pub fn remove_order(&mut self, order_id: Uuid) {
+        self.orders.retain(|o| o.id != order_id);
+    }
+    /// removes the first order from the sorted list representing the best order in the orderbook.
+    pub fn take_order(&mut self) -> Option<Order> {
+        match self.orders.is_empty() {
+            true => None,
+            false => Some(self.orders.remove(0)),
```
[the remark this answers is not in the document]
holzeis: refactored with a `BTreeMap`

coordinator/src/orderbook/mod.rs
```diff
@@ -30,120 +41,317 @@ pub mod websocket;
...
+impl OrderbookSide {
+    pub fn new() -> Self {
+        Self { orders: vec![] }
     }
-    pub fn remove(&mut self, order_id: Uuid) {
-        self.0.retain(|o| o.id != order_id)
+    /// adds the given order to the orderbook.
+    pub fn add_order(&mut self, order: Order) {
+        self.orders.push(order);
+        self.sort(order.direction);
```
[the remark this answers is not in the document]
holzeis: refactored with a `BTreeMap`

coordinator/src/orderbook/db/orders.rs
holzeis: This is only done if the user deletes his order. If he deletes the order which has already been partially matched. the order quantity is updated to the matched amount and the state is set to `Taken`. I found this to be the most correct way of handling a delete order request of an already partially matched order.

Would you still add an optional `used_quantity` field which is only set when the user deletes his order? and what should be the state of the order `Taken` or `Deleted`?

coordinator/src/orderbook/mod.rs
bonomat: I would not allow this and only allow for changing the quantity. Updating the price makes things more complicated. E.g. on BitMEX this is also not possible.

coordinator/src/orderbook/db/orders.rs
bonomat: Ah ok, I think I misunderstood this part then.

coordinator/src/orderbook/mod.rs
bonomat: Got it, thanks.

coordinator/src/orderbook/mod.rs
holzeis: Happy to change that, but I don't see the complexity. Can you explain?

holzeis: I think its not used at the moment, but we have an `OrderbookMessage`, so I've added it preliminary.

coordinator/src/orderbook/mod.rs
holzeis: I've added a `commit_matched_orders` function, that only removes (takes) the orders from the orderbook if the quantity is enough.

coordinator/src/orderbook/mod.rs
bonomat: it's a simplification: 
- no need to resort lists on the server side nor on the client side if the order was updated
- what do you do if the order was already partially matched?

holzeis: > - no need to resort lists on the server side nor on the client side if the order was updated

I switched the implementation to a BTreeMap so the sorting is inherent to the data structure.

> - what do you do if the order was already partially matched?

Hmm.. that would be indeed a bit more complex. 

We would need to add a check that only as much quantity can be reduced as still open.

Other than that I guess I would simply update the order. But maybe we don't do updates at all and simply delete and insert, like @luckysori suggested?

bonomat: > Other than that I guess I would simply update the order. But maybe we don't do updates at all and simply delete and insert, like @luckysori suggested?

yes, I would say so.

holzeis:
@luckysori @bonomat this PR turned into more refactoring than originally anticipated, although not all issues are addressed yet I think its in a good spot to get into main, before tackling limit orders.

Below shows the new design of the orderbook.

- The orderbook component is solely responsible for the order. It's now the only place where the order gets manipulated.
- The match executor takes a matched order and stores that into a matches table. this table keeps record if the execution actually succeeded or not. The orderbook doesn't care anymore if the order execution fails or succeeds. From the point of view of the orderbook, the order has been taken. Eventually we might want to re-add a failed order, but for now we simply ignore it.
- The new design also goes a step forward towards decoupling the orderbook from the coordinator.
- The `OderState::Matched` is now deprecated as the matches table takes care of tracking if an order / match is still pending or completed (filled or failed).

Screenshot 2024-06-07 at 14 26 07

There are certainly still some minor things that I'd like to clean up, but its in a good state for your review.

PR #2541, opened 2024-05-15
Let the coordinator open single-funded DLC channels

luckysori:
With the `TradeAction::OpenSingleFundedDlcChannel`, the coordinator can choose to open a DLC channel with a trader, without using any trader UTXOs. This is useful if the coordinator wants to onboard trading through alternate means, such as Lightning.

`rust-dlc` now allows to configure the TX fee split between the offer party and the accept party.

The feature has been tested locally, but is currently disabled as we need a different flow to trigger the creation of a single-funded DLC channel.

Also, the DLC channel is now 0-conf: the trader can keep trading as soon as the fund transaction is found in mempool.

https://github.com/get10101/10101/assets/9418575/c8457061-b249-4ad1-b137-db85f9a81ee7

In the video you can see that we still have to fund the on-chain wallet, because the app still thinks that on-chain funds are needed to open a channel. But once the channel is created, you can see that the on-chain funds were not used!

---

See https://github.com/get10101/rust-dlc/pull/18 for the matching PR in `rust-dlc`.

### DIFF
--- a/coordinator/src/trade/mod.rs
+++ b/coordinator/src/trade/mod.rs
@@ -58,6 +58,8 @@ pub mod websocket;
 
 enum TradeAction {
     OpenDlcChannel,
+    #[allow(dead_code)]
+    OpenSingleFundedChannel,
     OpenPosition {
         channel_id: DlcChannelId,
         own_payout: u64,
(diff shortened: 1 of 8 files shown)


holzeis:
Nicely done 💪

I think there is not much left to get this ready 🙂

mobile/native/src/dlc/node.rs
```diff
@@ -540,7 +540,7 @@ impl Node {
         match self
             .inner
             .dlc_manager
-            .accept_channel(&channel_id)
+            .accept_channel(&channel_id, dlc::FeeConfig::AllOffer)
```
holzeis, line 543: 🤔 Don't we need to separate between the two flows here? or don't we support the `EvenSplit` flow any more from the app?

holzeis: I think you will have to look into the offered channel and use the corresponding `FeeConfig`

coordinator/src/trade/mod.rs
```diff
@@ -98,6 +100,18 @@ pub struct TradeExecutor {
     notifier: mpsc::Sender<OrderbookMessage>,
 }
 
+/// The funds the trader will need to provide to open a DLC channel with the coordinator.
+///
+/// We can extend this enum with a `ForTradeCost` variant to denote that the trader has to pay for
+/// everything except for transaction fees.
+enum TraderRequiredLiquidity {
+    /// Pay for margin, collateral reserve, order-matching fees and transaction fees.
```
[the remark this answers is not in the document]
luckysori: Yeah, I had an earlier version with other variants, but they were unused and I did not want to add any more dead code.

holzeis (approved):

PR #2544, opened 2024-05-16
Fund channel flow

bonomat:
Ready for review :)


bonomat:
Updated sneak peak: 


https://github.com/get10101/10101/assets/224613/d401f55a-b858-4196-a475-4eb3d8b35937

Open TODOs: 

- [ ] make it work without wallet balance: for this we need to change the validations on the order creation screen
- [ ] cleanup 😬

bonomat:
Final version before review :) 

https://github.com/get10101/10101/assets/224613/b3e9f395-bba1-444f-8a04-67893f062b03

mobile/native/src/max_quantity.rs
```diff
@@ -40,7 +39,21 @@ pub fn max_quantity(
...
         Amount::from_sat(channel_trade_constraints.max_counterparty_balance_sats);
-    let max_trader_balance = Amount::from_sat(channel_trade_constraints.max_local_balance_sats);
+
+    // If the trader has a channel, his max balance is the channel balance and we continue,
+    // otherwise we can return here as the max amount to trade depends on what the coordinator can
+    // provide
+    let max_trader_balance = if channel_trade_constraints.is_channel_balance {
+        Amount::from_sat(channel_trade_constraints.max_local_balance_sats)
+    } else {
+        return Ok(Decimal::from_f32(calculations::calculate_quantity(
```
holzeis, line 49: 🤔 what if the user is funding with the 10101 wallet?

mobile/native/src/api.rs
```diff
@@ -895,3 +902,24 @@ impl From<xxi_node::commons::ReferralStatus> for ReferralStatus {
...
+
+#[tokio::main(flavor = "current_thread")]
+pub async fn submit_unfunded_channel_opening_order(
+    funding_address: String,
+    order: NewOrder,
+    coordinator_reserve: u64,
+    trader_reserve: u64,
+) -> Result<()> {
+    let funding_address = Address::from_str(funding_address.as_str())?.assume_checked();
+    tracing::info!("I'm here, before starting the watch");
```
holzeis, line 914: 🙃 me too

mobile/native/src/api.rs
```diff
@@ -895,3 +902,24 @@ impl From<xxi_node::commons::ReferralStatus> for ReferralStatus {
...
+
+    unfunded_orders::submit_unfunded_channel_opening_order(
+        funding_address,
+        order,
+        coordinator_reserve,
+        trader_reserve,
+    )
+    .await?;
+    tracing::info!("I'm here, I've finished the watch");
```
holzeis, line 923: 🙃 me too

mobile/native/src/event/api.rs
```diff
@@ -30,6 +30,7 @@ pub enum Event {
     BackgroundNotification(BackgroundTask),
     Authenticated(TenTenOneConfig),
     DlcChannelEvent(DlcChannel),
+    FundingChannelNotification(FundingChannelTask),
```
holzeis, line 33: 🤔 do we really need a different kind of notification here? Why not use a background task?

Looks to me that's exactly what it is.

mobile/native/src/event/api.rs
```diff
@@ -190,3 +195,25 @@ pub struct Balances {
...
     pub off_chain: Option<u64>,
 }
+
+#[frb]
+#[derive(Clone)]
+pub enum FundingChannelTask {
+    Pending,
+    Funded,
+    Failed(String),
+    OrderCreated(String),
```
holzeis, line 205: That's a different task that you can subscribe to. Async order pending.

The rest is equal to the task status states.

mobile/lib/features/trade/channel_creation_flow/channel_funding_screen.dart
```diff
@@ -0,0 +1,452 @@
...
+          ),
+        )
+      ],
+    );
+  }
+
+  Column buildInfoBox(FundingChannelTaskStatus? value, FundinType selectedBox) {
+    String transactionStatusText = "Waiting for payment...";
+    String transactionStatusInformationText =
+        "Please wait. If you leave now, your position won’t be opened when the funds arrive.";
```
holzeis, line 304: 🤔 if the user would want to switch to another app to send the onchain funds this message might be misleading.

holzeis:
It's a bit weird that one jumps from the modal to a screen. What do you think about showing the new screens also as modals?

holzeis:
Also did you test this with smaller devices?

holzeis:
❓ Do we still need to show the QR code when the trade is executing? I think I would prefer if the user sees directly the position tab.

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
```diff
@@ -0,0 +1,512 @@
...
+        maxOnChainSpending - orderMatchingFees - fundingTxFeeWithBuffer - channelFeeReserve;
+
+    maxCounterpartyCollateralSats = (maxCounterpartyCollateral.sats * counterpartyLeverage).toInt();
+
+    fundWithWalletEnabled = maxUsableOnChainBalance.sats >= ownTotalCollateral.sats;
+
+    // We add this so that the confirmation slider can be enabled immediately
+    // _if_ the form is already valid. Otherwise we have to wait for the user to
+    // interact with the form.
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
holzeis, line 148: 🤔 that looks weird. Is that really the way to go?

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
```diff
@@ -0,0 +1,512 @@
...
+                    ),
+                    const SizedBox(height: 10),
+                    Center(
+                      child: Text.rich(
+                        textAlign: TextAlign.justify,
+                        TextSpan(
+                          children: [
+                            TextSpan(
+                              text:
+                                  "🔔 This is your first trade and you do not have a channel yet. "
```
holzeis, line 224: In the video this is very hard to read. I also think its too much text, maybe the title is enough?

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
```diff
@@ -0,0 +1,512 @@
...
+                          padding: const EdgeInsets.only(top: 15, bottom: 5, left: 10, right: 10),
+                          child: Column(
+                            mainAxisAlignment: MainAxisAlignment.center,
+                            children: [
+                              ValueDataRow(
+                                  type: ValueType.amount,
+                                  value: ownTotalCollateral,
+                                  label: 'Channel size'),
+                              FeeExpansionTile(
+                                  label: "Fee*",
```
holzeis, line 370: 🔧 The asterix isn't pointing to anything.

mobile/native/src/unfunded_orders.rs
```diff
@@ -0,0 +1,123 @@
...
+                        .map(|(_, amount)| amount.to_sat())
+                        .sum::<u64>();
+
+                    if total_unspent_amount_received >= needed_trader_reserve {
+                        tracing::info!(
+                            amount = total_unspent_amount_received.to_string(),
+                            address = funding_address.to_string(),
+                            "Address has been funded enough"
+                        );
+                        // TODO: notify frontend
```
holzeis, line 56: I think that todo is done.. isn't it?

mobile/native/src/max_quantity.rs
bonomat: His max quantity still depends on the coordinator's quantity. We have a frontend validation which does not let him chose to fund with his internal wallet if the selected channel size is too big.

mobile/native/src/api.rs
bonomat: upsi 😬

mobile/native/src/event/api.rs
bonomat: I didn't pick the `BackgroundTask` to keep things separated.

mobile/lib/features/trade/channel_creation_flow/channel_funding_screen.dart
bonomat: Could be. Do you have a proposal to change this?

bonomat:
> ❓ It's a bit weird to me that one jumps from the modal to a screen. What do you think about showing the new screens also as modals?

I agree, I switched to an extra screen because the modals were a pita to design.

bonomat:
> ❓ Did you test this with smaller devices?

Yes, tested iPhone SE and looks ok.

bonomat:
> ❓ Do we still need to show the QR code when the trade is executing? I think I would prefer if the user sees directly the position tab.

Probably not. I think there is lots we could improve. For example, I wouldn't show the popup dialog here at all but it's kinda impossible to control 😬

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
bonomat: I've copied this from the old screen 🙈

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
bonomat: Feel free to propose something else :)

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
bonomat: Fair, I wanted to indicate that one can click on it.

mobile/native/src/unfunded_orders.rs
bonomat: Jap :)

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
```diff
@@ -0,0 +1,512 @@
...
+    tentenoneConfigChangeNotifier = context.read<TenTenOneConfigChangeNotifier>();
+    var tradeConstraints = tentenoneConfigChangeNotifier.channelInfoService.getTradeConstraints();
+
+    DlcChannelService dlcChannelService =
+        context.read<DlcChannelChangeNotifier>().dlcChannelService;
+
+    quantity = widget.tradeValues.quantity;
+    leverage = widget.tradeValues.leverage;
+    liquidationPrice =
+        widget.tradeValues.liquidationPrice == null ? 0.0 : widget.tradeValues.liquidationPrice!;
```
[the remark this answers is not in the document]
bonomat: 👍

mobile/native/src/event/api.rs
holzeis: Hmm.. can you explain? From what do you want to keep it separate?

mobile/lib/features/trade/channel_creation_flow/channel_configuration_screen.dart
holzeis: > maybe the title is enough?

😅

mobile/native/src/event/api.rs
bonomat: From the flow how the `BackgroundTask` is being handled as I think I like to keep things separated. 
Imho, everything every event is a _background task_ as it's executed asynchronously in the _background_. 

Anyways, I have no hard feelings around this (💔). If you think I should merge it into the `BackgroundTask`, I'm happy to do so.

holzeis: No need, we can still change it, if we think it gets to annoying to maintain.

crates/xxi-node/src/node/mod.rs
```diff
@@ -335,6 +338,72 @@ impl<D: BdkStorage, S: TenTenOneStorage + 'static, N: Storage + Sync + Send + 's
...
+                            vout = index,
+                            "Failed at checking tx status {error:?}"
+                        );
+                    }
+                }
+            } else {
+                tracing::error!(
+                    txid = tx.txid.to_string(),
+                    address = address.to_string(),
+                    "Output not found. This should not happen, but if, it indicates something is wrong.");
```
luckysori, line 384:
```suggestion
                    "Output not found. This should not happen, but if it does, it indicates something is wrong.");
```

mobile/native/src/event/api.rs
```diff
@@ -125,6 +129,7 @@ impl Subscriber for FlutterSubscriber {
             EventType::ServiceHealthUpdate,
             EventType::ChannelStatusUpdate,
             EventType::BackgroundNotification,
+            EventType::FundingChannelNotification,
```
luckysori, line 132: I only spent 1 hour yesterday debugging because I forgot to add by new `EventType` ❤️

mobile/native/src/lib.rs
```diff
@@ -36,3 +36,4 @@ pub use report_error::report_error_to_coordinator;
 )]
 mod bridge_generated;
 mod position;
+pub mod unfunded_orders;
```
luckysori, line 39: 🙃 Should go with the other `pub mod`s in this file.

mobile/native/src/max_quantity.rs
```diff
@@ -126,7 +139,7 @@ pub fn max_quantity(
     Ok(max_quantity)
 }
 
-/// Calculates the max quantity for the given input parameters. If an on-chai fee estimate is
+/// Calculates the max quantity. If an on-chai fee estimate is
```
luckysori, line 142:
```suggestion
/// Calculates the max quantity. If an on-chain fee estimate is
```

mobile/native/src/unfunded_orders.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::trade::order;
+use crate::trade::order::api::NewOrder;
+use anyhow::Result;
+use bitcoin::Address;
+use bitcoin::Amount;
+use futures::FutureExt;
+use std::time::Duration;
+use xxi_node::commons::ChannelOpeningParams;
+
+pub(crate) async fn submit_unfunded_channel_opening_order(
```
luckysori, line 14: 🙃 Every channel that is not yet open is unfunded. We could go for `submit_unfunded_wallet_channel_opening_order` or `submit_channel_opening_order_without_utxos`.

mobile/native/src/unfunded_orders.rs
```diff
@@ -0,0 +1,123 @@
...
+    let node = state::get_node().clone();
+    let bdk_node = node.inner.clone();
+    event::publish(&EventInternal::FundingChannelNotification(
+        FundingChannelTask::Pending,
+    ));
+    let runtime = crate::state::get_or_create_tokio_runtime()?;
+    let (future, remote_handle) = async move {
+        event::publish(&EventInternal::FundingChannelNotification(
+            FundingChannelTask::Pending,
+        ));
```
luckysori, line 30: ❓ Why do you need to send `Pending` so many times?

mobile/native/src/unfunded_orders.rs
```diff
@@ -0,0 +1,123 @@
...
+            FundingChannelTask::Funded,
+        ));
+
+        if let Err(error) = bdk_node.sync_on_chain_wallet().await {
+            tracing::error!("Failed at syncing wallet {error:?}")
+        }
+
+        let balance = bdk_node.get_on_chain_balance();
+        tracing::debug!(balance = balance.to_string(), "Wallet synced");
+
```
luckysori, line 82: 🔧 The second log message can contradict the first one.

Should we just loop until the on-chain balance is enough? We confirmed that the UTXOs are there, but we can also confirm that the balance is up to date.

luckysori: Alternatively, we could try to do it without the manual sync. It might not be necessary if we have broken out of the loop above.

mobile/native/src/unfunded_orders.rs
```diff
@@ -0,0 +1,123 @@
...
+                tracing::error!("Failed at submitting order {error:?}");
+                event::publish(&EventInternal::FundingChannelNotification(
+                    FundingChannelTask::Failed("Failed at posting the order".to_string()),
+                ));
+            }
+        }
+    }
+    .remote_handle();
+
+    // We need to store the handler which will drop any old handler if present.
```
luckysori, line 116:
```suggestion
    // We need to store the handle which will drop any old handle if present.
```

luckysori (approved):
LGTM! I did not delve into the Flutter code, but the screens look like a good starting point!

mobile/native/src/event/api.rs
bonomat: Sorry to hear ... this literally happened to me as well :/

mobile/native/src/unfunded_orders.rs
bonomat: good question 🙈

mobile/native/src/unfunded_orders.rs
bonomat: I figured syncing the wallet is more expensive than looking up just a single address. But you are right, this would be an alternative. 

> Alternatively, we could try to do it without the manual sync. It might not be necessary if we have broken out of the loop above.

Unfortunately we need to sync the wallet once we found the address holds enough funds as otherwise we might fail later on because the wallet thinks it doesn't have enough funds.

luckysori: > I figured syncing the wallet is more expensive than looking up just a single address. But you are right, this would be an alternative.

Right. My first suggestion was to sync in a loop _after_ the first loop has exited successfully. In theory, you only need to sync once, but there is an error path that we are currently just logging. I think if we fail to sync we want to keep trying until the UTXOs that we think should be there appear in the wallet.

> Unfortunately we need to sync the wallet once we found the address holds enough funds as otherwise we might fail later on because the wallet thinks it doesn't have enough funds.

Okay, so one successful sync is necessary.

PR #2550, opened 2024-05-16
fix: Remove multiple order update to open

holzeis:
Obviously the second set order state to `Open` should have been removed. 🤦


bonomat (approved):
LGTM

PR #2553, opened 2024-05-21
fix: small layout bug

bonomat:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>

### DIFF
--- a/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
+++ b/mobile/lib/features/trade/trade_bottom_sheet_tab.dart
@@ -349,12 +349,12 @@ class _TradeBottomSheetTabState extends State<TradeBottomSheetTab>
                       }
                     }),
                 const SizedBox(width: 55),
+                const Spacer(),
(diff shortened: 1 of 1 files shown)


holzeis:
🤔 is this really a bug? Is the label now jumping with the amount? Or is it already doing that?

bonomat:
Don't you see the big space between `Fee:` and `17,880`? Or are you saying this was by design, because I find it quite weird 😅

image

holzeis:
> Don't you see the big space between `Fee:` and `17,880`? Or are you saying this was by design, because I find it quite weird 😅
> 
> 
> 
> image
> 
> 

I think it's by design so that the label is not jumping when the fee amount is changing.

bonomat:
> I think it's by design so that the label is not jumping when the fee amount is changing.

Lol, and here I sit, thinking this is ugly as a hat full of bums.

PR #2557, opened 2024-05-21
fix: Ignore missing order id when rejecting a rollover offer

### LINKED ISSUE #2554: Rejecting a `RolloverOffer` fails due to missing order_id
When rejecting a dlc channel in the state `RenewOffered` we are missing a handling for a Rollover scenario. In this case we do not have an order and we fail on setting the order to failed, thus not sending the reject message.

https://github.com/get10101/10101/blob/567266a215cf6c3da42c8ed16156afa3900bef3e/mobile/native/src/dlc/node.rs#L657-L676

holzeis:
In case of a rollover we do not have an order id or and order in filling. As a result this logic failed when trying to reject a rollover offer.

The solution is still hacky as with the correct context information we shouldn't even try to set the order to failed, but that context is missing when we are reconnecting in the `on_connect` function and finding a dlc channel in state `RenewOffered`.

fixes #2554

### DIFF
--- a/coordinator/src/node.rs
+++ b/coordinator/src/node.rs
@@ -481,9 +481,14 @@ impl Node {
                                     "DLC Channel settle offer has been rejected. Setting position to back to open."
                                 );
 
-                        db::positions::Position::update_closing_position(
+                        db::positions::Position::update_position_state(
                             &mut connection,
                             node_id.to_string(),
+                            vec![
+                                // the closing price doesn't matter here.
+                                PositionState::Closing { closing_price: 0.0 },
+                                PositionState::Rollover,
+                            ],
                             PositionState::Open,
                         )?;
                     }
--- a/mobile/native/src/trade/order/handler.rs
+++ b/mobile/native/src/trade/order/handler.rs
@@ -322,20 +322,28 @@ pub(crate) fn order_failed(
     let order = match order_id {
         None => get_order_in_filling(),
         Some(order_id) => db::get_order(order_id),
-    }?
-    .with_context(|| format!("Could not find order. order_id = {order_id:?}"))
+    }
     .inspect_err(|e| {
+        // it doesn't matter that we send here an async trade failed even though we do not exactly
+        // know what kind of background task failed, because the error screen looks always the same.
         event::publish(&EventInternal::BackgroundNotification(
             BackgroundTask::AsyncTrade(TaskStatus::Failed(format!("{e:#}"))),
         ))
     })?;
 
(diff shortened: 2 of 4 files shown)


mobile/native/src/trade/order/handler.rs
```diff
@@ -322,20 +322,31 @@ pub(crate) fn order_failed(
...
 
     let task_status = TaskStatus::Failed(format!("{error:#}"));
-    let task = match order.reason {
-        OrderReason::Manual => BackgroundTask::AsyncTrade(task_status),
-        OrderReason::Expired => BackgroundTask::Expire(task_status),
-        OrderReason::CoordinatorLiquidated | OrderReason::TraderLiquidated => {
-            BackgroundTask::Expire(task_status)
+    let task = match order {
+        Some(order) => {
+            let task_status = TaskStatus::Failed(format!("{error:#}"));
```
bonomat, line 337: This is redundant as we have `task_status` already above.
```suggestion
```

coordinator/src/node.rs
```diff
@@ -481,9 +481,14 @@ impl Node {
                                     "DLC Channel settle offer has been rejected. Setting position to back to open."
                                 );
 
-                        db::positions::Position::update_closing_position(
```
bonomat, line 484: `update_closing_position` is now unused and can be removed.

coordinator/src/node.rs
```diff
@@ -481,9 +481,14 @@ impl Node {
...
+                        db::positions::Position::update_position_state(
                             &mut connection,
                             node_id.to_string(),
+                            vec![
+                                // the closing price doesn't matter here.
+                                PositionState::Closing { closing_price: 0.0 },
```
bonomat, line 489: Can you explain this please: Shouldn't this only deal with positions which are currently in `Rollover`?

bonomat:
I'm not sure I understand this fix.

coordinator/src/node.rs
holzeis: We are missing a bit of context here to be exact. I guess a typed reject would help here.

Anyways this part is processed after the `Reject` has been executed.

E.g. if we are processing a reject of a `SettleOffer` the channel state will have reverted to `Established`, but the same is true if a `RolloverOffer` or a `RenewOffer` gets rejected.

Since we don't know we simply update any position of that user in `Closing` (SettleOffer) or `Rollover` (RolloverOffer)` back to `Open`

bonomat: Got it thank you for the explanation :)

bonomat (approved):

PR #2560, opened 2024-05-22
feat: Onboarding through a hodl invoice

holzeis:
Introduces a lightning onboarding by leveraging an external lnd instance to manage a hodl invoice flow.

TODO:

- [ ] Submit pre image when posting the order after the lightning funding event has been received.
- [ ] Settle the invoice with the pre image received from the app.
- [ ] Initiate a one sided channel opening with the app.


crates/lnd-bridge/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rust:1.77 as builder
```
bonomat, line 1: Why run and build it in a container?
This will slow down our CI builds significantly.

crates/xxi-node/src/commons/mod.rs
```diff
@@ -77,6 +79,13 @@ impl User {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct HodlInvoiceParams {
+    pub trader_pubkey: PublicKey,
+    pub amt_sats: u64,
+    pub r_hash: String,
```
bonomat, line 86: why `r_hash` and not simply `hash`?

holzeis: To be aligned with the lnd api model.

crates/lnd-bridge/Dockerfile
holzeis: This was a left over from when I used a builder image. This can be removed.

bonomat:
https://github.com/get10101/10101/assets/224613/0195bc3b-4d14-4bca-8d73-dfd42ca4c7fa

🚀

crates/lnd-bridge/Dockerfile
bonomat: Not sure if you understood me: building and running the mock in a container will slow down CI (provided we write tests for this).

holzeis: oh got ya. I thought you were referring to the `as builder` part. don't you think the CI will cache the build image as well? if not we could manually push an image and use this instead in the docker compose. It shouldn't change to often anyways.

holzeis:
@bonomat I continued working on this and mainly tested the feature.

- Added a way to cancel running watch tasks
- Fixed the hash encoding (needs to be url safe as otherwise it could be interpreted as path)
- Moved the pre-image argument also to the channel opening params (combined it with the flag)
- Reworked the channel opening params addressing the todo to introduce an internal model. This model can now hold the external funding amount. Based on that amount I've added a check when opening the position to reject if the amount is not sufficient. I think this is crucial as everything else would be a huge vulnerability.

However, we need to introduce a minimum trader reserve for that as otherwise price movements could also result into a rejected order. (we need margin orders). I tried to add this to the ui, but failed, would be great if you could take over.

holzeis:
@luckysori probably best to review the whole PR.

coordinator/migrations/2024-05-23-000406_hodl_invoice_table/up.sql
```diff
@@ -0,0 +1,10 @@
+create table if not exists hodl_invoices
+(
+    id               SERIAL PRIMARY KEY       NOT NULL,
+    trader_pubkey    TEXT                     NOT NULL REFERENCES users (pubkey),
+    hash             TEXT                     NOT NULL,
+    amount_sats      BIGINT                   NOT NULL,
+    pre_image        TEXT,
+    timestamp        timestamp WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
holzeis, line 8: nit: if we have a updated at this would make more sense to be named `created_at`.

coordinator/src/trade/mod.rs
```diff
@@ -372,24 +362,65 @@ impl TradeExecutor {
...
+                        Ordering::Less => external_funding_sat - margin_trader - order_matching_fee,
+                        Ordering::Equal => 0,
+                        // TODO(holzeis): Introduce margin orders to directly use the
+                        // external_funding_sats for the position instead of failing here. We need
+                        // to do this though as a malicious actor could otherwise drain us.
+                        //
+                        // Note, we add a min trader reserve to the external funding to ensure that
+                        // minor price movements are covered.
+                        Ordering::Greater => bail!("Not enough external funds to open position."),
+                    };
```
luckysori, line 394: 🔧 It would be simpler to use 

```rust
external_funding_sat.checked_sub(margin_trader + order_matching).context("Not enough external funds to open position")?
```

luckysori (commented):
I reviewed d81a770e22623f3cc9169e159327f1eed2eb0fdd and the logic seems sound.

If the external funding is insufficient, I think we should fail the whole thing earlier though. It seems to me like we have all the information available as soon as we get the match from the orderbook in `process_new_market_order`, so we don't have to wait until we call `open_dlc_channel`.

holzeis:
> I reviewed [d81a770](https://github.com/get10101/10101/commit/d81a770e22623f3cc9169e159327f1eed2eb0fdd) and the logic seems sound.
> 
> If the external funding is insufficient, I think we should fail the whole thing earlier though. It seems to me like we have all the information available as soon as we get the match from the orderbook in `process_new_market_order`, so we don't have to wait until we call `open_dlc_channel`.

Thanks for you feedback, I've adapted the code, moved the check early and also now subtract the funding tx reserve and the channel fee reserve from the trader reserve.

I think this is now ready to go, but I will wait for your final review and approval.

coordinator/src/trade/mod.rs
```diff
@@ -372,24 +392,43 @@ impl TradeExecutor {
...
-                    + margin_trader
-                    + collateral_reserve_trader
-                    // If the trader doesn't bring their own UTXOs, including the order matching fee
-                    // is not strictly necessary, but it's simpler to do so.
-                    + order_matching_fee,
-                0,
-                dlc::FeeConfig::AllOffer,
-            ),
-        };
+        let (offer_collateral, accept_collateral, fee_config, collateral_reserve_trader) =
```
luckysori, line 395: 🔧 I think the last element of the tuple is the same for both branches, so you can omit it.

coordinator/src/trade/mod.rs
```diff
@@ -231,13 +234,44 @@ impl TradeExecutor {
...
                 .context("Failed to open DLC channel")?;
             }
-            TradeAction::OpenSingleFundedChannel => {
+            TradeAction::OpenSingleFundedChannel { external_funding } => {
                 let collateral_reserve_coordinator = params
                     .coordinator_reserve
                     .context("Missing coordinator collateral reserve")?;
-                let collateral_reserve_trader = params
-                    .trader_reserve
-                    .context("Missing trader collateral reserve")?;
```
luckysori, line 240: 🔧 Shouldn't we calculate the `trader_reserve` before we get here so that we can read it from the `params` like we do with the `coordinator_reserve`?

crates/tests-e2e/src/lnd_mock.rs
```diff
@@ -0,0 +1,50 @@
+use anyhow::Context;
+use anyhow::Result;
+use reqwest::Client;
+use serde::Serialize;
+
+/// A wrapper over the coordinator HTTP API.
```
luckysori, line 6: 🔧 Copy-paste error.

crates/xxi-node/src/commons/message.rs
```diff
@@ -41,6 +41,11 @@ pub enum Message {
         order_id: Uuid,
         error: TradingError,
     },
+    PaymentReceived {
```
luckysori, line 44: 🙃 Maybe we should talk about Lightning payment here (and in other places) to disambiguate.

mobile/native/src/backup.rs
```diff
@@ -67,9 +67,6 @@ impl Subscriber for DBBackupSubscriber {
 
     fn events(&self) -> Vec<EventType> {
         vec![
-            EventType::PaymentClaimed,
-            EventType::PaymentSent,
-            EventType::PaymentFailed,
```
luckysori, line 72: 👀

mobile/native/src/lib.rs
```diff
@@ -36,4 +36,6 @@ pub use report_error::report_error_to_coordinator;
     unused_qualifications
 )]
 mod bridge_generated;
+mod hodl_invoice;
 mod position;
+mod unfunded_channel_opening_order;
```
luckysori, line 41: 🙃 Hmmm, we should align the terminology between coordinator and app.

mobile/native/src/unfunded_channel_opening_order.rs
```diff
@@ -0,0 +1,126 @@
...
+    trader_reserve: u64,
+    estimated_margin: u64,
+    order_matching_fee: u64,
+) -> anyhow::Result<ExternalFunding, Error> {
+    let node = get_node();
+    let bitcoin_address = node.inner.get_new_address()?;
+
+    let fees = Amount::from_sat(order_matching_fee)
+        + crate::dlc::estimated_fee_reserve()?
+        + (crate::dlc::estimated_funding_tx_fee()?);
```
luckysori, line 39: ❓ Superfluous brackets?

mobile/native/src/watcher.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .sum::<u64>();
+
+                if total_unspent_amount_received >= funding_amount.to_sat() {
+                    tracing::info!(
+                        amount = total_unspent_amount_received.to_string(),
+                        address = funding_address.to_string(),
+                        "Address has been funded enough"
+                    );
+
+                    break 'watch_address;
```
luckysori, line 71: ❓ Do we need the labels?

luckysori (approved):
LGTM, I think this should work!

coordinator/src/lib.rs
```diff
@@ -106,3 +107,10 @@ pub fn decimal_from_f32(float: f32) -> Decimal {
 pub fn f32_from_decimal(decimal: Decimal) -> f32 {
     decimal.to_f32().expect("Decimal to fit into f32")
 }
+
+#[derive(Clone, Copy, Debug)]
+pub struct ChannelOpeningParams {
+    pub trader_reserve: Amount,
+    pub coordinator_reserve: Amount,
+    pub external_funding: Option<Amount>,
```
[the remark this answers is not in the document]
holzeis: hmm.. I kind of like the option better because it is a bit more expressive than automatically implying that a zero amount is a `OpenSingleFundedChannel`

```
match params.external_funding {
    Some(external_funding) => {
        TradeAction::OpenSingleFundedChannel { external_funding }
    }
    None => TradeAction::OpenDlcChannel,
}
```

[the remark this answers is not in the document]
luckysori: But what does it mean to have a 0-valued `external_funding` then?

coordinator/src/orderbook/async_match.rs
```diff
@@ -119,6 +119,7 @@ async fn process_pending_match(
                 },
                 trader_reserve: channel_opening_params.map(|c| c.trader_reserve),
                 coordinator_reserve: channel_opening_params.map(|c| c.coordinator_reserve),
+                external_funding: None,
```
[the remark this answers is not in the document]
holzeis: good catch. Yes its possible, although unlikely. The user could close the app after we have matched the order, but before we have sent the dlc proposal.

coordinator/src/lib.rs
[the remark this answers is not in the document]
holzeis: Probably a bug. Any value below 250k is probably a bug.

coordinator/src/routes.rs
```diff
@@ -602,3 +609,61 @@ async fn post_error(
...
+    .map_err(|error| {
+        AppError::InternalServerError(format!("Could not process hodl invoice {error:?}"))
+    })?;
+
+    // watch for the created hodl invoice
+    invoice::spawn_invoice_watch(
+        state.tx_orderbook_feed.clone(),
+        state.lnd_bridge.clone(),
+        invoice_params,
+    );
```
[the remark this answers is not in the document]
holzeis: I haven't considered it. But, besides having less logic in the api (which is a good thing) I don't see a lot of advantages. The actor would still have to spawn a new task on every hodl invoice to monitor, wouldn't he?

coordinator/src/trade/mod.rs
holzeis: yes, thank you that was a left over from a previous change.

crates/xxi-node/src/commons/message.rs
holzeis: I renamed it to `LnPaymentReceived`

mobile/native/src/unfunded_channel_opening_order.rs
holzeis: yes

mobile/native/src/watcher.rs
holzeis: No, I've removed it.

mobile/native/src/lib.rs
holzeis: what do you mean? the coordinator doesn't know of an unfunded channel opening order.

coordinator/src/trade/mod.rs
holzeis: The reason why I am recalculating the trader_reserve is because we can't trust the client on these params, so if we are going the `SingleFundedChannelOpening` path we only, use the funds received via lightning and ignore the params.

Note, I think its also not a good idea to blindly trust the client about the coordinators reserve. We should probably at least verify if the values make sense. I'll leave a FIXME.

mobile/native/src/lib.rs
luckysori: Yeah, I think you are right. I was confused at first because this is implicitly about a single-funded channel, which is a different name.

Maybe it would be clearer with `unfunded_wallet_channel_opening_order`? The name would be annoyingly long though.

coordinator/src/routes.rs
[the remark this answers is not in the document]
luckysori: Yeah, I think you are right. We would still need a task per invoice.

coordinator/src/trade/mod.rs
luckysori: What I mean is that I would like to recalculate the trader reserve before calling `TradeExecutor::execute`. Then we don't need the `external_funding` to be passed any deeper.

> Note, I think its also not a good idea to blindly trust the client about the coordinators reserve. We should probably at least verify if the values make sense. I'll leave a FIXME.

I agree with that. We could end up locking up more or less funds than desired.

holzeis: > What I mean is that I would like to recalculate the trader reserve before calling `TradeExecutor::execute`. Then we don't need the `external_funding` to be passed any deeper.

Hmm.. that would make the api a bit less convenient as the caller would need to implement the logic for handling the channel opening params before calling execute.

That would mean we'd have to duplicate the logic for "async match". Do you think it's worth it?

coordinator/src/trade/mod.rs
```diff
@@ -378,17 +412,19 @@ impl TradeExecutor {
...
+                        + collateral_reserve_coordinator.to_sat()
+                        + margin_trader
+                        + collateral_reserve_trader
+                        // If the trader doesn't bring their own UTXOs, including the order matching fee
+                        // is not strictly necessary, but it's simpler to do so.
+                        + order_matching_fee,
+                    0,
+                    dlc::FeeConfig::AllOffer,
+                )
+            }
```
luckysori, line 427: 🙃 I think nothing actually changed here.

coordinator/src/trade/mod.rs
luckysori: I guess it's debatable, so none of this is a deal breaker at all.

What I dislike is leaking the knowledge of the external funding any further than necessary, and the fact that the `trader_reserve` field is ignored, when it could be used to carry this information from the outside.

And I think we could avoid code duplication by reusing a helper function.

luckysori (approved):

coordinator/src/trade/mod.rs
holzeis: jap only the brakets have been added 🙈

coordinator/src/trade/mod.rs
holzeis: Yeah i think you are right.. I opted to leave this now, since I wanted to go forward with this PR. I think we can address it the next time we touch this part. (which will be probably be very soon :))

PR #2562, opened 2024-05-24
fix: return only wallet unspent utxos

bonomat:
Before we were printing all known utxos, even not from our own wallet.


luckysori (approved):
Thanks!

holzeis (approved):

PR #2564, opened 2024-05-24
fix: Unreserve utxos on wallet sync

bonomat:
on each blockchain sync we clear the lcoked utxos. This is fine, because we are syncing with the blockchain and are learning about all acutally spent utxos.

### DIFF
--- a/crates/xxi-node/src/node/wallet.rs
+++ b/crates/xxi-node/src/node/wallet.rs
@@ -114,6 +114,10 @@ impl<D: BdkStorage, S: TenTenOneStorage, N: Storage + Send + Sync + 'static> Nod
         .await
         .expect("task to complete")?;
 
+        // On sync, we unlock our locked utxos, because we are syncing with the blockchain we will
+        // find already spent utxos and release those which were locked unnecessarily.
+        self.wallet.locked_utxos.lock().clear();
+
         Ok(())
     }
 


crates/xxi-node/src/node/wallet.rs
```diff
@@ -71,6 +71,9 @@ impl<D: BdkStorage, S: TenTenOneStorage, N: Storage + Send + Sync + 'static> Nod
     /// Sync the state of the on-chain wallet against the blockchain.
     pub async fn sync_on_chain_wallet(&self) -> Result<()> {
         let client = &self.blockchain.esplora_client_async;
+        // On sync, we unlock our locked utxos, because we are syncing with the blockchain we will
+        // find already spent utxos and release those which were locked unnecessarily.
+        self.wallet.locked_utxos.lock().clear();
```
luckysori, line 76: 🔧 I think it would make more sense to move this to end of this function. The sync is fast, but it's safer to unlock these once the wallet state has been updated.

bonomat:
@luckysori : feel free to take it over. I should work on something else 😬

crates/xxi-node/src/node/wallet.rs
holzeis: fixed it.

holzeis (approved):

PR #2570, opened 2024-05-27
feat: Sync wallet when contract is not confirmed yet.

holzeis:
This is covering an edge case where the user opens a dlc channel and wants to close the position within the sync window of the coordinator.

It allows to instantly close the position without having to manually sync the wallet after the dlc channel has been created.

---

It's probably not the most important feature for prod, but it makes testing a bit easier... Happy to close this PR if the reviewer thinks that this is an unnecessary overhead.

### DIFF
--- a/coordinator/src/trade/mod.rs
+++ b/coordinator/src/trade/mod.rs
@@ -804,8 +804,13 @@ impl TradeExecutor {
         trade_params: &TradeParams,
         resize_action: ResizeAction,
     ) -> Result<()> {
-        if !self.node.inner.is_dlc_channel_confirmed(&dlc_channel_id)? {
-            bail!("Underlying DLC channel not yet confirmed");
+        if !self
+            .node
+            .inner
+            .check_if_signed_channel_is_confirmed(position.trader)
+            .await?
+        {
+            bail!("Underlying DLC channel not yet confirmed.");
         }
 
         let peer_id = trade_params.pubkey;
@@ -1045,8 +1050,13 @@ impl TradeExecutor {
         trade_params: &TradeParams,
         channel_id: DlcChannelId,
     ) -> Result<()> {
-        if !self.node.inner.is_dlc_channel_confirmed(&channel_id)? {
-            bail!("Underlying DLC channel not yet confirmed");
+        if !self
+            .node
+            .inner
+            .check_if_signed_channel_is_confirmed(position.trader)
+            .await?
+        {
+            bail!("Underlying DLC channel not yet confirmed.");
         }
 
         let closing_price = trade_params.average_execution_price();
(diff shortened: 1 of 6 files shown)


coordinator/src/trade/mod.rs
```diff
@@ -672,8 +672,13 @@ impl TradeExecutor {
...
     ) -> Result<()> {
-        if !self.node.inner.is_dlc_channel_confirmed(&dlc_channel_id)? {
-            bail!("Underlying DLC channel not yet confirmed");
+        if !self
+            .node
+            .inner
+            .is_dlc_channel_confirmed(&dlc_channel_id)
+            .await?
+        {
+            bail!("Did not see funding tx in mempool yet.");
```
luckysori, line 681: 🔧 I don't like this change because it assumes that the DLC channel will always be confirmed as soon as the transaction hits mempool, but this may not always be the case.

I think the original error message is fine, but if you want better error messages you could adapt the `is_dlc_channel_confirmed` API to return more information to build a better error message here.

luckysori (commented):
I'm not a fan of this :(

I think it makes sense to sync eagerly in some cases. But when I call `is_dlc_channel_confirmed` I don't always want the sync to trigger e.g. in certain tests.

I would be okay with defining a _separate_ API with similar behaviour, to be called in certain spots.

coordinator/src/trade/mod.rs
holzeis: > 🔧 I don't like this change because it assumes that the DLC channel will always be confirmed as soon as the transaction hits mempool, but this may not always be the case.

I thought this was the case. When do we have to wait for a confirmation?

holzeis:
> I think it makes sense to sync eagerly in some cases. But when I call is_dlc_channel_confirmed I don't always want the sync to trigger e.g. in certain tests.

hmm.. Can you explain in what tests you would not want to eagerly sync if the dlc channel is signed, but the contract is not confirmed? I thought this could only be the case if we are missing a sync, because use 0-conf everywhere.

> I would be okay with defining a separate API with similar behaviour, to be called in certain spots.

I can definitely do that, but I guess I will replace all calls to the current `is_dlc_channel_confirmed` function, so maybe a rename would be enough?

coordinator/src/trade/mod.rs
luckysori: What I mean is that the error message now reflects an implementation detail: that the number of required confirmations is always 0. And the implementation detail may change in the future, but this error message is independent of that and could easily get stale.

luckysori (approved):
I do dislike the way we are sneakily triggering the on-chain sync from a method called `is_dlc_channel_confirmed`, so I would prefer something a more explicit name.

But this is definitely useful in a few spots, so I don't want to block the PR.

coordinator/src/trade/mod.rs
holzeis: Fair enough! I will adapt the message.

holzeis:
I will see to rename the function to make it clearer.

PR #2572, opened 2024-05-28
fix: Prevent multiple next button clicks

holzeis:
The lnd node is running separately on the test vm, which can lead to a delay when requesting a new invoice.

This patch disables the button and indicates to the user that the request is being processed.

https://github.com/get10101/10101/assets/382048/f62af30d-f12e-4b93-a9cf-1d43c03547b1


luckysori (approved):
Thanks!

PR #2573, opened 2024-05-28
feat: Handle error when requesting a hodl invoice

holzeis:
If we can't receive a hodl invoice the app gracefully handles that scenario and shows an according message instead of the invoice.

https://github.com/get10101/10101/assets/382048/0998f5d2-a3a1-4782-9716-bc84544edb8c


bonomat (approved):
Mhm, it only says now: coming soon if we couldn't get the invoice.
But I guess this is good enough.

holzeis:
> Mhm, it only says now: coming soon if we couldn't get the invoice.
> 
> But I guess this is good enough.

I wanted to do something with else but I wasn't able to design something nice so I reused the coming soon image. 🙈

PR #2574, opened 2024-05-28
fix: Watch for specific r_hash

holzeis:
This PR fixes two issues.

1. Any paid invoice that has been created by the coordinator will produce a `LnPaymentReceived` event. Before that change we would have simply assumed that this was the payment we've been waiting for. But it could have happened that the users went back and forth multiple times and thus created multiple invoices. So if an old invoice would have been paid, we would have triggered the order creation which would have failed, because the shared pre-image would have been incorrect. We do now verify that the received r_hash is equal to the watched r_hash.

2. On every creation of a hodl invoice we subscribed the `InvoiceWatcher` for `LnPaymentReceived` events. These subscribers lived forever and never got cancelled, producing various error messages since the receiver has already been gone. The invoice watcher is now subscribed at the app startup and the broadcast::Sender is stored in the apps state - so that only a single Subscriber can do the job.

### DIFF
--- a/crates/lnd-bridge/examples/mock.rs
+++ b/crates/lnd-bridge/examples/mock.rs
@@ -5,6 +5,7 @@ use axum::extract::ws::WebSocketUpgrade;
 use axum::extract::Extension;
 use axum::extract::Json;
 use axum::extract::Query;
+use axum::extract::State;
 use axum::http::HeaderMap;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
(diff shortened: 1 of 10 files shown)


crates/lnd-bridge/examples/mock.rs
```diff
@@ -169,15 +183,18 @@ async fn cancel_invoice(
...
+) -> impl IntoResponse {
     let message = serde_json::to_string(&InvoiceResult {
         result: Invoice {
             memo: "".to_string(),
             expiry: 0,
             amt_paid_sat: 0,
             state: InvoiceState::Accepted,
             payment_request: "".to_string(),
-            r_hash: "".to_string(),
+            r_hash: state.read().expect("").to_string(),
```
luckysori, line 197: 🙃 Just allow unwrap in this crate or in this file if you don't want to deal with `expect`.

luckysori (approved):
Nice, LGTM!

PR #2576, opened 2024-05-29
feat: store some metrics in db

bonomat:
I thought this could come handy.
For start, we are storing on-chain balance in the DB every hour.


luckysori (approved):
LGTM.

PR #2578, opened 2024-05-29
chore: Make rotation direction of rotating icon configurable

holzeis:
That was killing me 🙈 

|	**before**	|	**after**	|
|:---:|:---:|
|	wrong-rotation	|	reversed-rotation	|


luckysori (approved):
Hahaha, I didn't notice, but thanks!

PR #2585, opened 2024-05-30
feat: Only settle invoice if proposal was successful

holzeis:


coordinator/src/trade/mod.rs
```diff
@@ -173,6 +207,66 @@ impl TradeExecutor {
...
+        })
+        .await??
+        .context("Missing pre_image")?;
+
+        let pre_image = PreImage::from_url_safe_encoded_pre_image(&pre_image)?;
+
+        self.node
+            .lnd_bridge
+            .settle_invoice(pre_image.get_base64_encoded_pre_image())
+            .await?;
```
luckysori, line 230: 🔧 I guess we would like to update the DB only if we were able to settle the invoice with the LND bridge. Could we use a transaction here?

coordinator/src/trade/mod.rs
```diff
@@ -173,6 +207,66 @@ impl TradeExecutor {
...
+            spawn_blocking({
+                let pool = self.node.pool.clone();
+                move || {
+                    let mut conn = pool.get()?;
+                    db::last_outbound_dlc_message::delete(&mut conn, &trader)?;
+
+                    anyhow::Ok(())
+                }
+            })
+            .await??;
```
luckysori, line 264: 🔧 This is a bit scary. Imagine failing to delete the message or deleting the wrong message.

Could we instead insert the channel-opening offer only _after_ we settle the invoice?

luckysori (approved):

coordinator/src/trade/mod.rs
holzeis: It shouldn't matter as the error handling will set the hodl_invoice to `Failed`.

coordinator/src/trade/mod.rs
holzeis: I don't think that's easily possible.. you are right that we could end up deleting an unrelated last outbound dlc message. The consequences though should be insignificant, since this is only used for recovery reasons.

holzeis: Also note, we are only doing this when the dlc channel is offered. So any `last_outbound_dlc_message` would anyways be useless.

PR #2588, opened 2024-05-30
fix: Do not create dlc protocol entry for force closed channels

holzeis:
Force closing a channel should not get recorded as regular collab close dlc protocol.

This lead to an issue where the dlc channel did not get updated properly because the contract was missing.


bonomat (approved):
Makes sense.

PR #2590, opened 2024-05-31
feat: Cancel hodl invoice if dlc proposal fails

### LINKED ISSUE #2586: Cancel hodl invoice if dlc proposal failed
If the dlc channel proposal fails we do not settle the invoice. Additionally we could cancel the invoice to fail the payment backwards.

### LINKED ISSUE #2589: Set timed out hodl invoices to canceled
We generate a hodl invoice everytime the user gets to the fund channel screen. This invoice might not get paid an time out.

After the timeout the invoice will be automatically canceled and we get the canceled event.

At the moment that should set the invoice to failed, which is also not working because the hash is received in the wrong encoding.

- [x] Properly map the r_hash value to the encoding we have in the database
- [x] Set the invoice state to `Canceled` if it timed out.

https://github.com/get10101/10101/blob/main/coordinator/src/node/invoice.rs

holzeis:
fixes #2586 
fixes #2589

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2569,7 +2569,6 @@ version = "2.4.5"
 dependencies = [
  "aes-gcm-siv",
  "anyhow",
- "base64 0.21.7",
  "bdk",
  "bdk_file_store",
  "bip21",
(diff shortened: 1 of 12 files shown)


bonomat (approved):
LGTM

PR #2591, opened 2024-05-31
chore: Set hodl invoice to `Settled` on lnd event

holzeis:
It's better to update our state only if the same state has been applied to lnd.

### DIFF
--- a/coordinator/src/db/hodl_invoice.rs
+++ b/coordinator/src/db/hodl_invoice.rs
@@ -69,6 +69,19 @@ pub fn get_r_hash_by_order_id(conn: &mut PgConnection, order_id: Uuid) -> QueryR
         .get_result(conn)
 }
 
+/// Returns the pre image of the hodl invoice associated with the order id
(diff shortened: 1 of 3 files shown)


coordinator/src/db/hodl_invoice.rs
```diff
@@ -69,6 +69,16 @@ pub fn get_r_hash_by_order_id(conn: &mut PgConnection, order_id: Uuid) -> QueryR
         .get_result(conn)
 }
 
+pub fn get_pre_image_by_order_id(
```
bonomat, line 72: Uh, the API confused me. It returns `QueryResult<Option<String>>` which I would expect it means that if no pre-image was found, it returns `None`, however, the `get_result` will return an `Err(NotFound)` if it no element was found (by `order_id`) but it will return `None` if the hodl_invoice's pre-image was null...

I had to think unnecessarily long about this... maybe a rust doc could help? :D

bonomat (approved):
LGTM

PR #2594, opened 2024-06-03
fix: don't change app bar colour when scrolling

### LINKED ISSUE #2217: App bar changes color when scrolling down
IMG_4715

Restioson:
Fixes #2217


holzeis (approved):
Nice 👍

PR #2595, opened 2024-06-03
fix: Stop welcome screen overflowing

### LINKED ISSUE #2405: Rendering overflow on welcome screen
image

Restioson:
Fixes #2405

### DIFF
--- a/mobile/lib/common/scrollable_safe_area.dart
+++ b/mobile/lib/common/scrollable_safe_area.dart
@@ -3,14 +3,17 @@ import 'package:flutter/cupertino.dart';
 class ScrollableSafeArea extends StatelessWidget {
   const ScrollableSafeArea({
     required this.child,
+    this.bottom = true,
     Key? key,
   }) : super(key: key);
 
+  final bool bottom;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


mobile/lib/common/scrollable_safe_area.dart
```diff
@@ -3,14 +3,17 @@ import 'package:flutter/cupertino.dart';
...
     Key? key,
   }) : super(key: key);
 
+  final bool bottom;
   final Widget child;
 
   @override
   Widget build(BuildContext context) {
     return SafeArea(
+        bottom: bottom,
```
holzeis, line 16: ❓ what is this doing?

Restioson: This is to pass it through from `this.bottom`. It's set in the welcome screen. If it's not needed there though it could be removed

Restioson: > Do we need this if the area is anyways scrollable?

I simply kept it as it was in the original and I was not sure what behaviour or display may change with removing it. I'm happy to remove it if you'd like.

FWIW the page did not fit on my phone as-is so it's probably good that it's scrollable now :sweat_smile:

PR #2596, opened 2024-06-04
feat: regtest is dead, long live signet

bonomat:
Note: the hidden faucet button does not work, you will need to use the official mutinynet faucet for this https://faucet.mutinynet.com/

Before this gets merged, I'll need to update the environment settings on this repository so that the latest build binaries are being deployed to the new machine.


justfile
```diff
@@ -8,11 +8,12 @@ maker_log_file := "$PWD/data/maker/regtest.log"
 pubspec := "$PWD/mobile/pubspec.yaml"
 
 # public regtest constants
-public_regtest_coordinator := "03507b924dae6595cfb78492489978127c5f1e3877848564de2015cd6d41375802@34.32.62.120:9045"
-public_regtest_electrs := "http://34.32.62.120:3000"
+public_regtest_coordinator := "021eddb3062fddef2c24f16f6de55fbd78ab985b74b41858719def46321d4be7d0@146.0.73.175:9045"
+public_regtest_electrs := "http://146.0.73.175:3003"
```
luckysori, line 12: 🔧 I guess we should rename these variables to use `signet` instead of `public_regtest`.

luckysori (approved):
The changes LGTM, but I guess there is a chance we missed changing something somewhere :D

justfile
bonomat: done

bonomat:
> The changes LGTM, but I guess there is a chance we missed changing something somewhere :D

there is only one way to find out 😬

PR #2598, opened 2024-06-04
feat: use mutinynet faucet

### LINKED ISSUE #2597: Implement signet faucet for signet app
After https://github.com/get10101/10101/pull/2596 lands we will have to use mutinynet faucet for funding our app (https://faucet.mutinynet.com/). The form looks very simple, so we should be able to easily bind this into our app.

bonomat:
Resolves #2597


holzeis (approved):
Nice, can you also do that for the lightning invoice 🙂

bonomat:
> Nice, can you also do that for the lightning invoice 🙂

Tracked here: https://github.com/get10101/10101/issues/2600

PR #2602, opened 2024-06-04
fix: build ios archive was missing

bonomat:
Looks like the [daily task ](https://github.com/get10101/10101/actions/workflows/testflight-for-review.yml) never worked because we did not build the ios archive 🙈 . 

Should be fixed now.


holzeis (approved):
LGTM

Issue #2430, opened 2024-04-16
The app shouldn't fail to start if a electrs is down


holzeis:
It happened on production that the bitcoin node stopped and thus electrs was also not available. This lead for at least one user to the app crashing.

bonomat:
Related to #2431

Restioson:
It seems, to me, that more than just electrs being down is required. My app has launched and I can interact with it despite not having electrs. Verifying this through these kinds of logs after running `docker stop electrs` and then starting the app
```
2024-06-03T14:44:05.288081485Z ERROR native::dlc: On-chain sync failed: Reqwest(reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(3000), path: "/scripthash/1ddc89617c14192a6fcda71f16e5275e2abe01362b7ee96bc0da82db8d7f6e0a/txs", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })) })
```

holzeis:
> Unless @holzeis was referring to some other loss of functionality rather than just crashing?

Two months ago the app was crashing. If it doesn't do this anymore the bug seems to be closed and we can close this ticket.

Restioson:
I've now got it to fail as such:
```
flutter: [INFO] TIME: 2024-06-04T07:53:17.475029Z Successfully stored the contact:  .
2024-06-04T07:53:17.476421442Z  INFO native::api: Creating a new seed in "/home/restioson/.local/share/finance.get10101.app/regtest/seed"
2024-06-04T07:53:17.476810904Z  INFO xxi_node::seed: No seed found. Generating new seed
flutter: [DEBUG] TIME: 2024-06-04T07:53:17.477887Z Registering user with  &
2024-06-04T07:53:17.50028326Z DEBUG native::trade::users: Registering user pubkey="02ce86347e39bc41fbbfff4a4973e4ae50fab42ed2fb0984b0e5657f41b4cb6edb" contact="" referral_code="" version="2.4.5"
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: FfiException(RESULT_ERROR, Failed to register beta program with coordinator

Caused by:
    0: error sending request for url (http://127.0.0.1:8000/api/users): error trying to connect: tcp connect error: Connection refused (os error 111)
[...]

PR #2603, opened 2024-06-04
fix: Make onboarding a bit more robust when coordinator is down

Restioson:
This PR adds two related commits which do the following:
1. Show an error to the user if the coordinator is down (or unavailable) during onboarding instead of just hanging
2. If the coordinator is down during onboarding, try to register them for the beta program the next time they launch the app instead of just leaving them unregistered.

This might address #2430 somewhat, but I'm currently unclear exactly what the issue with #2430 is.

### DIFF
--- a/mobile/lib/common/routes.dart
+++ b/mobile/lib/common/routes.dart
@@ -39,7 +39,7 @@ GoRouter createRoutes() {
           path: LoadingScreen.route,
           pageBuilder: (context, state) => NoTransitionPage<void>(
             child: LoadingScreen(
-              future: state.extra as Future<void>?,
+              task: state.extra as LoadingScreenTask?,
             ),
           ),
         ),
--- a/mobile/lib/features/welcome/loading_screen.dart
+++ b/mobile/lib/features/welcome/loading_screen.dart
@@ -1,13 +1,15 @@
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter_native_splash/flutter_native_splash.dart';
+import 'package:get_10101/common/global_keys.dart';
 import 'package:get_10101/common/scrollable_safe_area.dart';
 import 'package:get_10101/common/snack_bar.dart';
-import 'package:flutter_native_splash/flutter_native_splash.dart';
 import 'package:get_10101/backend.dart';
 import 'package:get_10101/features/welcome/error_screen.dart';
 import 'package:get_10101/features/welcome/onboarding.dart';
 import 'package:get_10101/features/trade/trade_screen.dart';
 import 'package:get_10101/features/wallet/wallet_screen.dart';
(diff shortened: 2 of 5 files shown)


holzeis:
I think we can't delay the user registration to the next start as the user might trade once the coordinator is back online. This will result in errors if the user is not registered.

Restioson:
> I think we can't delay the user registration to the next start as the user might trade once the coordinator is back online. This will result in errors if the user is not registered.

Is the user able to create a position whilst the coordinator is offline? Or is this in the case of the coordinator coming back whilst the user is still online?

Restioson:
Perhaps this could be addressed by checking if the user is registered every time before they open a position?

holzeis:
> Is the user able to create a position whilst the coordinator is offline? Or is this in the case of the coordinator coming back whilst the user is still online?

If the coordinator is coming back, while the user is online.

holzeis:
> Perhaps this could be addressed by checking if the user is registered every time before they open a position?

I wouldn't make it too complicated.. I think your solution of simply retrying to register, when the app starts the next time is fine.

holzeis:
@Restioson This feels a bit weird. I would go for the warning we already show when you open the app after you have registered. "Coordinator is not available" but show the app.

https://github.com/get10101/10101/assets/382048/0ef791b4-af9e-496e-926a-39defebedc19

Restioson:
@holzeis: I've gone for showing both warnings. Now, it will still show that you have failed to register for beta but also continue to wallet and then show that the coordinator is offline. I did not want to simply silence the failure to register for beta as I think the user should still know about it.

holzeis:
> @holzeis: I've gone for showing both warnings. Now, it will still show that you have failed to register for beta but also continue to wallet and then show that the coordinator is offline. I did not want to simply silence the failure to register for beta as I think the user should still know about it.

Can you share a video.. I would like to see how this looks like. There was also the issue of a hanging loading screen.

Restioson:
For future reference: @holzeis and I discussed that the additional warning about beta signup should be removed, since there is no observable effect of this failure (it's only internal). If this changes, like if we add an email notification or the like, we could consider reintroducing it as an extra line on the main 'coordinator down' warning.

Restioson:
I changed it to a snackbar just to make it consistent with the import seed which does the same thing when it's LoadingScreenTask fails. I did it in such a way that I can remove it very easily though, on purpose :smile:

holzeis:
> I changed it to a snackbar just to make it consistent with the import seed which does the same thing when it's LoadingScreenTask fails. I did it in such a way that I can remove it very easily though if you think it should go though @holzeis, on purpose 😄

Sounds reasonable to me, can you share a video?

Restioson:
[Kazam_screencast_00002.webm](https://github.com/get10101/10101/assets/6688948/a3ae7920-781c-4c04-970d-32a4010e39c0)

PR #2613, opened 2024-06-05
Display next funding rate below candlestick chart

luckysori:
Thanks to @holzeis for the inspiration!

image

### DIFF
--- a/mobile/lib/features/trade/trade_screen.dart
+++ b/mobile/lib/features/trade/trade_screen.dart
@@ -72,12 +72,6 @@ class TradeScreen extends StatelessWidget {
                       price: Usd.fromDouble(price ?? 0.0),
                     );
                   }),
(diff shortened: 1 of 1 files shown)


bonomat (commented):
Any reason why you didn't put it into the top bar where we have Bid/Ask?

luckysori:
> Any reason why you didn't put it into the top bar where we have Bid/Ask?

It doesn't really fit, even on a bigger phone screen, so it had to be displayed in two rows which didn't look perfect next to the bid and ask.

I don't mind if you propose something else. I only did this because @holzeis really didn't like the original approach. But I think it looks pretty good now.

bonomat:
Happy to leave it for now but I think it is a bit weird to have it in the middle of the screen. Also, I wouldn't have added the text but just the rate. 😬

PR #2615, opened 2024-06-06
Add `unique_end_date` constraint to `funding_rates` table

luckysori:
We only need one funding rate entry per `end_date`.

Without this change we were getting a new duplicate entry per maker restart.

### DIFF
--- a/coordinator/src/funding_fee/db/funding_rates.rs
+++ b/coordinator/src/funding_fee/db/funding_rates.rs
@@ -1,5 +1,4 @@
 use crate::schema::funding_rates;
-use anyhow::bail;
 use anyhow::Result;
 use diesel::prelude::*;
 use rust_decimal::prelude::FromPrimitive;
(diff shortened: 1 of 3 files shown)


holzeis:
What will happen if we get a duplicate entry. Is the error simply logged?

luckysori:
> What will happen if we get a duplicate entry. Is the error simply logged?

Good question.

An error would be logged, but it would also mean that any other rows in the batch would be ignored. I've updated the patch to fix this.

coordinator/src/funding_fee/db/funding_rates.rs
```diff
@@ -39,6 +39,8 @@ pub fn insert_funding_rates(
 
     let affected_rows = diesel::insert_into(funding_rates::table)
         .values(funding_rates)
+        .on_conflict(funding_rates::end_date)
+        .do_nothing()
         .execute(conn)?;
 
     if affected_rows == 0 {
```
holzeis, line 46: 🤔 Maybe we don't want this either?

luckysori: I agree, I've removed it now :D

holzeis (approved):

PR #2617, opened 2024-06-06
feat: higher leverages

bonomat:
A simple feature allowing us to configure higher leverages through the coordinator's settings.


holzeis (approved):
LGTM 👍

PR #2618, opened 2024-06-06
feat: allow to configure coordinator leverage

bonomat:
With this patch we introduce a new table: `coordinator_leverages`.
It contains two fields:
- trader_leverage
- coordinator_leverage

i.e. we can define a coordinator leverage for each possible trader_leverage. For simplcitiy reasons we only have round numbers in the database and no floating points, hence, when looking up a coordinator leverage for a trader leverage, we need to round. If no match is found, we fall back and take the max allowed leverage of the coordinator. This allows us to not have an entry from 1 - 100 but stop at e.g. 10 with a max coordinator leverage of 10. If a user then goes a 100x, we still pick leverage 10.

There is one downside of this patch: The coordinator's reserve now depends on the trader's leverage which we might not want.

### DIFF
--- a/coordinator/src/db/coordinator_leverages.rs
+++ b/coordinator/src/db/coordinator_leverages.rs
@@ -1,0 +1,72 @@
+use crate::schema::coordinator_leverages;
+use anyhow::Result;
+use diesel::dsl::max;
+use diesel::ExpressionMethods;
+use diesel::OptionalExtension;
+use diesel::PgConnection;
+use diesel::QueryDsl;
+use diesel::QueryResult;
+use diesel::Queryable;
+use diesel::RunQueryDsl;
+use rust_decimal::prelude::ToPrimitive;
+use std::collections::HashMap;
+
+#[derive(Queryable, Debug, Clone, PartialEq)]
+#[diesel(table_name = coordinator_leverages)]
+pub(crate) struct CoordinatorLeverage {
+    pub id: i32,
+    pub trader_leverage: i32,
+    pub coordinator_leverage: i32,
+}
+
+pub(crate) fn get_all(conn: &mut PgConnection) -> QueryResult<HashMap<u8, u8>> {
+    let all = coordinator_leverages::table.load::<CoordinatorLeverage>(conn)?;
+    Ok(all
+        .iter()
+        .map(|lev| {
+            (
+                lev.trader_leverage.to_u8().expect("to fit"),
+                lev.coordinator_leverage.to_u8().expect("to fit"),
+            )
+        })
+        .collect())
+}
+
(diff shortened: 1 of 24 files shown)


coordinator/src/db/coordinator_leverages.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// We assume we only have round numbers in the db and hence round `trader_leverage`. If no item can
+/// be found we take the max available coordinator_leverage
+pub fn find_coordinator_leverage_by_trader_leverage(
+    connection: &mut PgConnection,
+    trader_lev: f32,
+) -> f32 {
+    find_by_trader_leverage(connection, &trader_lev).unwrap_or_else(|err| {
+        tracing::error!("Failed at loading leverage from db. Falling back to 2.0 {err:#}");
+        2.0
```
luckysori, line 45: ❓I guess we don't need this?

coordinator/src/orderbook/websocket.rs
```diff
@@ -217,6 +218,15 @@ pub async fn websocket_connection(stream: WebSocket, state: Arc<AppState>) {
                         Ok(_) => {
                             let liquidity_options =
                                 db::liquidity_options::get_all(&mut conn).unwrap_or_default();
+                            let coordinator_leverages =
+                                db::coordinator_leverages::get_all(&mut conn).unwrap_or_default();
+                            // TODO(bonomat): we should make this configurable as well. Ideally, we
+                            // can differentiate between a _leverage_ when opening a position and a
+                            // _multiplier_ when opening a channel.
+                            let default_coordinator_leverage =
```
luckysori, line 226: ❓Why do we need a default coordinator leverage?

mobile/native/src/channel_trade_constraints.rs
```diff
@@ -31,6 +30,48 @@ pub struct TradeConstraints {
     pub order_matching_fee_rate: f32,
     /// Total collateral in the dlc channel, none if [`is_channel_balance`] is false.
     pub total_collateral: Option<u64>,
+    /// Defines what leverage the coordinator will take depending on what the trader takes
+    ///
+    /// Unfortunately our version of flutter_rust_bridge does not support hashmaps yet
+    pub coordinator_leverages: Vec<CoordinatorLeverage>,
+    /// The leverage the coordinator will take if none is provided in `coordinator_leverages`
```
luckysori, line 37: ❓Do we really need to cater for this situation by adding a default? If `coordinator_leverages` is empty I think something went very wrong and we shouldn't allow trading anyway.

luckysori (commented):
I like the simplicity of this approach, but I don't get why we need the default coordinator leverage.

Issue #1933, opened 2024-01-31
Fix layout on small devices


bonomat:
On a small device, e.g. a iPhone SE, the welcome carousel does not look too god as components are overlapping.

Restioson:
Is our target for screen resolution an iPhone SE (1334 × 750)? I'm happy to try and take this on so long as I know what resolution we are aiming for

PR #2631, opened 2024-06-17
fix: don't overflow UI on small screens

Restioson:
Addresses #1933 partially. This _does_ remove all overflow on small mobile screens that I could find, but it doesn't make it _nice_ necessarily in all cases (see below...)

Unfortunately, there is still one issue which will be harder to resolve:
- When the screen is very small, the position/order list in the trade tab correctly scrolls
- However, the TradingView chart takes up the lion's share of the space, meaning that you are viewing the position/order list through a pinhole, which is very frustrating and hard to see
- The ideal UX solution is to scroll the entire page - not just the position list. This way, the trading view widget will eventually disappear if you scroll down far enough.
- However, this is difficult. We can't just wrap the entire thing in a scroll widget because TabBarView also creates a scroll widget, so it won't work properly. If we remove the scroll widget in TabBarView, it has unconstrained height and refuses to lay itself out. IntrinsicHeight is flat-out disallowed it seems(?). The correct solution is a NestedScrollView, but this is a bit of a pain to implement.

Because it may take a bit longer to fix the trade screen, I've opened this PR in the meantime, as these fixes are separate.

### DIFF
--- a/mobile/lib/common/custom_qr_code.dart
+++ b/mobile/lib/common/custom_qr_code.dart
@@ -19,24 +19,27 @@ class CustomQrCode extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return SizedBox.square(
-      dimension: dimension,
-      child: QrImageView(
-        data: data,
-        eyeStyle: const QrEyeStyle(
-          eyeShape: QrEyeShape.square,
-          color: Colors.black,
+    return AspectRatio(
+      aspectRatio: 1,
+      child: SizedBox.square(
+        dimension: dimension,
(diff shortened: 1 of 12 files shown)


holzeis:
Can you add before and after screenshots?

Regarding the trading view. How about adjusting its height proportionally to the available space?

Restioson:
> Can you add before and after screenshots?
I can do this, though nothing should change really 
> Regarding the trading view. How about adjusting its height proportionally to the available space?
I think that could be a good solution! We can revisit it and make the scroll nice and fancy later if we want.

Issue #2630, opened 2024-06-17
Failed rollover if the wallet hasn't synced yet


holzeis:
If a position gets opened very close to the rollover window, it could happen that the app still hasn't synced yet, thus leaving the underlying contract in `signed` instead of `confirmed`.  

When opening a position we check if the underlying contract is `confirmed` we should do the same when processing a rollover. Note, I am referring here to the app side, as the coordinator already implements this check.

PR #2633, opened 2024-06-17
fix: Only accept a rollover offer if the contract has been confirmed.

holzeis:
Syncs the wallet eagerly if the channel is not yet confirmed.

If the contract is not confirmed, the rollover offer gets rejected. This fixes an issue where the app otherwise risk ending up in an intermediate state as it can't accept the `RenewConfirm` message.

fixes https://github.com/get10101/10101/issues/2630

### DIFF
--- a/coordinator/src/bin/coordinator.rs
+++ b/coordinator/src/bin/coordinator.rs
@@ -263,11 +263,10 @@ async fn main() -> Result<()> {
         node.inner.oracle_pubkey,
     );
     let _handle = rollover::monitor(
-        pool.clone(),
+        node.clone(),
         node_event_handler.subscribe(),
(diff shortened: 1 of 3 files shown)


holzeis:
@luckysori Any idea why 9635806fc4c065520d50588befbf30a4577fef53 doesn't fix the issue?

Even though I pass the check if confirmed contract check when processing the `RolloverOffer` processing the `RolloverConfirm` message still fails on an invalid state error. 

```
Failed to handle RolloverConfirm message from 02dd6abec97f9a748bf76ad502b004ce05d1b2d1f43a9e76bd7d85e767ffb022c9: Invalid state: Invalid state Contract { state: "signed" } expected Confirmed.
```

PR #2646, opened 2024-08-08
fix: fastlane

bonomat:


bonomat:
Giving testnet release a try

PR #2649, opened 2024-08-09
feat(bitmex-client): introduce getter for instrument from bitmex

bonomat:
This allows us to fetch funding rates etc from bitmex.


holzeis (approved):
LGTM 👍

PR #2660, opened 2024-10-28
Use blockstream's public API for clients

luckysori:
Since we are winding down the 10101 coordinator, apps will only work if they use an alternative API, such as Blockstream's public API.


luckysori:
Merging this because I want to test that it works.

PR #2664, opened 2024-10-28
Use `ConfirmationTarget::Background` to claim settle TX outputs

luckysori:
There is actually no protocol pressure to claim these outputs quickly, so it seems foolish to pay a possibly very high price for these claim transactions.

See https://github.com/get10101/rust-dlc/commit/906cb4d422dbd90bf2fe35991f2a709801c250e7.


luckysori:
Going ahead with the merge to prepare another app release.

